PR #7, opened 2017-06-03
small text update

danielskatz:
clarify that we don't actually build a DAG


sophie-bui:
@all-contributors
please add @danielskatz for code and content
please add @yadudoc for review

Issue #115, opened 2018-03-08
Tweak documentation organization


annawoodard:
The documentation is already in awesome shape-- this is just a minor thing that occurs to me while reading through it. We currently have class documentation under `Developer documentation` -> `Developer documentation`. I think we should reorganize that a bit to remove the redundancy. We currently have a very nice layout:

1. Quickstart
2. User guide (includes overview, detailed descriptions of various components, FAQ)
3. Developer documentation

As a new user, you'd probably see 'Developer documentation' and assume that was for Parsl devs only. But class documentation for user-facing classes is clearly relevant for them too. Sometimes, especially users that are relatively familiar with Parsl, but want to quickly figure out what the options are for a particular class want to quickly look up the class documentation and not have to search through the detailed descriptions. I vote to add a fourth section, after the User guide, called `Class documentation` or similar, which only contains the class documentation.

annawoodard:
Note that switching to making the class documentation more user-facing has an added benefit: it will not need to be synced with the documentation. For example, [here](https://parsl.readthedocs.io/en/latest/userguide/apps.html#special-keywords-arguments) the documentation is out-of-sync with the class signature. We should just make the docstring in the class signature clear enough that it does not need to be repeated there.

annawoodard:
@yadudoc I'm thinking we can split up the class documentation into classes that are relevant for users (`PythonApp`, etc) which go in the user class documentation, and classes that aren't ('Appbase`, etc), which go in the developer documentation. The point is to make it easy for users to find the class documentation which is relevant to them. Does that sound reasonable?

yadudoc:
I'd take a second opinion on this. Several projects like ipyparallel, boto, and even the C Programming Language book start with a beginners section followed by the reference section. My worry is that we'll make the userguide needlessly tedious by pushing class docs into it. For eg, classes documentation like the DFK contain info on methods that the user is never expected to use. The pythonApp class is simple because it has very few methods.

At it's core parsl does just a few things, I hope that the userguide is easy to run through and I worry that for a new user adding class docs may not help in the initial reading.

If the doc is outdated, that's a doc bug. The user facing docs still have to be written and maintained, and we realistically can rely on the class docs only as a supplement, esp for a new user. 

Perhaps we could either link, or include just specific interfaces rather than the class definition ?

annawoodard:
> I'd take a second opinion on this. Several projects like ipyparallel, boto, and even the C Programming Language book start with a beginners section followed by the reference section. 

[...]

PR #129, opened 2018-03-13
Tweak docs #115

annawoodard:
@yadudoc I think I might not have been explaining my proposal clearly enough, so since this was fast to implement I just went ahead and did it to show you what I'm thinking. You can see a demo of the compiled docs [here](http://www.crc.nd.edu/~awoodard/ci/_build/html/index.html).

This just adds a reference index for user-facing classes. This is the standard in most software I've used. 

I also tweaked the organization structure so that the full table of contents is viewable on a 13" screen (my laptop), and redundancies are removed (we had 'quickstart` listed twice and 'developer documentation' listed twice). Other minor changes: I added links to the source code, and enabled cross-referencing such that Parsl classes surrounded by backticks link to their class definition. Eventually we should try to use cross-referencing throughout the docstrings.

Note that I did not modify the developer documentation (I prefer we keep PRs as atomic as possible), though I had to add `noindex` in a few places so that the user class cross-references do not point to the dev docs. (I prefer the dev documentation just be folded into the docstrings and we use `sphinx-apidoc` to create stubs for everything and link from the dev docs, but I'll give @yadudoc a break from my hassling for a bit before I try to make the case for that :smile:)


annawoodard:
@yadudoc I've changed the name to 'Reference Guide' as suggested and split errors out into a separate table, so I think this is ready to go unless there is more feedback.

PR #150, opened 2018-03-21
Add note about git hooks to contributing docs

annawoodard:


yadudoc:
Looks like we have another travis test stalling. Here's the raw logs : https://api.travis-ci.org/v3/job/356533671/log.txt

Issue #106, opened 2018-03-06
Automatic checkpointing


annawoodard:
Checkpointing is currently manually triggered by the user. It would be convenient if users could specify automatic checkpointing at a given frequency and/or after each task completes.

As requested by @djf604.

yadudoc:
From the discussion with @djf604, this should have high priority given user triggered checkpointing has limited usability. I can start adding functionality to support this with the following modes (copied over from the previous issue : #43 

1. (eager) checkpoint at the completion of every task or
2. (lazy) at configurable intervals (say ~1hr).
3. (atExit) write out a checkpoint when the DFK exits, capturing early termination from app failures,
uncaught exceptions, keyboard interrupts etc. This won't work for cases when the DFK is killed abruptly (SIGKILL), machine powered off etc.

annawoodard:
@yadudoc Sounds good to me. I vote for a minor tweak to the naming scheme. I would guess without documentation that "lazy checkpointing is only performed when it is needed", not "lazy checkpointing is performed at regular intervals". Here's a possibility:

```
checkpoint=None # default, no checkpointing
checkpoint='periodic' # configurable intervals
checkpoint='task_exit'
checkpoint='dfk_exit'
```

PR #151, opened 2018-03-21
Auto checkpoint #106

yadudoc:
This brings in automatic checkpointing as described in #106 
Added following automated checkpointing methods

* task_exit
* dfk_exit
* periodic (with user specified intervals)

This also contains user docs, and tests fort the functionality added.


.travis.yml
```diff
@@ -17,5 +17,6 @@ os:
 script:
     - export PARSL_TESTING="true"
     - pip install -r test-requirements.txt
-    - nosetests -v parsl/tests/test_threads parsl/tests/test_data parsl/tests/test_checkpointing
+    - nosetests -v parsl/tests/test_threads parsl/tests/test_data
+    - for test in parsl/tests/test_checkpointing/test*; do nosetests -v $test; done
```
annawoodard: Why is this necessary? Nose will traverse through directory levels. Also, maybe we should change `-v` -> `-vx` so things fail faster.

docs/devguide/dev_docs.rst
```diff
@@ -96,7 +96,7 @@ DataFlowKernel
 ==============
 
 .. autoclass:: parsl.dataflow.dflow.DataFlowKernel
-   :members:  __init__, submit, cleanup, load_checkpoints, config, handle_update, launch_task, _count_deps, _count_all_deps, sanitize_and_wrap, _load_checkpoints
+   :members:  __init__, submit, cleanup, load_checkpoints, checkpoint, config, handle_update, launch_task, _count_deps, _count_all_deps, sanitize_and_wrap, _load_checkpoints
```
annawoodard: Isn't this just all the members? Unless we're trying to purposely exclude members, I think we should consider
```
:members:
```
instead of
```
:members:  __init__, submit, cleanup, load_checkpoints, checkpoint, config, handle_update, launch_task, _count_deps, _count_all_deps, sanitize_and_wrap, _load_checkpoints
```
because it is easier to maintain so that we don't have to remember to add a member here if we add one to the code. That applies everywhere...

docs/userguide/checkpoints.rst
```diff
@@ -18,46 +18,92 @@ will checkpoint all results that have updated since the last checkpoint. When lo
...
-``DataFlowKernel.checkpoint()`` function is invoked. The checkpoint is written
-out to the ``runinfo/RUN_ID/checkpoint/`` directory.
+Checkpointing works in the following modes :
 
-Example code follows:
+1. ``task_exit`` : In this mode, each app is checkpointed when it has completed or
+   failed (after retries if enabled). This mode reduces the risk of losing information
+   from completed tasks to a minimum.
+
+   >>> config["globals"]["checkpoint"] = 'task_exit'
```
annawoodard: Let's consider reserving the use of `>>>`. It is the Python interpreter prompt and the standard way of running [doc tests](https://docs.python.org/2/library/doctest.html) so people may think they should be able to cut and paste directly into a prompt to test something. By not using this prompt here we avoid possible confusion and could eventually support doc tests.

parsl/dataflow/dflow.py
```diff
@@ -94,12 +96,27 @@ def __init__(self, config=None, executors=None, lazyErrors=True, appCache=True,
             self.lazy_fail = self._config["globals"].get("lazyErrors", lazyErrors)
             self.fail_retries = self._config["globals"].get("retries", retries)
             self.flowcontrol = FlowControl(self, self._config)
+            self.checkpoint_method = self._config["globals"].get("checkpoint",
+                                                                 checkpointMethod)
+            if self.checkpoint_method == "periodic":
+                print("Setting periodic")
```
annawoodard: I vote not to mix print and logging statements

parsl/dataflow/dflow.py
```diff
@@ -46,7 +45,7 @@ class DataFlowKernel(object):
     """
 
     def __init__(self, config=None, executors=None, lazyErrors=True, appCache=True,
-                 rundir=None, retries=0, checkpointFiles=None):
+                 rundir=None, retries=0, checkpointFiles=None, checkpointMethod=None):
```
annawoodard: can we call this `checkpointMode` instead of `checkpointMethod`? This gives 'method' a double meaning so it is confusing.

parsl/dataflow/dflow.py
```diff
@@ -587,15 +621,18 @@ def checkpoint(self):
...
+                    logger.debug("Task {} checkpointed".format(task_id))
+
+        self.checkpointed_tasks += count
 
-        end = time.time()
         if count == 0:
-            logger.warn(
-                'No tasks checkpointed, please ensure caching is enabled')
+            if self.checkpointed_tasks == 0:
+                logger.warn("No tasks checkpointed, please ensure caching is enabled")
```
annawoodard: Can we just check if caching is enabled automatically and return an error if someone tries to instantiate the DFK this way?

parsl/tests/test_checkpointing/test_dfk_exit.py
```diff
@@ -0,0 +1,67 @@
+''' Testing python apps
```
annawoodard: This is not a helpful docstring 😄, I prefer to just omit it in this case.

parsl/dataflow/flow_control.py
```diff
@@ -141,8 +151,94 @@ def close(self):
...
+
+    '''
+
+    def __init__(self, callback, *args, interval=5):
+        ''' Initialize the flowcontrol object
+        We start the timer thread here
+
+        Args:
+             - dfk (DFK object) : DFK object to track parsl progress
+             - config (Dict) : Config dict structure
```
annawoodard: `DFK object` -> `DataFlowKernel`
`Dict` -> `dict`

If you use the actual class names sphinx will cross-reference them (it may need some tweaking to the config, I will submit it as a PR).

parsl/dataflow/flow_control.py
```diff
@@ -141,8 +151,94 @@ def close(self):
         self._thread.join()
 
 
+class Timer(object):
+    '''This timer is a simplified version of the FlowControl timer.
+    This timer does not employ notify events.
+
+    This is based on the following logic :
+
+    >>> BEGIN (INTERVAL, THRESHOLD, callback) :
```
annawoodard: Same comment about the Python interpreter prompt.

Also, this is just personal preference, so no pressure to actually change this: I prefer to avoid toy code in documentation in most cases. The actual code this toy code is describing is only ~2x as long, and Python syntax is so simple it is basically toy code. But sometimes toy code cannot convey all of the proper details, so in the end you read the toy code and the real code and have to do the additional work of mapping between the two. So I prefer to use docs for exposition. Just my two cents.

.travis.yml
yadudoc: Agreed on -v -> -vx.
Nosetests is doing something unusual when running these tests from a directory. They pass when called one by one. This was easier that debugging what nose is doing.

docs/userguide/checkpoints.rst
yadudoc: This is paste'able into the interpreter and is valid code. I don't understand the problem.
What I want to do is hilight code snippets assuming that users will copy paste it.

parsl/dataflow/dflow.py
yadudoc: This is harder. We can test for appCaching being enabled at the top_level, but we won't be able to find out if specific apps have caching enabled. This is why that warning is added.

annawoodard: Sounds good, thanks for explaining.

parsl/dataflow/dflow.py
```diff
@@ -94,12 +96,26 @@ def __init__(self, config=None, executors=None, lazyErrors=True, appCache=True,
             self.lazy_fail = self._config["globals"].get("lazyErrors", lazyErrors)
             self.fail_retries = self._config["globals"].get("retries", retries)
             self.flowcontrol = FlowControl(self, self._config)
+            self.checkpoint_method = self._config["globals"].get("checkpointMode",
```
annawoodard: @yadudoc where we have properties can they always mirror the option? That way we don't have to keep track of the mapping?

annawoodard: In other words `checkpoint_mode` instead of `checkpoint_method`

Issue #118, opened 2018-03-09
Add linting for docstrings


annawoodard:
Now that we have an awesome setup due to @yadudoc that plugs code linting into Travis, I think it makes sense to add linting for our docstrings. It can be done via:

```
pep257 --ignore=D100,D101,D102,D103,D104
```

Or if you want more info:
```
pep257 --ignore=D100,D101,D102,D103,D104  --explain --source
```
The full list of errors can be found [here](http://pep257.readthedocs.io/en/latest/error_codes.html). Resolving this issue should also include fixing the current things pep257 is complaining about. I can volunteer for this, since I'm currently going through the code to get familiar with it.

benclifford:
Installing this: ```pip install flake8-docstrings``` makes the existing flake8 CI call check docstrings for pep257 compliance. There are 2416 violation reports as of now.

PR #160, opened 2018-03-23
Lint docstrings #118

annawoodard:
This partially addresses #118. I had started working on it, but it is low-priority so I haven't finished it yet. I want to merge what I have so far now in order to reduce effort later keeping the branch synced.


annawoodard:
FYI @yadudoc I've tried to be careful to only include minor changes, the minimal required to fix the docstring error. So hopefully reviewing this shouldn't be too arduous even though there are many changed files

Issue #71, opened 2018-01-19
Support for Globus transfer?


dlebauer:
It would be great to have support for Globus transfer within Parsl. We are interested in, e.g., transferring 10s-1000sGB from a server at NPCF to an XSEDE resource such as Comet or Bridges for (re)processing.

We aren't currently using Parsl but the abstraction of queue managers is an attractive feature. Globus transfer would be another particularly useful feature, that we will need to implement somehow. The docs say Globus support will appear "in the near future" and I see it listed in the roadmap. However, I can't find an open issue so I am opening one. 

My questions are: 

1. What is the status of this feature?
2. How might it be implemented?

kylechard:
We have started working on Globus integration but don't yet have a timeline for when it will be released. 

The basic model is built around a Globus file abstraction that will allow scripts to be written with references to remote (Globus accessible data). Parsl will handle the staging of data to the endpoint/worker node based on the site configuration.  As part of this work we're adding Globus Auth support to authenticate through the Parsl script. 

Do you have any thoughts or requirements for how you'd like Globus to be integrated? We could also put together a proof of concept by using Parsl + the Globus SDK to get a better idea of how this could/should work.

dlebauer:
We have a meeting Feb 2 with your group that @danielskatz set up. We can discuss then, but I wanted to get the discussion started so that we can make the most of that time. For requirements, I'll defer to the expert, @max-zilla, but here is a start:

* we want to be able to transfer data, run scripts, then return output data via globus. There is also a lot of communication with APIs.
* between July-Nov we would like to process about a PB of data this way.
* we also want users to be able to run and develop pipelines on XSEDE and other resources using our data
* @zongyangli is the first person working out a workflow for transferring data from the #terraref endpoint to comet for offline processing: https://github.com/terraref/computing-pipeline/issues/385
* the file abstraction sounds nice. From the user perspective, is it similar to how THREDDS works, where I can just replace a file path with a url?

yadudoc:
Explicit staging support is ready for v0.5.0, but implicit staging is deferred for 0.6.0

yadudoc:
Explicit staging is now supported in master along with Globus support.

PR #161, opened 2018-03-23
Globus #71

lukaszlacinski:
Add:
1) File with the 'globus' scheme,
2) Globus Auth support (Globus class),
3) DataManager class,
4) explicit staging in/out,
5) a simple test that stages in a file with unsorted string from a Globus shared endpoint shared with 'public' to the 'UChicago RCC Midway' endpoint, sorts the strings on Midway, and stages out a files with sorted strings to go#ep1/~/,
6) extends data.rst (Data Management) to describe explicit staging with an example code cut out from test_data/test_explicit_staging.py


lukaszlacinski:
> I'm not sure I completely follow this explanation. The total time for processing + waiting on staging will be the same, no?
Parsl can run tasks concurrently. I see an opportunity for DataManager to handle file staging in a smart way to minimize waiting time, startup time, etc. For example, DataManager can start a file transfer before a task that is in the middle of the worklfow requests it. Or, if many tasks (n) need a different file from the same Globus endpoint, all of the file transfers can be combined in one Globus transfer to avoid startup time for n-1 transfers.

PR #175, opened 2018-03-29
Adding the fspath function to the File object to support opening of f…

### LINKED ISSUE #174: PathLike behavior for File
With the recent changes to the DataFuture behavior and the File based model that we are moving to in general, we want to have File behave as closely to old behavior as possible so that there's least amount of code breakage. This is one such case:

Earlier we could do this since the result of a DataFuture was a string
> f = open(DataFuture1.result(), 'r') 

Now the result of a DataFuture is a File, which can't simply be passed to open:
> f = open(File, 'r') <-- Fails 

This can be fixed by making File follow the [os.PathLike](https://docs.python.org/3/library/os.html#os.PathLike) interface

kylechard:
…iles. Fixes #174.

### DIFF
--- a/parsl/data_provider/files.py
+++ b/parsl/data_provider/files.py
@@ -15,7 +15,7 @@ from parsl.data_provider.data_manager import DataManager
 logger = logging.getLogger(__name__)
 
 
-class File(object):
+class File(str):
     """The Parsl File Class.
 
     This is planned to be a very simple class that simply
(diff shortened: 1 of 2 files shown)


annawoodard:
@yadudoc @kylechard: Not sure if this would be worth the convolution, but subclassing `File` from `str` would allow for `open(File, ...)` and work on Python 3.5+.

PR #184, opened 2018-04-02
Fix ipp tests

annawoodard:
This is partial (there are still other issues to be addressed, but I wanted to go ahead and merge what I've got so far).

### DIFF
--- a/parsl/tests/test_ipp/bash_app.py
+++ b/parsl/tests/test_ipp/bash_app.py
@@ -12,7 +12,7 @@ import argparse
 # parsl.set_stream_logger()
 
 workers = IPyParallelExecutor()
-dfk = DataFlowKernel(workers)
+dfk = DataFlowKernel(executors=[workers])
 
 
 @App('bash', dfk)
--- a/parsl/tests/test_ipp/bash_app.py
+++ b/parsl/tests/test_ipp/bash_app.py
@@ -64,7 +64,7 @@ def test_parallel_for(n=10):
 
     start = time.time()
     for i in range(0, n):
-        d[i], _ = echo_to_file(inputs=['Hello World {0}'.format(i)],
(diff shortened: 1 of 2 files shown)


parsl/tests/test_ipp/bash_app.py
```diff
@@ -12,12 +12,15 @@
 # parsl.set_stream_logger()
 
 workers = IPyParallelExecutor()
-dfk = DataFlowKernel(workers)
+dfk = DataFlowKernel(executors=[workers])
```
yadudoc: Sorry these tests have not been updated!

parsl/tests/test_ipp/bash_app.py
```diff
@@ -12,12 +12,15 @@
...
 
 
 @App('bash', dfk)
 def echo_to_file(inputs=[], outputs=[], stderr='std.err', stdout='std.out'):
-    pass
+    res = ""
+    for i, o in zip(inputs, outputs):
```
yadudoc: This assumesj len(inputs) == len(outputs).
Might be safer to do : 
return "for i in {0}; do echo {1} &> $i ; done".format(" ".join(outputs), " ".join(inputs) )

parsl/tests/test_ipp/bash_app.py
```diff
@@ -64,12 +67,12 @@ def test_parallel_for(n=10):
 
     start = time.time()
     for i in range(0, n):
-        d[i], _ = echo_to_file(inputs=['Hello World {0}'.format(i)],
```
yadudoc: sorry! ancient code.

parsl/tests/test_ipp/bash_app.py
annawoodard: Good point @yadudoc, I've updated it.

PR #201, opened 2018-04-10
File pickling #200

### LINKED ISSUE #200: Passing Files breaks over IPP
The new File class contains a dictionary that maps DataFutures for each site to which it is being staged and contains a reference to the DataManager. Neither of these are pickle-able.

So if we do something like this :+1: 
```
data = File("foo.txt")
fu = remote_app(inputs=[data])
fu.result() # <--- We'll get an error from here
```
Here's the relevant piece from the exception traceback :
```
File "/usr/local/lib/python3.5/dist-packages/ipyparallel/serialize/serialize.py", line 112, in serialize_object
   buffers.insert(0, pickle.dumps(cobj, PICKLE_PROTOCOL))
TypeError: can't pickle _thread.lock objects
```

I believe that the File object is the best place to hold the Future information about itself, and that would give us the opportunity to do smarter file staging in the future. So I propose that we fix this with a custom pickler for the File class.

This is blocker for 0.5.0.

yadudoc:
Adding some minimal code to support custom pickling for File objects such that they can be dispatched via IPP.
Added regression tests.
Fixes #200

### DIFF
--- a/parsl/data_provider/files.py
+++ b/parsl/data_provider/files.py
@@ -90,7 +90,6 @@ class File(str):
         return self.dman.stage_out(self)
 
     def set_data_future(self, df, site=None):
-        print("Setting datafuture")
         self.data_future[site] = df
 
     def get_data_future(self, site):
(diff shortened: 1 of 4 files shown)


lukaszlacinski:
Also tested using apps that stage in and out files. All completed successfully.

yadudoc:
Fixes #200

PR #214, opened 2018-04-19
Also offer BibTeX citation snippet

katrinleinweber:
Hello :-)

would this be of interest as well? It's the same snippet that Zenodo exports.

Cheers!

### DIFF
--- a/CITATION.md
+++ b/CITATION.md
@@ -1,1 +1,21 @@
 Babuji, Yadu, Brizius, Alison, Chard, Kyle, Foster, Ian, Katz, Daniel S., Wilde, Michael, & Wozniak, Justin. (2017, August 30). Introducing Parsl: A Python Parallel Scripting Library. Zenodo. https://doi.org/10.5281/zenodo.853492
+
+or
+
+```{tex}
+@misc{babuji_yadu_2017_891533,
+  author       = {Babuji, Yadu and
+                  Brizius, Alison and
+                  Chard, Kyle and
+                  Foster, Ian and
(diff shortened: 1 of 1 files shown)


danielskatz:
This seems like a good addition to me.

Alternatively, we could include [BibTeX](https://zenodo.org/record/853492/export/hx)

or a copy of what's in Export on the bottom right of https://zenodo.org/record/853492#.WtiO1y-ZOJQ

I'm not sure which would be best.

yadudoc:
This is a super useful addition @katrinleinweber !

danielskatz:
@yadudoc @annawoodard  Did you read my comment above?

annawoodard:
> Alternatively, we could include "BibTeX"

I don't follow, this PR is bibtex.

> or a copy of what's in Export on the bottom right of https://zenodo.org/record/853492#.WtiO1y-ZOJQ
I think this would be nice but will take more time; we should open it as a separate issue.

danielskatz:
we could copy the "BibTex" link from the zenodo page, and let it resolve back to the Zenodo page itself

Or we could copy all the output options from the zenodo page.

danielskatz:
basically, my question is we should duplicate the bibtex info on our page or point (link) to what zenodo generates directly

annawoodard:
I would say that far and away bibtex will be the most common citation method, so I prefer putting the text directly as is done in this PR, because it saves people time. We can add a link separately, i.e., "for more options, visit here"

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #59, opened 2017-12-19
Circular dependency in Parsl


yadudoc:
Reported by @dfj604 

The following code will hang due to a circular dependency : 
```
@App('python', dfk)
def App_A (x):
    import time
    time.sleep(0.2)
    return x*2

@App('python', dfk)
def AppSum (inputs=[]):
    return sum(inputs)

# Creating a list of App_Futures
app_futures = [ App_A(i) for i in range(0,2) ]
# Sum depends on the future results from the previous step
app_futures.extend([AppSum(inputs=app_futures)])
# However by extending app_futures with the future from AppSum, AppSum
# now depends on itself and will block forever.

print(app_futures)
[i.result() for i in app_futures]
```

While this maybe not ideal code from the user, this is something parsl should try to protect against.
Making a deep copy of the args and kwargs to immutable containers (say tuples) could help.

PR #218, opened 2018-04-19
Adding some troubleshooting info

yadudoc:
Provides documentaion fix for #59
Adding troubleshooting for network issues with IPP.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -185,3 +185,33 @@ Here's an example of running a python2.7 code as a bash application:
        return '''conda activate py2.7_env  # Use conda to ensure right env
        python2.7 my_python_app.py -arg {0} -d {1}
        '''.format(arg1, arg2)
+
+Parsl hangs
+^^^^^^^^^^^
+
+There are a few common situations in which a Parsl script might hang:
+
+1. Circular Dependency in code
+   If an `app` takes a list as an `input` argument and the future returned
+   is added to that list, it creates a circular dependency that cannot be resolved.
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -185,3 +185,33 @@ Here's an example of running a python2.7 code as a bash application:
...
+
+Parsl hangs
+^^^^^^^^^^^
+
+There are a few common situations in which a Parsl script might hang:
+
+1. Circular Dependency in code
+   If an `app` takes a list as an `input` argument and the future returned
+   is added to that list, it creates a circular dependency that cannot be resolved.
+   This situation is described `here <https://github.com/Parsl/parsl/issues/59>`_ in more detail
```
annawoodard: detail -> detail.

docs/faq.rst
```diff
@@ -185,3 +185,33 @@ Here's an example of running a python2.7 code as a bash application:
...
+
+2. Workers requested are unable to contact the Parsl client due to one or
+   more issues listed below :
```
annawoodard: below : -> below:

docs/faq.rst
```diff
@@ -185,3 +185,33 @@ Here's an example of running a python2.7 code as a bash application:
...
+
+   * Parsl client does not have a public IP (e.g. laptop on wifi)
```
annawoodard: Should we add advice here for this case?

yadudoc: What options do we have besides getting a publicIp (plugging into a network that does publicIPs) or ssh'ing to a machine that has a publicIp ?

annawoodard: Exactly, let's add that.

Issue #230, opened 2018-04-25
`task_exit` checkpointing erases repeatedly truncates checkpoint file during run


benclifford:
With `config['globals']['checkpointMode'] = 'task_exit'`, the checkpoint file remains small, and sometimes shrinks, sometimes even to 0.

It looks like the file opening mode is wrong. I have a one line fix.

```
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 90 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 0 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 0 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 0 Apr 25 10:13 tasks.pkl
benc@dogger:~/parsl/desc-skeleton/runinfo/107/checkpoint$ ls tasks.pkl -l
-rw-r--r-- 1 benc benc 0 Apr 25 10:13 tasks.pkl
[...]

PR #231, opened 2018-04-25
Change checkpoint file modes to append. Fixes issue #230.

benclifford:
Previous behaviour truncated and overwrote the checkpoint
file at each checkpoint, although it was intended to
append.

The previous mode, wb+, means to open the file for both reading and
writing, truncating if necessary - some documentation refers to
this as "updating", which may be the source of confusion.

The new mode, ab, means to open the file for writing, but without
truncating, and positioned initially at the end of the file.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -606,7 +606,7 @@ class DataFlowKernel(object):
             pickle.dump(state, f)
 
         count = 0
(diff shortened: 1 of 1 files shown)


yadudoc:
Holy cow! Thanks for catching this @benclifford

annawoodard:
I didn't want to hold this up so I opened a separate issue to add a regression test for this in #233

PR #236, opened 2018-04-25
Remove comparison to `Parameter.empty`

### LINKED ISSUE #226: Pandas DataFrames as Default Values
Parsl does not work with Pandas DataFrames as the default value to a keyword argument.

For example,

```python
data = pd.DataFrame({'x': [None,2,[3]]})
@App('python', dfk)
def f(d=data):
    return d.sum()
f().result()
```
yields an error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-23-cba3cd6e0846> in <module>()
      3 def f(d=data):
      4     return d.sum()
----> 5 f().result()

~/miniconda3/lib/python3.6/site-packages/parsl/app/app_factory.py in __call__(self, *args, **kwargs)
     73                                  walltime=self.walltime,
     74                                  cache=self.cache,
---> 75                                  fn_hash=self.func_hash)
     76         return app_obj(*args, **kwargs)
     77 

~/miniconda3/lib/python3.6/site-packages/parsl/app/python_app.py in __init__(self, func, executor, walltime, cache, sites, fn_hash)
     16         This bit is the same for both bash & python apps.
     17         """
---> 18         super().__init__(func, executor, walltime=walltime, sites=sites, exec_type="python")
     19         self.fn_hash = fn_hash
     20         self.cache = cache

~/miniconda3/lib/python3.6/site-packages/parsl/app/app.py in __init__(self, func, executor, walltime, sites, cache, exec_type)
     47         self.kwargs = {}
     48         for s in sig.parameters:
---> 49             if sig.parameters[s].default != Parameter.empty:
     50                 self.kwargs[s] = sig.parameters[s].default
     51 

~/miniconda3/lib/python3.6/site-packages/pandas/core/generic.py in __nonzero__(self)
   1119         raise ValueError("The truth value of a {0} is ambiguous. "
   1120                          "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
-> 1121                          .format(self.__class__.__name__))
   1122 
   1123     __bool__ = __nonzero__

ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().```

annawoodard:
This check fails for classes which do not have `__eq__` implemented. I
do not think we actually need to do the check, so I am removing it.
Fixes #226.

Note I'm currently re-organizing the tests, so I'm just dropping the regression test in the top testing directory for now.

### DIFF
--- a/parsl/app/app.py
+++ b/parsl/app/app.py
@@ -4,7 +4,7 @@ The APP class encapsulates a generic leaf task that can be executed asynchronous
 
 """
 import logging
-from inspect import signature, Parameter
+from inspect import signature
 
 # Logging moved here in the PEP8 conformance fixes.
 logger = logging.getLogger(__name__)
@@ -44,10 +44,6 @@ class AppBase (object):
         self.cache = cache
 
         sig = signature(func)
-        self.kwargs = {}
-        for s in sig.parameters:
-            if sig.parameters[s].default != Parameter.empty:
-                self.kwargs[s] = sig.parameters[s].default
 
         self.stdout = sig.parameters['stdout'].default if 'stdout' in sig.parameters else None
         self.stderr = sig.parameters['stderr'].default if 'stderr' in sig.parameters else None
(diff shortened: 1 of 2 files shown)


annawoodard:
This is wrong, I will re-open to clarify the PR is changing.

Issue #221, opened 2018-04-23
DataFlowKernel cleanup is not aggressive enough


benclifford:
dfk.cleanup() can take a very long time to shutdown.

It looks like before shutting down, all queued apps that do not have unsatisfied dependencies are executed. (subsequent apps with dependencies do not execute before shutdown)

In my example, this means that a workflow will continue to run for several hours draining a pile of 15 minutes apps before shutdown; it wasn't initially apparent that the subsequent failure to continue with the workflow was because cleanup was in progress.

My example code sets itself up like this:

    workers = parsl.ThreadPoolExecutor(max_workers=5)
    dfk = parsl.DataFlowKernel(executors=[workers])

yadudoc:
@benclifford Here you are creating the worker pool explicitly, and the DFK is not responsible for resource handling. So you end up with cleanup(

benclifford:
By "explicitly", does that mean "declared in the python file rather than in the json config file?"

annawoodard:
I vote that we remove this mode of declaring the dfk, if it results in different behavior:
```
workers = parsl.ThreadPoolExecutor(max_workers=5)
dfk = parsl.DataFlowKernel(executors=[workers])
```
It is true that in some cases users may wish to have more control over executors, as this mode allows, but I think that can be solved with class-based configuration (which will allow us to remove the provider factory) as proposed in #133.

benclifford:
I changed to this:

``
from parsl.configs.local import localThreads as config
dfk = parsl.DataFlowKernel(config=config)
``

It looks like it still drains some of the queued apps over a period of time, but now also throws an exception as some (but not all) of the apps complete:

```
RuntimeError: cannot schedule new futures after shutdown
```

I'll note that in both cases, the `dfk.cleanup()` call returns very fast (less than 0.1 seconds) and job execution continues even after it has returned.

yadudoc:
Okay, I'm tagging this as a bug.

benclifford:
Attached is the workflow that is giving me the long-slowdown problem.

This should print the string `X12` to console (indicating that drk has returned from cleanup) very early on in the run, and then continue running stuff for a minute or so afterwards.

[fake_drp_pipeline.txt](https://github.com/Parsl/parsl/files/1940185/fake_drp_pipeline.txt)

yadudoc:
It looks like this might still be an issue with threads.

PR #237, opened 2018-04-25
Task checkpointing #234

### LINKED ISSUE #234: Race condition in task checkpointing
With `config["globals"]["checkpointMode"] = "task_exit"`, the DFK attempts to checkpoint every task at task completion. However in the case of fast apps without dependencies, the task returns before the AppFuture is created. This results in a race condition between the checkpoint method and the task update.

If the checkpoint method is called first we get :
```
  File "test_regression_233.py", line 38, in <module>
    test_checkpoint_availability()
  File "test_regression_233.py", line 27, in test_checkpoint_availability
    original = run_checkpointed([])
  File "test_regression_233.py", line 18, in run_checkpointed
    x = cached_rand(i)
  File "/home/yadu/src/parsl/parsl/app/app_factory.py", line 76, in __call__
    return app_obj(*args, **kwargs)
  File "/home/yadu/src/parsl/parsl/app/python_app.py", line 46, in __call__
    **kwargs)
  File "/home/yadu/src/parsl/parsl/dataflow/dflow.py", line 499, in submit
    task_id, func, *new_args, **kwargs)
  File "/home/yadu/src/parsl/parsl/dataflow/dflow.py", line 322, in launch_task
    exec_fu.add_done_callback(partial(self.handle_update, task_id))
  File "/usr/lib/python3.5/concurrent/futures/_base.py", line 376, in add_done_callback
    fn(self)
  File "/home/yadu/src/parsl/parsl/dataflow/dflow.py", line 218, in handle_update
    self.checkpoint(tasks=[task_id])
  File "/home/yadu/src/parsl/parsl/dataflow/dflow.py", line 617, in checkpoint
    if self.tasks[task_id]['app_fu'].done() and \
AttributeError: 'NoneType' object has no attribute 'done'
```

yadudoc:
Fixes #234 

DFK relies on internal task status instead of futures, avoiding a race-condition.
Updates the task_exit test to look at the checkpoint files rather than logs.


parsl/dataflow/dflow.py
```diff
@@ -606,10 +621,13 @@ def checkpoint(self):
...
         count = 0
+
         with open(checkpoint_tasks, 'ab') as f:
-            for task_id in self.tasks:
-                if self.tasks[task_id]['app_fu'].done() and \
-                   not self.tasks[task_id]['checkpoint']:
+            for task_id in checkpoint_queue:
+
+                if not self.tasks[task_id]['checkpoint'] and \
+                   self.tasks[task_id]['status'] in (States.done, States.failed, States.dep_fail):
```
annawoodard: I don't quite follow why we would checkpoint failed tasks? For example, what if there was a transient issue causing tasks to fail, the user stops their script and fixes the issue, then re-runs their script?

parsl/tests/test_cleanup/test_cleanup.py
```diff
@@ -0,0 +1,34 @@
...
+        f = slow_double(i)
+        round_1.append(f)
+
+    round_2 = []
+    for i in round_1:
+        f = slow_double(i)
+        round_2.append(f)
+
+    dfk.cleanup()
+
```
annawoodard: I think we need to assert something here?

parsl/dataflow/dflow.py
benclifford: I think checkpointing failures is the wrong thing to do (in most cases?)

parsl/tests/test_cleanup/test_cleanup.py
yadudoc: Well, I don't know what to assert yet. I think the mistake was in this test getting added to this PR.
This was for figuring out the #221 behavior.

parsl/dataflow/dflow.py
yadudoc: Makes sense. There's so much external, non-trackable things that will influence apps.

I'll change this behavior.

parsl/tests/test_cleanup/test_cleanup.py
yadudoc: This (incomplete) test should've gone in on a separate PR. I was watching for cleanup behavior, and there isn't something to assert on here yet.

parsl/dataflow/dflow.py
```diff
@@ -572,12 +578,14 @@ def cleanup(self):
...
 
-    def checkpoint(self):
+    def checkpoint(self, tasks=None):
         """Checkpoint the dfk incrementally to a checkpoint file.
 
         When called, every task that has been completed yet not
         checkpointed is checkpointed to a file.
 
+        Kwargs:
+            - tasks (List of task ids) : List of task ids to checkpoint. Default=None
```
benclifford: clarify what happens in None case (assuming that the intention isn't to checkpoint nothing...)

parsl/tests/test_sites/test_ec2/ec2.py
```diff
@@ -1,91 +0,0 @@
-ec2OnDemand = {
```
benclifford: is this EC2 stuff relevant to the PR?

yadudoc: It's not I keep pushing these changes to the wrong branch instead of master. While not related, can we keep it so that it'll get to master ?

parsl/tests/test_checkpointing/test_regression_239.py
```diff
@@ -0,0 +1,85 @@
...
+            while f:
+                tasks.append(pickle.load(f))
+                print
+        except EOFError:
+            pass
+        print("Tasks from cache : ", tasks)
+        assert len(tasks) == 1, "Expected {} checkpoint events, got {}".format(1, len(tasks))
+
+
+def test_checkpointing_at_dfk_exit():
```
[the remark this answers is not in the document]
yadudoc: Agreed. Removing dfk_exit test.

parsl/tests/test_checkpointing/test_regression_239.py
```diff
@@ -0,0 +1,85 @@
+from parsl import App, DataFlowKernel
+import pickle
+import time
+
+
+def run_checkpointed(n=5, mode="task_exit"):
```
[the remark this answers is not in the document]
yadudoc: Done

parsl/dataflow/dflow.py
```diff
@@ -573,12 +578,16 @@ def cleanup(self):
 
         logger.info("DFK cleanup complete")
 
-    def checkpoint(self):
+    def checkpoint(self, tasks=None):
```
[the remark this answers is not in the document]
yadudoc: With checkpointing mode `task_exit` the checkpoint method is called at the completion of each task. Earlier, we were iterating over the entire list of tasks to identify what is to be checkpointed, when we really only needed to checkpoint a specific task.

Issue #50, opened 2017-12-12
hide dfk


danielskatz:
If 99% of users and usages don't use the dfk variable, but simply pass it in and out of various functions, we should hide it by default.

Can we do this in a backward compatible way?  (make dkf an optional arg)

This probably requires major documentation/tutorial changes.

PR #252, opened 2018-05-01
Allow default dfk #50

annawoodard:
This PR adds a singleton class for loading the DFK which allows for more natural importing of apps. These changes were necessary in order to allow the testing framework to swap DFKs in and out such that we can factorize the testing code and the testing configurations, and cycle through such that each test is run on each configuration. Note it is possible to load a DFK either from a file containing a configuration or from a local DFK instance; I have only documented the latter way in the docstring, trying to keep things simple for now. Here is a preview of the new documentation:

screen shot 2018-05-01 at 10 26 50 am

### DIFF
--- a/docs/userguide/examples/library.py
+++ b/docs/userguide/examples/library.py
@@ -1,0 +1,5 @@
+from parsl import App
+
+@App('python')
+def increment(x):
+    return x + 1
--- a/docs/userguide/examples/run_increment.py
+++ b/docs/userguide/examples/run_increment.py
@@ -1,0 +1,9 @@
+import parsl
+from config import local_threads
+from library import increment
+
+parsl.load(local_threads)
+
+for i in range(5):
+    print('{} + 1 = {}'.format(i, increment(i).result()))
+
--- a/docs/userguide/importing.rst
+++ b/docs/userguide/importing.rst
@@ -1,0 +1,36 @@
+.. _codebases
+
+Importing Parsl apps
+--------------------
+
+It may be convenient to define Parsl apps separately from the definition of the
+:class:`~pars.dataflow.dflow.DataFlowKernel`, or in libraries of apps which are
+intended to be imported by other modules. For this reason, the
+:class:`~parsl.dataflow.dflow.DataFlowKernel` is an optional argument to the
+:func:`~parsl.app.app.App` decorator. If the
+:class:`~pars.dataflow.dflow.DataFlowKernel` is not passed to the
+:func:`~parsl.app.app.App` decorator, a configuration must be loaded using
+:meth:`parsl.load <parsl.DataFlowKernelLoader.load>` prior to calling the app.
+
+The configuration can be defined in the Parsl script, or elsewhere before being imported.
+As an example of the latter, consider a file called ``config.py`` which contains the
+following definition:
+
+.. literalinclude:: examples/config.py
+
+In a separate file called ``library.py``, we define:
+
+.. literalinclude:: examples/library.py
+
+Putting these together in a third file called ``run_increment.py``, we load the
+configuration from ``config.py`` before calling the ``increment`` app:
+
+.. literalinclude:: examples/run_increment.py
+
+Which produces the following output::
+
+    0 + 1 = 1
+    1 + 1 = 2
+    2 + 1 = 3
+    3 + 1 = 4
+    4 + 1 = 5
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -736,3 +736,36 @@ class DataFlowKernel(object):
             raise BadCheckpoint("checkpointDirs expects a list of checkpoints")
 
         return self._load_checkpoints(checkpointDirs)
+
+
+class DataFlowKernelLoader(object):
+    """Manage which DataFlowKernel is active.
+
+    This is a singleton class containing only class methods. You should not
+    need to instantiate this class.
+    """
+
+    _dfk = None
+
+    @classmethod
+    def load(cls, config):
+        """Load a DataFlowKernel.
+
+        Args:
+            - config (dict) : Configuration to load. This config will be passed to a
+              new DataFlowKernel instantiation which will be set as the active DataFlowKernel.
+        Returns:
+            - DataFlowKernel : The loaded DataFlowKernel object.
+        """
+        if cls._dfk is not None:
+            raise RuntimeError('Config has already been loaded')
+        cls._dfk = DataFlowKernel(config=config)
+
+        return cls._dfk
+
+    @classmethod
+    def dfk(cls):
+        """Return the currently-loaded DataFlowKernel."""
+        if cls._dfk is None:
+            raise RuntimeError('Must first load config')
+        return cls._dfk
--- a/docs/devguide/dev_docs.rst
+++ b/docs/devguide/dev_docs.rst
@@ -88,7 +88,7 @@ Exceptions
 
 .. autoclass:: parsl.app.errors.DependencyError
 
-.. autoclass:: parsl.dataflow.error.DataFlowExceptions
+.. autoclass:: parsl.dataflow.error.DataFlowException
 
 .. autoclass:: parsl.dataflow.error.DuplicateTaskError
 
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -10,6 +10,7 @@ Reference Guide
     parsl.app.app.App
     parsl.app.futures.DataFuture
     parsl.dataflow.futures.AppFuture
+    parsl.dataflow.dflow.DataFlowKernelLoader
     parsl.data_provider.files.File
 
 .. autosummary::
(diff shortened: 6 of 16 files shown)


parsl/dataflow/dflow.py
```diff
@@ -717,3 +718,58 @@ def load_checkpoints(self, checkpointDirs):
...
+            - DataFlowKernel : The loaded DataFlowKernel object.
+        """
+        if isinstance(payload, DataFlowKernel):
+            cls._dfk = payload
+        else:
+            spec = importlib.util.spec_from_file_location("", payload)
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
+            if hasattr(module, 'config'):
+                cls._dfk = DataFlowKernel(config=module.config, **kwargs)
```
yadudoc: I don't think this is a good/clean idea.  Why would we not allow the payload to be a config ? I might be missing some advantage to this method.

parsl/dataflow/dflow.py
```diff
@@ -717,3 +718,58 @@ def load_checkpoints(self, checkpointDirs):
...
+        """If there is no active DataFlowKernel, load one.
+
+        Args:
+            - payload (str or DataFlowKernel) : Payload to load in case there is no active DataFlowKernel.
+              If a string is passed, it should be the path to a file which defines a dictionary named `config`;
+              this will be passed to the DataFlowKernel constructor.
+            - **kwargs : Arbitrary keyword arguments which are passed to the DataFlowKernel constructor.
+
+        """
+        if cls._dfk is None:
```
yadudoc: Should we return a bool, so that whoever calls can know if this succeeded ?

docs/userguide/examples/library.py
```diff
@@ -0,0 +1,5 @@
+from parsl import App
+
+@App('python')
```
yadudoc: So, `@App('python')` , `@App('python', dfk)` and `@App('python', dfk=DFK)` are all legal now ? This does look better.

parsl/dataflow/dflow.py
annawoodard: I don't follow. If it doesn't succeed, it will raise an error, which the caller can handle. That seems cleaner to me than checking against a bool. In the former case, the error-handling code would only need to be traversed if there's an error, whereas a check against a bool would have to be traversed every time.

parsl/dataflow/dflow.py
annawoodard: > Why would we not allow the payload to be a config ?

To make sure I understand, are you asking, "why would we allow the payload to be anything other than a config"? The main motivation was not to force users to always have the config defined in a separate file; they might want to define it locally and then import apps. Here are some alternatives:

- Payload could accept a string (the filename to load) or a dict (a locally-defined config)
- We could split this into two methods (naming suggestions welcome, maybe `load_file` and `load_config`?)

docs/userguide/examples/library.py
annawoodard: 👍

parsl/dataflow/dflow.py
annawoodard: @yadudoc: agreed, I've switched to just loading the config dict and rebased onto master.

docs/userguide/examples/library.py
benclifford: one more variable assignment to get @python

parsl/dataflow/dflow.py
```diff
@@ -736,3 +736,51 @@ def load_checkpoints(self, checkpointDirs):
...
+        Args:
+            - config (dict) : Configuration to load. This config will be passed
+              to a new DataFlowKernel instantiation which will be set as the active
+              DataFlowKernel.
+            - **kwargs : Arbitrary keyword arguments which are passed to the
+              DataFlowKernel constructor.
+
+        """
+        if cls._dfk is None:
+            cls.load(config, **kwargs)
```
benclifford: Don't silently ignore a load if this is already initialised: probably throw an exception - it's a horrible user experience to spend 3 days wondering why your configuration is being ignored.

docs/userguide/importing.rst
```diff
@@ -0,0 +1,36 @@
...
+
+It may be convenient to define Parsl apps separately from the definition of the
+:class:`~pars.dataflow.dflow.DataFlowKernel`, or in libraries of apps which are
+intended to be imported by other modules. For this reason, the
+:class:`~parsl.dataflow.dflow.DataFlowKernel` is an optional argument to the
+:func:`~parsl.app.app.App` decorator. If the DFK is not passed to the
+:func:`~parsl.app.app.App` decorator, it must be loaded using
+:meth:`DataFlowKernelLoader.load <parsl.dataflow.dflow.DataFlowKernelLoader.load>` prior
+to importing the app.
+
```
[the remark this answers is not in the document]
benclifford: From a brief look at the code, it looks like the default-dfk resolution happens at the time of making a call to an `@App` annotated function, not at the point of declaring the `@App`-annotated function; in which case you can declare functions, and initialise later, as long as it's happened before executing. (I haven't actually tried that).

If that's the case, libraries of `@App`s can be imported at the top of a file, like other imports - which I think is a nicer style.

docs/userguide/examples/library.py
annawoodard: Agreed, but let's wait on that for a separate PR

docs/userguide/importing.rst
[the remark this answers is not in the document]
annawoodard: Good catch @benclifford-- I made that change (having the resolution executed at calltime instead of declaration time) and forgot to change it in the docs. I'll update this.

docs/userguide/examples/run_increment.py
```diff
@@ -0,0 +1,10 @@
+from parsl.dataflow.dflow import DataFlowKernelLoader
+
+from config import local_threads
+DataFlowKernelLoader.load(local_threads)
```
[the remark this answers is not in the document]
annawoodard: > mandatory init operation.

It's not mandatory (you can still pass the DFK to the app decorator instead), but I agree it's long. I like your suggestion and will implement it. One thing to consider is that you *can* pass kwargs which are passed on to the DFK, so you'd need to look at the DFK docs to understand what is possible there. I am hoping when we re-work the config we can eliminate most of that, so that the main interface the user needs to deal with is the `Config` class. So for now I'm fine with getting rid of `DataFlowKernel` in the name here but still saying "these kwargs will be passed to the DFK", with the understanding that when the configs are tweaked we should be able to remove the reference to the DFK and just say "pass a `Config`", which the user can read about in the `Config` docs)

parsl/dataflow/dflow.py
annawoodard: I don't want to throw an exception-- the point of this method is that it should only load if nobody else has loaded first. If you want to load no matter what, you should use `load` instead of `set_default`. Would more verbosity here be enough? ("Configuration has already been loaded; default will have no effect" and "Using default configuration").

docs/userguide/examples/run_increment.py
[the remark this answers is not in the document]
annawoodard: On second thought, to keep things simple, I'll just ditch passing the kwargs, and assume the user will manipulate the config dict beforehand if they need to.

parsl/dataflow/dflow.py
benclifford: The specific failure case (that I have encountered on other projects) I am trying to prevent is:

A user, while hacking round on their application code, end up sticking in a bunch of inits in all different files.

Then six months later, when they've forget that they've done that, they modify one to include some different config options.

Then they spend a long time trying to understand why their config options aren't being applied.

I don't want people to load no matter what - I want to stop people putting multiple init calls in in the first place - or more specifically, specifying multiple configurations, and having one of them used based on the possibly fairly arbitrary choice of which of their modules was imported first.

At the very least output a warning, though I think it is insufficient.

annawoodard: Understood. The reason I put this in in the first place was in order for `pytest` to have control over the DFK while still being able to run tests interactively. Now that the resolution is executed at calltime instead of declaration time, that isn't necessary, so I'll just remove this method, which I think should address your concern.

annawoodard:
@benclifford sorry for the excessive rebasing, as we were discussing on Slack, I'm trying to keep "doing X", "undoing X, doing Y"-type commits out of the history where possible. Next time I'll put the unrelated commits in a separate PR, or at least before the relevant commits so that only the actually-changed commits are updated. I've implemented your requests, you can see the relevant diff here: 1ec7d04...3f0fbac. A few notes:

- To avoid using `global`s I wanted to retain the singleton class, but I've concealed it from the user by importing `DataFlowKernelLoader.load` in `parsl.__init__.py` such that the user can just call `parsl.load`. The motivation for doing that over writing a wrapper function was so that the method documentation only needs to appear once (note it is accessible via `>>>help(parsl.load)`).
- Because `init` might be confused with `__init__`, I went with `parsl.load` instead of `parsl.init`. Let me know if you have a strong preference.

Thanks for your feedback! Here's a preview of the docs:

screen shot 2018-05-04 at 11 02 08 am

benclifford:
looks good

PR #258, opened 2018-05-03
Do not raise a `KeyError` if `globals` is missing

### LINKED ISSUE #257: `globals` should be an optional config field
Running over a config without `globals`, I see:
```
    def make_rundir(config=None, path=None):
        """When a path has not been specified, make the run directory.

        Creates a rundir with the following hierarchy:
            ./runinfo <- Home of all run directories
              |----000
              |----001 <- Directories for each run
              | ....
              |----NNN

        Kwargs:
            - path (str): String path to a specific run dir
                   Default : None.
        """
        try:
            if not path:
                path = None
>           elif config["globals"].get('runDir', None):
E           KeyError: 'globals'

../dataflow/rundirs.py:25: KeyError
```

annawoodard:
Fixes #257.

### DIFF
--- a/parsl/dataflow/rundirs.py
+++ b/parsl/dataflow/rundirs.py
@@ -22,7 +22,7 @@ def make_rundir(config=None, path=None):
     try:
         if not path:
             path = None
-        elif config["globals"].get('runDir', None):
+        elif config.get("globals", {}).get('runDir'):
             path = config["globals"]['runDir']
 
         if not path:


parsl/dataflow/rundirs.py
```diff
@@ -22,7 +22,7 @@ def make_rundir(config=None, path=None):
     try:
         if not path:
             path = None
-        elif config["globals"].get('runDir', None):
+        elif config.get("globals", {}).get('runDir'):
```
yadudoc: :+1:  Makes sense, at the point where rundirs are created, we haven't set global defaults yet.

Issue #94, opened 2018-02-10
Rerun from last checkpoint


kylechard:
It would be nice to have a simple way of telling parsl to run from the last checkpoint. At the moment the developer has to work out what the last checkpoint file was.

yadudoc:
I've made this explicit since at the time the feature was developed it wasn't quite clear what modes users would use besides run from the last available checkpoint. Then there's the question of multiple workflows in the same folder and disambiguation of which checkpoint to use.

yadudoc:
Pushing to 0.6.0 pending limited external interest.

annawoodard:
Note from today's meeting with @djf604: this feature would be useful in his workflow.

annawoodard:
What do we mean by 'last checkpoint'? Currently, this would mean checking for the last checkpoint in the rundir specified by the current script. There are various reasons a user might want to change rundirs-- those reasons may not always be coupled to where they want the cache to be. For me, since one of the exciting features of caching is that it would allow us to save work _between_ runs, it does not make sense to couple the caching to the rundir. I would prefer the cache to be saved in a separate directory. We could enforce a(n adjustable) maximum cache size to control how far back is cached.

benclifford:
There's (slightly buggy) boilerplate code in the documentation to find the last checkpoint from the runinfo directory.

Most immediately, I think that should turn into a helper function inside parsl, so you can say something like:

`dfk = parsl.DataFlowKernel(config=config, checkpointFiles = parsl.last_checkpoint())`

and expect it to work (including if there are no checkpoints yet...)

annawoodard:
> The default should be for the current script as doing it for many scripts could be confusing.

I'm not totally convinced by this, but since it is still possible to manually share checkpoints between scripts, I'll go with the consensus view for now.

PR #259, opened 2018-05-04
(for discussion) Move boilerplate for last checkpoint into code from docs into helper

benclifford:
(not for pushing to master)

The docs previously contained example code that can be pasted, but
as it is non-trivial, it is better placed as a helper function.

This commit also enhances that code to deal with the case where
no checkpoints exist yet - for example, on a first run.

However, this PR does not find all the needed checkpoints, I think, because they are made incrementally, and this will only find the most recent increment (?)

### DIFF
--- a/parsl/utils.py
+++ b/parsl/utils.py
@@ -24,3 +24,37 @@ def get_version():
         logger.exception("Unable to determine code state")
 
     return version
+
+def get_last_checkpoint():
+  """Finds the checkpoint from the last run, if one exists.
+
+  Note that checkpoints are incremental, and this helper will not find
+  previous checkpoints from earlier than the most recent run. It probably
+  should be made to do so.
+
+  Returns:
+     - a list suitable for checkpointFiles parameter of DataFlowKernel
+       constructor, with 0 or 1 elements
+   
+  """
+  print("BENC: get last checkpoint")
+
+  if(not(os.path.isdir('runinfo/'))):
+    print("BENC: rundir does not exists")
+    return []
+
+  dirs = sorted(os.listdir('runinfo/'))
+
+  if(len(dirs) == 0):
+    print("BENC: no run directories in runinfo")
+    return []
+
+  last_runid = dirs[-1]
+  last_checkpoint = os.path.abspath('runinfo/{0}/checkpoint'.format(last_runid))
+
+  if(not(os.path.isdir(last_checkpoint))):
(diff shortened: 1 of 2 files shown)


parsl/utils.py
```diff
@@ -24,3 +24,37 @@ def get_version():
...
+    return []
+
+  dirs = sorted(os.listdir('runinfo/'))
+
+  if(len(dirs) == 0):
+    print("BENC: no run directories in runinfo")
+    return []
+
+  last_runid = dirs[-1]
+  last_checkpoint = os.path.abspath('runinfo/{0}/checkpoint'.format(last_runid))
```
annawoodard: I am worried about hardcoding `runinfo` here, when the run directory is configurable. Maybe read this off of `DataFlowKernel.rundir` and warn that it will only find the last checkpoint, if it used the same rundir?

parsl/utils.py
```diff
@@ -24,3 +24,37 @@ def get_version():
         logger.exception("Unable to determine code state")
 
     return version
+
+def get_last_checkpoint():
+  """Finds the checkpoint from the last run, if one exists.
+
+  Note that checkpoints are incremental, and this helper will not find
+  previous checkpoints from earlier than the most recent run. It probably
```
annawoodard: I think we need to clarify that this will only find checkpoints that happen to be in `rundir`, unless we switch to reading off `DataFlowKernel.rundir`, in which case we should clarify that this will only find checkpoints that use the same rundir as the current script.

Also we may want to allow one to specify `n` last checkpoints, rather than just the most recent?

annawoodard:
Linking the related issue for reference: #94.

parsl/utils.py
```diff
@@ -24,3 +24,63 @@ def get_version():
...
+    if(not(os.path.isdir('runinfo/'))):
+        return []
+
+    dirs = sorted(os.listdir('runinfo/'))
+
+    if(len(dirs) == 0):
+        return []
+
+    last_runid = dirs[-1]
+    last_checkpoint = os.path.abspath('runinfo/{0}/checkpoint'.format(last_runid))
```
annawoodard: The updates look good-- I'm still concerned that we are hardcoding the checkpoint directory here though.

benclifford:
https://github.com/Parsl/parsl/issues/273 contains some related discussion

PR #275, opened 2018-05-14
Log the commandline that will be invoked in a bash app

benclifford:


parsl/app/bash_app.py
```diff
@@ -44,6 +44,8 @@ def remote_side_bash_executor(func, *args, **kwargs):
         logging.error("[{}] Caught exception during cmd_line resolution: {}".format(func_name, e))
         raise e
 
+    logging.debug("Executable: %s", executable)
```
yadudoc: This will log the command line on the remote side, which may not be super useful as we don't automatically bring these logs back (which is something we need to fix).

benclifford: yeah, that's a separate thing I've been wondering about, and I will open a separate issue on.

however, it's useful when the remote side is actually local.

yadudoc: Ok, then.

Issue #298, opened 2018-05-23
Tests rely on `at exit` behaviour which does not happen until far in the future.


benclifford:
Running `pytest --basic`, I see a lot of heavy stuff created by tests but then left around to be cleaned up by "at exit" cleanup, which doesn't happen until the whole test suite has run.

For example, with --log-cli=DEBUG,

    swift_t.py                 105 DEBUG    [RUNNER] Queue is empty

is logged every few seconds even after all the tests have completed and pytest shutdown is happening - presumably from an earlier swift_t test.

I also see a lot of this:

    2018-05-23 20:34:44 parsl.dataflow.strategy:155 [DEBUG]  Site:local_threads Status:STATIC

... which might be from a lot of DFKs sitting round doing nothing.

Also, a bunch of `ipcontrollers` are left and (supposedly) cleaned up by at exit behaviour (see another issue/PR I have yet to write).

It would probably be better if all of these things were cleaned up more explicitly: with explicit DFK cleanup in each test, and with that DFK cleanup really cleaning up everything it should.

annawoodard:
Sorry for the delay, I have a small PR for running `cleanup` after every test. It does not completely solve this but reduces the number of processes. I will submit it shortly.

annawoodard:
This was fixed in #301.

PR #301, opened 2018-05-23
Call cleanup after each test

annawoodard:
This improves the situation for #298. I see far fewer peak ipengine
processes, but still more than I would expect if cleanup was working as
expected after each test.


benclifford:
Seems the right thing to do.

Being picky, maybe closedown the dfk the fixture actually created rather than the one the global dfk state holds - although the hope probably is that they're the same, they potentially might not be if some test has done a parsl.clear/load.

Issue #133, opened 2018-03-16
Switch to class-based configuration


annawoodard:
This was discussed at the meeting yesterday.

annawoodard:
For example, instead of
```python
singleNode = {
  "sites": [
    {"site": "Local_IPP",
      "auth": {
        "channel": "ssh",
        "hostname": "swift.rcc.uchicago.edu",
        "username": os.environ['MIDWAY_USERNAME'],
        "scriptDir": "/scratch/{0}/parsl".format(USERNAME)
      },
      "execution": {
        "executor": "ipp",
        "provider": "slurm",
        "block": {  # Definition of a block
           "nodes": 1,            # of nodes in that block
           "taskBlocks": 1,       # total tasks in a block
           "initBlocks": 1,
           "maxBlocks": 1,
            "options": {
               "partition": "westmere",
               "overrides": '''module load python/3.5.2+gcc-4.8'''
            }
        }
      }
    }
 ],
 "globals": {"lazyErrors": True}
}
```
we could have
```python
from parsl.config import Config
from libsubmit.providers.slurm.slurm import Slurm

user = os.environ['MIDWAY_USERNAME']

config = Config(
    [
        Site(
            'local_IPP',
            channel=SSHChannel(
                hostname='swift.rcc.uchicago.edu',
                username=user,
                script_dir='/scratch/{0}/parsl'.format(user)
            ),
            provider=Slurm(
                executor='ipp',
                nodes=1,
                task_blocks=1,
                init_blocks=1,
                max_blocks=1,
                partition='westmere',
                overrides=''''module load python/3.5.2+gcc-4.8'''
            )
        )
    ],
    globals={'lazy_errors': True}
)
```
One advantage would be that if someone handed off the above config to you, finding the relevant documentation interactively would be more natural:
```
>>> help(Slurm)

Help on class Slurm in module libsubmit.providers.slurm.slurm:

class Slurm(libsubmit.providers.provider_base.ExecutionProvider)

| Slurm Execution Provider
|
| This provider uses sbatch to submit, squeue for
| status and scancel to cancel jobs. The sbatch
| script to be used is created from a template file
| in this  same module.
```

Additionally, the documentation could be written more clearly, and defaults could be specified in the constructor signature (and so would be self-documenting) (left is current, right is proposed option)
[...]

PR #305, opened 2018-05-31
Add implicint staging in; schemes: ftp, http, https, globus

lukaszlacinski:


parsl/dataflow/dflow.py
```diff
@@ -437,6 +429,14 @@ def sanitize_and_wrap(task_id, args, kwargs):
     def submit(self, func, *args, parsl_sites='all', fn_hash=None, cache=False, **kwargs):
         """Add task to the dataflow system.
 
+        If the app task has the sites attributes not set (default=='all')
+        the task will be launched on a randomly selected executor from the
+        list of executors. This behavior could later be updates to support
```
annawoodard: updates -> updated

parsl/tests/test_globus/test_implicit_staging_ftp.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+import parsl
+from parsl.app.app import App
+from parsl.data_provider.files import File
+from parsl.tests.configs.local_threads_globus import config
+
+parsl.local(config)
```
annawoodard: `local` -> `load` (and subsequent instances)

parsl/tests/test_globus/test_implicit_staging_ftp.py
```diff
@@ -0,0 +1,49 @@
...
+    Create a remote input file (ftp) that points to file_test_cpt.txt.
+    """
+
+    unsorted_file = File('ftp://ftp.uconn.edu/48_hour/file_test_cpt.txt')
+
+    # Create a local file for output data
+    sorted_file = File('sorted.txt')
+
+    f = sort_strings(inputs=[unsorted_file], outputs=[sorted_file])
+    f.result()
```
annawoodard: To be on the safe side, can we add an assertion here that we're getting what we expect? (applies to subsequent instances)

parsl/data_provider/data_manager.py
```diff
@@ -110,39 +127,102 @@ def _get_globus_site(self, site_name=None):
...
-
-        if file.scheme == 'file':
-            f = self.submit(self._file_transfer_in, file)
-        elif file.scheme == 'globus':
-            f = self.submit(self._globus_transfer_in, file, globus_ep)
+    def _file_stage_in_app(self):
+        return App("python", self.dfk, sites=['data_manager'])(self._file_stage_in)
 
-        from parsl.app.futures import DataFuture
+    def _file_stage_in(self, outputs=[]):
```
annawoodard: I don’t follow why `_file_stage_in/_file_stage_out` are no ops. Say, for example, you’re running the same app on two sites, Site A and Site B. You want results to be sent to `/home/foobar` on Site A. If you’re running on Site B, you use globus transfer, if you’re running on site A, you just want to use `mv`.  Is this possible? I think a test that has two sites, one with (for example) globus transfer, another with local transfer, and a task that runs on both sites would clarify this for me.

parsl/tests/test_globus/test_implicit_staging_https.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+import parsl
+from parsl.app.app import App
+from parsl.data_provider.files import File
+from parsl.tests.configs.local_threads_globus import config
```
annawoodard: It looks like this file hasn't been added.

parsl/data_provider/data_manager.py
```diff
@@ -110,39 +127,102 @@ def _get_globus_site(self, site_name=None):
...
+            - file (File) : file to stage in
+            - site_name (str|List) : a site the file is going to be staged in to.
                                 If the site argument is not specified for a file
                                 with 'globus' scheme, the file will be staged in to
                                 the first site with the "globus" key in a config.
         """
 
         if file.scheme == 'file':
-            site_name = None
+            stage_in_app = self._file_stage_in_app()
```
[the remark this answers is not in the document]
annawoodard: I do favor refactoring, but I don't think it is urgent, I'd be happy to just open an issue and look at it later.

It is not clear to me how a staging_app_factory actually simplifies this. I would favor (as usual) a class-based system, where the possible schemes each have a dedicated `Scheme` class (not sure about the name). That way in the `Config` we can list the possible classes. Each subclass would have all of the possible options documented in the constructor (see the example I already posted [here](https://github.com/Parsl/parsl/issues/133), though the naming needs work). Currently the only documentation we have is [here](http://parsl.readthedocs.io/en/latest/userguide/data.html#configuration), which needs some work, and we don't have any examples which actually use the configuration.

parsl/data_provider/data_manager.py
annawoodard: I don't think I am doing a good job of communicating my concern. Let's take a few minutes to talk about this at the office tomorrow morning and then I'll follow up with a summary here.

parsl/data_provider/data_manager.py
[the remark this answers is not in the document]
annawoodard: (To clarify, each subclass would implement its own `get_stage_in_app` or whatever)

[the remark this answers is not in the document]
yadudoc: If you are saying that we should just have subclasses of files/schemes, I agree that it would be cleaner solution.

parsl/tests/test_globus/test_implicit_staging_https.py
lukaszlacinski: https://github.com/Parsl/parsl/commit/49d9594b2d6ac782a9edf4776c4521db9133a622 has been added by you.

annawoodard: (facepalm) Thanks @lukaszlacinski sorry I missed that! 😄

parsl/data_provider/data_manager.py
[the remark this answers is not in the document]
annawoodard: Awesome! Refactoring might take more time than the other more trivial feedback we have on this PR, but it's not really critical for functionality, so I think we can safely defer it for later, unless @lukaszlacinski is chomping at the bit to do it now.

parsl/tests/test_globus/test_implicit_staging_ftp.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+import parsl
+from parsl.app.app import App
+from parsl.data_provider.files import File
+from parsl.tests.configs.local_threads_globus import config
```
annawoodard: Do we really need the Globus config for non-globus tests? They will be skipped because globus isn't configured, but I think they should be able to run automatically.

parsl/dataflow/config_defaults.py
```diff
@@ -71,6 +71,21 @@ def update_config(config, rundir):
                    }
     }
 
+    # Add DataManager thread pool
+    data_manager_site = {
```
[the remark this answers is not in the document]
annawoodard: Agreed, but we might be missing something. @lukaszlacinski can you clarify for us?

annawoodard:
@lukaszlacinski it looks like there are two issues with `test_immutable_config`: 1), we actually _are_ changing the config, so we need to account for that in the test, and 2) we need to check if there is already a `data_manager` site, and not add one in that case. Looking at the test, I see many duplicate `data_manager` sites:
```
{'globals': {'lazyErrors': True, 'retries': 2},
 'sites': [{'auth': {'channel': None},
   'execution': {'block': {'initBlocks': 0,
     'maxBlocks': 10,
     'minBlocks': 0,
     'parallelism': 0,
     'taskBlocks': 4},
    'executor': 'threads',
    'maxThreads': 6,
    'provider': None},
   'site': 'local_threads'},
  {'auth': {'channel': None},
   'execution': {'executor': 'threads', 'maxThreads': 8, 'provider': None},
   'site': 'data_manager'},
  {'auth': {'channel': None},
   'execution': {'executor': 'threads', 'maxThreads': 8, 'provider': None},
   'site': 'data_manager'},
... etc ....
}
```

parsl/dataflow/dflow.py
```diff
@@ -337,6 +313,22 @@ def launch_task(self, task_id, executable, *args, **kwargs):
...
+        Args:
+            - site (str) : site where the app is going to be launched
+            - args (List) : Positional args to app function
+            - kwargs (Dict) : Kwargs to app function
+        """
+
+        inputs = kwargs.get('inputs', [])
+        for idx, f in enumerate(inputs):
+            if isinstance(f, File) and f.is_remote():
+                inputs[idx] = f.stage_in([site])
```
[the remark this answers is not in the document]
lukaszlacinski: How can Parsl know if there's a shared-fs available or not?

parsl/data_provider/data_manager.py
annawoodard: Talking with @yadudoc cleared this up for me. The answer to my question ('What happens if you are running on two sites?') is that multisite + data staging is not supported. @lukaszlacinski We need to add a warning that complains if the user tries to run a config with multiple sites + data management. I will open a separate issue to do that later and assign myself, because it should be done in the constructor of the new `Config` class.

lukaszlacinski: @annawoodard Multisite is supported. A remote input file is staged in on a site where an app that needs the file is going to be launched.

yadudoc: We are better off advertising only single remote site support. I don't think implicit staging has been fleshed out to support data movement across multiple sites. For eg. AppB on SiteB consumes files from AppA on SiteA. It is unclear in this model if we'll stage all output data to local and then stage out.

parsl/dataflow/dflow.py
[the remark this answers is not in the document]
yadudoc: User has to indicate that to us. Or (in the future) the user is running on a site that has been cataloged.

lukaszlacinski:
When I run pytest it returns success if 'test_checkpointing/' and 'test_regression/' are removed. If 'test_checkpointing/' or 'test_regression/' are run before 'test_staging/', 'test_staging/test_implicit_staging_ftp.py' fails with the error:

`    def submit(self, fn, *args, **kwargs):
        with self._shutdown_lock:
            if self._shutdown:
>               raise RuntimeError('cannot schedule new futures after shutdown')
E               RuntimeError: cannot schedule new futures after shutdown

/usr/lib/python3.5/concurrent/futures/thread.py:107: RuntimeError
`

parsl/dataflow/dflow.py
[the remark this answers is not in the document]
lukaszlacinski: We can add a flag to the 'data' section for each site, e.g. 'shared-filesystem', that is True by default. For sites without shared file system, a user sets the flag to False to turn off _add_input_deps() function and turn on staging in/out in {BashApp|PythonApp}.__call__() or so.

PR #322, opened 2018-06-06
Adding support for logging to ElasticSearch

yadudoc:
We are logging to a test elasticsearch machine. Kibana is available here:
https://search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-east-1.es.amazonaws.com/_plugin/kibana/app/kibana#/discover?_g=()&_a=(columns:!(_source),index:c95a5540-644d-11e8-90be-3b2ab570ca29,interval:auto,query:(language:lucene,query:''),sort:!(timestamp,desc))


parsl/dataflow/dflow.py
```diff
@@ -50,7 +51,8 @@ class DataFlowKernel(object):
     """
 
     def __init__(self, config=None, executors=None, lazyErrors=True, appCache=True,
-                 rundir=None, retries=0, checkpointFiles=None, checkpointMode=None):
+                 rundir=None, retries=0, checkpointFiles=None, checkpointMode=None,
+                 db_logger=get_db_logger(level=logging.CRITICAL)):
```
annawoodard: Consider allowing the user to disable DB logging by setting this to `None`. Currently it will only be disabled if they don't have `cmreslogging` installed. I think the default should be to have DB logging disabled (the average user is not going to have an AWS instance set up).

parsl/db_logger.py
```diff
@@ -0,0 +1,74 @@
...
+class OptionalModuleMissing(Exception):
+    ''' Error raised a required module is missing for a optional/extra provider
+    '''
+
+    def __init__(self, module_names, reason):
+        self.module_names = module_names
+        self.reason = reason
+
+    def __repr__(self):
+        return "Unable to Initialize logger.Missing:{0},  Reason:{1}".format(
```
annawoodard: Initialize -> initialize

parsl/db_logger.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Returns
+    -------
+    logging.logger object
+
+    Raises
+    ------
+
+    """
+    logger = logging.getLogger("ParslElasticsearch")
```
annawoodard: I favor logger = logging.getLogger(__file__), to be consistent with other modules

parsl/db_logger.py
```diff
@@ -0,0 +1,74 @@
...
+                  index_name="parsl.campaign"):
+    """
+    Parameters
+    ----------
+    host : str, optional
+        URL to the elasticsearch cluster. Skip the http(s)://
+    port : int, optional
+        Port to use to access the elasticsearch cluster
+    level : logging.LEVEL, optional
+        This module will log *only* if logging.INFO or lower is provided.
```
annawoodard: I don't think a configurable logging level is needed here. In traditional logging, it makes sense to control the verbosity. In the Elasticsearch case, we want everything to go to the DB. I think decisions about how to display info to the user should be done at the Kibana level.

parsl/db_logger.py
```diff
@@ -0,0 +1,74 @@
...
+    index_name : str, optional
+        Index name to use for elasticsearch
+
+    Returns
+    -------
+    logging.logger object
+
+    Raises
```
annawoodard: You can omit this if it's empty

parsl/db_logger.py
```diff
@@ -0,0 +1,74 @@
...
+        if not _es_logging_enabled:
+            raise OptionalModuleMissing(
+                ['cmreslogging'], "Logging to ElasticSearch requires the cmreslogging module")
+
+        handler = CMRESHandler(hosts=[{'host': host,
+                                       'port': port}],
+                               use_ssl=True,
+                               auth_type=CMRESHandler.AuthType.NO_AUTH,
+                               es_index_name="my_python_index",
+                               es_additional_fields={'Campaign': "test", 'Username': "yadu"})
```
annawoodard: can we make the username configurable?

annawoodard:
Another thing I forgot to mention is that we need a way to differentiate between parsl runs. As it is, everything is going to get dumped together. But I don't think that should hold up this PR, let's just include it in the issue we're making for Connor.

parsl/dataflow/dflow.py
yadudoc: This is hacky at best. I've set the db_logger to only require the `cmreslogging` module and send out log messages only if logging level is INFO or lower. I think this could use some redesign.

parsl/db_logger.py
yadudoc: We should be comparing this to the get_*_logger methods in parsl/__init__.py where we name the logger "parsl". This doesn't really add much anyway.

parsl/db_logger.py
yadudoc: Sorry this isn't very clear. This logger only logs to Elasticsearch when the level is set to INFO or lower. This is not a traditional logger, and the logging level results in binary behavior.
```
if LEVEL > logging.INFO:
      return NullLogger
else
      return cmreslogger
```

annawoodard: Right, since the purpose of the level is switching on whether or not we enable the cmsrellogger, why not just have the adjustable thing be a bool `enable_es_logging` instead of a configurable logging level? Then the purpose of the knob is stated explicitly in the name.

annawoodard: What is the point of logging levels in the context of `cmreslogger`? I think we want everything to be sent-- we should always use the same logging level.

parsl/db_logger.py
yadudoc: Sorry, that was unintentional. The logger already captures and reports that.

yadudoc:
These changes are all included and improved upon in #371. Closing.

PR #334, opened 2018-06-11
Change test FTP server address

lukaszlacinski:
Changes ftp server used by the ftp staging test from ftp://ftp.cs.brown.edu which it turned out to be not too reliable to ftp://ftp.iana.org/.


annawoodard:
👏  Thanks @lukaszlacinski

PR #335, opened 2018-06-12
Class based configs #133

### LINKED ISSUE #133: Switch to class-based configuration

### LINKED ISSUE #253: Make the config shown in the debug log reflect the actual config
For example:

```
config = ....
dfk = DataFlowKernel(config=config, retries=2)
```

The config printed in the debug log will not reflect the fact that the `retries` keyword arg had an effect, which is what the user cares about.

Fixing this should be straightforward after #133 is implemented.

### LINKED ISSUE #296: Make `Controller` parameter 'interfaces' configurable
Currently the `Controller` parameter `interfaces` is silently ignored and set to `ip=*`.

### LINKED ISSUE #297: Remove `reuse_controller` argument to the IPP constructor
The `reuse_controller` argument to the IPP constructor currently has no effect. This info is specified in the `Controller` constructor and can be removed from the IPP constructor. This will be fixed as part of the changes needed for #133.

### LINKED ISSUE #303: Naming maxThreads/max_threads/max_workers
The class definition for the ThreadPoolExecutor uses `max_workers` as a kwarg. The config options for threads specify `maxThreads`. We need to update all of these to use `max_threads` to be consistent going forward.

### LINKED ISSUE #309: Make tests work better as examples
reported by @yadudoc 

The site tests require user-specific configuration ('username', 'overrides', etc). If a test isn't configured, we want pytest to skip it. The way we currently deal with this is to collect all of the user-specific options in `user_opts.py`, and if the dictionary there does not contain an entry for the site, we skip testing that site. This works well for testing (as you configure sites, more site-specific tests will be run each time you run pytest). One drawback is that it divides the relevant config information between two files, which makes it harder to pass on a configuration as an example to a user. It might be better to remove `user_opts.py` and explicitly include a `pytest.skip` in each site configuration along with dummy `FIXME` example fields. When one completes a configuration they can remove the skip option. In that way, you could point a user directly to the example configuration and say "change all of the `FIXME` fields to the appropriate value."

annawoodard:
This PR switches us from dictionary configurations to using a `Config` class.

Notes:

- I did not add exhaustive validation-- I think that should be saved for a separate issue. But, critically, now all validation can go in the respective constructor such that misconfigured configs will be complained about as early as possible.
- This PR introduces further mixing of docstring styles (moving more of our docstrings to NumPy style, which is parsed properly by sphinx.) It was more convenient to adjust things as I saw them. We can continue adjusting the remaining docstrings to match moving forward.
- I have gotten rid of 'site' everywhere in favor of 'executor', as discussed previously.
- Because executors are now constructed the config, I've moved setting attributes and validation to the constructor, and added a method 'start' (better name ideas welcome-- spinup?) to the executors which contains the dynamic stuff (starting IPP workers, starting the thread pool, etc). This is so that the DFK can control when this happens.
- This PR removes the magic 'options' dict and replaces with documented keyword args.
- This PR introduces a new class `RepresentationMixin` which can be added to existing classes to add a `__repr__` method. The current behavior is to print the object instantiation with all arguments and keyword arguments (including defaults). I hope this will reduce confusion about what options are actually being applied. It also has the benefit that it can be copy-and-pasted to tweak options. If we think it is too verbose, I could switch to only displaying any keyword arguments which were different from the default.

This PR depends on changes to libsubmit. Parsl/libsubmit#57 should be reviewed first and a point release cut. Then I will update `requirements.txt` in this PR with the new point release, which should fix CI.

Fixes #133, fixes #253, fixes #296, fixes #297, fixes #303, fixes #309, fixes Parsl/libsubmit#9, fixes Parsl/libsubmit#48. Fixes (partially) #297-- we still need to update the doc images to complete it.


parsl/dataflow/dflow.py
```diff
@@ -49,100 +44,69 @@ class DataFlowKernel(object):
...
+                    config.checkpoint_period = "00:30:00"
+                h, m, s = map(int, config.checkpoint_period.split(':'))
+                checkpoint_period = (h * 3600) + (m * 60) + s
+                self._checkpoint_timer = Timer(self.checkpoint, interval=checkpoint_period)
+            except Exception as e:
+                logger.error("invalid checkpoint_period provided:{0} expected HH:MM:SS".format(period))
+                self._checkpoint_timer = Timer(self.checkpoint, interval=(30 * 60))
 
+        if any([x.managed for x in config.executors]):
+            self.flowcontrol = FlowControl(self, self._config)
```
yadudoc: The self._config arg can be taken out.

parsl/dataflow/dflow.py
```diff
@@ -598,20 +558,15 @@ def cleanup(self):
...
 
         for executor in self.executors.values():
-            if executor.scaling_enabled:
-                job_ids = executor.execution_provider.resources.keys()
+            if executor.scaling_enabled and executor.managed:
+                job_ids = executor.provider.resources.keys()
                 executor.scale_in(len(job_ids))
 
-            # We are not doing shutdown here because even with block=False this blocks.
             executor.shutdown()
```
yadudoc: We shouldn't call executor shutdown on non managed executors. Just fix indent here and if block

parsl/executors/ipp.py
```diff
@@ -22,15 +60,81 @@ class IPyParallelExecutor(ParslExecutor):
...
+        self.managed = managed
+
+    def start(self):
+        self.controller.profile = self.label
+        self.controller.ipython_dir = self.run_dir
+        self.controller.start()
+
+        self.engine_file = self.controller.engine_file
+
+        with wait_for_file(self.controller.client_file, seconds=20):
```
yadudoc: the wait should be 2 mins or so to account for the worst behaving system.

parsl/tests/configs/beagle_multinode.py
```diff
@@ -10,42 +10,29 @@
 ========================================================================
```
yadudoc: This config is broken. Remove.

parsl/tests/configs/beagle_multinode_MPI.py
```diff
@@ -10,42 +10,34 @@
...
+from parsl.config import Config
+from parsl.executors.ipp import IPyParallelExecutor
 from parsl.tests.user_opts import user_opts
+from parsl.tests.utils import get_rundir
 
-if 'beagle' in user_opts:
-    info = user_opts['beagle']
-else:
-    pytest.skip('beagle user_opts not configured', allow_module_level=True)
+config = Config(
```
yadudoc: This config is broken. Let's remove this.

parsl/tests/configs/beagle_single_node.py
```diff
@@ -7,41 +7,33 @@
...
+                channel=SSHChannel(
+                    hostname='login4.beagle.ci.uchicago.edu',
+                    username=user_opts['beagle']['username'],
+                    script_dir="/lustre/beagle2/{}/parsl_scripts".format(user_opts['beagle']['username'])
+                ),
+                nodes_per_block=1,
+                tasks_per_node=1,
+                init_blocks=1,
+                max_blocks=1,
+                launcher='srun',
```
yadudoc: Launcher should be `aprun`.

parsl/tests/configs/cc_in2p3_local_multinode.py
```diff
@@ -10,38 +10,30 @@
...
+            label='cc_in2p3_local_multinode',
+            provider=GridEngine(
+                channel=LocalChannel(
+                    script_dir=user_opts['cc_in2p3']['script_dir']
+                ),
+                nodes_per_block=4,
+                tasks_per_node=2,
+                init_blocks=1,
+                max_blocks=1,
+                launcher='srun',
```
yadudoc: Doesn't need a launcher.

parsl/tests/configs/cc_in2p3_local_single_node.py
```diff
@@ -7,39 +7,30 @@
...
+            label='cc_in2p3_local_single_node',
+            provider=GridEngine(
+                channel=LocalChannel(
+                    script_dir=user_opts['cc_in2p3']['script_dir']
+                ),
+                nodes_per_block=1,
+                tasks_per_node=1,
+                init_blocks=1,
+                max_blocks=1,
+                launcher='srun',
```
yadudoc: Shouldn't specify launcher.

parsl/tests/configs/cooley_local_single_node.py
```diff
@@ -1,41 +1,26 @@
...
+config = Config(
+    executors=[
+        IPyParallelExecutor(
+            label='cooley_local_single_node',
+            provider=Cobalt(
+                nodes_per_block=1,
+                tasks_per_node=1,
+                init_blocks=1,
+                max_blocks=1,
+                walltime="00:05:00",
```
yadudoc: We need account info for Cobalt/cooley

parsl/tests/configs/cooley_ssh-il_single_node.py
```diff
@@ -1,40 +1,30 @@
...
+                channel=SSHInteractiveLoginChannel(
+                    hostname='cooleylogin1.alcf.anl.gov',
+                    username=user_opts['cooley']['username'],
+                    script_dir="/home/{}/parsl_scripts/".format(user_opts['cooley']['username'])
+                ),
+                nodes_per_block=1,
+                tasks_per_node=1,
+                init_blocks=1,
+                max_blocks=1,
+                walltime="00:05:00",
```
yadudoc: Please check to make sure we are not missing `account` or `partition` options for cobalt

parsl/tests/configs/cori_ipp_multinode.py
```diff
@@ -9,40 +9,26 @@
...
-if 'cori' in user_opts:
-    info = user_opts['cori']
-else:
-    pytest.skip('cori user_opts not configured', allow_module_level=True)
+config = Config(
+    executors=[
+        IPyParallelExecutor(
+            label='cori_ipp_multinode',
+            provider=Slurm(
+                'regular',
```
yadudoc: we need `launcher : 'srun'` for multinode launches on cori

parsl/tests/configs/ec2_single_node.py
```diff
@@ -11,42 +11,31 @@
...
     ],
-    "globals": {
-        "lazyErrors": True,
-        "runDir": get_rundir()
-    },
-    "controller": {
-        "publicIp": '*'
-    }
-}
+    controller=Controller(public_ip='*'),
```
yadudoc: this is most likely wrong. Defaults should be used here rather.

Issue #344, opened 2018-06-20
Implement local file staging


annawoodard:
Currently, in order to have parsl transfer a local file for you, you must either:

1) set up an FTP server locally and use File(’ftp:...),
2) set up an HTTPS server locally and use File(’https:...),
3) set up a GridFTP server locally (e.g. Globus Connect Server or Globus Connect Personal) and use File(’globus:...).

We need a simple way of transferring a file that exists locally which does not require setting up a server. Probably using `channel.push_file` and `channel.pull_file` would cover 90% of our use cases (the caveat would be it wouldn't work if workers did not have shared filesystem access). It also has the advantage of already being authenticated.

benclifford:
This is doable with rsync as of #1378.

There might still be value in a channel based provider so I'll leave this open.

PR #356, opened 2018-06-26
Add local staging #344

annawoodard:
This is branched from `class-based-configs-#133`, the only new commit here is df84cb0. This is a quick hack because I need this functionality for the CHEP workflow. At the dev meeting yesterday, we decided that @lukaszlacinski should focus on documentation for already-merged staging functionality for the 0.6.0 release. We can merge this PR, but not advertise the functionality until he's had time to test/improve/think more deeply about this (we should not consider this a fix for #344 until then), and we've added documentation.


parsl/data_provider/data_manager.py
```diff
@@ -140,8 +142,12 @@ def stage_in(self, file, executor):
     def _file_stage_in_app(self):
         return App("python", executors=['data_manager'])(self._file_stage_in)
 
-    def _file_stage_in(self, outputs=[]):
-        pass
+    def _file_stage_in(self, executor_label, outputs=[]):
+        file = outputs[0]
+        executor = self.executors[executor_label]
+        if isinstance(executor, IPyParallelExecutor):
+            executor.provider.channel.push_file(file.filepath, executor.working_dir)
```
yadudoc: @annawoodard I'm a bit confused here. This would only work for sites where the login node and compute nodes have a shared file system. Am I missing something obvious ?

PR #370, opened 2018-07-03
Fixes #204. Duplicate log line bug

### LINKED ISSUE #204: Duplicate log lines when using IPP
The following code fragment emits log lines to STDERR with the default format style even though no logging is requested. I suspect that `ipyparallel` is doing something dirty.

    from parsl import DataFlowKernel
    from parsl.configs.local import localIPP as config

    dfk = DataFlowKernel(config=config)
    dfk.cleanup()

The above code with the minor change of using threads will not emit the log lines.

    from parsl import DataFlowKernel
    from parsl.configs.local import localThreads as config

    dfk = DataFlowKernel(config=config)
    dfk.cleanup()

Please help test by running this with the latest parsl code.

yadudoc:
The problem arises when a strategy thread calls status on IPP executor.
So, the strategy thread explicitly mutes the handler attached to the rootLogger
by setting the logLevel to ERROR.

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -116,6 +116,8 @@ class Strategy(object):
         self.strategies = {None: self._strategy_noop, 'simple': self._strategy_simple}
 
         self.strategize = self.strategies[self.config.strategy]
+        self.logger_flag = False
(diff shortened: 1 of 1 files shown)


annawoodard:
YESSSSSSSSSSSSSSSSSS

Issue #169, opened 2018-03-27
Implement implicit staging


annawoodard:
Currently users have to call

```
dfu = file.stage_in()
dfu.result()
```

It would be preferable to call `result()` automatically. Related to #162.

yadudoc:
@annawoodard We should just remove explicit staging. If we call result() automatically, we block the workflow, it becomes serial.
Trying to handle externally launched futures in the DFK is pretty ugly, and I would recommend we fix this by adding implicit staging.

annawoodard:
Sounds good @yadudoc

PR #372, opened 2018-07-04
Implicit staging #169

lukaszlacinski:
Adds implicit staging out for executors with shared file systems. Updates data.rst to reflect the latest changes:
- added 'ftp', 'http', https' schemes
- hidden explicit staging in/out (although still can be used)
- added implicit staging in/out

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -196,8 +196,8 @@ class DataFlowKernel(object):
 
         # Submit _*_stage_out tasks for output data futures that correspond with remote files
         if (self.tasks[task_id]['app_fu'] and
-            self.tasks[task_id]['status'] == States.done and
-            self.tasks[task_id]['executor'] != 'data_manager'):
+                self.tasks[task_id]['status'] == States.done and
+                self.tasks[task_id]['executor'] != 'data_manager'):
             for dfu in self.tasks[task_id]['app_fu'].outputs:
                 f = dfu.file_obj
                 if isinstance(f, File) and f.is_remote():
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -195,7 +195,8 @@ class DataFlowKernel(object):
                 self.checkpoint(tasks=[task_id])
 
         # Submit _*_stage_out tasks for output data futures that correspond with remote files
-        if self.tasks[task_id]['executor'] != 'data_manager':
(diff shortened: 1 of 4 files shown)


lukaszlacinski:
I do not like the long condition at https://github.com/Parsl/parsl/blob/6363ac7e0be88a3a5676ab167247e3f36e9733ee/parsl/dataflow/dflow.py#L198, but I cannot find a good way to get rid of this.

parsl/dataflow/dflow.py
```diff
@@ -194,6 +194,13 @@ def handle_update(self, task_id, future, memo_cbk=False):
             if self.checkpoint_mode is 'task_exit':
                 self.checkpoint(tasks=[task_id])
 
+        # Submit _*_stage_out tasks for output data futures that correspond with remote files
+        if self.tasks[task_id]['executor'] != 'data_manager':
+            for dfu in self.tasks[task_id]['app_fu'].outputs:
+                f = dfu.file_obj
+                if isinstance(f, File) and f.is_remote():
+                    f.stage_out(self.tasks[task_id]['executor'])
```
yadudoc: What happens if stage_out cannot happen, say there's no staging method available ?

parsl/dataflow/dflow.py
```diff
@@ -197,7 +197,10 @@ def handle_update(self, task_id, future, memo_cbk=False):
         # Submit _*_stage_out tasks for output data futures that correspond with remote files
         if (self.tasks[task_id]['app_fu'] and
                 self.tasks[task_id]['status'] == States.done and
-                self.tasks[task_id]['executor'] != 'data_manager'):
+                self.tasks[task_id]['executor'] != 'data_manager' and
+                self.tasks[task_id]['func_name'] != '_file_stage_in' and
```
yadudoc: We should add a task type "stage_in" , "stage_out" tag rather than check func_names

yadudoc:
@lukaszlacinski I think these changes look good. Let me know how soon you can make these changes, I can also lend a hand to get this merged quickly :)

parsl/dataflow/dflow.py
lukaszlacinski: DataManager will throw an Exception,https://github.com/Parsl/parsl/blob/677c451dc418ad41e44af41efe405001e716ac7a/parsl/data_provider/data_manager.py#L209

PR #378, opened 2018-07-10
Add explicit errors for unknown url schemes

benclifford:
This is more a request for discussion on error handling:

When a URL is used with an unknown URL scheme, what should happen staging-wise?

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -136,6 +136,8 @@ class DataManager(ParslExecutor):
             stage_in_app = self._globus_stage_in_app()
             app_fut = stage_in_app(globus_ep, outputs=[file])
             return app_fut._outputs[0]
+        else:
+            raise Exception('Staging in with unknown file scheme {} is not supported'.format(file.scheme))
 
     def _file_stage_in_app(self):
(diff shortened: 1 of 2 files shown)


parsl/data_provider/data_manager.py
```diff
@@ -214,6 +216,8 @@ def stage_out(self, file, executor):
             globus_ep = self._get_globus_endpoint(executor)
             stage_out_app = self._globus_stage_out_app()
             return stage_out_app(globus_ep, inputs=[file])
+        else:
```
yadudoc: Shouldn't this have been caught at File init ?

benclifford:
I dug into this some more and added another check, in commit 56265fa. Also see issue #382 which reports that unknown URL schemes are silently ignored.

My main concern here is fragility of the error path:

 * when users use an unknown staging scheme, as is likely to happen when performing exploratory coding: how do I discover if a "foo" url scheme is supported? well, let's run some code with a "foo" URL - and it "sort of" works

 * when a developer adds in a new URL scheme. There are several places where the supported schemes are enumerated in code (in various ways) and I think it is important for future-developer-friendliness to fail properly rather than silently when they don't update one of those places

PR #403, opened 2018-07-22
Attempting to simplify parallelism section

yadudoc:


docs/userguide/execution.rst
```diff
@@ -116,30 +113,18 @@ Parsl provides a simple user-managed model for controlling elasticity.
 It allows users to prescribe the minimum
 and maximum number of blocks to be used on a given executor as well as
 a parameter (*p*) to control the level of parallelism. Parallelism
-is expressed as the ratio of resource capacity and active tasks.
+is expressed as the ratio of task execution capacity and active tasks (running tasks + dependency resolved but waiting tasks)
```
kylechard: Not sure we should consider queued tasks as active. Perhaps best to split this up?

yadudoc: This is clearer.

annawoodard: Maybe available tasks would be less confusing than active tasks? Active tasks sounds like "tasks that are running".

kylechard: My concern is the wording. I don't consider a task that is ready to run is "active." So we could instead say something like:  ... task execution capacity to active tasks (running tasks) and available tasks (those that have had their dependencies met but are waiting to be executed).

danielskatz: an alternative to "dependency resolved but waiting tasks" would be "ready to run but queued"

docs/userguide/execution.rst
```diff
@@ -113,7 +113,8 @@ Parsl provides a simple user-managed model for controlling elasticity.
 It allows users to prescribe the minimum
 and maximum number of blocks to be used on a given executor as well as
 a parameter (*p*) to control the level of parallelism. Parallelism
-is expressed as the ratio of task execution capacity and active tasks (running tasks + dependency resolved but waiting tasks)
+is expressed as the ratio of task execution capacity and active tasks (running tasks)
```
annawoodard: Sorry to be nitpicky, but now that we've broken it up, why not call running tasks running tasks instead of active tasks?

PR #408, opened 2018-07-23
Fix incorrect variable name logging checkpoint_period parsing problem

benclifford:


benclifford:
note that this changes behaviour when there's a parsing problem: previously this threw an exception (due to the error being fixed) but now a log message is logged and execution continues.

I think I prefer the exception-throwing behaviour.

PR #414, opened 2018-07-25
Use an empty list instead of None for default executors list.

benclifford:
Without this, a type exception is thrown when the code tries to
iterate over the non-iterable None default value.


yadudoc:
This is not recommended due to issues with how python lists in params are pass-by-reference leading to behavior like this : https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
It is preferable to specify types with annotations (https://www.python.org/dev/peps/pep-3107/) that is supported in Python3, and error out in the body. We've got many instances in the code that do this, but I'm more worried about this since users might try to add/remove executors dynamically.

PR #415, opened 2018-07-25
Tighten scale_in signatures and docstrings

benclifford:


parsl/executors/base.py
```diff
@@ -40,9 +40,11 @@ def scale_out(self, *args, **kwargs):
         pass
 
     @abstractmethod
-    def scale_in(self, *args, **kwargs):
+    def scale_in(self, count):
         """Scale in method.
 
+        Cause the executor to reduce the number of workers by count.
```
annawoodard: @yadudoc does this actually increase the number of workers by count, or the number of blocks by count?

PR #449, opened 2018-08-07
adding initial codemeta file - needs all contributors added

danielskatz:


danielskatz:
also need to fix orcids for @mjwilde @jmjwozniak & Ian

danielskatz:
I updated the PR to not have this conflict, but it still is shown - I'm not sure how to mark the conflict as resolved.

PR #456, opened 2018-08-09
Merge libsubmit repository into parsl repo.

benclifford:
This does not attempt to merge the packaging/release process, the testing or the documentation, but brings the two repositories together so that further merge work may proceed (hopefully more easily) in a single repository.

Notably, this will break anything which relies on libsubmit being at a particular github URL - probably documentation and libsubmit testing.


benclifford:
My proposal is the following intentionally-asynchronous steps:

1) we merge libsubmit and parsl into the same repo, with submit as a subdir; but have it still exist as two separate packages still. Anything which uses the libsubmit github repo URL explicitly will need updating (our CI and our doc generation, mostly?).

What we get at this stage is the ability to make changes to both projects atomically: breaking changes that would need something edited in both parsl and libsubmit can, at this stage, form a single commit.

Releases would still happen (if indeed a release happens before further steps) as two separate package preparations, with a `parsl` and a `libsubmit` release.

I propose this stage doesn't last long.

2) People perform further merge work piece-wise.

I played with this by hand, and it looked mostly straightforward - but not entirely so. Mostly there was some fiddling with how the docs link together.

With a single repo in place by step [1], this can happen when someone who loves packaging wants to rearrange the packaging; and when someone who loves docs wants to rearrange the docs, without there being too much of a subsequent "big cutover" between repos.

So step 1 merges the version control, which is intended to make merging of other things in step 2 much easier and less dependent on a big switchover.

annawoodard:
I vote to do everything in a single step. My reasoning is that splitting things into two steps, while being more flexible in terms of how long we have to perform the work, will result in more integrated work over all. I don't mind volunteering to do the merge in a single fell swoop, if we agree on a convenient time to do it-- for example, I think we should merge the mpi executor into the main branch beforehand.

benclifford:
I am fine with someone else doing this entirely in one big blast. I was mostly trying to start an incremental process that might be able to make progress rather than always being punted to the future. Alas, that got punted to the future.

annawoodard:
Superseded by #508

PR #462, opened 2018-08-14
Add simpler instructions for starting jupyter

annawoodard:


annawoodard:
Thanks @yadudoc. I've clarified the text to provide instructions for starting both a single notebook or jupyter lab.

PR #464, opened 2018-08-14
added a deep copy method to copy config. Fixes #438

### LINKED ISSUE #438: Add a custom `copy` method for configs
An unfortunate side effect of our recent switch to class-based configurations is confusing behavior when copying configurations. I think what most people want when they do something like (simplest case as an example):
```
In [1]: from parsl.config import Config
In [2]: from parsl.executors.threads import ThreadPoolExecutor
In [3]: threads = ThreadPoolExecutor(max_threads=4)
In [4]: config_a = Config(executors=[threads])
In [5]: config_b = config_a
In [6]: config_b.executors[0].max_threads = 5
In [7]: config_a.executors[0].max_threads
Out[7]: 5
```

is for `config_b` to be a deep copy of `config_a`, which would have required:
```
In [8]: import copy
In [9]: threads = ThreadPoolExecutor(max_threads=4)
In [10]: config_a = Config(executors=[threads])
In [11]: config_b = copy.deepcopy(config_a)
In [12]: config_b.executors[0].max_threads = 5
In [13]: config_a.executors[0].max_threads
Out[13]: 4
```

One possibility is to encourage people to use `copy.deepcopy` instead of `=`. Another possibility is (assuming many people will miss details in the documentation) to override `__copy__` to always make a deep copy. I don't love the idea of overriding `__copy__` (because it would result in unexpected behavior for someone thinking carefully about the operation), but I think it would lower the total amount of unexpected behavior.

ZhuozhaoLi:


benclifford:
I have a vague sense of unease with #438 / #464 with how much deepcopy actually solves the problem of copying a copying.

A "config" contains actual executors - not "configuration for executors" (right?)

What happens when i deep copy a config that contains excecutors (or other stateful stuff) that has been merrily starting threads/launching things and updating its internal state to refer to those things? (when I'm deep in a run and at that point decide I want to initialise a new DFK).

deepcopy at that point has similar problems as re-using a config, maybe.

annawoodard:
That's a great point, @benclifford. Nothing dynamic happens in the config constructors-- that is controlled by the DFK with the `start` method. But the correct way to do this is probably to instantiate a new object with the same calling arguments as the original. @ZhuozhaoLi you can search through the code for the `signature` module to see places where we do similar things.

benclifford:
potentially could deepcopy the config at the point that it goes into the DFK - that might solve any uncertainties we have about people modifying the config during a run anyway?

annawoodard:
I support @benclifford's recommendation (and forget introspection)-- to clarify, I think this could just be in the constructor, which sets an attribute `self.original` or whatever with a deepcopy.

annawoodard:
Agreed, that's much better. [Here](https://github.com/Parsl/parsl/blob/master/parsl/dataflow/dflow.py#L72) is where I think we'd need the deepcopy. We'd still want the dedicated `copy` method though I think for the reasons outlined in the original issue.

ZhuozhaoLi:
> Agreed, that's much better. Here is where I think we'd need the deepcopy. We'd still want the dedicated copy method though I think for the reasons outlined in the original issue.

After trying some potential solutions, I do not think we should add a custom deepcopy method to make deepcopy of the config, . Based on the [documentation](https://docs.python.org/3/library/copy.html) that `deepcopy() does not copy types like module, method, stack trace, stack frame, file, socket, window, array, or any similar types.` 

In the config (even we just specify the configuration), it already instantiates a new ThreadPoolExecutor, which includes the threading.Lock() and hence disallows the use of deepcopy. Making `deepcopy` and `load(config)` in this case returns error `TypeError: can't pickle _thread.lock objects` (same error occurs [here](https://github.com/tensorflow/tensorflow/issues/14189#issuecomment-341633152)).

Below is the sample code you may try that returns the error.

```
import parsl
from parsl.executors.threads import ThreadPoolExecutor
import copy

threads = ThreadPoolExecutor(max_threads=4)
config_a = Config(executors=[threads])
parsl.load(config_a)
parsl.clear()

config_c = copy.deepcopy(config_a)
parsl.load(config_c)
parsl.clear()
```
Probably, to allow 'copy', we should create another identical config instead? Do you have any other viable suggestions? @benclifford @annawoodard

annawoodard:
There are probably very good reasons _not_ to do it this way, which @benclifford should point out to us, but this is the first thing that occurs to me:

```
import parsl
from parsl.executors.threads import ThreadPoolExecutor
import copy

threads = ThreadPoolExecutor(max_threads=4)
config_a = Config(executors=[threads])
parsl.load(config_a)
parsl.clear()

config_c = eval(str(config_a))
parsl.load(config_c)
parsl.clear()
```

annawoodard:
This was a really valiant effort on @ZhuozhaoLi's part, but it looks like we need to re-think this a bit and come back to it later.

PR #465, opened 2018-08-14
Improve strategy logging

annawoodard:
Considering a config with two IPP executors `midway-1` and `midway-2`, this changes the logging line in `strategy.py` from
```
2018-08-14 18:01:26 parsl.dataflow.strategy:260 [DEBUG] Tasks: 0 Slots: 0 Parallelism: 1
2018-08-14 18:01:26 parsl.dataflow.strategy:260 [DEBUG] Tasks: 0 Slots: 0 Parallelism: 1
```
to

```
2018-08-14 18:01:26 parsl.dataflow.strategy:260 [DEBUG]  
executor      active tasks    running blocks    submitting blocks    pending blocks    connected engines
----------  --------------  ----------------  -------------------  ----------------  -------------------
midway-1                 0                 0                    0                 1                    0
midway-2                 0                 0                    0                 1                    0
```

Logging a table in this way is a bit heterodox but it seemed like the clearest way to display info when there are multiple executors, and a not-terrible way when there is a single executor. Feedback on whether this makes things clearer or more confusing is appreciated. Another big change here is logging the number of connected engines directly-- hopefully this will close the 'check controller logs, see that it says connected and then ends which means there are actually no connected engines' loop.

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -2,8 +2,6 @@ import logging
 import time
 import math
 
-import tabulate
-
 from parsl.executors.ipp import IPyParallelExecutor
 
 logger = logging.getLogger(__name__)
@@ -164,8 +162,6 @@ class Strategy(object):
             - kind (Not used)
         """
 
-        headers = ['executor', 'active tasks', 'running blocks', 'submitting blocks', 'pending blocks', 'connected engines']
-        data = []
         for label, executor in self.dfk.executors.items():
             if not executor.scaling_enabled:
                 continue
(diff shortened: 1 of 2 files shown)


annawoodard:
After discussion on Slack, I've removed the table. The log now reads:

```
2018-08-14 20:06:19 parsl.dataflow.strategy:191 [DEBUG]  Executor midway-1 has 0 active tasks, 0/0/1 running/submitted/pending blocks, and 0 connected engines
2018-08-14 20:06:19 parsl.dataflow.strategy:191 [DEBUG]  Executor midway-2 has 0 active tasks, 0/0/1 running/submitted/pending blocks, and 0 connected engines
```

PR #474, opened 2018-08-20
Fix data.rst - storage_access should be a list

lukaszlacinski:
It corrects the 'Data management' documentation page where storage_access in an example should be a list instead of a GlobusSchema object.


parsl/tests/configs/local_threads_globus.py
```diff
@@ -14,10 +14,10 @@
     executors=[
         ThreadPoolExecutor(
             label='local_threads_globus',
-            storage_access=GlobusScheme(
+            storage_access=[GlobusScheme(
                 endpoint_uuid=user_opts['globus']['endpoint'],
                 endpoint_path=user_opts['globus']['path']
-            ),
+            )[,
```
annawoodard: [ -> ]

Issue #472, opened 2018-08-20
Change Local Database Strategy from Many Writers to a Single Writer


ConnorPigg:
Currently, local database logging follows the same pattern as Elasticsearch logging in that each worker connects directly to log storage to write their own resource usage. This causes issues with the SQLite database that only handles single writers. The proposal is to change the workers to send information back to the machine that started the Parsl run and have that be the sole writer of both resource and status updates to the database. This will ideally solve the inconsistencies with local database logging and be more in line with what other successful projects that include task monitoring do.

Issue #471, opened 2018-08-20
Generic Monitoring Improvements


ConnorPigg:
What follows is a list of improvements too small to warrant individual issues.

- [x] Allow resource monitoring frequency to be configurable
- [x] Chand runid to be prettier
- [x] Change table names in local logging to be prettier
- [x] Make timestamp format consistent
- [x] Handle race condition if a task resource update occurs before a task status update
- [x] Lower precision of information reported and make the format consistent
- [x] Change database logging configuration to match the new configuration class styles.
- [x] Document the monitoring code
- [x] Default local database logging to SQLite inside the rundir
- [x] Log user and host information to local workflows database database

Issue #470, opened 2018-08-20
Missing Database Status Information


ConnorPigg:
With both the Elasticsearch and local database monitoring implementations there is frequent information loss. This is most notable by observing complete missing resource usage information about certain tasks but is also apparent when observing missing status updates about task state transitions.

I suspect for Elasticsearch the missing resource information occurs due to errors with the handler and psutil library. Inconsistent reentrant errors from the logging handler and psutil access are examples. Missing status update information seems less obvious.

For the local logging implementation, I suspect concurrent writing to SQLite to be the cause. This would explain why status information is lost as an active resource update can abort or interfere with task status updates.

For Elasticsearch, some error/exception squashing occurs to try to log what information is available. For local database logging, I hope changing the strategy from multiple writers to a single writer will avoid the issue.

ConnorPigg:
One bug causing the missing final task status update in local database logging was poor logic that was fixed inside the logging handler.

PR #493, opened 2018-09-13
Monitoring improvements

ConnorPigg:
This stands to be v1 for local logging. It addresses issues #472 #471 and #470. It uses a tornado server to receive log messages over HTTP which then acts as a single writer to a database. The feature appears to accurately log workflow, task status, and task resource information to a local database. ~It unintentionally streams log messages to the console.~ It assumes workers have the ability to connect to the machine running the main Parsl script over HTTP.


parsl/monitoring/app_monitor.py
```diff
@@ -5,59 +5,87 @@
...
     simple = ["cpu_num", 'cpu_percent', 'create_time', 'cwd', 'exe', 'memory_percent', 'nice', 'name', 'num_threads', 'pid', 'ppid', 'status', 'username']
+    # values that can be summed up to see total resources used by task process and its children
     summable_values = ['cpu_percent', 'memory_percent', 'num_threads']
 
-    logger = get_db_logger(enable_es_logging=False) if db_logger_config is None else get_db_logger(**db_logger_config)
-    logger.info("starting monitoring for {} on {}".format(pid, os.getpid(), extra={'task_id': task_id, 'task_run_id': run_id}))
     pm = psutil.Process(pid)
     pm.cpu_percent()
 
     def to_mb(some_bytes):
```
benclifford: my personal taste would be to log the raw byte count, and deal with presentation units elsewhere. not a super big deal though. mostly I've found it irksome in practice dealing with monitoring systems where it's not clear what some number is: does my process occupy 20Gb or 20Mb? (when either might be plausible so i can't guess from context).

parsl/monitoring/app_monitor.py
```diff
@@ -5,59 +5,87 @@
...
             try:
-                d['psutil_process_disk_write'] += to_mb(child.io_counters().write_bytes)
-                d['psutil_process_disk_read'] += to_mb(child.io_counters().read_bytes)
+                d['psutil_process_disk_write'] = to_mb(pm.io_counters().write_bytes)
+                d['psutil_process_disk_read'] = to_mb(pm.io_counters().read_bytes)
             except psutil._exceptions.AccessDenied:
-                pass
-        logger.info("test", extra=d)
-        time.sleep(5)
+                # occassionally pid temp files that hold this information are unvailable to be read so set to zero
```
benclifford: is this a bug? in this code? in some other code?

parsl/monitoring/app_monitor.py
```diff
@@ -5,59 +5,87 @@
...
+                    d['psutil_process_disk_read'] += to_mb(child.io_counters().read_bytes)
+                except psutil._exceptions.AccessDenied:
+                    # occassionally pid temp files that hold this information are unvailable to be read so add zero
+                    d['psutil_process_disk_write'] += 0
+                    d['psutil_process_disk_read'] += 0
+
+            to_be_rounded = ['psutil_process_memory_virtual', 'psutil_process_memory_resident', 'psutil_process_cpu_percent', 'psutil_process_memory_percent',
+                             'psutil_process_disk_write', 'psutil_process_disk_read']
+            for k in to_be_rounded:
+                d[k] = round(d[k], 2)
```
benclifford: i'm not a fan of rounding deep inside stuff. i'm not going to really argue, though.

parsl/monitoring/app_monitor.py
```diff
@@ -5,59 +5,87 @@
...
+    Wrap the Parsl app with a function that will call the monitor function and point it at the correct pid when the task begins.
+    """
     def wrapped(*args, **kwargs):
-        p = Process(target=monitor, args=(os.getpid(), task_id, db_logger_config, run_id))
+        p = Process(target=monitor, args=(os.getpid(), task_id, db_logger_config_object, run_id))
         p.start()
         try:
             return f(*args, **kwargs)
         finally:
+            # Note: I believe finally blocks are not called if the python parsl process is killed with certain SIGs which may leave the monitor as a zombie
```
benclifford: If the parsl process is killed with `SIGKILL`, then no one gets any chance to shut down nicely (including leaving all kinds of stuff running, not just this)

Other signals (SIGTERM, SIGINT) I think will run this finaliser (or at least, if not, other parts of parsl also have problems...)

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+# TODO: expand to full set of info
+def create_task_status_table(meta):
+    return Table(
+          'task_status', meta,
+          Column('task_id', Text, sa.ForeignKey('task.task_id'), nullable=False),
+          Column('task_status', Text, nullable=False),
+          Column('task_status_name', Text, nullable=False),
+          # Column('timestamp', Text, nullable=False, primary_key=True),
```
benclifford: remove dead code

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+          Column('timestamp', Text, nullable=False),
+          Column('task_run_id', Text, sa.ForeignKey('workflows.task_run_id'), nullable=False),
```
benclifford: these columns should just be `run_id` ?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+          Column('task_fail_count', Text, nullable=False),
+          Column('task_fail_history', Text, nullable=True),
+    )
+
+
+def create_workflow_table(meta):
```
benclifford: this creates the `task` table, right? not the `workflow` table? so rename to create_task_table?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+          Column('task_stdin', Text, nullable=True),
+          Column('task_stdout', Text, nullable=True),
+    )
+
+
+def create_task_resource_table(meta):
+    return Table(
+          'task_resources', meta,
+          Column('task_id', Text, sa.ForeignKey('task.task_id'), nullable=False),
+          # Column('timestamp', Text, nullable=False, primary_key=True),
```
benclifford: remove dead code

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+                                self.meta.create_all(con)
+                                con.execute(task_resource_table.insert().values(**{k: v for k, v in info.items() if k in task_resource_table.c}))
+                            # if this resource table already exists, just insert the update
+                            else:
+                                task_resource_table = self.meta.tables['task_resources']
+                                con.execute(task_resource_table.insert().values(**{k: v for k, v in info.items() if k in task_resource_table.c}))
+                            return
+
+            except Exception as e:
+                print(str(e))
```
benclifford: log this at ERROR level rather than printing it, along with an explanation (eg your code comment at the start of the loop) so that someone reading the log doesn't have to read the source code. Log the number of tries with it as that info is easily available.

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+            failed = False
+            try:
+                with self.eng.connect() as con:
+                    # formating values to convert from python or parsl to db standards
+                    info['task_fail_history'] = str(info.get('task_fail_history', None))
+                    info['timestamp'] = time.strftime('%Y-%m-%d %H:%M:%S', time.localtime(record.created))
+                    run_id = info['task_run_id']
+
+                    # if workflow or task has completed, update their entries with the time.
+                    if 'time_completed' in info.keys() and info['time_completed'] != 'None':
```
benclifford: should that be `'None`` or None, the python none-value? if the former, that's weird...

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+                        workflows = self.meta.tables['workflows']
+                        up = workflows.update().values(time_completed=info['time_completed']).where(workflows.c.task_run_id == run_id)
+                        con.execute(up)
+                        return
```
benclifford: does that mean if the first call to this is a completed call (which might or might not happen depending on how other stuff in parsl goes...), then none of the other records will get recorded?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+                    if 'tasks_completed_count' in info.keys():
+                        workflows = self.meta.tables['workflows']
+                        up = workflows.update().values(tasks_completed_count=info['tasks_completed_count']).where(workflows.c.task_run_id == run_id)
+                        con.execute(up)
+                    if 'tasks_failed_count' in info.keys():
+                        workflows = self.meta.tables['workflows']
+                        up = workflows.update().values(tasks_failed_count=info['tasks_failed_count']).where(workflows.c.task_run_id == run_id)
+                        con.execute(up)
+
+                    # create workflow table if this is a new run without one
```
benclifford: change comment to "create task table" ?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+                        workflow = create_workflow_table(self.meta)
+                        self.meta.create_all(con)
+
+                    # check to make sure it is a task log and not just a workflow overview log
+                    if info.get('task_id', None) is not None:
+                        # if this is the first sight of the task in the workflow, add it to the workflow table
+                        if len(con.execute(self.meta.tables['task'].select(self.meta.tables['task'].c.task_id == info['task_id'])
+                                               .where(self.meta.tables['task'].c.task_run_id == run_id)).fetchall()) == 0:
+                            if 'psutil_process_pid' in info.keys():
+                                # this is the race condition that a resource log is before a status log so ignore this resource update
```
benclifford: So specifically here, I'm losing a log line of resource usage information, and hoping that next time round, the task row will have been created?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+        standard_log_info = ['name', 'msg', 'args', 'levelname', 'levelno', 'pathname', 'filename', 'module', 'exc_info', 'exc_text', 'stack_info', 'lineno',
+                             'funcName', 'created', 'msecs', 'relativeCreated', 'thread', 'threadName', 'processName', 'process']
+        for t in range(self.retries):
+            try:
+                info = {k: str(v) for k, v in record.__dict__.items() if not k.startswith('__') and k not in standard_log_info}
+                bod = 'log={}'.format(json.dumps(info))
+                http_client.fetch(self.addr, method='POST', body=bod, request_timeout=self.request_timeout)
+            except Exception as e:
+                # Other errors are possible, such as IOError.
+                print("Error: " + str(e))
```
benclifford: log, not print

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+        """ Set up various behavior of the attempt to connect to the logging server. Not currently configurable. """
+        logging.Handler.__init__(self)
+        self.addr = web_app_host + ':' + str(web_app_port)
+        self.request_timeout = request_timeout
+        self.retries = retries
+        self.on_fail_sleep_duration = on_fail_sleep_duration
+
+    def emit(self, record):
+        """ Open up an HTTP connection to the logging server. This is currently blocking.
+        It should not matter for the resource monitors but could hold up the DFK logs if this connection/sending takes a lot of time. """
```
benclifford: does this blocking only hold up logging, or will it hold up workflow progression in general?

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+        http_client = httpclient.HTTPClient()
```
benclifford: wonder about http connection sharing - does this create a new HTTP/TCP connection for every single call of `emit`? if so, does sharing an http client make that better?

parsl/monitoring/db_logger.py
```diff
@@ -30,23 +32,85 @@ def emit(self, record):
...
 
-def get_db_logger(host='search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-east-1.es.amazonaws.com',
-                  port=443,
-                  enable_es_logging=False,
-                  index_name="my_python_index",
-                  version='1.0.0',
+class LoggerConfig():
+    """ This is a config class for creating a logger. """
+    def __init__(self,
+                 host='search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-east-1.es.amazonaws.com',
```
benclifford: is that a useful default? or should the code throw an exception saying this parameter is mandatory when needed?

(alternatively: aree arbitrary parsl users expected to often use that particular instance? and if so, get a proper DNS name set up for it)

parsl/monitoring/db_logger.py
```diff
@@ -57,24 +121,40 @@ def get_db_logger(host='search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-
...
-                                       'port': port}],
-                               use_ssl=True,
+        handler = CMRESHandler(hosts=[{'host': db_logger_config_object.host,
+                                       'port': db_logger_config_object.port}],
+                               use_ssl=db_logger_config_object.enable_ssl,
                                auth_type=CMRESHandler.AuthType.NO_AUTH,
-                               es_index_name=index_name,
+                               es_index_name=db_logger_config_object.index_name,
                                es_additional_fields={
                                    'Campaign': "test",
```
benclifford: what's this campaign?

parsl/monitoring/db_logger.py
```diff
@@ -57,24 +121,40 @@ def get_db_logger(host='search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-
...
+        logger.setLevel(logging.INFO)
+        logger.addHandler(handler)
+    elif db_logger_config_object.logger_type == 'local_database' and is_logging_server:
+        # add a handler that will take logs being recieved on the server and log them to the database
+        handler = DatabaseHandler(db_logger_config_object.eng_link)
+        # use the specific name generated by the server or the monitor wrapper
+        logger = logging.getLogger(logger_name)
         logger.setLevel(logging.INFO)
         logger.addHandler(handler)
     else:
```
benclifford: you should never hit this else, right? because you've gone over all the valid cases of logger_type? (if so, then throw an exception indicating that the code is broken if you hit it rather than silently failing)

parsl/monitoring/logging_server.py
```diff
@@ -0,0 +1,42 @@
+import tornado.ioloop
+import tornado.web
+import json
+from parsl.monitoring.db_logger import get_db_logger
+
+
+class MainHandler(tornado.web.RequestHandler):
+    """ A handler for all requests/logs sent to the logging server."""
+    def initialize(self, db_logger_config_object):
+        """This function is called on every request which is not ideal but __init__ does not appear to work."""
```
benclifford: what does that mean?

parsl/monitoring/logging_server.py
```diff
@@ -0,0 +1,42 @@
...
+        self.flush()
+
+    def post(self):
+        """
+        Defines responses to post requests for the / ending. Receives logs from workers and main dfk in the body of the post request.
+        Should be log=json.dumps(info). Then writes this info to the database using the database handler.
+        This needs to be a quick function so that the server can accept other requests and may be a bottle neck for incoming logs.
+        """
+        arg = json.loads(self.get_body_argument('log'))
+        try:
```
benclifford: can this initialisation of `self.application.logger` happen in an initialisation function?

parsl/dataflow/dflow.py
```diff
@@ -84,25 +88,49 @@ def __init__(self, config=Config()):
         self.usage_tracker.send_message()
 
         # ES logging
-        self.db_logger_config = config.db_logger_config
-        self.db_logger = get_db_logger(enable_es_logging=False) if self.db_logger_config is None else get_db_logger(**self.db_logger_config)
-        self.workflow_name = str(inspect.stack()[1][1])
+        self.tasks_completed_count = 0
+        self.tasks_failed_count = 0
+        self.db_logger_config_object = config.db_logger_config
+        if self.db_logger_config_object is not None and self.db_logger_config_object.logger_type == 'local_database'\
```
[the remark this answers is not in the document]
ConnorPigg: That makes sense, changed it back.

parsl/dataflow/dflow.py
```diff
@@ -84,25 +88,49 @@ def __init__(self, config=Config()):
...
+                and self.db_logger_config_object.eng_link is None:
+            # uses the rundir as the default location. Should report to users?
```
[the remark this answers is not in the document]
ConnorPigg: Removed the question at the end.

parsl/dataflow/dflow.py
```diff
@@ -84,25 +88,49 @@ def __init__(self, config=Config()):
...
+            self.logging_server = multiprocessing.Process(target=logging_server.run, kwargs={'db_logger_config_object': self.db_logger_config_object})
+            self.logging_server.start()
+        else:
+            self.logging_server = None
+        workflow_info = {
+                'python_version': sys.version_info,
+                'parsl_version': get_version(),
+                'libsumbit_version': libsubmit.__version__,
+                "time_began": str(self.time_began.strftime('%Y-%m-%d %H:%M:%S')),
+                'time_completed': str(self.time_completed),
```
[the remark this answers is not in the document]
ConnorPigg: Replaced with str(None) as the time_completed should be None before DFK cleanup.

parsl/monitoring/app_monitor.py
ConnorPigg: That makes sense. I don't have a preference and am happy to remove it. It was initially included here to prepare the data for direct presentation via Elasticsearch/Kibana.

parsl/monitoring/app_monitor.py
ConnorPigg: Good question. I do not know the answer for exactly why this happens but from my experience, it seems that psutil is not always successful at gathering certain information about every process. This is one of a few unreliable points in the collection process. My approach was that it was relatively harmless to take this simple approach of ignoring failed attempts to gather the information and take what we can get when it is available with the assumption that over the course of a workflow the general idea of what is happening will still be clear. It is likely useful to explore through testing how accurate this resource collecting process is. As far as I can tell it seems rather representative of what I would expect but I do not have any hard evidence for its performance or accuracy.

parsl/monitoring/app_monitor.py
ConnorPigg: This is similar to the change to MB (also the rounding of the times in the DFK logs), it was done in an effort to clean up the presentation through Elasticsearch. I am happy to remove this rounding.

parsl/monitoring/db_local.py
ConnorPigg: It is weird. This is the database handler which receives the logs from an HTTP request. The result is that everything was converted to a string to be sent over HTTP thus this is the string ''None''. I'm sure there are better ways of sending these logs over HTTP but this was how I got it to work. Maybe for a v2 I could switch away from all strings?

parsl/monitoring/db_local.py
ConnorPigg: In general, I would describe the motto of this monitoring to make no promises on the success of a log/information being recorded but to try hard to capture what information it can. From that perspective, I guess there could be a weird run where nothing is recorded.

If the workflow completed call is the first received then it means the DFK cleanup() has been called and the logging server will be shut down and I can only speculate on exactly what would and would not get recorded.

ryanchard:
The run_id is currently a combination of the script name and the minute of the hour it was submitted. Could we change this to a uuid so we can reliably look up previous executions and inspect their performance?

parsl/monitoring/db_local.py
ConnorPigg: Yes, this drops a resource a log line. I wouldn't expect this to be often unless resource_loop_sleep_duration is very small. I do not know of a more practical response to this situation.

parsl/monitoring/db_local.py
ConnorPigg: While the logging server is handling a log, all other logs are queued holding up logging. When a log from the DFK (for example a task status update) is submitted, it blocks while it lines up in the server's queue and will block until the request is handled or it has failed too many times. To avoid this blocking, I could try to implement the asynchronous HTTP request which I am currently unfamiliar with. In action, I'm not sure how worrisome or impactful this blocking is but I think it is important to have considered/mentioned.

parsl/monitoring/db_local.py
ConnorPigg: Yes, it does create a new connection with every call to emit. I am unfamiliar with connection sharing and do not know what impact it would have on performance.

parsl/monitoring/db_logger.py
ConnorPigg: I agree this is a bad choice for a default. I have made some changes to these defaults and tried to raise errors for some incorrect configurations. What level of checking would you recommend for these inputs?

parsl/monitoring/db_logger.py
ConnorPigg: Not sure, it was there in the initial Elasticsearch logger example and so I left it. From what I could tell it is an identifier that can be used with Elasticsearch but it might be arbitrary or unnecessary.

parsl/monitoring/logging_server.py
ConnorPigg: With my attempts, I could not get __init__ to work with this handler in order to pass in and set the config information. Instead, there is this initialize method that to my understanding is called at the beginning of every HTTP request using the arguments from the request and additional arguments specified during the Applications configuration and start. In this sense, it is not ideal to have to keep setting something that stays constant.

parsl/monitoring/app_monitor.py
annawoodard: I agree with Ben

parsl/monitoring/app_monitor.py
annawoodard: I agree with @benclifford

parsl/monitoring/db_logger.py
```diff
@@ -30,23 +32,90 @@ def emit(self, record):
         pass
 
 
-def get_db_logger(host='search-parsl-logging-test-2yjkk2wuoxukk2wdpiicl7mcrm.us-east-1.es.amazonaws.com',
-                  port=443,
-                  enable_es_logging=False,
-                  index_name="my_python_index",
-                  version='1.0.0',
+class LoggerConfig():
```
annawoodard: I would prefer to call this a `MonitoringConfig` to make it clearer to users who associate logging with "something that writes text messages out to `parsl.log`" that this is specifically dealing with monitoring. Additionally, we might want to stick viz options in the same configuration, in which case LoggingConfig would make less sense. For similar reasons I vote `logger_type` -> `database_type` since both options (local or elasticsearch) are types of databases.

parsl/monitoring/db_local.py
```diff
@@ -0,0 +1,218 @@
...
+          Column('task_executor', Text, nullable=False),
+          Column('task_fn_hash', Text, nullable=False),
+          Column('task_time_started', Text, nullable=False),
+          Column('task_time_completed', Text, nullable=True),
+          Column('task_memoize', Text, nullable=False),
+          Column('task_inputs', Text, nullable=True),
+          Column('task_outputs', Text, nullable=True),
+          Column('task_stdin', Text, nullable=True),
+          Column('task_stdout', Text, nullable=True),
+    )
```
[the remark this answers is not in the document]
ConnorPigg: The task_status table contains updates about when tasks change state ie from running to complete and has multiple status updates per workflow task.
The task table contains information that is static and known when first seeing the task ie name, inputs, etc. and has one row per workflow task.
The idea for separating is that they to reduce duplicate information by not having every status update contain the static information.

parsl/monitoring/db_logger.py
```diff
@@ -30,23 +32,90 @@ def emit(self, record):
...
+            Determines whether to use Elasticsearch logging or local database logging, defaults to 'local_database' and accepts 'elasticsearch'.
+        index_name : str, optional
+            Used with Elasticsearch logging, the name of the index to log to.
+        logger_name : str, optional
+            Used with both Elasticsearch and local db logging to define naming conventions for loggers.
+        eng_link : str, optional
+            Used with local database logging, SQLalchemy engine link to define where to connect to the database. If not set, DFK init will use a sqlite3
+            database inside the rundir.
+        version : str, optional
+            Optional workflow identification to distinguish between workflows with the same name, not used internally only for display to user.
```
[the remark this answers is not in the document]
ConnorPigg: This can be done with the uuid run_id that is unique to each workflow. I believe the suggestion for having a simple version was to potentially give users something easier to work with that they could control.

parsl/monitoring/db_local.py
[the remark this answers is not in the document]
annawoodard: Gotcha, makes perfect sense-- thanks for the clarification!

PR #524, opened 2018-09-25
All working files in rundir #457

### LINKED ISSUE #442: Make there be a single `script_dir` definition
Reported by @djf604.

It is confusing that we have `script_dir` defined both in the `Provider` and the `Channel`. They do have two different meanings: one is the place we store the scripts _locally_, the other is _remotely_. I vote we remove the confusion by hardcoding the `Provider` script_dir to be a directory within `Config.run_dir`.

### LINKED ISSUE #457: All run-related working files should go in the rundir unless otherwise configured
At the moment, run-related working files appear in various places, including in the hidden directories .scripts and .ipp_controller.

These should all go in the relevant rundir, when that directory is available.

At present, I'm finding it confusing to find relevant files: both because I have to remember there are several places to look, and because many of the automatically named files do not correlate easily with parsl runs that happen close in time/simultaneously.

### LINKED ISSUE #521: LocalProvider Initialized with Default Arguments, even when using TorqueProvider
See Parsl/libsubmit#74

annawoodard:
Unless otherwise configured, this makes all run-related files go into
the run_dir (fixes #457). It also eliminates the script_dir keyword
argument in the Provider (instead, it is hard-coded to be inside the
run_dir); fixes #442. Having it configurable adds no benefit and a lot
of confusion. By default, if the local run_dir is accessible on the
remote machine, output files will be put there; otherwise they will be
put in a run_dir with the same number on the remote side. This should
also fix Parsl/libsubmit#74 (correspondingly, fix #521).

For example, this config:
```
config = Config(
    executors=[
        IPyParallelExecutor(
            provider=SlurmProvider(
                'westmere',
                channel=SSHChannel(
                    hostname='swift.rcc.uchicago.edu',
                    username=user_opts['midway']['username'],
                ),
                overrides=user_opts['midway']['overrides']
            ),
            label='midway_ipp',
        )
    ]
)
```
produces this locally
```
~/ci/parsl/parsl/tests >ls -l runinfo/000
total 24
drwxr-xr-x  2 awoodard  staff    64 Sep 24 22:40 extensions
drwxr-xr-x  3 awoodard  staff    96 Sep 24 22:40 local_submit_scripts
-rw-r--r--  1 awoodard  staff  1625 Sep 24 22:41 midway_ipp.controller.err
-rw-r--r--  1 awoodard  staff     0 Sep 24 22:40 midway_ipp.controller.out
drwxr-xr-x  2 awoodard  staff    64 Sep 24 22:40 nbextensions
-rw-r--r--  1 awoodard  staff  4590 Sep 24 22:41 parsl.log
drwxr-xr-x  6 awoodard  staff   192 Sep 24 22:40 profile_midway_ipp
```
and this remotely
```
[midway-login1] ~ >ls -l runinfo/000/
total 0
drwxrwxr-x 2 annawoodard annawoodard 4096 2018-09-24 22:41 engine_logs
-rw-rw-r-- 1 annawoodard annawoodard  334 2018-09-24 22:41 ipengine.1b906104-2691-4180-8fa5-26440d05e116.json
drwxrwxrwx 2 annawoodard annawoodard 4096 2018-09-24 22:41 remote_submit_scripts
```
whereas the local_ipp config produces
```
(parsl_py36) [MacBook-Pro] ~/ci/parsl/parsl/tests >ls -l runinfo/000/
total 40
drwxr-xr-x  3 awoodard  staff    96 Sep 24 22:49 engine_logs
drwxr-xr-x  2 awoodard  staff    64 Sep 24 22:49 extensions
-rw-r--r--  1 awoodard  staff   334 Sep 24 22:49 ipengine.d8118096-45ba-4488-980e-528b304554a7.json
-rw-r--r--  1 awoodard  staff  2460 Sep 24 22:49 local_ipp.controller.err
-rw-r--r--  1 awoodard  staff  1044 Sep 24 22:49 local_ipp.controller.out
drwxr-xr-x  2 awoodard  staff    64 Sep 24 22:49 nbextensions
-rw-r--r--  1 awoodard  staff  4128 Sep 24 22:49 parsl.log
drwxr-xr-x  6 awoodard  staff   192 Sep 24 22:49 profile_local_ipp
drwxr-xr-x  4 awoodard  staff   128 Sep 24 22:49 submit_scripts
```


parsl/executors/ipp.py
```diff
@@ -69,7 +66,6 @@ class IPyParallelExecutor(ParslExecutor, RepresentationMixin):
     def __init__(self,
                  provider=LocalProvider(),
                  label='ipp',
-                 engine_file='~/.ipython/profile_default/security/ipcontroller-engine.json',
```
yadudoc: Good catch.

parsl/executors/ipp.py
```diff
@@ -112,26 +112,21 @@ def start(self):
         self.launch_cmd = command_composer(self.engine_file, self.engine_dir, self.container_image)
         self.engines = []
 
-        if self.provider:
```
yadudoc: This was left in to allow for users to potentially set provider=None, but no one would use that anymore.

parsl/executors/ipp.py
```diff
@@ -156,8 +158,8 @@ def compose_launch_cmd(self, filepath, engine_dir, container_image):
 {1}
 EOF
 
-mkdir -p '.ipengine_logs'
-ipengine --file=ipengine.{uid}.json {debug_option} >> .ipengine_logs/$JOBNAME.log 2>&1
+mkdir -p 'engine_logs'
```
yadudoc: Can you confirm that the engine_dir is actually used in this launch script ?

annawoodard: Yes, it is the directory that the batch jobs changes to before creating the `engine_logs` directory.

parsl/executors/ipp.py
annawoodard: Ah, thanks for the clarification!

PR #550, opened 2018-10-09
Do not use a global data manager

annawoodard:


parsl/data_provider/data_manager.py
```diff
@@ -47,13 +48,18 @@ class DataManager(ParslExecutor):
     to it, and DataFutures are returned.
     """
 
-    default_data_manager = None
-
     @classmethod
     def get_data_manager(cls, max_threads=None, executors=None):
```
benclifford: `max_threads` and `executors` are spurious arguments now - they're passed in to the constructor elsewhere now.

parsl/data_provider/data_manager.py
```diff
@@ -47,13 +48,18 @@ class DataManager(ParslExecutor):
...
-        if cls.default_data_manager is None:
-            cls.default_data_manager = DataManager(max_threads=max_threads, executors=executors)
-        return cls.default_data_manager
+        """Return the DataManager of the currently loaded DataFlowKernel.
+
+        Note that an error will be raised if one tries to create a File before
```
benclifford: this comment maybe belongs on `File`, not here.

It is (should be?) possible to create a file attached to an arbitrary DFK, not just the current globally loaded one, and this phrasing doesn't reflect that.

parsl/data_provider/data_manager.py
annawoodard: good catch, thanks ben!

parsl/data_provider/data_manager.py
annawoodard: agreed, i'll move it and clarify

benclifford:
It looks like this might break `parsl/tests/test_data/test_file.py` sometimes: that test does not initialise a DFK/DataManager and instead relies on a global data manager being available at all times - something this PR gets rid of.

I think the test is broken, not the PR.

I also think the test might pass sometimes, because it might acquire a global data manager from previous runs?

PR #552, opened 2018-10-09
Fix AppTimeout exceptions

benclifford:
Previously the supplied constructor parameter was used in a confused
fashion: in one place referred to as exitcode, in another used to
hold an exception.

This commit removes that parameter.


benclifford:
I haven't seen anything use the retries from this exception.

I also haven't seen anything explicitly catch it by type anywhere in the codebase.

Issue #559, opened 2018-10-11
MonitoringConfig should inherit from RepresentationMixin


annawoodard:
The `MonitoringConfig` should inherit from `RepresentationMixin`, so that the parsl config printed in logs is copy and pastable. This will make

```
Config(
    app_cache=True,
    checkpoint_files=None,
    checkpoint_mode=None,
    checkpoint_period=None,
    data_management_max_threads=10,
    executors=[IPyParallelExecutor(
        container_image=None,
        controller=Controller(
            interfaces=None,
            ipython_dir='~/.ipython',
            log=True,
            mode='auto',
            port=None,
            port_range=None,
            profile='default',
            public_ip=None,
            reuse=False
        ),
        engine_debug_level=None,
        engine_dir=None,
        label='ipp',
        managed=True,
        provider=LocalProvider(
            channel=LocalChannel(
                envs={},
                script_dir=None,
                userhome='/home/annawoodard/parsl/parsl/tests'
            ),
            init_blocks=4,
            label='local',
            launcher=SingleNodeLauncher(),
            max_blocks=10,
            min_blocks=0,
            nodes_per_block=1,
            parallelism=1,
            tasks_per_node=1,
            walltime='00:15:00'
        ),
        storage_access=[],
        working_dir=None
    )],
    lazy_errors=True,
    monitoring_config=<parsl.monitoring.db_logger.MonitoringConfig object at 0x7f1b340634e0>,
    retries=0,
    run_dir='runinfo',
    strategy='simple',
    usage_tracking=True
)
```

be, for example,

```
Config(
    app_cache=True,
    checkpoint_files=None,
    checkpoint_mode=None,
    checkpoint_period=None,
    data_management_max_threads=10,
    executors=[IPyParallelExecutor(
        container_image=None,
        controller=Controller(
            interfaces=None,
            ipython_dir='~/.ipython',
            log=True,
            mode='auto',
            port=None,
            port_range=None,
            profile='default',
            public_ip=None,
            reuse=False
        ),
        engine_debug_level=None,
        engine_dir=None,
        label='ipp',
        managed=True,
        provider=LocalProvider(
            channel=LocalChannel(
                envs={},
                script_dir=None,
                userhome='/home/annawoodard/parsl/parsl/tests'
            ),
            init_blocks=4,
            label='local',
            launcher=SingleNodeLauncher(),
            max_blocks=10,
            min_blocks=0,
            nodes_per_block=1,
            parallelism=1,
[...]

PR #560, opened 2018-10-12
Added RepresentationMixin to monitoring config.

ConnorPigg:
Also included a test script to manually check monitoring behavior runs. This script is simply an example of monitoring usage and does not assert anything but should be expected to work. Addresses issue #559.


annawoodard:
Fixes #559

Issue #564, opened 2018-10-13
Save function name to DB


annawoodard:
It looks like we are saving `task_fn_hash` to the DB. For tasks that have caching disabled, `task_fn_hash = func.__name__`. I think we need to save func.__name__ in a separate column, in case caching is enabled. This is available in `DataFlowKernel.tasks['func_name']`.

PR #570, opened 2018-10-15
Monitoring db naming

ConnorPigg:
Closes issue #564 and changes time-stamping to avoid string formatting to preserve precision. Pushes formating to UI.


parsl/dataflow/dflow.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, config=Config()):
         workflow_info = {
                 'python_version': sys.version_info,
                 'parsl_version': get_version(),
-                "time_began": str(self.time_began.strftime('%Y-%m-%d %H:%M:%S')),
+                "time_began": str(self.time_began),
```
annawoodard: Can we make float columns be `sqlalchemy.Float` type? Then we don't need to convert this to a string.

parsl/dataflow/dflow.py
```diff
@@ -409,7 +409,7 @@ def launch_task(self, task_id, executable, *args, **kwargs):
             executable = app_monitor.monitor_wrapper(executable, task_id, self.monitoring_config, self.run_id)
         exec_fu = executor.submit(executable, *args, **kwargs)
         self.tasks[task_id]['status'] = States.running
-        self.tasks[task_id]['time_started'] = str(datetime.now().strftime('%Y-%m-%d %H:%M:%S'))
+        self.tasks[task_id]['time_submitted'] = time.time()
```
annawoodard: Can we move this `time.time()` call to the very start of this method? In my view, `time_submitted` should be the time the _user_ submitted the task. For measuring latency we want to be able to measure how long it takes us to go round trip (user submits task -> user receives task output).

parsl/dataflow/dflow.py
```diff
@@ -280,7 +280,7 @@ def handle_update(self, task_id, future, memo_cbk=False):
                 final_state_flag = True
                 self.tasks_failed_count += 1
 
-                self.tasks[task_id]['time_completed'] = str(datetime.now().strftime('%Y-%m-%d %H:%M:%S'))
+                self.tasks[task_id]['time_completed'] = time.time()
```
annawoodard: I vote to change `time_completed` to `time_returned`, to distinguish when the task finished running on the worker from when the task was returned to the DFK. Soon we should add a `time_completed` which will refer to when the task finished running on the worker.

parsl/dataflow/dflow.py
ConnorPigg: Currently, all log information is being converted to strings by the handler that sends the log over HTTP inside db_local.py RemoteHandler. I convert them to strings here to be explicit about this. The reason for this conversion had to do with a problem while sending the information over HTTP.

parsl/dataflow/dflow.py
ConnorPigg: That makes sense and I will change this in the code.

parsl/dataflow/dflow.py
annawoodard: Got it, thanks for the clarification!

PR #580, opened 2018-10-21
Factor launch code

benclifford:
Prior to this commit, there were two ways in which a task would be launched,
depending on whether there were dependencies pending when the task was
created or not.

These two ways were similar but not identical.

This commit makes a single `launch_if_ready` method which captures one of those
launch behaviours - the case where a launch had been deferred.

This commit also changes the initial submission to always submit as if
dependencies were pending, and then immediately calls `launch_if_ready`.

This makes the code easier to follow, I think, as tidying up the separation
of creating a non-running task first, and then separately visiting it
to decide if it can be run.

There is a (hopefully minor) increased expense in visiting all of the dependencies
twice to see if they are done, on task creation (as launch_if_ready always checks
dependencies, even if task creation has just done so).

This refactoring also means that `launch_if_ready` can be called *anywhere* where
there is a possibility that a task has become ready for launch, without needing
further co-ordination of that launching.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -345,7 +345,6 @@ class DataFlowKernel(object):
                 # Acquire a lock, retest the state, launch
                 with self.task_launch_lock:
                     if self.tasks[task_id]['status'] == States.pending:
-                        self.tasks[task_id]['status'] = States.running  # launch_task also does this though? so maybe unnecessary?
                         exec_fu = self.launch_task(
                             task_id, self.tasks[task_id]['func'], *new_args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


benclifford:
After some rambling on slack, I'm concerned that there is a potential deadlock - already in master, but made more likely by this PR.

 launch a task (with or without the launch lock) which is cached, and so calls handle_update immediately (still within launch_task)- which maybe makes a dependant task ready to run - which gets launched by launch_task, within the outer launch_task.

In the current master case, the outer layer of this (buggily) doesn't take the lock, but even in that case I wonder if this causes a problem if we have several layers of dependency - so that we have multiple layers of handle_update/launch_task on the call stack.

PR #581, opened 2018-10-21
put failure states into constants

benclifford:
NOT TESTED: the usage tracking change

### DIFF
--- a/parsl/dataflow/usage_tracking/usage.py
+++ b/parsl/dataflow/usage_tracking/usage.py
@@ -10,7 +10,7 @@ import sys
 import platform
 import multiprocessing as mp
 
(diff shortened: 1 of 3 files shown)


parsl/dataflow/usage_tracking/usage.py
```diff
@@ -185,7 +185,7 @@ def construct_end_message(self):
 
         site_count = len([x for x in self.dfk.config.executors if x.managed])
 
-        failed_states = (States.failed, States.dep_fail)
+        failed_states = FINAL_FAILED_STATES
```
annawoodard: I slightly prefer search-and-replace `failed_states -> FINAL_FAILED_STATES` but don't want to hold up the PR for that

Issue #461, opened 2018-08-12
Parsl slows down after hundreds of thousands of jobs 


reidmcy:
I have a script that requires millions of jobs to be run on parsl and have noticed that after a day or so of running the time per jobs starts increasing. The jobs are independent, nearly identical and randomly ordered so I don't thinks its related to them. I've also changed from `Remote_IPP` to `IPyParallelExecutor` locally to `IPyParallelExecutor` on SLURM and disabled `app_cache`.

Batching jobs has helped reduce the slowdown so it is likely something internal to parsl causing the issue.

My config is:

```
rccNode = Config(
    executors=[
        IPyParallelExecutor(
            label='midway_ipp',
            provider=SlurmProvider(
                'broadwl',
                channel=LocalChannel(),
                #launcher=SrunLauncher(),
                init_blocks=1, # 1
                min_blocks=0, # 0
                max_blocks=10, # 10
                nodes_per_block=1, # 4
                tasks_per_node=28, # 12
                walltime="00:30:00",
                overrides="#SBATCH --mem-per-cpu=2000..."
            ),
        ),
    ],
    app_cache=False,
)
```

reidmcy:
![2018_08_11-23_23_53_count_recent_minutes_cdf](https://user-images.githubusercontent.com/4342916/44007166-7b875fc8-9e56-11e8-9584-9806f4a2bf3b.png)

This shows jobs run vs time, with a restart in the middle and an OLS. You can see that after the restart the line is much steeper, i.e. jobs are running faster.

I think the simplest way to reproduce the issue would be to run one of the test examples in a loop a million times, but I have not tested it.

MalcolmUnWang:
This is interesting. It appears that the run-vs-time line is pretty straight before you restart it. So I assume the running speed drops during the first a hundred thousand and keeps the slower speed before a restart? My first guess is the problem of memory-cache locality. Longer execution results in bigger memory usage, and results in the loss of locality of cache, which causes longer memory access time for the master to pull tasks. Can you show me your code and machine environment you were working on? @reidmcy 

@yadudoc @kylechard @annawoodard If that is a common case, I think a solution might come up with to solve this slow-down. Probably by refreshing where slows us down and not necessary to keep in memory every hundred thousand tasks.

reidmcy:
@MalcolmUnWang the slowdown is nonlinear, it's just not obvious in that picture. Here is a longer time window:
[...]

PR #583, opened 2018-10-22
unify launch plus callback unlocking

benclifford:


annawoodard:
I think this should fix #461

annawoodard:
[graph-size.pdf](https://github.com/Parsl/parsl/files/2503698/graph-size.pdf)

Here is a plot showing performance improvement due to this PR.

benclifford:
needs #580 merging first (or more specifically, #580 is part of this that i'd like to appear on master separately)

benclifford:
On digging into callback behaviour a bit more and looking at retry behaviour (which we aren't testing!), it looks like callbacks get lost.

I'll dig into this a little more.

annawoodard:
superseded by #600

PR #590, opened 2018-10-24
Worker lost #577

### LINKED ISSUE #577: Fix 'ManagerLost' error in htex
The following test fails on Midway:

```
from requests import get

import parsl
from parsl.providers.slurm.slurm import SlurmProvider
from parsl.launchers import SimpleLauncher

htex = Config(
    executors=[
        HighThroughputExecutor(
            label="midway_htex",
            worker_debug=True,
            cores_per_worker=1,
            public_ip=get('https://api.ipify.org').text,
            provider=SlurmProvider(
                'broadwl',
                init_blocks=1,
                min_blocks=1,
                nodes_per_block=1,
                tasks_per_node=28,
                overrides='source ~/.bashrc; source activate parsl',
                walltime= "00:30:00",
                launcher=SimpleLauncher()
            ),
        )
    ],
    strategy=None,
)

parsl.set_stream_logger()
parsl.load(htex)

@python_app
def sleep(seconds):
    import time
    time.sleep(seconds)

def call_sleep(size):
    tasks = [sleep(0) for _ in range(0, size)]
    for task in tasks:
        task.result()

call_sleep(10000)
```

yadudoc:
Support for result batching and fixes #577

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -3,9 +3,9 @@
 
 from concurrent.futures import Future
 import logging
-import uuid
 import threading
 import queue
+import pickle
 from multiprocessing import Process, Queue
 
 from ipyparallel.serialize import pack_apply_message  # ,unpack_apply_message
@@ -132,6 +132,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
         self.tasks = {}
         self.cores_per_worker = cores_per_worker
 
+        self._task_counter = 0
         self.public_ip = public_ip
         self.worker_ports = worker_ports
         self.worker_port_range = worker_port_range
@@ -220,10 +221,14 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
         The `None` message is a die request.
         """
         logger.debug("[MTHREAD] queue management worker starting")
+        result_counter = 0
+
         while True:
             try:
-                msg = self.incoming_q.get(timeout=1)
-                logger.debug("[MTHREAD] get has returned")
+                msgs = self.incoming_q.get(timeout=1)
+                result_counter += len(msgs)
+                # logger.debug("[MTHREAD] get has returned {}".format(len(msgs)))
+                # result_counter += 1
 
             except queue.Empty as e:
                 logger.debug("[MTHREAD] queue empty")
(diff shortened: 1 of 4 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -331,7 +338,8 @@ def submit(self, func, *args, **kwargs):
         Returns:
               Future
         """
-        task_id = uuid.uuid4()
+        self._task_counter += 1
```
benclifford: Can `submit` be called from multiple threads? Because `+=` is maybe not threadsafe: https://stackoverflow.com/questions/1717393/is-the-operator-thread-safe-in-python

If this is true, fix critical for functionality

parsl/executors/high_throughput/executor.py
```diff
@@ -240,32 +245,34 @@ def _queue_management_worker(self):
...
-                    try:
-                        tid = msg['task_id']
-                    except Exception as e:
-                        raise BadMessage("Message received does not contain 'task_id' field")
+                    for serialized_msg in msgs:
+                        try:
+                            msg = pickle.loads(serialized_msg)
+                            tid = msg['task_id']
+                        except Exception as e:
+                            raise BadMessage("Message received does not contain 'task_id' field")
```
benclifford: That's not the only reason we'd get an Exception? Most likely pickle.loads() would throw something too?

Test membership explicitly or fix the error message, for the sake of confused people later.

(fix not critical for functionality)

parsl/executors/high_throughput/interchange.py
```diff
@@ -127,7 +131,7 @@ def __init__(self,
 
         self.heartbeat_thresh = heartbeat_period * 2
 
-    def get_tasks(self, count, socks):
+    def _get_tasks(self, count, socks):
```
benclifford: please pick two better names rather than using the same name twice

(this is not critical for functionality)

parsl/executors/high_throughput/executor.py
```diff
@@ -220,10 +221,14 @@ def _queue_management_worker(self):
         The `None` message is a die request.
         """
         logger.debug("[MTHREAD] queue management worker starting")
+        result_counter = 0
```
benclifford: Is this used for anything?

parsl/executors/high_throughput/zmq_pipes.py
```diff
@@ -24,9 +27,15 @@ def __init__(self, ip_address, port_range):
         self.port = self.zmq_socket.bind_to_random_port("tcp://{}".format(ip_address),
                                                         min_port=port_range[0],
                                                         max_port=port_range[1])
+        self.poller = zmq.Poller()
+        self.poller.register(self.zmq_socket, zmq.POLLOUT)
 
     def put(self, message):
```
benclifford: This is (now more explicitly) blocking (for the zmq_socket to be ready), which causes HighThroughputExecutor.submit to be similarly blocking.

The documentation for concurrent.futures.Executor suggests (but is not explicit) that submit should be non-blocking, and return a future "quickly"; and other code possibly relies on that behaviour.

What are the conditions in which this code will block?

Should there be a logged warning if we are blocking here? For example, if the first poll does not return the socket as available?

parsl/executors/high_throughput/interchange.py
```diff
@@ -175,6 +179,48 @@ def get_tasks(self, count, socks):
...
+
+        Parameters:
+        -----------
+        kill_event : threading.Event
+              Event to let the thread know when it is time to die.
+        """
+        logger.info("TASK_PULL_THREAD] Starting")
+        task_counter = 0
+        poller = zmq.Poller()
+        poller.register(self.task_incoming, zmq.POLLIN)
```
benclifford: `poller` is unused?

parsl/executors/high_throughput/executor.py
yadudoc: We already get a lock in the DFK at the launch stage to avoid this.

parsl/executors/high_throughput/executor.py
yadudoc: True, pickling could fail. I'm adding a variant of the BadMessage exception for the unpickling failure. At this stage we are only unpickling {'task_id' : <>, 'result' : <serialized buffer>, 'exception' : <serialized buffer>} messages from the managers, so if this fails, we definitely got a bad message.

parsl/executors/high_throughput/interchange.py
yadudoc: Removed dead code.

parsl/executors/high_throughput/executor.py
yadudoc: Not anymore, was used for debug. Removed.

parsl/executors/high_throughput/zmq_pipes.py
yadudoc: Correct. This call is now a blocking call to ensure that the receiving end has space for an incoming message. If we do not do this, the receiver can overflow and the socket can go bad. There's a debug message that now explains why some tasks take a few more ms to go out.

But moving to this model means that we can still sustain 2000Tasks/s pump rate without having the receiver fall over.

parsl/executors/high_throughput/interchange.py
yadudoc: Changed in updated code.

PR #595, opened 2018-10-26
Default dfk #280

### LINKED ISSUE #280: Make parsl.load() load the thread pool config
We currently begin the user guide with a discussion of apps, futures, etc, but do not show a complete example (including the configuration) until significantly later in the documentation. I think it is reasonable to wait until later for a detailed discussion of config options, but in that case, we should consider removing the DFK from the examples and loading a default one (`ThreadPoolExecutor`) if none has been loaded.

ConnorPigg:
Closes #280 with a simple change. Also, I tagged on the change to the automatic workflow naming strategy used by monitoring.


parsl/tests/integration/test_parsl_load_default_config.py
```diff
@@ -0,0 +1,43 @@
+import parsl
```
benclifford: This test doesn't get run by the CI, but this is the sort of simple feature that should be tested there.

There's a pattern in `.travis.yml` in the root directory of the project, `parsl/tests/test*/test*` about which tests to run, and this test doesn't run because `/integration/` doesn't match `test*`.

I think you might need to make a test function called `test_*` inside this file, for pytest to discover how to run the test: https://docs.pytest.org/en/latest/goodpractices.html#test-discovery

annawoodard:
> I think it's better to split two different features into two PRs: two PRs that look really simple is easier to understand later on than one PR that changes two things at once.

Agreed-- it also de-couples disparate work, so that a problem with A doesn't prevent B from being merged.

PR #602, opened 2018-10-26
Update kube provider #478

### LINKED ISSUE #478: Update Kubernetes provider to new config system
Reported by @ryanchard.

We forgot to update the Kubernetes provider with the 0.6 changes.

ZhuozhaoLi:
Fix #478

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -10,7 +10,6 @@ from parsl.utils import RepresentationMixin
 
 try:
     from kubernetes import client, config
-    config.load_kube_config()
     _kubernetes_enabled = True
 except (ImportError, NameError, FileNotFoundError):
     _kubernetes_enabled = False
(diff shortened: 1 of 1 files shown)


ZhuozhaoLi:
> iku

Yes, the main change was tested on Petrelkube before. Today i added the new "overrides" keyword arg and tested on minikube.

parsl/providers/kubernetes/kube.py
```diff
@@ -1,86 +1,107 @@
...
-from parsl.providers.kubernetes.template import template_string
-from parsl.providers.error import OptionalModuleMissing
-
+from parsl.utils import RepresentationMixin
 
 try:
-    from kubernetes import client
-    from kubernetes import config as kube_config
+    from kubernetes import client, config
+    config.load_kube_config()
```
yadudoc: Do we need to load the kube config here ? This try catch will set _kube_enabled to false for 2 different reasons and the user won't know if there was just a problem with the load_kube_config.

ZhuozhaoLi: Moved `config.load_kube_config()` to init function in KubernetesProvider

ZhuozhaoLi:
Moved `config.load_kube_config()` to init function in KubernetesProvider.
load_kube_config in kubernetes lib will catch the config's exception.

Issue #313, opened 2018-06-05
Wait for all tasks to finish before shutting down the DFK


annawoodard:
We should not shut down the DFK until all tasks have finished. We could add a `wait` or `join` method which the user could call to wait until pending tasks are completed, although I am interested if someone could describe a use case for calling such a function anywhere except the end of the script, in a system where dependencies are taken care of automatically. I favor assuming that the user definitely wants all of their tasks to finish, so we automatically call such a function at the end of the script, and do not require the user to do so.

benclifford:
> although I am interested if someone could describe a use case for calling such a function anywhere except the end of the script, in a system where dependencies are taken care of automatically.

Where I have multiple terminal nodes that aren't all gathered together into a single combining app, and I want to (for example, interactively) keep my executors running (or some other context?), but I want to be able to know that the jobs I've launched have definitely output all their files, then either I contrive a join myself, or I use this DFK-provided join. Then a bit later, I can launch another set of tasks.

annawoodard:
This is related to #221.

yadudoc:
When the dfk.<wait/join> is called there is one tricky case to account for:
Any task launched on a threadPool, could potentially launch more tasks and so the list of tasks will expand while we iterate on it. Iterating on a list that expands asynchronously is generally unsafe. In the future we might even want to allow remote apps to launch tasks somehow.

A potentially interesting side-effect of having dfk.<wait/join> is that we could switch defer task launches until wait it called, at which point the dfk can do resource planning on the available DAG and then do smarter launches. We would be able to borrow some cool features from systems that do static planning on workflow DAGs

annawoodard:
I like @benclifford's implicit-optimized-scheduling proposal.

Optimizing scheduling is more complicated than simply providing a wait/join, so I'm going to split that into its own issue, so that we don't need to wait for the complicated thing before having the simple thing.

yadudoc:
@annawoodard  Might have a solution already.

Issue #222, opened 2018-04-23
End of workflow behaviour to discard some unexecuted work is surprising


benclifford:
When reaching the end of the main Python script of a workflow, `dfk.cleanup()` happens implicitly. This causes some unexecuted apps to be discarded.

I think less surprising behaviour is that when you reach the end of the main Python script, any work that has been asked for actually gets done before exit.

The present correct thing to do in a workflow, I think(?), is record which Apps you have launched, and wait for all of them at the end. I would like to not have to do that.

annawoodard:
Agreed. I think we would want to add a `dfk.wait()`-type function. If the process receives sigterm or sigkill, then we call `dfk.cleanup()` and don't ewait, otherwise we wait.

benclifford:
My current proposal is that at DFK cleanup:

i) puts parsl in a state where it refuses to create new AppFutures
ii) does a wait on every app future in the dfk.tasks structure

Every app known to exist will be waited for; code which subsequently tries to launch a new app will be refused (for example, getting an exception saying the DFK is shutting down)

benclifford:
crossref #313

benclifford:
That latter problem was fixing in PR #612.

PR #610, opened 2018-10-31
Wait for all tasks to complete during DFK cleanup

benclifford:
This partially addresses #222 #313, enough that this should be usable in practice.

There is a potential race here with tasks that are submitted
after DFK cleanup has begun:

  * a task on a local threaded executor might be running and
    submit such a task

  * I'm unclear if data stageout tasks appear as regular
    tasks

The second of those points is probably the most immediate
to be concerned about.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -714,6 +714,22 @@ class DataFlowKernel(object):
         if not self.cleanup_called:
             self.cleanup()
 
+    def cleanup_wait_for_all_tasks(self):
+        """Waits for all tasks in the task list to be completed, by waiting for their
+        AppFuture to be completed. This method will not necessarily wait for any tasks
+        added after cleanup has started (such as data stageout?)
+        """
+
+        logger.info("Waiting for all remaining tasks to complete")
+        for task_id in self.tasks:
+            # .exception() is a less exception throwing way of
+            # waiting for completion than .result()
+            fut = self.tasks[task_id]['exec_fu']
+            if not fut.done():
+                logger.debug("Waiting for task {} to complete".format(task_id))
+                fut.exception()
+        logger.info("All remaining tasks completed")
+
     def cleanup(self):
         """DataFlowKernel cleanup.
 
@@ -731,6 +747,8 @@ class DataFlowKernel(object):
         if self.cleanup_called:
             raise Exception("attempt to clean up DFK when it has already been cleaned-up")
         self.cleanup_called = True
+
+        self.cleanup_wait_for_all_tasks()
 
         self.log_task_states()
 


yadudoc:
Data stage-outs are launched as tasks, I believe after the parent task is done.

parsl/dataflow/dflow.py
```diff
@@ -714,6 +714,22 @@ def atexit_cleanup(self):
         if not self.cleanup_called:
             self.cleanup()
 
+    def cleanup_wait_for_all_tasks(self):
```
annawoodard: consider naming this `wait_for_all_tasks` since it's not doing any particular cleanup

parsl/dataflow/dflow.py
```diff
@@ -714,6 +714,22 @@ def atexit_cleanup(self):
...
+        """
+
+        logger.info("Waiting for all remaining tasks to complete")
+        for task_id in self.tasks:
+            # .exception() is a less exception throwing way of
+            # waiting for completion than .result()
+            fut = self.tasks[task_id]['exec_fu']
+            if not fut.done():
+                logger.debug("Waiting for task {} to complete".format(task_id))
+                fut.exception()
```
annawoodard: If a task raises an exception, will that be logged anywhere else? If not, maybe we should do so here.

yadudoc: We shouldn't switch to an eager fail behavior here, unless the user requested it. We should log the exception and wait for the next one.
```
try:
     fut.result()
except Exception as e:
     logger.exception("Task raised this :{}".format(e))
```

benclifford: It's not an eager fail as implemented in the PR at the moment - fut.exception() doesn't throw an exception even if the future has an exception rather than a real result. (I think)

annawoodard: I think, if we agree it makes sense to log an exception, we should go with Yadu's suggestion to use `result()` to avoid confusion about what `.exception()` does

benclifford: we don't really have a way to detect "lost" tasks at the moment (whatever that means? tasks that the workflow has never waited on?)

we can get a list of every task ever, and a list of every task that hasn't completed at DFK shutdown, but neither of those is a particularly good candidate for outputting exceptions from - anything that failed before that won't have an exception logged here.

So maybe we should rely on the existing logging (or not logging) of exceptions that happen if you don't ask for a result, so that it is consistent before/after shutdown?

annawoodard: Ben correctly points out that logging the error here doesn't make sense, it will be a random subset of tasks depending on whether it is done or not.

annawoodard: yep, agreed

PR #612, opened 2018-11-01
Make DFK shutdown wait for app futures, not exec futures

benclifford:
This was my original intended behaviour of #610

A task does not always have an exec_fu, and in some cases in
testing, the final wait was failing rather than waiting,
because it could not find an exec_fu to wait for.


benclifford:
The race condition I mentioned above seems to be caused by lazy fail exception problem that should be fixed by PR #614

PR #614, opened 2018-11-02
Deal with spurious callback exceptions.

benclifford:
Exceptions were appearing in strange places with lazyFail=False.

On investigation, two unexpected things were happening. Each of the two commits on this branch fixes one of those things:

* handle_update was throwing an exception in the case on a non-lazy fail, rather than `return`ing. In general, it doesn't make sense for a callback to throw an exception as there is no where for the exception to go. One commit on this branch replaces that explicit raise with a `return`.

EXCEPT:

* if a callback fires straight away when added to a future, and that callback throws an exception, then (sometimes?) the exception propagates to up through add_done_callback call and out into the code containing that. That gives a race condition where sometimes a failing task (if it fails fast enough) will cause an exception there, and sometimes not (if it fails too slowly); in the latter case, the exception disappears into the void.

The second commit on this branch puts explicit try/catch-log-discard blocks round places where handle_update is called, to give "exceptions in callback handles don't come out of add_done_callback calls" behaviour (as I thought was originally happening).

Either of these commits would be sufficient to solve the specific problem I was experiencing with lazy fail exceptions, but both together make things a bit more resilient in general.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -441,8 +441,14 @@ class DataFlowKernel(object):
         hit, memo_fu = self.memoizer.check_memo(task_id, self.tasks[task_id])
         if hit:
             logger.info("Reusing cached result for task {}".format(task_id))
-            self.handle_exec_update(task_id, memo_fu)
-            self.handle_app_update(task_id, memo_fu, memo_cbk=True)
+            try:
+                self.handle_exec_update(task_id, memo_fu)
+            except Exception as e:
+                logger.error("handle_exec_update raised an exception {} which will be ignored".format(e))
+            try:
+                self.handle_app_update(task_id, memo_fu, memo_cbk=True)
+            except Exception as e:
+                logger.error("handle_app_update raised an exception {} which will be ignored".format(e))
             return memo_fu
 
         executor_label = self.tasks[task_id]["executor"]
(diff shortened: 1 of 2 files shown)


benclifford:
this hung in  parsl/tests/test_checkpointing/test_regression_233.py  on python 3.6. Unclear to me if that is related to PR or something else.

benclifford:
I ran test_regression_233 test a bunch of times on my laptop. It hung once in many hundreds. I also ran master, and it hung 0 times in many hundreds.

My gut feeling is that that the test hung due to some other race condition, but other PRs exist that peturb our race condition ecosystem, so I think we should merge this (and them) and move on.

benclifford:
crossref #1257 discussion

hategan:
Perhaps another option would be to add a @callback decorator that ensures that exceptions thrown by the wrapped method are very clearly reported.

I think that deadlocks and failure to make progress are notoriously hard to troubleshoot. In an async system, like parsl, in which progress relies on the correct functioning of callbacks, failure of callbacks is a clear source of problems. Ensuring that such faults are reported and, perhaps, integrated with tests such that callback faults cause tests to fail, should be something we strive for.

benclifford:
I think that approach would also have value where threads/processes are launched and exceptions can fall out the top of those too - that has been a problem in parsl too.

Issue #103, opened 2018-02-28
Report running state via AppFutures and DataFutures


yadudoc:
Reported by @djf604

Currently we only have 3 states reported by future types:
1. Not done
2. Done
3. Done with Exception.

For long running applications, potentially with several stages it would be useful to report additional states. This should be queriable in non-blocking way from the Future types. For instance, we could 
support a set of reported state transitions like : `pending`, `launched`, `running`, `completed`, `failed`

benclifford:
I discussed this with a few people. My concern is that we don't have a clear idea on what the right state model to expose is at the moment. My proposal is to expose a function that returns some kind of state as a string that may vary over parsl versions, so that we can more easily explore what is useful to represent.

yadudoc:
Additional states need more discussion/design effort. Let's push this to 0.8.0.

benclifford:
Another thing I was wondering is what people want to do with more exposed/interesting state: if it is something like log it to a console / group it by identical states, then we could return a string and have great flexibility in how we proceed; if we're trying to do something with more structured data, there's a rathole of data structure design/backward compatibility constraints to fall down.

annawoodard:
A string sounds fine to me

benclifford:
When the monitoring DB is present, task information relevant to this might be more readily available now.

PR #624, opened 2018-11-06
Expose task state as a method of AppFuture

benclifford:
Partly addresses #103 - it exposes the task state
to end uses via the AppFuture interface; but the
task state is not particularly rich information.


yadudoc:
@benclifford and I had a chat about this and decided that it would be useful to have a broader discussion of what states might be useful to expose to the user that we already can tease out from the DFK. In addition we also need to consider the perf cost of fine grained state reporting from the new executors.

benclifford:
Another thing I was wondering is what people want to do with more exposed/interesting state: if it is something like log it to a console / group it by identical states, then we could return a string and have great flexibility in how we proceed; if we're trying to do something with more structured data, there's a rathole of data structure design/backward compatibility constraints to fall down.

benclifford:
I think this feature is worth revisiting soonish, in the context of the richer information now available (sometimes) in the monitoring DB. One view is that this call would give a view into the relevant part of the monitoring DB information.

yadudoc:
@benclifford What do we want to do about this. I'm okay with merging this, the code is fine. As covered in the discussion we do not yet have the richer state info that might be in the monitoring db available yet in the dfk side.

PR #628, opened 2018-11-06
Adding nbsphinx to test-requirements for readthedocs.

yadudoc:
Readthedocs will now switch to using test-requirements.txt instead of requirements.txt


annawoodard:
You might also need sphinx_rtd_theme

PR #640, opened 2018-11-09
rename overrides with scheduler_options and worker_init. fix #336

### LINKED ISSUE #336: Consider renaming 'overrides'
In our configuration, `overrides` sounds a bit like something that is going to be run _instead_ of something else, rather than something that is going to get prepended to the command. We should consider something like `prepend` or `prologue`.

ZhuozhaoLi:
* change the `overrides` in `/parsl/providers` directory
    * change the `overrides` of the example configs for docs in `/parsl/configs/` directory
    * change the `overrides` of the test configs in `/parsl/tests/configs` directory
    * fix #336


parsl/configs/beagle_single_node.py
```diff
@@ -34,7 +34,8 @@
                 init_blocks=1,
                 max_blocks=1,
                 launcher=AprunLauncher,
-                overrides='OVERRIDES',     # Please replace OVERRIDES with your overrides
+                scheduler_options='SCHEDULER_OPTIONS',     # Please replace SCHEDULER_OPTIONS with your scheduler_options
```
annawoodard: Sorry I didn't catch this in the previous PR, but maybe as an example this should just be an empty string? So that the example would be closer to something that works out-of-the-box, since there may be cases where one does not need to add their own scheduler options. Same with`worker_init`. I don't have a strong opinion about this though, feel free to ignore.

parsl/configs/cori_ipp_multinode.py
```diff
@@ -36,7 +36,8 @@
                 tasks_per_node=2,
                 init_blocks=1,
                 max_blocks=1,
-                overrides='OVERRIDES',     # Please replace OVERRIDES with your overrides
+                scheduler_options='SCHEDULER_OPTIONS',     # Please replace SCHEDULER_OPTIONS with your scheduler_options
+                worker_init='WORKER_INIT',     # Please replace WORKER_INIT with your worker_init
                 launcher=SrunLauncher,
```
annawoodard: I know this isn't part of your PR, but `SrunLauncher` should be `SrunLauncher()`, no?

parsl/configs/osg_ipp_multinode.py
```diff
@@ -22,8 +22,8 @@
                 tasks_per_node=1,
                 init_blocks=4,
                 max_blocks=4,
-                overrides='Requirements = OSGVO_OS_STRING == "RHEL 6" && Arch == "X86_64" &&  HAS_MODULES == True',
-                worker_setup='WORKER_SETUP',     # Please replace WORKER_SETUP with your worker setup
+                scheduler_options='Requirements = OSGVO_OS_STRING == "RHEL 6" && Arch == "X86_64" &&  HAS_MODULES == True',
+                worker_init='WORKER_init',     # Please replace WORKER_INIT with your worker_init
```
annawoodard: either `worker_init='WORKER_INIT'` or (my preference) `worker_init=''`

parsl/providers/cobalt/cobalt.py
```diff
@@ -48,8 +48,10 @@ class CobaltProvider(ClusterProvider, RepresentationMixin):
         Account that the job will be charged against.
     queue : str
         Torque queue to request blocks from.
-    overrides : str
-        String to append to the Torque submit script on the scheduler.
+    scheduler_options : str
+        String to prepend to the submit script to the scheduler.
+    worker_init : str
+        Command to be run before running a task, such as 'module load Anaconda; source activate env'.
```
annawoodard: We have two different docstrings, but really these two args are going to the same place and doing the same thing (being prepended to the submit script to the scheduler). I am sure there was a good reason for splitting it up like this, but I can't recall now what it was. I've reached my quota of opinining on this issue, so I recuse myself from promoting any particular course of action, while noting that I find this choice a bit confusing. In other words.... ignore this paragraph, sorry for the noise.

Also, "Command to be run before running a task" might make people think this executed once per task, while it is really once per batch script (even though it is technically before running a task). Maybe "Command to be run before starting a worker".

PR #645, opened 2018-11-12
[NOT FOR MERGE] Fast fail DESIGN DISCUSSION / HACKERY

benclifford:
This passes tests on my laptop, but I haven't tested to see if the
abort functionality works at all.

There is a _fast_abort flag in the DFK, and a method fast_abort() which sets it (and may trigger other abort behaviour).

It is set in two places: when non-lazy-fails trigger, and when ctrl-C is pressed.

It should do various things to make a shutdown happen "quite fast" but not necessarily instantly - sane cleanup must still happen. Exactly what needs to happen, and how, is open for discussion.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -8,6 +8,7 @@ import random
 import threading
 import inspect
 import sys
+import signal
 import multiprocessing
 import time
 
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -151,6 +151,8 @@ class DataFlowKernel(object):
         self._checkpoint_timer = None
         self.checkpoint_mode = config.checkpoint_mode
 
+        self._fast_abort = False
+
         data_manager = DataManager(max_threads=config.data_management_max_threads, executors=config.executors)
         self.executors = {e.label: e for e in config.executors + [data_manager]}
         for executor in self.executors.values():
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -786,14 +865,22 @@ def cleanup(self):
...
 
         self.log_task_states()
 
         # Checkpointing takes priority over the rest of the tasks
         # checkpoint if any valid checkpoint method is specified
         if self.checkpoint_mode is not None:
-            self.checkpoint()
+
+            # controversial in benc's mind
+            if not self._fast_abort:
```
annawoodard: This seems reasonable to me

parsl/dataflow/dflow.py
```diff
@@ -190,6 +192,71 @@ def __init__(self, config=Config()):
...
+           involving this DFK should throw an exception if they cannot
+           be completed immediately (for example, all unset AppFutures
+           should get an exception status immediately)
+        """
+        logger.warn("Fast-abort activated")
+        self._fast_abort=True
+        # what needs to happen here in addition to the rest of the
+        # DFK failing fast as it notices?
+
+    def set_termination_handler(self):
```
annawoodard: Where is the idea to call this from?

benclifford: For now, at the start, roughtly where you might call set stream logger.

Adding signal handlers by default is a bit of an awkward thing to do for a library (in the same way that setting logging by default is).

In the case of a standalone workflow, it might make sense to do this automatically at the start of DFK initialisation, but I'm not sure when running in a notebook.

annawoodard: Yikes. I see your point, but I also think this could be a bit obtuse to explain to new users, and something people could easily forget to add to their scripts. I'd estimate that in 99% of the cases we encounter (in standalone scripts, at least), people will have wanted to set these signal handlers. Maybe, if we're not willing to add signal handlers by default, we need to undo the automatic call to `wait_for_current_tasks` we added in #610, and instead rely on users to call that themselves. We could add a shortcut `parsl.wait()` or similar so the user wouldn't have to interact with the DFK directly.

yadudoc:
Ben to close this one.

Issue #558, opened 2018-10-11
Implement heuristics for determining IP with exex/htex


annawoodard:
Issue #466 lists a number of heuristics which might be implemented, and their upsides/downsides. Discussion and further heuristics should be added there.

benclifford:
crossref #466

Issue #466, opened 2018-08-15
heuristics for public_ip setting - bad user experience when using multi-login-node clusters


benclifford:
At present our recommendation is roughly to hardcode a submit machine IP in configurations, for ipp engines to call back to a submitting ipp controller.

This is a terrible user experience when a user regularly logs into one of several login machines, as is pretty common on large clusters/machines, and so has a frequently changing callback address.

There are a bunch of heuristics we could use, none of which is perfect in all situations - in some cases the suitable callback IP isn't even configured on the submitting machine (!)

My recommendation then is to provide one or more helpers that can be used (perhaps by default?) but easily overridden.

heuristics which have worked in some places:

* `/sbin/ip route get 8.8.8.8 | awk '{print $NF;exit}'`    - [midway] - relies on route to 8.8.8.8 going out of an address that is contactable from everywhere we might want to (not always the case, I think, in the case of a headnode and worker nodes with a private network, where worker nodes don't have a route out of the network)

* `hostname`    - this actually returns a hostname not an IP, but it turns out the public_ip field does not need to be an IP. [cori] - needs that returned hostname to resolve sensibly everywhere we might try to resolve it

*  `public_ip=get('https://api.ipify.org').text` - https://stackoverflow.com/questions/2311510/getting-a-machines-external-ip-address-with-python  - address returned might not be address of machine, if going through a NAT. This might be a blessing (eg in some circumstances on EC2) or a curse (if the things we're trying to talk to is this side of the NAT)

annawoodard:
agreed!

benclifford:
#646 should resolve this

PR #646, opened 2018-11-13
Implement submit host address heuristics (#558, #466)

benclifford:
This PR implements the three heuristics referred to in #466 without giving preference to any of them.

The term 'address' is used instead of IP or especially not 'public ip' because that isn't always what is wanted/needed.

I tested this on cori.nersc.gov against HighThroughputExecutor:

* running parsl on a login node, all three methods work

* running parsl on a worker node, `address_by_hostname` and `address_by_route` work, and `address_by_query` gives the wrong address which causes silent hangs. The address returned in the `address_by_query` case is an internet-facing address of a NAT box at NERSC.

I haven't thought much about the right way to document this, or to test it.

Example usage is where you have a `public_ip=` setting, say (for example) `public_ip=address_by_query()`


annawoodard:
Fixes #466, fixes #558

PR #650, opened 2018-11-20
Remove implicit wait for outstanding tasks at DFK shutdown.

benclifford:
In practice, this turned out to be awkward when, for example, ctrl-C
was pressed and parsl spent a long time waiting (or sometimes hung).

This commit intends that the same wait code is executed explicitly
at the place in the workflow code that the wait is desired (usually
the end). Ctrl-C and other abnormal exits will not cause this wait
to happen.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -786,8 +786,6 @@ class DataFlowKernel(object):
             raise Exception("attempt to clean up DFK when it has already been cleaned-up")
         self.cleanup_called = True
 
-        self.wait_for_current_tasks()
-
         self.log_task_states()
 
         # Checkpointing takes priority over the rest of the tasks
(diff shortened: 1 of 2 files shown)


parsl/dataflow/dflow.py
```diff
@@ -1015,6 +1013,10 @@ def load(cls, config=None):
 
         return cls._dfk
 
+    @classmethod
```
annawoodard: Should we add a shortcut `wait_for_current_tasks = DataFlowKernelLoader.wait_for_current_tasks` in `__init__.py`?

annawoodard:
Also, as @benclifford points out, this needs a bit of documentation.

parsl/dataflow/dflow.py
yadudoc: Done

PR #651, opened 2018-11-20
deal with exception stacktraces better - needs tidyup

benclifford:


benclifford:
I've pulled the RemoteException part of this out into a separate tidier PR #662

PR #652, opened 2018-11-21
Fix broken test runner

benclifford:
The `for` loop in travis.yml affected by this PR was buggy, I think.

Previous to this PR, the loop would terminate silently at the first test that ended with exit code 5, which is `parsl/tests/test_flowcontrol/test_doc_config.py`

I think this is because the value of `$?`, which is compared with the number 5, is not the exit code expected (the pytest exit code), but is the exit code of the last thing executed (which is the command `[[ "$?" != 0 ]]`).

This PR introduces a new named variable to store the desired $? value, and uses that in future comparisons.

It looks like this broke in PR #525.

With that change in place, several tests that are now run fail; this PR puts a skip annotation on those  so that this PR can at least get all the other tests running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ script:
     - export PARSL_TESTING="true"
     - pip install -r test-requirements.txt
     - flake8 parsl/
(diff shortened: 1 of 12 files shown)


annawoodard:
(The approval is obviously pending skipping the rest of the broken tests. We already have a 'fix broken tests' issue.)

benclifford:
@annawoodard  please re-approve and merge

Issue #660, opened 2018-11-29
Slurm provider sets `ntasks-per-node=1` for htex


annawoodard:
Currently the user needs to remember to set `worker_init='#SBATCH --exclusive'` if they are using htex with the slurm provider (which isn't documented). We should deal with this under-the-hood.

PR #661, opened 2018-11-29
Adding slurm exclusive kwarg

yadudoc:
This PR adds a new KWarg "exclusive" to the SlurmProvider. Htex and Exex assume nodes are exclusive and that assumption is exposed more clearly via setting the default to True.

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -59,6 +59,9 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         String to prepend to the #SBATCH blocks in the submit script to the scheduler.
     worker_init : str
         Command to be run before starting a worker, such as 'module load Anaconda; source activate env'.
+    exclusive : bool (Default = True)
+        Requests nodes which are not shared with other running jobs. Executors such as HighThroughputExecutor
+        and ExtremeScaleExecutor assume that the nodes are exclusive.
     launcher : Launcher
         Launcher for this provider. Possible launchers include
         :class:`~parsl.launchers.SingleNodeLauncher` (the default),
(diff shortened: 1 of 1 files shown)


annawoodard:
I'm worried this is a bit too aggressive. The point of setting ntasks-per-node in the submit script was just to make sure not all processes were restricted to a single core. I prefer to just set `ntasks-per-node` in the submit script appropriately rather than always default to exclusive (which can drastically increase queue times, there may be cases where you don't mind sharing even though you want one manager per node). For ipp `ntasks-per-node=tasks_per_node`, for htex `ntasks-per-node=cores_per_node / cores_per_worker + 1` (for the manager), no?

annawoodard:
Note this PR addresses #660

yadudoc:
The general assumption is that we provision the whole node and do the partitioning. These changes make those assumptions explicit. Specifying the # of workers in this situation is straightforward. I think that this is the more common/preferred mode. 

The core issue is that HTEX assumes control of the whole node, so it starts workers with the right # of cores/mem. In this model -ntasks-per-node is not known ahead of time. So you'll have to set it manually every time partitions are switched.

parsl/providers/slurm/slurm.py
```diff
@@ -59,6 +59,9 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         String to prepend to the #SBATCH blocks in the submit script to the scheduler.
     worker_init : str
         Command to be run before starting a worker, such as 'module load Anaconda; source activate env'.
+    exclusive : bool (Default = True)
+        Requests nodes which are not shared with other running jobs. Executors such as HighThroughputExecutor
+        and ExtremeScaleExecutor assume that the nodes are exclusive.
```
ZhuozhaoLi: Based on our discussion, we should remove htex and exex from the docstring to avoid confusion.

Issue #530, opened 2018-10-05
Implement scaling in + MPIX


annawoodard:

yadudoc:
Currently htex's `cores_per_worker` option prevents the strategy component from determining the number of workers without additional information from the executors. In the current setup, we set tasks_per_node=1 due to lack of information, and this can result in an overestimation of resources required. We need to address this along with support for scaling_in.

benclifford:
cross-ref #2627 #2120 #3019

PR #663, opened 2018-11-30
Updating block semantics #557

### LINKED ISSUE #557: Re-examine block semantics
In the context of the new `HighThroughputExecutor` and `ExtremeScaleExecutor`, the concept of a block and how it maps to workers/nodes/cores/batch jobs is even more confusing than before.

From Yadu:
> For the exex 1 block = 1 mpi_worker_pool
> For the htex 1 block = potentially many nodes, each with 1 process_worker_pool
> so there is some difficulty with all the configs that specify tasks_per_node which should now be handled by the executor rather than the launcher.

We need to think carefully about this and make sure everything is consistent. I vote that we favor consistency/clarity and the new executors over backwards compatibility and addressing limitations of IPP.

yadudoc:
This PR moves the `tasks_per_node` kwarg from the provider definition to the executors so as to add flexibility in how worker counts can be specified by the user. The executors now call provider.submit() with the `tasks_per_node` kwarg at runtime.

Fixes #557 

This PR requires extensive testing and updates to example configs. I'm putting this up for review while testing is going on in parallel.

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -177,7 +177,15 @@ class Strategy(object):
             # FIXME probably more of this logic should be moved to the provider
             min_blocks = executor.provider.min_blocks
             max_blocks = executor.provider.max_blocks
-            tasks_per_node = executor.provider.tasks_per_node
+            if isinstance(executor, IPyParallelExecutor):
+                tasks_per_node = executor.workers_per_node
+            elif isinstance(executor, HighThroughputExecutor):
+                # This is probably wrong calculation, we need this to come from the executor
+                # since we can't know slots ahead of time.
+                tasks_per_node = 1
+            elif isinstance(executor, ExtremeScaleExecutor):
+                tasks_per_node = executor.ranks_per_node
+
             nodes_per_block = executor.provider.nodes_per_block
             parallelism = executor.provider.parallelism
 
--- a/parsl/executors/extreme_scale/executor.py
+++ b/parsl/executors/extreme_scale/executor.py
@@ -104,6 +104,9 @@ class ExtremeScaleExecutor(HighThroughputExecutor, RepresentationMixin):
 
     managed : Bool
         If this executor is managed by the DFK or externally handled.
+
+    ranks_per_node : int
+        Specify the ranks to be launched per node.
     """
 
     def __init__(self,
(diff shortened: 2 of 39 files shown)


parsl/tests/manual_tests/htex_local.py
```diff
@@ -16,10 +16,10 @@
                 channel=LocalChannel(),
                 init_blocks=1,
                 max_blocks=1,
-                tasks_per_node=1,  # For HighThroughputExecutor, this option should in most cases be 1
+                # tasks_per_node=1,  # For HighThroughputExecutor, this option should in most cases be 1
```
annawoodard: remove

parsl/dataflow/strategy.py
```diff
@@ -177,7 +177,15 @@ def _strategy_simple(self, tasks, *args, kind=None, **kwargs):
             # FIXME probably more of this logic should be moved to the provider
             min_blocks = executor.provider.min_blocks
             max_blocks = executor.provider.max_blocks
-            tasks_per_node = executor.provider.tasks_per_node
+            if isinstance(executor, IPyParallelExecutor):
+                tasks_per_node = executor.workers_per_node
+            elif isinstance(executor, HighThroughputExecutor):
```
annawoodard: Yikes, this is unfortunate, we should open an issue

yadudoc: Added a note on #530

parsl/tests/manual_tests/htex_local.py
yadudoc: Done

parsl/executors/extreme_scale/executor.py
```diff
@@ -137,9 +141,34 @@ def __init__(self,
...
+                                       cores_per_worker=self.cores_per_worker,
+                                       # This is here only to support the exex mpiexec call
+                                       ranks_per_node=self.ranks_per_node,
+                                       nodes_per_block=self.provider.nodes_per_block,
+                                       logdir="{}/{}".format(self.run_dir, self.label))
+        self.launch_cmd = l_cmd
+        logger.debug("Launch command: {}".format(self.launch_cmd))
+
+        self._scaling_enabled = self.provider.scaling_enabled
+        logger.debug("Starting HighThroughputExecutor with provider:\n%s", self.provider)
```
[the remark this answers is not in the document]
yadudoc: Fixed.

PR #673, opened 2018-12-03
Ad hoc cluster #670

### LINKED ISSUE #670: Running on a local cluster
I'd like to have parsl push jobs out to a small cluster we've got in the HEP group. About 120 cores, split over about 10 machines. There is no batch manager or similar.

The pilot model seemed appropriate. I was thinking to use the IPyParallelExecutor, and then have each on a seperate note.

However, I don't know what to use for a provider. LocalProvider seems like what I want - tho it needs to be on another machine. I tried to hack it by giving it an SSHChannel, but that didn't work.

Here is the configuration I've been playing with:

        remote_config = Config(
            executors=[
                IPyParallelExecutor(
                    label='remote_ipp',
                    provider=LocalProvider(
                        min_blocks=1,
                        init_blocks=1,
                        max_blocks=4,
                        tasks_per_node=2,
                        nodes_per_block=1,
                        parallelism=0.5,
                        channel=SSHChannel(hostname=m)
                    )
                ) for m in args.remote
            ]
        )
        parsl.load(remote_config)

If I remove the "channel=" line, btw, then my code runs just fine. All the jobs seem to execute without difficulty. Though the IPy executor generates a huge number of log messages!

I'm a totally newbee here - This is the first time I've used parsl. Very nice, btw, in about 2 hours I had something up and running on a single machine. Took me another 2 days to make it "robust". :-)

yadudoc:
The LocalProvider was not generalized enough to work with both the local and ssh channels.

This PR adds support for the LocalProvider + SSHChannel combination allowing users to launch pilots on ad-hoc clusters. There are still issues with cancelling remote workers due to issues with killing nested process trees safely. For now, ipp workers will timeout in a few minutes and cleanup automatically while HTEX workers linger around.

PR also adds example configs.

Fixes #670

### DIFF
--- a/parsl/tests/configs/htex_ad_hoc_cluster.py
+++ b/parsl/tests/configs/htex_ad_hoc_cluster.py
@@ -1,6 +1,5 @@
 from parsl.providers import LocalProvider
 from parsl.channels import SSHChannel
-from parsl.launchers import SingleNodeLauncher
 from parsl.executors import HighThroughputExecutor
 
 from parsl.config import Config
--- a/parsl/tests/configs/htex_ad_hoc_cluster.py
+++ b/parsl/tests/configs/htex_ad_hoc_cluster.py
@@ -1,0 +1,30 @@
+from parsl.providers import LocalProvider
+from parsl.channels import SSHChannel
+from parsl.launchers import SingleNodeLauncher
+from parsl.executors import HighThroughputExecutor
+
+from parsl.config import Config
+
+username = "yadunand"
+remotes = ['midway2-login1.rcc.uchicago.edu', 'midway2-login2.rcc.uchicago.edu']
+
+config = Config(
+    executors=[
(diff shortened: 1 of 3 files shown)


parsl/tests/configs/htex_ad_hoc_cluster.py
```diff
@@ -0,0 +1,29 @@
...
+
+from parsl.config import Config
+
+username = "yadunand"
+remotes = ['midway2-login1.rcc.uchicago.edu', 'midway2-login2.rcc.uchicago.edu']
+
+config = Config(
+    executors=[
+        HighThroughputExecutor(
+            label='remote_ipp_{}'.format(m),
```
annawoodard: ipp -> htex

annawoodard:
Thanks for testing this PR from @gordonwatts!

Issue #684, opened 2018-12-07
SSHChannel fails with host-based authentication


annawoodard:
Systems using host-based authentication (without a key or a password) fail with:
```
paramiko.ssh_exception.SSHException: No authentication methods available
```

Reported by @jmoon1506

PR #686, opened 2018-12-10
Host based ssh #684

annawoodard:
Tested by @jmoon1506.

### DIFF
--- a/parsl/channels/ssh/ssh.py
+++ b/parsl/channels/ssh/ssh.py
@@ -9,6 +9,10 @@ from parsl.utils import RepresentationMixin
 
 logger = logging.getLogger(__name__)
 
+class HostAuthSSHClient(paramiko.SSHClient):
+    def _auth(self, username, *args):
+        self._transport.auth_none(username)
+        return
 
 class SSHChannel(Channel, RepresentationMixin):
     ''' SSH persistent channel. This enables remote execution on sites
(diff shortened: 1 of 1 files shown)


benclifford:
If I merge this into my dev tree, `RepresentatioMixin` gets upset with `host_auth` not being stored in the object:

> AttributeError: class SSHChannel uses host_auth in the constructor, but does not define it as an attribute

annawoodard:
Great catch, thanks @benclifford.

parsl/channels/ssh/ssh.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, hostname, username=None, password=None, script_dir=None, envs
         self.password = password
         self.kwargs = kwargs
         self.script_dir = script_dir
+        self.host_auth = host_auth
 
         if host_auth:
```
ZhuozhaoLi: Should that be `if self.host_auth` then?

annawoodard: (shrug)

It is logically equivalent either way, I'm not sure if there's a reason to always do one over the other? I guess using the attribute as you suggest is better in case you move the code to a different method? We seem to mix elsewhere in the code.

yadudoc: @ZhuozhaoLi both ways are equivalent. Why don't we go with this unless there's something else we are missing.

PR #707, opened 2018-12-14
port #683 to slurm provider

benclifford:


benclifford:
tested with josephmoon on slack and it seems to work

PR #709, opened 2018-12-15
Add args/kwargs remote files as dependencies

### LINKED ISSUE #526: Stage in files passed as args and kwargs
In `DataFlowKernel._add_input_deps`, we only stage stuff in if it is in `inputs`, while we resolve any futures that are passed as args or kwargs. We should iterate over args and kwargs looking for files to stage in as well as `inputs`. In my opinion the only advantage of `inputs` is that it allows you to pass an arbitrary number of inputs at call time. Unless a user needs to fulfill that specific (and somewhat unusual) need, they should use args and kwargs instead, because they are more readable and easily documented.

ConnorPigg:
This looks through ```args``` and ```kwargs``` and stages in remote files in the manner that ```inputs``` list was processed.

Closes #526


benclifford:
With the test I just committed, I get this exception testing stage-in with a file in `args`. @ConnorPigg 

```
        for idx, f in enumerate(args):
>               args[idx] = f.stage_in(executor)
E               TypeError: 'tuple' object does not support item assignment
```

benclifford:
last commit failure was due to recent CI changes that are fixed on recent `master` - so I've merge in master to see what happens now.

PR #712, opened 2018-12-18
Bind DFKs later than usual

benclifford:


parsl/app/python.py
```diff
@@ -39,13 +39,18 @@ def __call__(self, *args, **kwargs):
                    App_fut
 
         """
+        dfk = self.data_flow_kernel
```
yadudoc: This line is not necessary, because the if block defines it later

benclifford:
fixed per @yadudoc 's comment. requires new review.

PR #715, opened 2018-12-21
Add GSS-API support

jmoon1506:
Needed to make this change to get SSHChannel working on a remote cluster with Kerberos/GSS-API authentication.

### DIFF
--- a/parsl/channels/ssh/ssh.py
+++ b/parsl/channels/ssh/ssh.py
@@ -26,7 +26,7 @@ class SSHChannel(Channel, RepresentationMixin):
 
     '''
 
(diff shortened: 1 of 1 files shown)


jmoon1506:
Oops found a bug.

benclifford:
whats a way to test this? I'm guessing something like:
- get some system with kerberos
- run `kinit`
- see that you cannot submit parsl jobs but can ssh to other hosts
- turn on `gssapi_auth=True`
- see that you can submit parsl jobs
?

jmoon1506:
that is correct, here's the setup script I use:

python3 -m pip install git+https://github.com/jmoon1506/parsl.git --user
python3 -m pip install python-gssapi --user
kinit

PR #716, opened 2018-12-21
Add GSS-API support

jmoon1506:
Needed to make this change to get SSHChannel working on a remote cluster with Kerberos/GSS-API authentication.

To test on a system with Kerberos:

python3 -m pip install git+https://github.com/jmoon1506/parsl.git --user
python3 -m pip install python-gssapi --user
kinit

### DIFF
--- a/parsl/channels/ssh/ssh.py
+++ b/parsl/channels/ssh/ssh.py
@@ -26,7 +26,7 @@ class SSHChannel(Channel, RepresentationMixin):
 
     '''
 
-    def __init__(self, hostname, username=None, password=None, script_dir=None, envs=None, skip_auth=False, **kwargs):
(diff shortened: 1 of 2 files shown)


jmoon1506:
Sorry about the duplicate PR - I rebased the commit so there's no way back to the other one.

benclifford:
That new python-gssapi dependency should probably be added to the packaging - see `extras_require` in `setup.py` for the way other optional dependencies are handled.

jmoon1506:
Okay, I'll write a new commit adding that

benclifford:
This failed one of the CI tests with a git problem (rather than a problem specifically with your code) so I hit restart.
```
$ cd Parsl/parsl
$ git fetch origin +refs/pull/716/merge:
fatal: Couldn't find remote ref refs/pull/716/merge
```

PR #725, opened 2019-01-03
HTEX error suppression

yadudoc:
The branch is poorly named. Here are the changes that are included in this PR:

* `suppress_failure` option added to Interchange that ignores *some* errors.
    When enabled, a message from a worker deemed lost is simply allowed to continue operation.
* `max_workers` option caps the # of workers launched by the managers on the remote side. Primarily for debugging.
* Fixes to operators (^ replaced with **) to make larger queues as expected.

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -106,6 +106,12 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
         cores to be assigned to each worker. Oversubscription is possible
         by setting cores_per_worker < 1.0. Default=1
 
+    max_workers : int
+        Caps the number of workers launched by the manager. Default: 99999
+
+    suppress_failure : Bool
+        If set, the interchange will suppress failures rather than terminate early. Default: False
+
     heartbeat_threshold : int
         Seconds since the last message from the counterpart in the communication pair:
         (interchange, manager) after which the counterpart is assumed to be un-available. Default:120s
@@ -127,8 +133,10 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
                  working_dir=None,
                  worker_debug=False,
                  cores_per_worker=1.0,
+                 max_workers=99999,
                  heartbeat_threshold=120,
                  heartbeat_period=30,
+                 suppress_failure=False,
                  managed=True):
 
         logger.debug("Initializing HighThroughputExecutor")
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -109,6 +109,12 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
         cores to be assigned to each worker. Oversubscription is possible
         by setting cores_per_worker < 1.0. Default=1
 
+    max_workers : int
+        Caps the number of workers launched by the manager. Default: 99999
```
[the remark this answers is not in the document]
yadudoc: Python doesn't have a int infinity. I can do a float inf which should work in our case.

parsl/executors/high_throughput/interchange.py
```diff
@@ -43,6 +43,19 @@ def __repr__(self):
         return "Task failure due to loss of worker {}".format(self.worker_id)
 
 
+class BadRegistration(Exception):
+    ''' A new Manager tried to join the executor with a BadRegistration message
+    '''
+    def __init__(self, worker_id, critical=False):
+        self.worker_id = worker_id
+        self.tstamp = time.time()
+        self.handled = "critical" if critical is True else "suppressed"
```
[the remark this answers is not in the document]
yadudoc: It is redundant, but it was clearer to me, when I wrote it. Will remove

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -444,6 +452,8 @@ def set_stream_logger(name='parsl', level=logging.DEBUG, format_string=None):
                         help="Number of cores assigned to each worker process. Default=1.0")
     parser.add_argument("-t", "--task_url", required=True,
                         help="REQUIRED: ZMQ url for receiving tasks")
+    parser.add_argument("--max_workers", default=float(int),
```
ZhuozhaoLi: 1. Should be `default=float('inf')`.
2. But even with `float('inf')`, it returns error. On line 490, `int(max_workers)` would return error like `cannot convert float infinity to int`. I guess default to a large enough number is enough though it is not ideal?

benclifford: Either:   use `None` as a default and check that appropriately later on;   or make this a required argument by adding `required = True`, at which point the default value becomes (I think) irrelevant.

Using arbitrary magic large numbers makes me sad.

yadudoc: I believe this should be resolved in code now. By default the user specifies resources/cores per worker, and max_workers is available to limit the # of workers launched. It defaults to float('inf') and when specified an int is expected.

PR #750, opened 2019-01-28
Fix stage-in return bug (for data manager tasks) introduced in PR709

benclifford:


benclifford:
... probably because  `https://testbed.petrel.host/test/public/unsorted.txt'`, used in the failing test, is broken at the moment and master tests are also failing for me.

PR #752, opened 2019-01-29
Remove vestigial FlowControl code for clarity

benclifford:
This code is apparently unused.


parsl/dataflow/dflow.py
```diff
@@ -182,11 +182,6 @@ def __init__(self, config=Config()):
                 logger.error("invalid checkpoint_period provided:{0} expected HH:MM:SS".format(config.checkpoint_period))
                 self._checkpoint_timer = Timer(self.checkpoint, interval=(30 * 60))
 
-        if any([x.managed for x in config.executors]):
```
yadudoc: Hey Ben, I believe we start a flowcontrol thread if any of the executors passed in are managed by Parsl. The flowcontrol thread is what i think calls the strategy piece at some interval.

benclifford:
Hm. I think you are right.

yadudoc:
I don't think we can remove this without redoing code touching strategy/flowcontrol. So let's scrap this PR ?

benclifford:
yep, i agree

Issue #755, opened 2019-01-31
TorqueProvider timeout with qstat and qsub commands


macintoshpie:
I've experienced timeouts by the TorqueProvider on NSCC. Specifically, it's timing out when executing the `qstat` [command](https://github.com/Parsl/parsl/blob/fa1977937cde63f5fb3f6463affd5c116410b235/parsl/providers/torque/torque.py#L114) in and `qsub` [command](https://github.com/Parsl/parsl/blob/fa1977937cde63f5fb3f6463affd5c116410b235/parsl/providers/torque/torque.py#L204) (haven't checked `qdel` in `cancel()`

Probably need to remove the wait time from these calls to allow use of `cmd_timeout` property of ClusterProvider.

benclifford:
I've experienced issues before with cluster providers where similar status commands took a long time rather than hanging, and did some similar sounding timeout hacking.

I'd like to just be clear if that is the situation that you are experiencing? For example, if you run qstat in a shell, how many seconds does it take to run?

yadudoc:
Fixed by #768

PR #768, opened 2019-02-06
refactor(torque.py): remove explicit time out values for execute_wait

macintoshpie:
Remove wait times from torque provider, thus using the base wait time.
Addresses issue #755

### DIFF
--- a/parsl/providers/torque/torque.py
+++ b/parsl/providers/torque/torque.py
@@ -77,7 +77,7 @@ class TorqueProvider(ClusterProvider, RepresentationMixin):
                  parallelism=1,
                  launcher=AprunLauncher(),
                  walltime="00:20:00",
-                 cmd_timeout=60):
+                 cmd_timeout=120):
         label = 'torque'
         super().__init__(label,
                          channel,
--- a/parsl/providers/torque/torque.py
+++ b/parsl/providers/torque/torque.py
@@ -76,7 +76,8 @@ class TorqueProvider(ClusterProvider, RepresentationMixin):
                  max_blocks=100,
                  parallelism=1,
                  launcher=AprunLauncher(),
-                 walltime="00:20:00"):
+                 walltime="00:20:00",
+                 cmd_timeout=10):
         label = 'torque'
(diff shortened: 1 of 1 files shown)


benclifford:
`execute_wait` doesn't seem to have a very clearly defined calling interface, which makes me think this change will not work with all channels - walltime doesn't seem to be optional on `local`.

I'm not sure which channels @macintoshpie tested this with - can you comment?

```
parsl/channels/base.py:    def execute_wait(self, cmd, walltime, envs={}, *args, **kwargs):
parsl/channels/local/local.py:    def execute_wait(self, cmd, walltime, envs={}):
parsl/channels/ssh/ssh.py:    def execute_wait(self, cmd, walltime=2, envs={}):
```

It probably makes sense to say that walltime can be optional as part of the definition of the base channel definition, and modify `local` accordingly.

benclifford:
I'll make a separate PR to tidy up the execute_wait interface, and then I think this PR will fit cleanly on top of that.

macintoshpie:
Yes, sorry, I hadn't tested everything yet. I had to update the call to super(). I followed the same usage/patters as in the slurm provider and it's working for me now.

parsl/providers/torque/torque.py
```diff
@@ -76,7 +76,8 @@ def __init__(self,
                  max_blocks=100,
                  parallelism=1,
                  launcher=AprunLauncher(),
-                 walltime="00:20:00"):
+                 walltime="00:20:00",
+                 cmd_timeout=10):
```
benclifford: my gut says this should be more like 30 seconds or a minute: the aim is to catch failures where the commands hang for a long long time, and I've definitely seem some queue systems take a very long time.

macintoshpie: I agree, I actually just had to update my instance's to 60 seconds to not time out

benclifford: in that case, make it even higher than 60 seconds by default -- the default (in my opinion) should be a decent amount (twice?) the time people are experiencing on reasonable systems.

PR #771, opened 2019-02-11
Add Workqueue executor

AndrewLitteken:
This is the initial barebones implementation of the WorkQueue executor.  It does not have any sort of project-name or project-password or log functionality set up at present.

The general idea is that it creates a shared task dictionary to relate Parsl tasks to their future, and a queue to send tasks to the worker.

At executor start, a thread is started to handle the WorkQueue operations, where the executor sends tasks through the queue to this thread.  The python executables are pickled to a file, and this file is sent along with the task to the worker to be executed with the worker script (`work_queue_worker.py`) that is sent along with each task.

The worker thread monitors the queue, and when it takes values from the queue it submits them, and then waits with a threshold of five seconds for a task before repeating.  When a task is completed, it unpickles the result, and updates the futures that were stored in the dictionary previously.

Some issues:
- How to identify the location of the worker script without absolute pathnames


AndrewLitteken:
To test this the I believe that the following would need to be done:
- Go to the [CCL Site](http://ccl.cse.nd.edu/software/workqueue/) to install Workqueue, make sure that the Python3 Version is installed.
- Make sure that the Workqueue library is in your python path
- For now there is a single absolute path included at line 55 of `parsl/executors/workqueue/executor.py`, that has the absolute path of the 
Workqueue worker code to send that would need to be changed.  This is something we need to work on generalizing.
- Import and use this config in your Parsl Script:
``` python
from parsl.executors import WorkQueueExecutor

config = Config(
    executors=[WorkQueueExecutor(port = your_port_here
        )]
)
```
- Once you are running the Parsl script, you can supply workers to WorkQueue on your local machine using: `work_queue_worker [options] server port`
- Provided there is a Condor Pool active you can try: `condor_submit_workers` but we haven't had the chance to test this yet
- There are additional cloud service (SGE and Torgue) tie-ins documented on the CCL software page

I think that that should be everything that needs to be changed as long as you are using this repository as the Parsl module.  It should be noted that we haven't tried installing with this new executor.

AndrewLitteken:
Hi @yadudoc I believe you should have permissions based on the permissions for the pull request. I think I'v fixed the linter issues, writing WorkQueue Logs and Parsl Logs are on our todo list as well.  I should be getting to it soon.

Where is it hanging? We have only tested on small examples.

AndrewLitteken:
@dthain

AndrewLitteken:
There is a new commit with updates for files with workqueue.  Here is a rundown with a more in depth explanation of what has been implemented so far

dthain:
I believe the intended link is missing.

AndrewLitteken:
Here is the presentation:
https://docs.google.com/presentation/d/1_JPssz0Kz1xsP0xhLIYF5EHfkEgQYUlKghi0ymTLhAA/edit?usp=sharing

parsl/data_provider/files.py
```diff
@@ -46,9 +46,13 @@ def __init__(self, url, dman=None, cache=False, caching_dir="."):
         self.scheme = parsed_url.scheme if parsed_url.scheme else 'file'
         self.netloc = parsed_url.netloc
         self.path = parsed_url.path
+        self.registered = False
```
benclifford: I'm pretty uncomfortable with `File` being modified to have what is pretty much executor-specific behaviour.

However I think we don't have a better abstraction at the moment so I can't really give any constructive direction here.  I'll try to formulate a more helpful opinion.

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,410 @@
...
+            function_data_loc = item["data_loc"]
+            function_result_loc = item["result_loc"]
+            function_result_loc_remote = function_result_loc.split("/")[-1]
+            function_data_loc_remote = function_data_loc.split("/")[-1]
+
+            input_files = item["input_files"]
+            output_files = item["output_files"]
+
+            # TODO Make this general
+            full_script_name = "/afs/crc.nd.edu/user/a/alitteke/parsl/parsl/executors/workqueue/workqueue_worker.py"
```
benclifford: The high throughput executor launches a script "process_worker_pool.py". It assumes that that script is in the current $PATH so can be launched without an explicit path - for example, we require users to have setup their virtualenv/conda environment in their job submission.

The install process makes that script available in the current path by placing it in the `script` block in `setup.py`.

Can your worker environment make the same assumption? If so, that might be a way to get rid of this path.

yadudoc: I've pushed fixes addressing this below.

yadudoc:
I've pushed some changes that fixes some superficial issues. I get logs from workqueue which is great, but I do still get `Workqueue task 1 failed with status -1` error message whenever I run apps. In the debug logs I see a few lines that don't make sense:
`2019/03/07 00:47:41.91 work_queue_python[18335] wq: Could not stat workqueue_worker.py: No such file or directory`
This happens despite workqueue_worker.py being available on the PATH.
There's another error message that looks serious, but is followed task status changes, so I'm unsure of the severity:
`Failed to send task 1 to worker borgmachine2 (127.0.0.1:34860).`

Full logs are available here: http://users.rcc.uchicago.edu/~yadunand/002.tar.gz

parsl/data_provider/files.py
```diff
@@ -46,9 +46,13 @@ def __init__(self, url, dman=None, cache=False, caching_dir="."):
...
+        self.shared = False
         self.filename = os.path.basename(self.path)
         self.dman = dman if dman else DataManager.get_data_manager()
         self.data_future = {}
+        self.task_path = None
+        self.in_task = False
```
benclifford: @AndrewLitteken:

There is a mechanism using `local_path` to allow a new worker-side path to be specified by a staging mechanism. I've made PR #808 which makes this usable with all file schemes.

It might be enough that you can use that instead of `task_path` and `in_task`: once you set `local_path`, then asking for `file.filepath` will return that overridden path, which is what you're doing here. I'm a little unsure if you need to separately track `in_task` here. The other staging mechanisms don't do so.

parsl/executors/workqueue/executor.py
dthain: I suggest the most foolproof way to go here is to have the WQ executor ship the script as an input file dependency from the Parsl install into the working directory of the task, then execute "./script".

yadudoc: We need to somehow guarantee that the parsl-client side and worker-side environments match. Otherwise we run into deserialization issues from python version mismatches, and missing library issues. If the assumption is that the worker-side env is a clean slate, we'll have to ship the entire env (conda tarball ?) or send in a script that will setup the env. In parsl we have a `worker_init` string that's used to execute an arbitrary string on the worker side to setup the worker env, will something like that work here ? Is there some option I can pass to the work_queue_worker so that the env is inherited ?

AndrewLitteken: There is an `init_command` argument string for the executor that can be specified to run something before `work_queue_worker.py` is run.  You can also specify environmental variables for the worker with the `env` argument on the executor

AndrewLitteken:
@yadudoc I can't find that line you are referencing where it can't find the worker.  Also, the following command is only failing because it couldn't satisfy all the dependences to send the task, so that is expected based on the previous error.

AndrewLitteken:
@yadudoc I’ve made some changes to the worker thread finding the script on the path by using `shutil.which`, and am now using the branch that @benclifford created to monitor the file paths.  This version should work if you have the same version of Parsl on both the master and worker machines, although there is no mechanism right now where Parsl is transferred.  Can you try this again and see if you run into any other problems?

yadudoc:
Wonderful! That did the trick. I'm not able to run a few different tests to completion.

parsl/data_provider/files.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, url, dman=None):
         self.filename = os.path.basename(self.path)
         self.dman = dman if dman else DataManager.get_data_manager()
         self.data_future = {}
+        self.is_a_parsl_file = True
```
AndrewLitteken: A few new things in this commit, we are working to remove any dependencies on outside libraries, including parsl, from the work queue worker since we do not want to assume that there is every machine has parsl installed on it.  Since in the runner we do not have access to the file, `isinstance(file_object, File)` does not work as expected on the runner.  This attribute is added so we can instead do `hasattr(file_object, "is_a_parsl_file)` in the easiest way.  We can add to this check by looking for other attributes to make sure that the object we are looking at is in fact a parsl file object. @nhazekam

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,438 @@
...
+            function_result_loc = item["result_loc"]
+            function_result_loc_remote = function_result_loc.split("/")[-1]
+            function_data_loc_remote = function_data_loc.split("/")[-1]
+
+            input_files = item["input_files"]
+            output_files = item["output_files"]
+
+            # TODO Make this general
+            # full_script_name = "/afs/crc.nd.edu/user/a/alitteke/parsl/parsl/executors/workqueue/workqueue_worker.py"
+            full_script_name = workqueue_worker.__file__
```
AndrewLitteken: In the current implementation of finding the script in the `bin` then using parsl to load the runner on the other side was another dependency that we were looking to remove.  These two lines of code allow us to find the file in the package, and then copy that file to the worker.

parsl/data_provider/files.py
benclifford: I'm fairly surprised that `File` deserialises at all if the parsl code isn't available - what kind of object do you get for file_object? (eg, what does `type(file_object)` return at the point you call hasattr(file_object, "is_a_parsl_file" ?)

AndrewLitteken: It returns <class parsl.data_provider.files.File>. From what we have found and read, pickling does store the properties of a class.

benclifford: so can you check the type that way rather than adding stuff into the `File` instances?

AndrewLitteken: We did try a couple different methods to check against the class.  But when we tried to check against that class with the full name, it required parsl to be present.  We also tried to send the file implementation with the worker script, the class was not resolved in the same way since we didn’t have the rest of parsl present.  We could check against the string returned by `type(file_object)` and just look for the name `parsl.data_provider.files.File` in the type name instead.

parsl/dataflow/dflow.py
```diff
@@ -479,17 +479,20 @@ def _add_input_deps(self, executor, args, kwargs):
...
-                inputs[idx] = f.stage_in(executor)
+            if isinstance(f, File):
+                if f.is_remote():
+                    inputs[idx] = f.stage_in(executor)
 
-        for kwarg, f in kwargs.items():
-            if isinstance(f, File) and f.is_remote():
-                kwargs[kwarg] = f.stage_in(executor)
+        for kwarg, potential_f in kwargs.items():
+            if isinstance(potential_f, File):
```
benclifford: What is this change from `and` to 2x `if` for?

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,438 @@
...
+            script_name = full_script_name.split("/")[-1]
+
+            # script_name = "workqueue_worker.py"
+            # full_script_name = shutil.which(script_name)
+            if full_script_name is None:
+                logger.error("Unable to find script {} on path, skipping task {}".format(script_name, parsl_id))
```
benclifford: Does this mean that the task is left hanging, not marked as failed at the parsl level?

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,438 @@
...
+
+        logger.debug("Creating Task {} with executable at: {}".format(task_id, function_data_file))
+        logger.debug("Creating Tasks {} with result to be found at: {}".format(task_id, function_result_file))
+
+        f = open(function_data_file, "wb")
+        fn_buf = pack_apply_message(func, args, kwargs,
+                                    buffer_threshold=1024 * 1024,
+                                    item_threshold=1024)
+        pickle.dump(fn_buf, f)
+        f.close
```
benclifford: Does that need `()` ?

parsl/executors/workqueue/executor.py
AndrewLitteken: I believe that it is here, this line might not be needed anymore yet either.  But I do need some guidance about how to handle task failure in Parsl

parsl/dataflow/dflow.py
AndrewLitteken: This is just an artifact of adding and removing code from the data flow manager, I'll change it make in the next commit

parsl/executors/workqueue/executor.py
benclifford: When you've got a successful result, you pass it back to parsl like this:
```
future.set_result(result)
```

If you want to indicate that the task has failed, instead set the future to an exception (the most python-like way):

```
future.set_exception(someexception)
```

or set the result using set_result to an instance of RemoteExceptionWrapper. This is a helper class that helps with some serialisation of exceptions from remote workers - so you might find it useful. Look in `executors/high_through/process_worker_pool.py` for an example of creating a RemoteExceptionWrapper.

Either of those methods for setting a failure will result in the exception you pass in popping out into the driving workflow script when future.result() is called.

AndrewLitteken:
The last commit is dealing with the fact that submissions to the executor were not happening concurrently with the waiting on the workqueue queue.  This was due to the fact that they were in the same process, but on different threads, so the workqueue wait was not relinquishing control of the thread.

Now, the submission and waiting on the workqueue queue happens in a different process, and has an incoming and outgoing queue to this process.  The incoming queue receives jobs from the executor, and the outgoing queue sends the serialized result to a collector thread running in the Parsl process where the results are deserialized and used to update the futures accordingly.

AndrewLitteken:
@yadudoc Could take a look to see if this is closer to the behavior you were expecting?

ZhuozhaoLi:
Hi @AndrewLitteken , I just tested with `test_scale.py` on the newest version. 

I think this looks like what we expect.

Maybe a small error: I encounter the exception below during dfk cleanup.
```
  File "/home/zhuozhao/.conda/envs/parsl-wq/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/zhuozhao/.conda/envs/parsl-wq/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/zhuozhao/.conda/envs/parsl-wq/lib/python3.6/site-packages/parsl-0.7.2-py3.6.egg/parsl/executors/workqueue/executor.py", line 251, in WorkQueueCollectorThread
    raise ExecutorError(executor, "Workqueue Submit Process is not alive")
parsl.executors.errors.ExecutorError: (<parsl.executors.workqueue.executor.WorkQueueExecutor object at 0x7f0146c3c0b8>, 'Workqueue Submit Process is not alive')
```

AndrewLitteken:
I'm trying to get this branch working on the Travis-CI, and I have a couple questions/problems:

1. What is the best/preferred way to install workqueue and cctools into the build? Right now I have a script that pulls the latest release from Github and adds it to the path, but there may be a better way.

2. On the current version, the htex tests are hanging, and I can't really figure out why.  I've tried to run them locally, and am having a problem there as well.  Any guidance here is appreciated.

benclifford:
That way of installing is probably ok.

Check you have the latest master merged with your branch - every now and then we make improvements in the reporting of errors.

Get a single test running locally with the thread executor and then try to get it running locay with htex. Running locally gives you more access to messing round with logging and the file system.  Call parsl.set_stream_logger() to get lots of logging output, and see if any interesting errors show up during your test run.

Ignoring the changes you made only in your executor, see what changes you made to the core of parsl - for example, start with parsl master, check the tests pass, then add those changes in one by one onto a separate branch and see if the tests pass. Consider them all as changes you're trying to justify adding to the parsl core unrelated to your executor and think how they would affect other code using those changes.

Paste interesting logs here.

AndrewLitteken:
I solved the problem for the htex tests, there was some path issue that prevented it from running.  However, now there is a problem when building that when I try to access the GitHub releases API to get the information about cctools, the Travis machine is sometimes rate limited.  This can be done by providing Travis with an access code,  which have to be encrypted through Travis.  I was trying to set the key to an environmental variable to do this as documented here: https://docs.travis-ci.com/user/environment-variables/#Encrypting-environment-variables, but I cannot since it cannot use secret information on a pull request from a branch in a different repository.  

For now, I just removed the call to Github Releases and specified a pull from a URL specifying the location of the correct tarball we need based on a cctools version. But, is there a way we can use an access token generated by/owned Parsl for this so it can be used as a setting in Travis for the builds?

AndrewLitteken:
Upon further discussion it's probably better to just use the current method since then the specified version can be used with confidence that there are no bugs that interfere with Parsl.  So the API key is probably not necessary.

AndrewLitteken:
I think all the parts to this pull request should be in.

benclifford:
I've previously commented on being not entirely happy with how files were handled in this PR, but that I didn't have a better story.

Now, I have a better story - which is PR #1123 (and many related PRs past and future) giving a more expressive API for data staging.

I think when both this PR #771 and all of the #1123 related bits are merged, we should examine how the two interact - but I don't think it should stop either being merged now.

dthain:
@benclifford could you elaborate a little more on the direction you are going with data staging?  I would like to understand whether Parsl and Work Queue have a compatible idea here.

The basic assumptions of WQ are that each task needs an explicit subset of the files available at the master.  So, each WQ task must declare its inpus and output files.  The master and worker collude to deliver the inputs to each task and bring the outputs back.  Within a given workflow, files are immutable, so they can be cached at the worker once created.  And, the master may schedule tasks to workers based on data already there.

To put this together, the Work Queue executor is depending upon annotations from Parsl in order to construct tasks with accurate input and output files.

benclifford:
@dthain Generally parsl files can be arbitrary URLs, not just paths on the submit side file system - for example, in parsl 0.8 you can stage in files from `http:`, `ftp:` and `globus:` URL schemes as well as the `file:` which is a bit vague in implementation but basically means files on the submit file system.

What I've been working on is a more flexible API to let you plug in different mechanisms for performing that staging, so that different execution environments can stage things differently. For example, in the case of workers with network access and no shared filesystem such as AWS EC2 workers, we want to perform staging on the same worker that is running the job (which is not how 0.8 does it).

With workqueue there are questions along the lines of: do the workers have network access and could stage in http files directly? or should they be staged to the submit side file system and then workqueue deals with getting them to the worker? if we are having the parsl tasks perform the staging, then how do we make the workqueue executor ignore those files?

I think (/hope) that there are no deeply fundamental problems here and that it will be more a case of figuring out who is going to perform which kinds of staging and where and how, and then writing a few lines of python in the right places to make that happen.

I also hope that this will be driven by me getting a more useful staging test suite in place, so that it's easier to understand what is meant to work but doesn't.

dthain:
@benclifford I think we are more or less on the same page; it seems like the transfers could be done at either the master or the worker, depending on the situation.  Master transfers are probably the right thing for HPC, where you don't want to overload the external network, but worker transfers would work well between EC2 and S3.

To answer your questions:

**1 - do the workers have network access and could stage in http files directly?**

As you know, It's hard to tell in the general case, so we try to rely on the master-worker connection since it's already established.  WQ already has some capability to specify input files as urls, basically anything that could be handled by curl.  (Although I see know that it isn't documented, so we need to fix that.)
 
**2 - or should they be staged to the submit side file system and then workqueue deals with getting them to the worker?**

That could work too -- if parsl does the download into its own cache, then we can just pluck the file out of that cache and send it to the worker.

**3  - if we are having the parsl tasks perform the staging, then how do we make the workqueue executor ignore those files?**

If you don't tell WQ about it, then it won't stage it.

benclifford:
This PR makes a bunch of modifications to the test suite to add in `File()` constructors where in some cases we pass in strings to be interpreted as files.

Modifying the test suite in that way makes me a little uncomfortable: the test suite says what should work, and if this provider doesn't work with that, then something is broken.

I'd like to dig into this a bit in the context of #1016. I guess that change is because this WQ executor is expecting File objects the whole time and breaking otherwise.

In that case, I think (maybe @annawoodard also has an opinion) that maybe parsl core should be made to let you write out a string in an input/out (as in the un-modified tests), and have it convert to a File for you right away, so that explicitly writing out `File()` constructors is not necessary.

dthain:
@benclifford this was something we discussed by telecon in the spring.
The basic issue here is that WQ needs to know what files the
task will actually use, so it can transport them to the execution node.
Otherwise the task won't run correctly.

If I recall, the sense of the discussion was that the tests ought to declare
file dependencies for the sake of the DFK as well as the executors,
and it was an accidental success that those tests worked correctly
in the case of a shared filesystem. So, it made sense to for @AndrewLitteken 
to update the tests to declare files.

Of course, that's for you (the Parsl team) to decide what the language semantics are.
If you can obtain the same file dependency info by some other means, it's all the same
to WQ.

kylechard:
The plan was that all of our tests would be updated to explicitly define files not just for WQ but also for other Parsl executors.

benclifford:
I think by the time that list reaches the executor (be it WQ or something else) it should be `File` objects  - other pieces of file management code also need this rather than raw strings too;  but I'm interested in whether a user should have to explicitly put in that `File` constructor or whether parsl should be doing that for them if handed a string - so that they write a string but `executors/workqueue/` gets a `File`.

annawoodard:
> I think by the time that list reaches the executor (be it WQ or something else) it should be File objects - other pieces of file management code also need this rather than raw strings too; but I'm interested in whether a user should have to explicitly put in that File constructor or whether parsl should be doing that for them if handed a string - so that they write a string but executors/workqueue/ gets a File.

I think we should decide on this as part of #1016. For now I agree with Kyle we can just explicitly define Files in the tests; as part of testing for #1016, if we decide raw strings are acceptable, we can add in some appropriate tests then.

Onward, to merging! Has anyone had a look at the failing tests?

AndrewLitteken:
I believe it is only failing on this test: `parsl/tests/test_docs/test_from_slides.py`.  From the looks of it, it is probably failing because WorkQueue cannot find the file that is being passed into the `cat` function.

AndrewLitteken:
Additionally, there are some linting and flake8 errors when doing `from work_queue import *` since it creates undefined names.  I'm not sure how to get around that.

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,577 @@
...
+
+                            reason += "\nTrace:\n" + t.output
+
+                            logger.debug("Workqueue runner script failed for task {} because {}\n".format(parsl_tid, reason))
+
+                        else:
+                            reason = "Workqueue system failure\n"
+
+                        msg = {"tid": parsl_tid,
+                               "result_recieved": False,
```
benclifford: swap `ie` to give received

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,577 @@
...
+                future.set_exception(RemoteExceptionWrapper(*future_update))
+
+    logger.debug("Exiting Collector Thread")
+    return
+
+
+class WorkQueueExecutor(ParslExecutor):
+    """Define the strict interface for all Executor classes.
+
+    This is a metaclass that only enforces concrete implementations of
```
benclifford: get rid of most of this docstring - it's the docstring for `ParslExecutor`

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,577 @@
...
+        self.worker_output = see_worker_output
+        self.full = True
+        self.cancel_value = multiprocessing.Value('i', 1)
+
+        if self.project_password is not None and self.project_password_file is not None:
+            logger.debug("Password File and Password text specified for WorkQueue Executor, only Password Text will be used")
+            self.project_password_file = None
+        if self.project_password_file is not None:
+            if os.path.exists(self.project_password_file) is False:
+                logger.debug("Password File does not exist, no file used")
```
benclifford: this log should be at least warning level

parsl/executors/workqueue/executor.py
```diff
@@ -0,0 +1,577 @@
...
+        self.working_dir = working_dir
+        self.used_names = {}
+        self.shared_files = set()
+        self.registered_files = set()
+        self.worker_output = see_worker_output
+        self.full = True
+        self.cancel_value = multiprocessing.Value('i', 1)
+
+        if self.project_password is not None and self.project_password_file is not None:
+            logger.debug("Password File and Password text specified for WorkQueue Executor, only Password Text will be used")
```
benclifford: this log should be at least warning level

parsl/executors/workqueue/install-workqueue.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# wget -O /tmp/cctools_latest https://api.github.com/repos/cooperative-computing-lab/cctools/releases/latest?access_token=$CCTOOLS_KEY
```
benclifford: get rid of these commented-out lines

parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -0,0 +1,131 @@
...
+    if file_type_string is None:
+        file_type_string = "<class 'parsl.data_provider.files.File'>"
+    if type_desc == file_type_string:
+        if parsl_file_obj.filepath in mapping:
+            parsl_file_obj.local_path = mapping[parsl_file_obj.filepath]
+
+
+if __name__ == "__main__":
+    name = "parsl"
+    # logger = logging.getLogger(name)
```
benclifford: dead code

parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -0,0 +1,131 @@
...
+    user_ns.update({fname: f,
+                    argname: args,
+                    kwargname: kwargs,
+                    resultname: resultname})
+
+    code = "{0} = {1}(*{2}, **{3})".format(resultname, fname,
+                                           argname, kwargname)
+
+    try:
+        # logger.debug("[RUNNER] Executing: {0}".format(code))
```
benclifford: dead code

benclifford:
@AndrewLitteken (or @dthain?) can you review the top PR description and adjust it to whatever you would like the commit message for this PR to be. I already changed the PR title. Once this has built against the latest master, someone will hit the merge button and this PR will be closed finally!

dthain:
Title is fine with me, thanks.

benclifford:
This is the initial barebones implementation of the WorkQueue executor.

The general idea is that it creates a shared task dictionary to relate Parsl tasks to their future, and a queue to send tasks to the worker.

At executor start, a thread is started to handle the WorkQueue operations, where the executor sends tasks through the queue to this thread. The python executables are pickled to a file, and this file is sent along with the task to the worker to be executed with the worker script (work_queue_worker.py) that is sent along with each task.

The worker thread monitors the queue, and when it takes values from the queue it submits them, and then waits with a threshold of five seconds for a task before repeating. When a task is completed, it unpickles the result, and updates the futures that were stored in the dictionary previously.

PR #774, opened 2019-02-14
Adding overrides to srun launcher

yadudoc:
Code changes simply replicate the overrides mechanism used by AprunLauncher.


parsl/launchers/launchers.py
```diff
@@ -218,6 +225,27 @@ class SrunMPILauncher(Launcher):
     at the same time. Workers should be launched with independent Srun calls so as to setup the
     environment for MPI application launch.
     """
+
+    def __init__(self, options=None):
```
ZhuozhaoLi: Seems SrunMPILauncher has two __init__ ? The first one should be removed.

PR #777, opened 2019-02-15
Usage tracking opt in model #639

### LINKED ISSUE #639: Make usage tracking opt-in

yadudoc:
Fixes #639


danielskatz:
I’m not so happy about this change, but if we are going to make it, I think we should also change the tutorials and examples to include turning tracking on

yadudoc:
Let's do the doc fixes separately, (tracked by https://github.com/Parsl/parsl/issues/807)

PR #797, opened 2019-03-01
Allow alternate modes to be specified for bash app stdout and stderr.

mjwilde:
Modes can be specified via a (file,mode) tuple as an alternative
option to a simple filename string.

Default mode remains "a+" for now.  Typical alternative fir simple
file overrwite is ('filename','w')

Also fixes error in exception handling for BadStfFileStream and
MissingOutputs exception.  Problem was that the exception superclass
needs to be initialized withthe same member variables as the
subclassed execption in order for it to pickle correctly in the
"remote" execution context.

Other exceptions have similar issues, but were not corrected, pending
review of the proposed fix.

The app_stdout.py pytest was extended to test the new options, and
corrected to initialize correctly (by not attempting to clear and
reset the dfk, which seemed to causing this test to file.

Note that only the local execution tests were tested at the time of
this commit.

### DIFF
--- a/parsl/tests/test_bash_apps/test_stdout.py
+++ b/parsl/tests/test_bash_apps/test_stdout.py
@@ -6,10 +6,6 @@ import pytest
 import parsl
 import parsl.app.errors as perror
 from parsl.app.app import App
-from parsl.tests.configs.local_threads import config
-
-parsl.clear()
-dfk = parsl.load(config)
 
 
 @App('bash')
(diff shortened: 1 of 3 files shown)


parsl/tests/test_bash_apps/test_stdout.py
```diff
@@ -67,6 +125,7 @@ def test_bad_stderr():
 
     if args.debug:
         parsl.set_stream_logger()
-
-    y = test_bad_stdout()
-    y = test_bad_stderr()
+    y = test_bad_stdout_specs()  # FIXME: Not sure how this should work when tested directly
```
benclifford: In general the __main__ tests don't always test everything - so if test_bad_stdout_specs() isn't going to work invoked this way, it is ok to leave it off the __main__ list.

PR #800, opened 2019-03-03
Fix standalone stdout test modified in PR 797

benclifford:
PR #797 correctly removed the global DFK initialisation, but
should have added a DFK initialisation into the `__main__`
block so that a DFK is available in non-pytest mode. This
commit adds that.

PR #797 also added an incorrect invocation of the parameterised
test_bad_stdout_specs. This commit removes that.

### DIFF
--- a/parsl/tests/test_bash_apps/test_stdout.py
+++ b/parsl/tests/test_bash_apps/test_stdout.py
@@ -6,6 +6,7 @@ import pytest
 import parsl
 import parsl.app.errors as perror
 from parsl.app.app import App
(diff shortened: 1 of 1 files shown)


mjwilde:
@benc #800 looks good to me.  That test could readily be added back into main() by explicitly coding the parameter sweep. Not sure how often these main() tests are used, given how easy pytest is to run. But happy to add that loop if it would be helpful.

benc:
I assume the mention was meant for @benclifford :-)

PR #803, opened 2019-03-05
Dynamic executors for DLHub

ZhuozhaoLi:
This PR aims to provide dynamic executor support for DLHub, where each executor is a machine learning model deployed on Kubernetes. Users may need to frequently upload their models in DLHub, and hence parsl needs to support dynamically adding executors. 

This PR exposes `add_executor` function in dfk to support this functionality. Basically, I just refactor the code a bit, and it does not affect any other functionalities.

I also added a test code to test dynamically adding executors.


benclifford:
I think it should be possible to add a test for this that can run in CI (eg local threads then add in htex?)

ZhuozhaoLi:
> DataManager expects a static list of executors to be supplied at initialisation time; this PR breaks that assumption.
> 
> I think the effect is that file staging will not work for dynamically added executors.
> 
> Maybe the right thing to do is for the DataManager to receive a reference to the DFK, and pull out executor information dynamically from there as needed, rather than maintaining its own copy.
> 
> That should probably be a separate PR, which I can work on now.

I agreed. I think file staging does not work for added executors. Currently I think DLHub does not require this feature, but this may be needed for the training.

benclifford:
see #805 for a PR that fixes this, and 36ca4d7fe86bafc5d5b68c5ab3273a1bc05d10a0 for a test that shows that it doesn't work in this PR (but does with #805 merged in)

ZhuozhaoLi:
> I think it should be possible to add a test for this that can run in CI (eg local threads then add in htex?)

Added a test in tests/sites and should be included in CI.

yadudoc:
@ZhuozhaoLi Can you check if my merge picked the right edits from this PR and Ben's ?

ZhuozhaoLi:
> @ZhuozhaoLi Can you check if my merge picked the right edits from this PR and Ben's ?

Yes, i think it is correct.

PR #808, opened 2019-03-08
Broaden scope in which local_path mechanism can be used for Files

benclifford:
It does not need to be restricted to specific URI schemes - any staging
mechanism should be able to specify an overridden local path.

This comes from a review of #771 which does something similar but not
quite the same as the URI-based staging system


benclifford:
@AndrewLitteken notes in https://github.com/Parsl/parsl/pull/771#issuecomment-472230993 that this PR is merged into PR #771 (which is what I originally created this PR for)

Issue #809, opened 2019-03-08
Non-deterministic 'DataManager' object has no attribute 'storage_access'


benclifford:
Sometimes when creating `File`s with Globus URLs, the creation fails with the below exception. This is non-deterministic, and I expect it is based on non-deterministic order at which `_get_globus_endpoint()` iterates over executors.

I think this was introduced by PR #805, which changes the list of executors used by the data manager, to now include itself as an executor.

```
  File "parsl/tests/test_staging/test_implicit_staging_globus.py", line 29, in test_implicit_staging_globus
    unsorted_file = File('globus://03d7d06a-cb6b-11e8-8c6a-0a1d4c5c824a/unsorted.txt')
  File "/home/benc/parsl/src/parsl/parsl/data_provider/files.py", line 53, in __init__
    self.dman.add_file(self)
  File "/home/benc/parsl/src/parsl/parsl/data_provider/data_manager.py", line 112, in add_file
    self._set_local_path(file)
  File "/home/benc/parsl/src/parsl/parsl/data_provider/data_manager.py", line 115, in _set_local_path
    globus_ep = self._get_globus_endpoint()
  File "/home/benc/parsl/src/parsl/parsl/data_provider/data_manager.py", line 121, in _get_globus_endpoint
    for scheme in executor.storage_access:
AttributeError: 'DataManager' object has no attribute 'storage_access'
```

benclifford:
fixed by #815

PR #810, opened 2019-03-08
Simplify staging indirections

benclifford:
This PR attempts to simplify access to the data manager and to the `Globus` object used for globus-based staging. It removes an assumption that there is only one executor. It simplifies some `local_path` setting in the Globus provider to look the same as other staging methods.  It should address bug #809.

This also allows the removal of the DataManger reference from Files, which fits in better with their role as objects which are serialised across the network and so shouldn't contain references to local process-like entities like DataManager.


parsl/dataflow/dflow.py
```diff
@@ -348,7 +348,7 @@ def handle_app_update(self, task_id, future, memo_cbk=False):
             for dfu in self.tasks[task_id]['app_fu'].outputs:
                 f = dfu.file_obj
                 if isinstance(f, File) and f.is_remote():
-                    f.stage_out(self.tasks[task_id]['executor'])
+                    self.data_manager.stage_out(f, self.tasks[task_id]['executor'])
```
yadudoc: Does the data_manager at this point defer if the file is to be fetched by worker_side mechanism ?

benclifford: I don't think it is the role of `File` to understand how it gets moved between places. I think it is the role of DFK and DataManager to know that.

At the moment, they do it by launching a stage out parsl task.

I'd imagine that this line is approximately the point in the code where the data manager or DFK would choose the right way to do the staging, and make a different call if it isn't using the existing separate-task-staging - for example annotating the job to be run/wrapping it in some wrapper function

benclifford: stage_out will need to move *before* the task is submitted to the executor, though: once a task has finished, it's too late to start deciding on extra things to do inside the task.

In the existing separate-stage-out-task implementation, I think that can be made to work because these will be tasks dependent on some Future - although a bit fiddly.

benclifford: and for per-worker stage-ins, there does not need to be any DataFuture mediating the execution order between the in-worker stage in sub-task and the main sub-task - so can hopefully pass in a plain File object (with no `localpath` set yet) and have the stage in sub-task set `localpath` on the far end.

yadudoc: Okay, this is consistent with my mental model where either the dfk/datamanager will have to do something like this at task launch time.

```pseudocode
If worker can access the file:
      do nothing
if worker can access file if brough to globus_ep:
      fetch with globus to dest ep via parsl side staging
if worker can access file only if brought to worker: # no shared-fs with ep
      fetch with globus to local ep via parsl side staging, if file is remote
      use worker side staging with task wrappers.
```

The `File` objects encode the information about different endpoints/shared-fs where they might have been fetched to.

We'll still need to expose methods to say move files to a particular site to user, ideally we shouldn't have to tell them about dfk/datamanager in order to do those.

benclifford: > We'll still need to expose methods to say move files to a particular site to user, ideally we shouldn't have to tell them about dfk/datamanager in order to do those.

 Is that you wanting users to be able to move files around in some way that isn't as an input/output of an App?

benclifford: Pretty much anything user-exposed needs a DFK: we already have mechanisms to avoid the user having to explicitly pass around a DFK (via `parsl.load`)

yadudoc: Just thinking about weird cases where the user just wants to be able to move files and not do any apps. Or have the files move to multiple places.

benclifford:
this PR has got out of order with other PRs from my datamanager patch stack, and one commit was already merged in #878.

I've opened #884 to contain one patch to do with simplifying the `File` staging code path, and will submit a PR later for the other patch.

Closing this without merge.

PR #820, opened 2019-03-14
0.7.2 release prep

yadudoc:
There's still an issue with MonitoringHub importing classes that do so automatic initialization that is being fixed.


yadudoc:
Issues with monitoring components raising exceptions when not explicitly imported are not resolved. This is ready to go out as `0.7.2`

PR #822, opened 2019-03-18
Remove un-needed D203 override from flake8

benclifford:


yadudoc:
@benclifford what did you mean by "no error is reported because of this test" ?

benclifford:
I mean test D203 was disabled, but there was no obvious reason for it to be disabled: re-enabling D203 (which this PR does) doesn't cause flake8 to fail. So making it enabled again will help us stay in line with the proper style guidelines.

PR #828, opened 2019-03-19
Kubenetes provider deployment name

ZhuozhaoLi:
This PR makes some minor changes to kube provider:

* enable customized deployment_name for the kube provider
* the timestamp is at the precision of millisecond (`time.time() * 1000`), as it was observed that the second-level precision is not enough for HTEX -- it deploys very fast and the deployment_name may be the same.

Tested on Petrelkube.


parsl/providers/kubernetes/kube.py
```diff
@@ -109,10 +111,14 @@ def submit(self, cmd_string, blocksize, tasks_per_node, job_name="parsl.auto"):
...
-            job_name = "{0}-{1}".format(job_name, time.time()).split(".")[0]
+            cur_timestamp = str(time.time() * 1000).split(".")[0]
+            job_name = "{0}-{1}".format(job_name, cur_timestamp)
 
-            self.deployment_name = '{}-{}-deployment'.format(job_name,
-                                                             str(time.time()).split('.')[0])
+            if not self.deployment_name:
+                self.deployment_name = '{}-deployment'.format(job_name)
+            else:
+                self.deployment_name = '{}-{}-deployment'.format(self.deployment_name,
```
yadudoc: Everytime you call submit, this logic will keep adding a timestamp to `self.deployment_name`. We should use a local variable here.

PR #829, opened 2019-03-19
Prefetch limits

yadudoc:
This PR brings two separate features from branch `htex_strategy`:
* Support for specifying a `prefetch_capacity` that limits the # of tasks that can be fetched over the available worker capacity on the managers.
* Support for identifying managers with block_ids, and identification via a registration message to the interchange.

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -164,7 +164,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
             raise ConfigurationError('Multiple storage access schemes are not supported')
         self.working_dir = working_dir
         self.managed = managed
-        self.blocks = []
+        self.blocks = {}
         self.tasks = {}
         self.cores_per_worker = cores_per_worker
         self.max_workers = max_workers
@@ -189,6 +189,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
                                "--task_url={task_url} "
                                "--result_url={result_url} "
                                "--logdir={logdir} "
+                               "--block_id={{block_id}} "
                                "--hb_period={heartbeat_period} "
                                "--hb_threshold={heartbeat_threshold} ")
 
@@ -427,14 +428,30 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
     @property
     def outstanding(self):
         outstanding_c = self.command_client.run("OUTSTANDING_C")
-        logger.debug("Got outstanding count: {}".format(outstanding_c))
+        # logger.debug("Got outstanding count: {}".format(outstanding_c))
         return outstanding_c
 
     @property
     def connected_workers(self):
         workers = self.command_client.run("MANAGERS")
-        logger.debug("Got managers: {}".format(workers))
+        # logger.debug("Got managers: {}".format(workers))
         return workers
+
+    def _hold_block(self, block_id):
+        """ Sends hold command to all managers which are in a specific block
+
+        Parameters
+        ----------
+        block_id : str
+             Block identifier of the block to be put on hold
+        """
+
+        managers = self.connected_workers
+
+        for manager in managers:
+            if manager['block_id'] == block_id:
+                logger.debug("[HOLD_BLOCK]: Sending hold to manager:{}".format(manager['manager']))
+                self.hold_worker(manager['manager'])
 
     def submit(self, func, *args, **kwargs):
         """Submits work to the the outgoing_q.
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -479,38 +506,60 @@ def scale_out(self, blocks=1):
         r = []
         for i in range(blocks):
             if self.provider:
-                block = self.provider.submit(self.launch_cmd, 1, 1)
-                logger.debug("Launched block {}:{}".format(i, block))
-                if not block:
+                external_block_id = len(self.blocks)
```
benclifford: `str`ingify the external_block_id here so that it's already the right type elsewhere in the code

parsl/executors/high_throughput/executor.py
```diff
@@ -479,38 +506,60 @@ def scale_out(self, blocks=1):
...
-                self.blocks.extend([block])
+                r.extend([str(external_block_id)])
+                self.blocks[str(external_block_id)] = internal_block
             else:
                 logger.error("No execution provider available")
                 r = None
         return r
 
-    def scale_in(self, blocks):
+    def scale_in(self, blocks, block_ids=[]):
```
benclifford: This parameter seems weird: either you'd want to specify a number of anonymous blocks to remove, or an explicit list of blocks to remove, but never both, especially if you specify both with conflicting meaning (eg a list of three IDs, and blocks=2).

Perhaps there should be no added parameter here, but the type of `blocks` should now be either an integer number of anonymous blocks to remove or a list of strings, being a list of identified blocks to remove. That better reflects the behaviour, rather than silently ignoring some parameters sometimes.

benclifford: What is the expected way in which a user of this would acquire a block IDs to pass in here? Something that knows how to look into the guts of htex?

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -72,6 +77,11 @@ def __init__(self,
              caps the maximum number of workers that can be launched.
              default: infinity
 
+        prefetch_capacity : int
+             Number of tasks that could be prefetched over available worker capacity. Default:100.
+             When there are a few tasks (<100) or when tasks are long running, this option should
+             be set to 0 for better load balancing.
+
```
benclifford: It's not immediately clear what to me the difference between `max_queue_size` and `prefetch_capacity`, so I'm going to assume that that isn't clear to other users. I'll have a bit more of a poke into the code to try to get my head round it.

parsl/executors/high_throughput/executor.py
```diff
@@ -417,15 +428,31 @@ def hold_worker(self, worker_id):
...
+        block_id : str
+             Block identifier of the block to be put on hold
+        """
+
+        managers = self.connected_workers
+
+        for manager in managers:
+            if manager['block_id'] == block_id:
+                logger.debug("[HOLD_BLOCK]: Sending hold to manager:{}".format(manager['manager']))
+                self.hold_worker(manager['manager'])
```
ZhuozhaoLi: This seems more like `hold_manager`?

parsl/executors/high_throughput/interchange.py
```diff
@@ -347,6 +349,7 @@ def start(self, poll_period=None):
                     # By default we set up to ignore bad nodes/registration messages.
                     self._ready_manager_queue[manager] = {'last': time.time(),
                                                           'free_capacity': 0,
+                                                          'max_capacity': 0,
```
ZhuozhaoLi: I think it should not be 0 for `max_capacity`?

annawoodard:
Note from @ZhuozhaoLi, default prefetch should be zero

parsl/executors/high_throughput/executor.py
yadudoc: Done

parsl/executors/high_throughput/executor.py
yadudoc: The mechanism is to call `executor.connected_workers` which returns something like this : 
```python
[{'manager': 'cd17f5ff5268', 'block_id': '0', 'worker_count': 4, 'tasks': 0, 'active': True}] ```

This would allow the strategy component to identify block to manager to worker mappings and determine what capacity is available/in-use.

File: parsl/executors/high_throughput/process_worker_pool.py (hunk 3 above)
Comment (yadudoc) [reply parsl/executors/high_throughput/process_worker_pool.py]: One of these are redundant. `max_queue_size` is unnecessary to expose to the user. It is the actual capacity that we'll have on the manager side.
I'm removing this as a configurable option.

Comment (yadudoc) [reply parsl/executors/high_throughput/process_worker_pool.py]: @benclifford please check the updated code and see if that resolves it. I've updated the code to have max_queue_size be the prefetch_capacity + worker_count.

File: parsl/executors/high_throughput/executor.py [hunk 6]
```diff
@@ -112,6 +112,11 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
     max_workers : int
         Caps the number of workers launched by the manager. Default: infinity
 
+    prefetch_capacity : int
+        Number of tasks that could be prefetched over available worker capacity. Default:100.
+        When there are a few tasks (<100) or when tasks are long running, this option should
+        be set to 0 for better load balancing.
```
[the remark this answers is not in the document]
yadudoc: Agreed. Setting it to 0.

parsl/executors/high_throughput/process_worker_pool.py
benclifford: that makes sense

parsl/executors/high_throughput/executor.py
```diff
@@ -479,38 +505,58 @@ def scale_out(self, blocks=1):
...
+
+        block_ids : list
+             List of specific block ids to terminate. Optional
+
         Raises:
              NotImplementedError
         """
-        to_kill = self.blocks[:blocks]
+        for block_id in block_ids:
+            self._hold_block(block_id)
```
benclifford: This does a hold on a to-be-killed block if that block is chosen through one block selection mechanism, but not through the other.

Is this needlessly inconsistent

parsl/executors/high_throughput/interchange.py
```diff
@@ -402,11 +405,15 @@ def start(self, poll_period=None):
...
                 while shuffled_managers and not self.pending_task_queue.empty():  # cf. the if statement above...
                     manager = shuffled_managers.pop()
-                    if (self._ready_manager_queue[manager]['free_capacity'] and
-                        self._ready_manager_queue[manager]['active']):
-                        tasks = self.get_tasks(self._ready_manager_queue[manager]['free_capacity'])
+                    tasks_inflight = len(self._ready_manager_queue[manager]['tasks'])
+                    real_capacity = min(self._ready_manager_queue[manager]['free_capacity'],
+                                        self._ready_manager_queue[manager]['max_capacity'] - tasks_inflight)
+
+                    if (real_capacity and self._ready_manager_queue[manager]['active']):
```
annawoodard: Can `real_capacity` be less than 0? If so, this will evaluate to True if `self._ready_manager_queue[manager]['active']` is True.

yadudoc:
#881 and #888 cover all the changes in this PR. Closing.

PR #830, opened 2019-03-19
Mem aware htex

yadudoc:


parsl/executors/high_throughput/executor.py
```diff
@@ -186,15 +212,24 @@ def initialize_scaling(self):
         debug_opts = "--debug" if self.worker_debug else ""
         max_workers = "" if self.max_workers == float('inf') else "--max_workers={}".format(self.max_workers)
 
+        if self.interchange_address == "localhost":
```
benclifford: As an operator, I've found it frustrating in the past to have non-network related stuff change in behaviour depending on a specified hostname - postgres does it sometimes, for example.

What is this `if` actually really meant to be testing?

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -102,8 +109,11 @@ def __init__(self,
         self.uid = uid
 
         cores_on_node = multiprocessing.cpu_count()
+        available_mem_on_node = round(psutil.virtual_memory().available / (2**30), 1)
+
         self.max_workers = max_workers
         self.worker_count = min(max_workers,
+                                math.floor(available_mem_on_node / mem_per_worker),
```
ZhuozhaoLi: I think this would raise exception if `mem_per_worker` is not set (default to 0)

parsl/executors/high_throughput/executor.py
```diff
@@ -93,6 +94,14 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
     worker_port_range : (int, int)
         Worker ports will be chosen between the two integers provided.
 
+    interchange_address : str | function
+        Address (str) at which the interchange will be reachable from compute nodes or an address
+        function from parsl.addresses that be executed on the interchange to determine a suitable
+        address. Only valid when remote_interchange=True.
```
ZhuozhaoLi: Description might need to be updated as we decide to remove `remote_interchange=True`

parsl/executors/high_throughput/executor.py
```diff
@@ -131,11 +145,13 @@ def __init__(self,
                  address="127.0.0.1",
                  worker_ports=None,
                  worker_port_range=(54000, 55000),
+                 interchange_address="localhost",
                  interchange_port_range=(55000, 56000),
                  storage_access=None,
                  working_dir=None,
                  worker_debug=False,
                  cores_per_worker=1.0,
+                 mem_per_worker=None,
```
ZhuozhaoLi: The default is 0 below.

parsl/providers/local/local.py
```diff
@@ -257,6 +257,12 @@ def current_capacity(self):
     def label(self):
         return self._label
 
+    def execute_wait(self, cmd, timeout=None):
+        t = self.cmd_timeout
+        if timeout is not None:
+            t = timeout
+        return self.channel.execute_wait(cmd, t)
```
ZhuozhaoLi: I think this method should also be included in other providers?

PR #832, opened 2019-03-21
Make it so mypy can make basic check of codebase

benclifford:
This PR fixes all of the errors that happen when trying to make mypy tests against the configuration and test files; and then makes that check happen in CI.

On my laptop, this additional testing takes 1m25s of realtime.

### DIFF
--- a/parsl/app/futures.py
+++ b/parsl/app/futures.py
@@ -220,7 +220,7 @@ if __name__ == "__main__":
         for item in nums:
             testfile.write("{0}\n".format(item))
 
-    foo = Future()
+    foo = Future()  # type: Future[str]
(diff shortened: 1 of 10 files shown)


parsl/app/futures.py
```diff
@@ -210,7 +210,7 @@ def testing_nonfuture():
         for item in nums:
             testfile.write("{0}\n".format(item))
 
-    foo = Future()
+    foo = Future()  # type: Future[str]
```
yadudoc: Does `Future[str]` mean a future that'll resolve to a string ?

benclifford: yes

PR #847, opened 2019-04-03
remove duplicate logging and change logging levels of few loggings

ZhuozhaoLi:
The changes in this PR:

1. Remove a duplicate logging line.
2. Change logging level of several lines from `debug` to `info` level, so that they won't slow down the interchange.


benclifford:
This PR is failing CI because:
```
pytest-xdist 1.28.0 has requirement pytest>=4.4.0, but you'll have pytest 3.7.1 which is incompatible.
```
which is possibly because `test-requirements.txt` has a restriction on `pytest` versions, but not on `pytest-xdist`.

This probably means that a clean build from `master` will also fail, and it would be appropriate to have a separate PR to fix this problem.

benclifford:
see #850 for a potential CI fix.

PR #852, opened 2019-04-05
Fix docs for visualization -- fixes #845

### LINKED ISSUE #845: Add necessary imports to Docs for monitoring example
Need to update the example config in https://parsl.readthedocs.io/en/stable/userguide/monitoring.html to include necessary imports.

ZhuozhaoLi:
fixes #845 and modify docs for using visualization server

### DIFF
--- a/docs/userguide/monitoring.rst
+++ b/docs/userguide/monitoring.rst
@@ -61,6 +61,10 @@ Install the visualization server::
 Once `viz_server` is installed, you can run the utility `parsl-visualize` in the directory with the
 `monitoring.db` sqlite file to launch a web page for the workflow visualization::
 
-   $ parsl-visualize sqlite:///<absolute-full-path-to-db>
+   $ parsl-visualize sqlite:///<absolute-path-to-db>
+
+For example, if the `monitoring.db` is at `/tmp/monitoring.db`, run the `parsl-visualize` as follows::
+
+   $ parsl-visualize sqlite:////tmp/monitoring.db
 
 .. warning:: Please note that visualization support is in `alpha` state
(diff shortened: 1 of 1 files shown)


docs/userguide/monitoring.rst
```diff
@@ -50,6 +59,8 @@ Install the visualization server::
    $ pip install git+https://github.com/Parsl/viz_server.git
 
 Once `viz_server` is installed, you can run the utility `parsl-visualize` in the directory with the
-monitoring.db sqlite file to launch a web page with the workflow visualization.
+`monitoring.db` sqlite file to launch a web page for the workflow visualization::
+
+   $ parsl-visualize sqlite:///<absolute-full-path-to-db>
```
yadudoc: We should point out the absurd 4 slashes necessary in a note. Eg: `sqlite:////tmp/monitoring.db`.
Absolute-path is sufficient, we don't need to say absolute-full-path

docs/userguide/monitoring.rst
```diff
@@ -23,11 +23,20 @@ Here's an example configuration that logs monitoring information to a local sqli
...
+    from parsl.addresses import address_by_hostname
+
+    import logging
+
     config = Config(
         executors=[
             HighThroughputExecutor(
                 label="local_htex",
                 cores_per_worker=1,
+                max_workers=4,
```
yadudoc: Aren't we missing the monitoring config here ? Say include something like : 
```python
threads_config = Config(
    executors=[
	HighThroughputExecutor(
            label="local_htex",
            cores_per_worker=1,
        )
    ],
    monitoring=MonitoringHub(
        hub_port=55055,
        logging_level=logging.INFO,
    ),
    strategy=None
)
```

ZhuozhaoLi: It is already there. You can see it if you view the whole file.

docs/userguide/monitoring.rst
ZhuozhaoLi: added new lines to explain that

Issue #851, opened 2019-04-04
parsl workflow issue: AttributeError: 'ManagerLost' object has no attribute 'reraise'


TomGlanzman:
A parsl (v0.7.2) job at NERSC running on Cori-login (for top-level script) and Cori-knl for the (single node) batch work via the HighThroughputExecutor ran into the situation where the running SLURM job ran out of time.  There had been four tasks running in batch on the same node at the time.  The top-level script was silent, so I waited.  After a time, not knowing what was happening, I hit ctrl-c, which produced the following traceback:

```
Traceback (most recent call last):                                                                                              
  File "/global/u1/d/descdm/tomTest/parslTest3/BFworkflow.py", line 135, in <module>                                            
    print("rc = ",job.result())                                                                                                 
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_11/python/miniconda3-4.5.4/envs/lsst-scipipe-fcd27eb/lib/python3.6/concurrent/futures/_base.py", line 432, in result                                                                                 
    return self.__get_result()                                                                                                  
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_11/python/miniconda3-4.5.4/envs/lsst-scipipe-fcd27eb/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result                                                                           
    raise self._exception                                                                                                       
  File "/global/homes/d/descdm/.local/lib/python3.6/site-packages/parsl/dataflow/futures.py", line 126, in parent_callback      
    res = executor_fu.result()                                                                                                  
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_11/python/miniconda3-4.5.4/envs/lsst-scipipe-fcd27eb/lib/python3.6/concurrent/futures/_base.py", line 425, in result                                                                                 
    return self.__get_result()                                                                                                  
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_11/python/miniconda3-4.5.4/envs/lsst-scipipe-fcd27eb/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result                                                                           
    raise self._exception                                                                                                       
  File "/global/homes/d/descdm/.local/lib/python3.6/site-packages/parsl/dataflow/dflow.py", line 244, in handle_exec_update     
    res = future.result()                                                                                                       
[...]

PR #856, opened 2019-04-10
Wrap interchange exceptions in RemoteExceptionWrapper

benclifford:
Prior to this commit, sometimes the serialised data structure contained raw `Exception` objects, rather than the expected `RemoteExceptionWrapper`.


benclifford:
This is buggy - see #851 - because `RemoteExceptionWrapper` isn't constructed properly.

PR #722, opened 2019-01-02
[WIP] continuing work on rearranging app futures

benclifford:

PR #865, opened 2019-04-12
Unify callback paths for execution and memo-based results

benclifford:
This removes some of the special casing for memoized results - instead result callbacks are handled the same whether the Future comes from execution or memoization.

This is from part of the PR #722 work on callbacks that I'm pulling into separate PRs.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -367,6 +367,12 @@ class DataFlowKernel(object):
                             task_id, self.tasks[task_id]['func'], *new_args, **kwargs)
 
                 if exec_fu:
+
+                    try:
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -367,6 +367,12 @@ def launch_if_ready(self, task_id):
                             task_id, self.tasks[task_id]['func'], *new_args, **kwargs)
 
                 if exec_fu:
+
+                    try:
+                        exec_fu.add_done_callback(partial(self.handle_exec_update, task_id))
```
yadudoc: We've had  issues with the order in which callbacks are added creating race-conditions/ breakages. Would be useful to run this through more rounds of testing than usual.

PR #874, opened 2019-04-14
Introduce mypy stub for typeguard, and type-check parsl.load()

benclifford:
Introduce mypy stub for typeguard, and type-check parsl.load()
    
This allows mypy to understand the type signatures of functions
decorated with @typeguard.typechecked, allowing the type signatures
on such functions to be used by both enforce and mypy.
    
Prior to this commit, adding @typeguard.typechecked removed the ability of mypy to
type-check calls to decorated functions, as it did not understand that
the type of such a decorated function is the same as the original
function. The added stub makes this declaration.
    
The test-suite mypy calls are modified to use that stub file.
    
The immediate motivation for this is to keep mypy passing when adding a
type annotation for parsl.load; that type annotation is also added in
this commit.
    
This commit fixes some of problem described in the commit message for
d7d9a255 about breaking mypy type checking for end user Configs in some
places, but by no means all of it.
    
However there is a mypy bug, https://github.com/python/mypy/issues/5398,
where decorated __init__ methods are not properly type-checked by mypy,
which means that mypy typechecking is still missing on any decorated
__init__ methods - which unfortunately is most of the user-facing type
checked code.
    
At time of writing, there are PRs open to fix this in mypy, so there
is hope; and that does not impede this PR being merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
     # This uses all of the configurations and tests as the base from which to
     # run mypy checks - these are likely to capture most of the code used in
     # parsl
-    - (for test in parsl/tests/configs/*.py parsl/tests/test*/test*; do mypy $test ; export MypyReturnCode=$? ; echo mypy return code is $MypyReturnCode ; if [[ "$MypyReturnCode" != 0 ]] ; then exit 1; fi; done ) ;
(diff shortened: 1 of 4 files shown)


benclifford:
i'd like #875 merged, and this pr updated for that first

PR #877, opened 2019-04-15
Make http staging test run with different configs, rather than fixed one

benclifford:
This means that http staging will be tested in CI with both thread local and htex - giving a test of staging in the submitting process and in a remote process.

It also means that it will be tested when a developer runs the tests manually with an arbitrary configuration.

Previously, this test ran only with a local threads config.


benclifford:
This fails in CI: at some point, something tries to import the original test module as `test_implicit_staging_https' - without proper qualifications. I think this is something to do with how pytest annotates some stuff, but I don't have a clear idea.

benclifford:
I think #951 might fix this.

PR #879, opened 2019-04-15
Remove methods that should come from Future superclass

benclifford:


annawoodard:
Is the hope that this doesn't change the behavior of anything?

benclifford:
Yes. There are some situations where I think the previous code would not work right (such as discarding callbacks) so the behaviour in general will be different but my intention is not in any code path that we actually use.

PR #883, opened 2019-04-16
Htex python version check #857

### LINKED ISSUE #857: HTEX interchange check python version mismatch to the micro level
Currently HTEX interchange check python version mismatch between worker and interchange to a micro level: e.g., `3.6.6 != 3.6.8`. I think it should be enough to only check the minor level.

https://github.com/Parsl/parsl/blob/2bc571c0a51ff04632548605d44314be7a97c8d0/parsl/executors/high_throughput/interchange.py#L358

ZhuozhaoLi:
This PR changes python version check in HTEX interchange from micro to minor level, e.g., now python 3.6.6 is compatible with python 3.6.8.

Fixes #857

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -371,7 +371,10 @@ class Interchange(object):
                                 logger.warning("[MAIN] Sent failure reports, unregistering manager")
                             else:
                                 logger.debug("[MAIN] Suppressing shutdown due to version incompatibility")
-
+                        else:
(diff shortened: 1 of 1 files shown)


benclifford:
My main concern here is that I don't understand what the version check is attempting to protect against - so I don't know if that protection remains in place.

yadudoc:
@benclifford The version check is to protect against managers with an incompatible python version connecting to the interchange. If there's a version mismatch, pickling/unpickling will fail, so we'd rather fail early in this case.

PR #888, opened 2019-04-16
Managers to report block_id on registration

yadudoc:
Support for identifying managers by block_id, and putting managers in a block on hold

Managers are given a block identifier at initialization that is reported back at registration.
This allows the executor to track tasks/worker slots at the block level.
The executor also has a _hold_block() function that puts all managers in a block on hold
preventing further task scheduling, and making it safe to terminate the block.

(cherry picked from commit dc94c0223bccd0159813e8425f787f624790bbae)

Extracted from #826 .

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -162,7 +162,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
             raise ConfigurationError('Multiple storage access schemes are not supported')
         self.working_dir = working_dir
         self.managed = managed
-        self.blocks = []  # type: List[Any]
+        self.blocks = {}  # type: Dict[str, str]
         self.tasks = {}  # type: Dict[str, Future]
         self.cores_per_worker = cores_per_worker
         self.max_workers = max_workers
@@ -185,6 +185,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
                                "--task_url={task_url} "
                                "--result_url={result_url} "
                                "--logdir={logdir} "
+                               "--block_id={{block_id}} "
                                "--hb_period={heartbeat_period} "
                                "--hb_threshold={heartbeat_threshold} ")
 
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -421,15 +422,31 @@ def hold_worker(self, worker_id):
...
 
     @property
     def connected_workers(self):
         workers = self.command_client.run("MANAGERS")
-        logger.debug("Got managers: {}".format(workers))
+        # logger.debug("Got managers: {}".format(workers))
         return workers
 
+    def _hold_block(self, block_id):
+        """ Sends hold command to all managers which are in a specific block
```
benclifford: Will there be multiple managers in the same block? In my head there was a 1:1 correspondence.

parsl/executors/high_throughput/executor.py
```diff
@@ -483,38 +500,58 @@ def scale_out(self, blocks=1):
...
-                self.blocks.extend([block])
+                r.extend([external_block_id])
+                self.blocks[external_block_id] = internal_block
             else:
                 logger.error("No execution provider available")
                 r = None
         return r
 
-    def scale_in(self, blocks):
+    def scale_in(self, blocks, block_ids=[]):
```
benclifford: Maybe this should be a single parameter: blocks. That should be either a number N, meaning kill N blocks of parsls choice, or a list of block_ids, meaning kill these blocks.

It never makes sense to pass both or neither - always exactly one (I think)

(It's the sum type Vs product type distinction from type theory if you like that sort of thing...)

parsl/executors/high_throughput/executor.py
```diff
@@ -483,38 +500,58 @@ def scale_out(self, blocks=1):
...
+
+        block_ids : list
+             List of specific block ids to terminate. Optional
+
         Raises:
              NotImplementedError
         """
-        to_kill = self.blocks[:blocks]
+        for block_id in block_ids:
+            self._hold_block(block_id)
```
benclifford: Why hold blocks when you specify the choice of blocks one way, but not when you specify the other way?

parsl/executors/high_throughput/executor.py
yadudoc: A block can have many nodes. Each node has 1 manager.

parsl/executors/high_throughput/executor.py
yadudoc: I'd have preferred having two kwargs `blocks_ids=[<list of block ids], block_count=<int>` but that would mean making changes everywhere to match the name changes everywhere.

I worry that overloading blocks to mean two different things based on type won't be very clear.

parsl/executors/high_throughput/executor.py
yadudoc: That's a bug. Fixing it now.

parsl/executors/high_throughput/executor.py
benclifford: As the code is now, you have to pass something for the blocks parameter, right? Even if it's going to get ignored because there is a block_ids list?

benclifford: Maybe adding =None to blocks will let it be used both positionally but also omitted if the block_ids keyword is used

benclifford: Maybe adding =None to blocks will let it be used both positionally but also omitted if the block_ids keyword is used

Issue #898, opened 2019-04-22
Manager on condor requests much-too-large number of tasks


lgray:
```
2019-04-22 10:48:04.460 interchange:393 [DEBUG]  [MAIN] Manager b'49250f3c86d1' requested 4294967295 tasks
2019-04-22 10:48:04.460 interchange:396 [DEBUG]  [MAIN] Manager b'49250f3c86d1' sends heartbeat
2019-04-22 10:48:04.461 interchange:401 [DEBUG]  [MAIN] leaving task_outgoing section
2019-04-22 10:48:04.461 interchange:406 [DEBUG]  Managers count (total/interesting): 16/11
2019-04-22 10:48:04.461 interchange:439 [DEBUG]  [MAIN] either no interesting managers or no tasks, so skipping manag
er pass
2019-04-22 10:48:04.462 interchange:456 [DEBUG]  [MAIN] entering bad_managers section
2019-04-22 10:48:04.462 interchange:472 [DEBUG]  [MAIN] leaving bad_managers section
2019-04-22 10:48:04.462 interchange:473 [DEBUG]  [MAIN] ending one main loop iteration
2019-04-22 10:48:04.462 interchange:334 [DEBUG]  [MAIN] starting task_outgoing section
2019-04-22 10:48:04.463 interchange:393 [DEBUG]  [MAIN] Manager b'49250f3c86d1' requested 40 tasks
2019-04-22 10:48:04.463 interchange:401 [DEBUG]  [MAIN] leaving task_outgoing section
2019-04-22 10:48:04.463 interchange:406 [DEBUG]  Managers count (total/interesting): 16/12
2019-04-22 10:48:04.463 interchange:439 [DEBUG]  [MAIN] either no interesting managers or no tasks, so skipping manager pass
```

Submitting some simple jobs to count the number of events in TTrees.
We are currently debugging this but some introspection would be appreciated.

benclifford:
That 4294967295 request is probably a spurious log message that isn't actually requesting that many tasks.

see `parsl/executors/high_throughput/interchange.py` where that message is actually a heartbeat code: 4294967295 == -1 == HEARTBEAT_CODE

The code looks like it is logging that number, then going down the heartbeat path not the request tasks path - so really that log statement should be moved elsewhere.

Lines like `Manager b'49250f3c86d1' requested 40 tasks` are more legitimate.


```
                    tasks_requested = int.from_bytes(message[1], "little")
[...]

PR #899, opened 2019-04-22
Tidy debug messages around heartbeat/task requests

benclifford:
Issue #898 shows a confusing log message, where a heartbeat
incorrectly logs that HEARTBEAT_CODE tasks have been requested,
in addition to logging a heartbeat.

This commit moves that log into the non-heartbeat path only, so
that exactly one of the heartbeat or task request messages will be
logged.

The phrasing of the heartbeat log message is modified to be in the
same sense as the task request message.


annawoodard:
Thanks @benclifford. I think this fixes #898.

lgray:
Yeah looks like it was this.

PR #902, opened 2019-04-23
Make `connected_workers` consistent

annawoodard:
This fix makes the `connected_workers` method return the number of connected workers. Note this is needed to more conveniently debug issues (to be posted shortly) with dynamic slot sizes on Condor with @lgray.

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -435,8 +435,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
 
     @property
     def connected_workers(self):
-        workers = self.command_client.run("MANAGERS")
-        # logger.debug("Got managers: {}".format(workers))
+        workers = self.command_client.run("WORKERS")
         return workers
 
     def _hold_block(self, block_id):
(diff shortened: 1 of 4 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -435,8 +435,7 @@ def outstanding(self):
 
     @property
     def connected_workers(self):
-        workers = self.command_client.run("MANAGERS")
-        # logger.debug("Got managers: {}".format(workers))
+        workers = self.command_client.run("WORKERS")
```
ZhuozhaoLi: A quick look is that, I think, we should have a new method `connected_workers`, and change the original `connected_workers` to `connected_managers`, instead of replacing.

This is because the method `_hold_block` below apparently uses `connected_workers` as the context of # of connected managers. There may be other places that use the same context.

annawoodard: Great catch, @ZhuozhaoLi. This bit of code was merged recently. I was working on an outdated branch and didn't catch this when I rebased to master. I'll create a new method as you suggest.

Issue #868, opened 2019-04-12
max_blocks not respected in SlurmProvider


TomGlanzman:
By specifying the following in my config:
```
            max_blocks=1, 
            min_blocks=1, 
```
I am finding *two* slurm jobs being submitted rather than one.  User wilde pointed out that by adding in `parallelism = 0.1` the extra slurm job would not be submitted.  Still, it seems that max_blocks should be respected.

Parsl 0.7.2
executor = HighThroughputExecutor
provider = SlurmProvider

annawoodard:
Thanks for the report, @TomGlanzman!

Looks like we were on the right track, but we never update `provisioned_blocks` anywhere:
[https://github.com/Parsl/parsl/blob/master/parsl/providers/slurm/slurm.py#L158](https://github.com/Parsl/parsl/blob/master/parsl/providers/slurm/slurm.py#L158)

yadudoc:
Related to #921

benclifford:
PR #907 claims to fix this

Issue #704, opened 2018-12-12
Fix strategy overallocations


annawoodard:
Reported by @gordonwatts

When using `LocalProvider` with `strategy='simple'`, the provider does not track the status of submitted blocks properly-- it therefore thinks it needs to submit more blocks, resulting in a run-away submission of blocks. This is especially bad for htex, where each manager is using ~100% CPU (see #599). 

```
2018-12-12 02:59:40 parsl.dataflow.strategy:125 [DEBUG]  Scaling strategy: simple
2018-12-12 02:59:42 parsl.dataflow.strategy:202 [DEBUG]  Executor remote_ipp_tev02.phys.washington.edu has 1 active tasks, 0/0/0 running/submitted/pending blocks, and 0 connected engines
2018-12-12 02:59:42 parsl.dataflow.strategy:256 [DEBUG]  Requesting 1 more blocks
2018-12-12 02:59:49 parsl.dataflow.strategy:202 [DEBUG]  Executor remote_ipp_tev02.phys.washington.edu has 1 active tasks, 0/0/0 running/submitted/pending blocks, and 2 connected engines
2018-12-12 02:59:49 parsl.dataflow.strategy:256 [DEBUG]  Requesting 1 more blocks
2018-12-12 02:59:55 parsl.dataflow.strategy:202 [DEBUG]  Executor remote_ipp_tev02.phys.washington.edu has 16 active tasks, 0/0/0 running/submitted/pending blocks, and 3 connected engines
2018-12-12 02:59:55 parsl.dataflow.strategy:256 [DEBUG]  Requesting 16 more blocks
2018-12-12 03:02:39 parsl.dataflow.strategy:202 [DEBUG]  Executor remote_ipp_tev02.phys.washington.edu has 31 active tasks, 0/0/0 running/submitted/pending blocks, and 19 connected engines
2018-12-12 03:02:39 parsl.dataflow.strategy:256 [DEBUG]  Requesting 31 more blocks
2018-12-12 03:13:47 parsl.dataflow.strategy:202 [DEBUG]  Executor remote_ipp_tev02.phys.washington.edu has 31 active tasks, 0/0/0 running/submitted/pending blocks, and 50 connected engines
2018-12-12 03:13:47 parsl.dataflow.strategy:256 [DEBUG]  Requesting 31 more blocks
[...]

PR #907, opened 2019-04-25
add constraints in strategy to cap the maximum blocks

ZhuozhaoLi:
Currently the strategy does not properly set the maximum blocks. This PR should fix multiple similar issues, e.g., #868, #704

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -259,6 +259,7 @@ class Strategy(object):
                     # logger.debug("Strategy: Case.2b")
                     excess = math.ceil((active_tasks * parallelism) - active_slots)
                     excess_blocks = math.ceil(float(excess) / (tasks_per_node * nodes_per_block))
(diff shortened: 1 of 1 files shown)


ZhuozhaoLi:
> Thinking about this a bit more: in the providers we check in various places if `self.provisioned_blocks >= self.max_blocks`, and if so, don't submit. But then we never increment provisioned blocks (which is why the protection has no effect at the moment). There may be a case for enforcing `max_blocks` both in the strategy and in the provider, but we should consider either fixing the provider check or removing it to make the code less confusing.

Agreed. I think we probably should only enforce `max_blocks` in the strategy. 
It looks like the `self.provisioned_blocks` do nothing in providers.
https://github.com/Parsl/parsl/blob/3fd299b09675de6fade59ce484669e82544f440c/parsl/providers/slurm/slurm.py#L158-L160

annawoodard:
We're factoring the provisioned blocks fix out into #921, so this is good to go.

PR #915, opened 2019-04-30
Make Travis CI fail fast

annawoodard:


benclifford:
This is failing in CI - I think because there are places where `pytest` returns a failure (value 5) in a situation that isn't a case where we want CI to fail.

annawoodard:
Closing-- this will be more straightforward to do once #984 is in.

PR #930, opened 2019-05-07
Log executable to stdout

annawoodard:


annawoodard:
The tests are failing because I've added unexpected output to standard out. For example,
```
@bash_app
    def echo_hello(stdout='echo-hello.stdout'):
        return 'echo "Hello, world!"'
```

will produce the following in `echo-hello.stdout`:
```
--> executable follows <--
echo "Hello, world!"
--> standard output follows <--
Hello, world!
```

I did this to avoid having to evaluate the executable in the main parsl process (in order to print it in `parsl.log`), and because I find this more readable. We can either evaluate the executable in the main parsl process and print to `parsl.log`, and leave the standard out as it was previously, or modify the failing tests. Any thoughts?

annawoodard:
Another option would be to write the executable to its own file, but that seems a bit overkill.

ZhuozhaoLi:
I vote to modify the failing tests.

annawoodard:
Thanks for the input everyone! I've changed to logging to stderr, so we should be good to go.

PR #940, opened 2019-05-14
Implement richer __repr__ for File which includes localpath

benclifford:


parsl/data_provider/files.py
```diff
@@ -44,7 +44,10 @@ def __str__(self):
         return self.filepath
 
     def __repr__(self):
-        return self.__str__()
+        if hasattr(self, 'local_path'):
+            return "<{} at %#x url={} localpath={}>".format(self.__class__, id(self), self.url, self.localpath)
```
annawoodard: What is the `%#x`?

benclifford: @annawoodard %#x formats `id(self)` as hex, which fits in with how other objects are __repr_-ed; for example:

```
>>> x = map(str,[1,2,3])
>>> x
<map object at 0x7f69ff01aac8>
>>> id(x)
140093226592968
>>> hex(id(x))
'0x7f69ff01aac8'
```

PR #944, opened 2019-05-16
Support for OAuth based SSH Channels

yadudoc:
We are using the `oauth-ssh` module from https://github.com/XSEDE/oauth-ssh/tree/master/client for this new channel.

I've only done very limited testing with this, so there are still rough edges and errors are probably not handled quite well. There's also a test script in `parsl/tests/manual_tests/test_oauth_ssh.py`  that should help with testing the channel on it's own.


parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,140 @@
...
+import socket
+
+from parsl.errors import OptionalModuleMissing
+from parsl.channels.ssh.ssh import SSHChannel
+
+try:
+    from oauth_ssh.ssh_service import SSHService
+    from oauth_ssh.oauth_ssh_token import find_access_token
+    _oauth_ssh_enabled = True
+except (ImportError, NameError, FileNotFoundError):
```
benclifford: What's the FileNotFoundError that might happen here? Is it something that isn't explicitly import related?

parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,140 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OAuthSSHChannel(SSHChannel):
+    """SSH persistent channel. This enables remote execution on sites
+    accessible via ssh. This channel uses Globus based OAuth tokens for authentication.
+    """
+
+    def __init__(self, hostname, username=None, script_dir=None, envs=None, port=22, **kwargs):
```
benclifford: is this explicit **kwargs necessary? I don't see it used in either `OAuthSSHChannel` or `SSHChannel`

parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,140 @@
...
+        ''' Initialize a persistent connection to the remote system.
+        We should know at this point whether ssh connectivity is possible
+
+        Args:
+            - hostname (String) : Hostname
+
+        KWargs:
+            - username (string) : Username on remote system
```
benclifford: these docs don't align with the actual function signature

parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,140 @@
...
+            self.service = SSHService(hostname, port)
+            self.transport = self.service.login(access_token, username)
+
+        except Exception:
+            logger.exception("Caught an in OAuth authentication step with {}".format(hostname))
+            raise
+
+        self.sftp_client = paramiko.SFTPClient.from_transport(self.transport)
+
+    def execute_wait(self, cmd, walltime=2, envs={}):
```
benclifford: not sure what that walltime actually sets - but other timeouts in parsl on that scale have turned out to be way too low in real usage.

parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,140 @@
...
+    from oauth_ssh.oauth_ssh_token import find_access_token
+    _oauth_ssh_enabled = True
+except (ImportError, NameError, FileNotFoundError):
+    _oauth_ssh_enabled = False
+
+
+logger = logging.getLogger(__name__)
+
+
+class OAuthSSHChannel(SSHChannel):
```
benclifford: UI-wise this might be better named GlobusSSHChannel, if that's really what this channel is for

setup.py
```diff
@@ -27,6 +27,7 @@
         'monitoring' : ['psutil', 'sqlalchemy', 'sqlalchemy_utils'],
         'aws' : ['boto3'],
         'kubernetes' : ['kubernetes'],
+        'globus_ssh' : ['globus-ssh>=0.5'],
```
benclifford: I'm not sure what is happening here: pip in my two dev envs (3.5 and 3.7) cannot find globus-ssh, and although google gives a link to it on pypy, that link resolves to a 404 for me at the moment: https://pypi.org/project/globus-ssh/

That means I can't install the required dependencies.

parsl/channels/oauth_ssh/oauth_ssh.py
yadudoc: It is now oauth-ssh with code hosted here : https://github.com/xsede/oauth-ssh
There's some branding thing going on.

setup.py
yadudoc: Updated to use the 'oauth-ssh' package.

parsl/channels/oauth_ssh/oauth_ssh.py
yadudoc: Updating this to 60s

benclifford:
This is failing with a mypy error at the moment as mypy cannot find type information for the oauth client.
Add a block to `mypy.ini` to tell it to ignore that module - that's how basically all other third party modules are treated there.

parsl/channels/oauth_ssh/oauth_ssh.py
```diff
@@ -0,0 +1,137 @@
...
+        except Exception:
+            logger.exception("Failed to find the access token for {}".format(hostname))
+            raise
+
+        try:
+            self.service = SSHService(hostname, port)
+            self.transport = self.service.login(access_token, username)
+
+        except Exception:
+            logger.exception("Caught an in OAuth authentication step with {}".format(hostname))
```
annawoodard: "Caught an in" -> "Caught an exception in"

parsl/channels/oauth_ssh/oauth_ssh.py
yadudoc: I don't think that's necessary. I believe that's for when modules might be coming from local dirs.

yadudoc: Fixed.

parsl/channels/oauth_ssh/oauth_ssh.py
yadudoc: Done.

PR #947, opened 2019-05-21
Make staging use the DFK associated with relevant data manager

benclifford:
Previously staging used the global default DFK

In real life I think that using the global DFK is likely to result in
non-deterministic obscure staging errors when multiple DFKs are in use
(for example, in notebooks) as staging tries to use a different DFK that
may be in some unexpected state.


parsl/data_provider/data_manager.py
```diff
@@ -149,30 +149,30 @@ def stage_in(self, file, executor):
 
         if file.scheme == 'ftp':
             working_dir = self.dfk.executors[executor].working_dir
-            stage_in_app = self._ftp_stage_in_app(executor=executor)
+            stage_in_app = self._ftp_stage_in_app(executor=executor, dfk=self.dfk)
```
annawoodard: Would we ever call this with `dfk != self.dfk`? If not, maybe we should remove the dfk keyword argument altogether and hardcode it the method definition?

benclifford: I think that would be a reasonable change -- I'll make a commit.

PR #962, opened 2019-05-22
Memory limits htex

yadudoc:
Applying new options to specify memory limits per worker for HTEX.


parsl/executors/high_throughput/.#process_worker_pool.py
```diff
@@ -0,0 +1 @@
+yadu@borgmachine2.26713:1557822238
```
ZhuozhaoLi: Should remove this.

parsl/executors/high_throughput/executor.py
```diff
@@ -152,6 +157,7 @@ def __init__(self,
                  working_dir: Optional[str] = None,
                  worker_debug: bool = False,
                  cores_per_worker: float = 1.0,
+                 mem_per_worker: float = 1.0,
```
ZhuozhaoLi: default to 1.0 or None? None seems better as the docstring mentions.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -115,9 +122,12 @@ def __init__(self,
         self.block_id = block_id
 
         cores_on_node = multiprocessing.cpu_count()
+        available_mem_on_node = round(psutil.virtual_memory().available / (2**30), 1)
+
         self.max_workers = max_workers
         self.prefetch_capacity = prefetch_capacity
         self.worker_count = min(max_workers,
+                                math.floor(available_mem_on_node / mem_per_worker),
```
ZhuozhaoLi: I think it should check if `mem_per_worker` is not 0 or none?

yadudoc: Fixed.

parsl/executors/high_throughput/executor.py
yadudoc: I've set this to Optional[float] = None

ZhuozhaoLi:
It seems something is wrong with the travis build right now. 

```
$ git diff --name-only FETCH_HEAD...$TRAVIS_BRANCH | grep -qvE '(\.rst$)|(^(docs))/' || {
    echo "Only docs were updated, stopping build process."
    exit
  }
  
Only docs were updated, stopping build process.
```

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -474,6 +490,8 @@ def set_stream_logger(name='parsl', level=logging.DEBUG, format_string=None):
                         help="Block identifier for Manager")
     parser.add_argument("-c", "--cores_per_worker", default="1.0",
                         help="Number of cores assigned to each worker process. Default=1.0")
+    parser.add_argument("-m", "--mem_per_worker", default=0,
+                        help="GB of memory assigned to each worker process. Default=0, no assignment")
```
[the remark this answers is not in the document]
yadudoc: Agreed. I'm fixing this to be consistent to use `None` to avoid memory in the calculation of worker count across interfaces.

test-requirements.txt
```diff
@@ -13,3 +13,4 @@ sqlalchemy
 sqlalchemy_utils
 mypy
 sqlalchemy-stubs
+psutil
```
benclifford: This is a requirement of HTEX, not of the test suite, I think? (process-worker-pool wants to import psutil...)

yadudoc: I've fixed this now.

PR #963, opened 2019-05-22
remove __del__ and merge it to close method in MonitoringHub

### LINKED ISSUE #827: 0.7.2 monitoring hangs in pytest
With the config added in benc-pytest-monitoring, 5762f434f518488881f3dae9cd18d29a25423c4c

pytest hangs after test completion:

```
$ pytest parsl/tests/test_bash_apps/test_apptimeout.py  --config parsl/tests/configs/local_threads_monitoring.py
============================= test session starts ==============================
platform linux -- Python 3.7.2+, pytest-4.0.2, py-1.8.0, pluggy-0.9.0
rootdir: /home/benc/parsl/src/parsl/parsl/tests, inifile: pytest.ini
plugins: xdist-1.26.1, ordering-0.6, forked-1.0.2, cov-2.6.1
collected 1 item                                                               

parsl/tests/test_bash_apps/test_apptimeout.py .                          [100%]

=========================== 1 passed in 3.65 seconds ===========================
```

The corresponding monitoring _hub.log shows:

```
2019-03-18 17:32:07.166 monitoring_hub:180 [INFO]  Monitoring Hub initialized
2019-03-18 17:32:07.774 monitoring_hub:232 [INFO]  Terminating Monitoring Hub
2019-03-18 17:32:07.774 monitoring_hub:235 [INFO]  Waiting Hub to receive all messages and terminate
2019-03-18 17:32:10.791 monitoring_hub:238 [INFO]  Received STOP from Hub
2019-03-18 17:32:10.791 monitoring_hub:241 [INFO]  Terminating Hub
2019-03-18 17:32:10.835 monitoring_hub:232 [INFO]  Terminating Monitoring Hub
2019-03-18 17:32:10.835 monitoring_hub:235 [INFO]  Waiting Hub to receive all messages and terminate

```

which looks as if the monitoring hub is trying to shut down twice.

Vague hypothesis: perhaps once when DFK is closed, and once when `__del__` is called by the garbage collector?

It's possible that this doesn't happen in manual tests because the garbage collector probably never tries to collect a MonitoringHub (or anything else) because the process is ready to exit; but if my hypothesis is right, this might cause similar hanging behaviour in interactive use.

ZhuozhaoLi:
fix #827.

Tested locally.


parsl/monitoring/monitoring.py
```diff
@@ -229,7 +229,7 @@ def send(self, mtype, message):
         self.logger.debug("Sending message {}, {}".format(mtype, message))
         return self._dfk_channel.send_pyobj((mtype, message))
 
-    def __del__(self):
+    def close(self):
```
yadudoc: I think there's value in having a __del__ method, for eg when something bad happens and close() isn't called.
We could easily avoid the double call to __del__ in #827 by adding a flag to avoid a second call.

ZhuozhaoLi: fixed. I now use a tag `monitoring_hub_active` to check if it is closed.

ZhuozhaoLi:
Maybe @benclifford  should submit a PR for his `benc-pytest-monitoring` branch mentioned at #827 after this merge.

parsl/monitoring/monitoring.py
```diff
@@ -245,7 +247,7 @@ def __del__(self):
             self.priority_msgs.put(("STOP", 0))
 
     def close(self):
-        return self.__del__()
+        self.__del__()
```
benclifford: I think the style for this should be `__del__` calls `close`, not `close` calls `__del__`

I'm still a little concerned about the ordering of events at process exit / garbage collection time: will we definitely get something back from the Hub? Or can it have been destroyed before __del__ is called? (i.e. what causes this __del__ to be called before whatever gets rid of the other end of the queue?)

ZhuozhaoLi: I switched the `__del__` and `close`. 

I think your concern is right. Currently it works okay so far. But we certainly need to make sure everything for monitoring is garbage collected later. I will create a new issue #963 to keep track of that.

PR #967, opened 2019-05-24
attempting to simplify configuration wording

danielskatz:


docs/userguide/configuring.rst
```diff
@@ -71,7 +88,7 @@ Here are a series of question to help formulate a suitable configuration:
 |                     | * `HighThroughputExecutor` |                        |
 +---------------------+----------------------------+------------------------+
 
-2. How many nodes do you have to execute them ? What task durations give good performance on different executors?
+2. How many nodes will you use to run them? What task durations give good performance on different executors?
```
danielskatz: I've tried not to use "task" here, but in this particular place, I'm not sure that "app" would make sense.

docs/userguide/configuring.rst
```diff
@@ -133,21 +150,22 @@ Here are a series of question to help formulate a suitable configuration:
...
-| Laptop/Workstation  | Cloud Resources          | None                               |
-+---------------------+--------------------------+------------------------------------+
-| Laptop/Workstation  | Clusters with no 2FA     | `SSHChannel`                       |
-+---------------------+--------------------------+------------------------------------+
-| Laptop/Workstation  | Clusters with 2FA        | `SSHInteractiveLoginChannel`       |
-+---------------------+--------------------------+------------------------------------+
-| Login node          | Cluster/Supercomputer    | `LocalChannel`                     |
-+---------------------+--------------------------+------------------------------------+
+4. Where will you run the main Parsl program, given that you already have determined where the apps will run?
+(This is needed to determine how to communicate between the Parsl program and the apps.)
```
danielskatz: It's not clear to the reader what happens over a channel.  Is this where inputs and outputs are passed?  containerized apps? files?  I'm not sure where we explain this, but it would be useful to probably say a little more here.

docs/userguide/configuring.rst
yadudoc: This sounds better. As it is, we've used app/task pretty interchangeably. Though I tend to refer to an app executing on a worker as a task.

docs/userguide/configuring.rst
yadudoc: I'm not sure if that level of detail is necessary/useful to the user while deciding what config pieces to use.

annawoodard: @danielskatz agreed, but can you split this off into either a PR which adds the proposed content or an issue to remind us to add these details? We don't want to hold up merging your nice improvements above by waiting on more work to be done.

PR #970, opened 2019-05-28
Support automatic task logging

annawoodard:


annawoodard:
@yadudoc No that was a mistake, but that commit has since been merged, so no point in pulling it out now.

PR #980, opened 2019-05-28
Manager errors #698

### LINKED ISSUE #698: Clarify warning message when a manager goes missing
Currently in the parsl log, if a manager goes missing, the user sees:
```
Remote exception description: b'da6348e1c3f1'
```

While in the interchange log you see the more helpful
```
2018-12-11 14:50:09 interchange:376 [WARNING]  [MAIN] Too many heartbeats missed for manager b'da6348e1c3f1'
```

We should improve the former.

yadudoc:
Adds __str__ magic methods missing from exceptions so that there's cleaner exception reporting in the event of system failures like: Manager dying, managers connecting with malformed registration messages etc. Fixes #698.

Fixes a KeyError raised due to a missing value in a dict.


parsl/executors/high_throughput/interchange.py
```diff
@@ -461,7 +473,7 @@ def start(self, poll_period=None):
                     logger.debug("[MAIN] Current tasks: {}".format(self._ready_manager_queue[manager]['tasks']))
                 logger.debug("[MAIN] leaving results_incoming section")
 
-            logger.debug("[MAIN] entering bad_managers section")
+            # logger.debug("[MAIN] entering bad_managers section")
```
annawoodard: prefer to remove dead code

parsl/tests/manual_tests/test_htex_worker_loss.py
```diff
@@ -0,0 +1,130 @@
...
+    x = platform(sleep=dur)
+    x.result()
+
+    # At this point we know that the worker is connected.
+    # Now we need to kill the network to mimic a n/w failure.
+    time.sleep(0.1)
+    print('*' * 80)
+    print("Manually kill the network to create a n/w failure")
+    print("We should see suppression of the failure")
+    print('*' * 80)
```
annawoodard: Where is the kill happening? I see a `provider.cancel` in `test_simple` but that isn't called. Is the idea the tester should do this by hand?

parsl/tests/manual_tests/test_htex_worker_loss.py
```diff
@@ -0,0 +1,130 @@
+import argparse
```
annawoodard: I am skeptical of tests that aren't plugged into CI-- they are of limited value if they are never run. We could consider re-designing this test to assert something, add the `@pytest.mark.local` decorator so that it is only run once, with the locally defined config, and move it to a pytest-accessible directory.

parsl/tests/manual_tests/test_htex_worker_loss.py
yadudoc: The kill needs to happen manually. You need to manually disconnect the work by cutting network connection.

parsl/tests/manual_tests/test_htex_worker_loss.py
yadudoc: This is not going into CI. To engineer the situation that triggers the failure condition is way too much work to try automating, so I'd rather defer on that but still have this test exist so that I may not have to redo this from scratch.

parsl/tests/manual_tests/test_htex_worker_loss.py
benclifford: This description should probably turn into a comment at the top of the test file for future readers.

PR #984, opened 2019-05-28
Add bash script for travis testing

annawoodard:


benclifford:
This is failing because of an awkward interaction between `set -e` and pytest's exit code 5.

pytest is exiting with error code 5 in some situations, and we have previously been saying this is ok previously - but `set -e` makes this fail before the script has a chance to ignore it.

`set -e` makes the script fail at any command failure, apparently, but travis only fails when the end result of a single travis step fails.

I'm not sure what the best bash idiom to fix this is.

parsl/tests/run.sh
```diff
@@ -0,0 +1,52 @@
...
+./parsl/executors/workqueue/install-workqueue.sh
+export PATH=$PATH:/tmp/cctools/bin
+export PYTHONPATH=/tmp/cctools/lib/python3.5/site-packages
+
+# mpi
+bash parsl/executors/extreme_scale/install-mpi.sh $MPI
+if [[ "$MPI" == "mpich"   ]]; then mpichversion; fi
+if [[ "$MPI" == "openmpi" ]]; then ompi_info;    fi
+
+pip install .[extreme_scale,monitoring]
```
benclifford: if this `pip install` is happening in this file, then the original `pip install .` in the `install` section at line 24 should probably move into this file too.

parsl/tests/run.sh
```diff
@@ -0,0 +1,52 @@
...
+rm -f .coverage
+
+pytest parsl -k "not cleannet" --config parsl/tests/configs/htex_local.py --cov=parsl --cov-append --cov-report= --random-order
+pytest parsl -k "not cleannet" --config parsl/tests/configs/local_threads.py --cov=parsl --cov-append --cov-report= --random-order
+
+# some of the site/ tests require more dependencies. These are installed here as needed,
+# so that the above tests happen with only the basic requirements installed.
+
+# workqueue
+./parsl/executors/workqueue/install-workqueue.sh
```
benclifford: this and MPI install at line 30 don't need to happen on each run if a user is running this script in a persistent environment - maybe this can be parameterised on the commandline to make it happen or not? If I'm going to use this on the commandline in my dev environment I want this to run the tests, not mess round with what is installed, most of the time.

parsl/tests/run.sh
benclifford: possibly if a user is running with `pip install -e` then having any kind of `pip install` in this script is going to be frustrating. see parameterisation comment for line 25?

annawoodard:
abandoning for now

PR #990, opened 2019-05-29
a few more small changes in conf docs

danielskatz:


docs/userguide/configuring.rst
```diff
@@ -1,17 +1,17 @@
 .. _configuration-section:
 
-Configuration
+Environment Configuration
```
yadudoc: Environment is so overloaded that I'd rather not use it here.

docs/userguide/configuring.rst
```diff
@@ -1,17 +1,17 @@
...
 There are very many different execution environments in which Parsl programs and their apps can run, and
-mamy of these enviroments have multiple options of how those programs and apps run, which makes
+mamy of these enviroments have multiple options of how those Parsl programs and apps run, which makes
 configuration somewhat complex, and also makes determining how to set up Parsl's configuration
 for a particular set of choices fairly complex, though we think the actual configuration
 itself is reasonable simple.
 
 Parsl offers an extensible configuration model through which the execution environment and
-communication with that environment is configured. Parsl is configured
+communication within that environment is configured. Parsl is configured
```
yadudoc: There are two types of communication that you might be configuring: the channels that determine how you connect to the resource (where communication with makes sense), and the executor itself over which the user doesn't really configure/control the communication. So this with is more correct that within.

docs/userguide/configuring.rst
danielskatz: But configuration by itself is not enough.  Would execution be better?

docs/userguide/configuring.rst
danielskatz: If the main parsl process is part of the environment we are configuring (which I think it is), then within seems more appropriate to me, since nothing is outside the environment.

yadudoc: I can see that, there is the environment as a whole and multiple envs within it.

docs/userguide/configuring.rst
yadudoc: This section deals with Parsl configuration, and that's more than execution. We deal with data management, aspects of how to connect/authenticate, monitoring of tasks, logging. Parsl configuration is redundant, so I'd prefer to leave this as "Configuration".

danielskatz: Can we say anything more specific than configuration?  I feel like this is too general

danielskatz: @yadudoc - If you don't think there's anything better we can say, then l'll drop this part, but let's merge this PR soon ...

benclifford: drop it out of the PR and we can merge the rest now and argue about configuration phrasing in a different venue/PR

danielskatz: ok, it's out.

PR #995, opened 2019-05-30
Remove non-critical modules from requirements.txt

### LINKED ISSUE #994: Remove non-critical modules from requirements.txt
When we merged in monitoring, all of the monitoring specific modules went into the requirements.txt file. Ideally those should only be in the optional_extras section only.
Right now we try to install pandas, plotly and a bunch of extra stuff that might not be useful.

annawoodard:
Fixes #994.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,10 +7,3 @@ tblib
 ipykernel
 requests
 paramiko
-pydot
-networkx
-Flask
-flask_sqlalchemy
-pandas
-plotly
-python-daemon


requirements.txt
```diff
@@ -7,10 +7,3 @@ tblib
 ipykernel
 requests
 paramiko
-pydot
-networkx
-Flask
-flask_sqlalchemy
-pandas
-plotly
-python-daemon
```
ZhuozhaoLi: this `python-daemon` is still useful, right?

ZhuozhaoLi: Ignore that ..

PR #1003, opened 2019-05-31
Prevent htex logs from becoming huge when sending blobs as arguments.

lgray:
Truncate long arguments in logs for htex, this was causing logs to explode in size and also become useless.

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -503,7 +503,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
 
         # handle people sending blobs gracefully
         args_to_print = args
-        if logging.getEffectiveLevel() >= logging.DEBUG:
+        if logger.getEffectiveLevel() >= logging.DEBUG:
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -501,7 +501,9 @@ def submit(self, func, *args, **kwargs):
         self._task_counter += 1
         task_id = self._task_counter
 
-        logger.debug("Pushing function {} to queue with args {}".format(func, args))
+        # handle people sending blobs gracefully
+        args_to_print = tuple([arg if len(repr(arg)) < 100 else (repr(arg)[:100] + '...') for arg in args])
```
yadudoc: What do you think about creating args_to_print conditional on debug logs being enabled?

lgray: Makes perfect sense. Done.

benclifford:
There's a non-deterministic problem with one of the components of parsl that we recently started test - that problem shows up in anything based on recent master.

PR #1007, opened 2019-06-03
Fix monitoring docs

ZhuozhaoLi:


docs/userguide/monitoring.rst
```diff
@@ -14,12 +14,12 @@ that query this database to create a web-based dashboard for the workflow.
 
 To enable workflow monitoring support install::
```
yadudoc: We should take this line off. We no longer need any additional install step for workflow monitoring support.

ZhuozhaoLi: Removed this line and the pip install line

PR #1009, opened 2019-06-03
Fix vis datetime precision #979

### LINKED ISSUE #979: Improve datatime precision in monitoring.db
e.g., `workflow_duration`, etc.

ZhuozhaoLi:
this PR 

1. converts workflow_duration from `86401.12432s` to `1 day, 00:00:01`
2. trims the microseconds for `workflow_version`
3. displays the datetime object without microseconds on viz server.

fix #979

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -114,7 +114,7 @@ class DataFlowKernel(object):
                     self.workflow_name = fname
                     break
 
-        self.workflow_version = str(self.time_began)
+        self.workflow_version = str(self.time_began.replace(microsecond=0))
         if self.monitoring is not None and self.monitoring.workflow_version is not None:
             self.workflow_version = self.monitoring.workflow_version
 
--- a/parsl/monitoring/visualization/views.py
+++ b/parsl/monitoring/visualization/views.py
@@ -9,6 +9,21 @@ from parsl.monitoring.visualization.plots.default.workflow_resource_plots import
 
 dummy = True
 
+import datetime
+
+
+def format_time(value):
+    if value is not None:
+        if isinstance(value, float):
+            return str(datetime.timedelta(seconds=round(value)))
+        else:
+            return value.replace(microsecond=0)
+    else:
+        return value
+
+
+app.jinja_env.filters['timeformat'] = format_time
+
 
 @app.route('/')
 def index():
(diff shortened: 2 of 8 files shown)


parsl/monitoring/visualization/views.py
```diff
@@ -9,6 +9,21 @@
...
 
+import datetime
+
+
+def format_time(value):
+    if value is not None:
+        if isinstance(value, float):
+            return str(datetime.timedelta(seconds=round(value)))
+        else:
+            return value.replace(microsecond=0)
```
benclifford: This is assuming that if value isn't a `float`, it is something which has a meaningful `replace` method. I have some defensive coding preference for something more like the below, where the code is a bit more explicit about all of the types it expects value to be:

```
if isinstance(v, type1):
   ...
elif isinstance(v, type2):
  ....
else:
  raise unknown type error
```

parsl/dataflow/dflow.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, config=Config()):
                     self.workflow_name = fname
                     break
 
-        self.workflow_version = str(self.time_began)
+        self.workflow_version = str(self.time_began.replace(microsecond=0))
```
benclifford: This information doesn't need destroying if it gets dropped in the UI later (which it does) ?

parsl/monitoring/visualization/views.py
ZhuozhaoLi: see if current changes make sense to you

benclifford: yes that looks basically what I was thinking. is returning that value rather than raising an exception the right thing to do? I guess the message will get printed on the page rather than crashing page generation?

Also on re-reading, the outer `if value is None` could fit in this multiway `if` too, rather than being a separate wrapping `if`: it's a case that considers all of the possible things that `value` could be - `None` is one such case, a `float` is another, and a `datetime` is another.

parsl/dataflow/dflow.py
ZhuozhaoLi: This `workflow_version` is a customizable variable in `monitoring` -- users can input something like a string. Thus, this value is stored in the database as `Text`, rather than `float` or `datetime` type.

I think we do not want to drop microsecond at the UI since we have no idea this is a customized string or a datetime string. So, I choose to destroy before it is commited to database.

benclifford: ok

parsl/monitoring/visualization/views.py
ZhuozhaoLi: good catch! I corrected the `if` statements.

I am not sure if printing a message is the right thing - but my thought is that a single incorrect time format should not crash the page and disallow users from viewing other things.

Issue #708, opened 2018-12-14
move_files abstraction in #683 and #707 is not the right abstraction


benclifford:
#683 and #707 provide a move_files parameter that stops ssh being used to copy a file over the top of itself.

this abstraction feels wrong to me.

it stems from an assumption that something submitted over a "remote" channel is also somehow on a "remote" filesystem (remote deliberately in "quotes").

in the cases that needed #683 and #707 this was not the case: job execution was "remote" but file access was "local" due to a shared file system.

annawoodard:
At the moment `move_files` is a keyword arg to the local provider and the slurm provider-- it should at least be consistent across all?

I got burned by this recently while preparing an example of an ad-hoc config for our docs. I was testing a case with a shared filesystem and had forgotten to disable `move_files`. If all we're trying to solve is not copying a file on top of itself, would it be sufficient to check if the the device number and inode number of the file matches on the 'local' and 'remote' side? If so, we could remove the 'move_files' keyword arg and reduce the number of options the user needs to configure correctly.

PR #1020, opened 2019-06-06
Remove the `move_files` keyword arg.

annawoodard:


parsl/channels/base.py
```diff
@@ -137,3 +138,10 @@ def abspath(self, path):
             Path for which the absolute path will be returned.
         """
         pass
+
+    def is_same_file(self, local, remote):
```
annawoodard: Oops just noticed this is missing a docstring, I'll add it after 0.8

benclifford:
Those values aren't globally scoped over all mounts of a shared file system. I thought I'd have to come up with some implausible corner case but it turns out:

```
bxc@cori02:~/pipeplay/tutwf> stat -c "%d-%i"  ~/pipeplay/tutwf/README.me
51-359900980
```

```
bxc@cori04:~> stat -c "%d-%i"  ~/pipeplay/tutwf/README.me
56-359900980
```

annawoodard:
Good point @benclifford. I'm closing this for now. Ben has me mostly convinced via slack that a declarative model is better than trying to figure this out for the user. I definitely think we need something better than `move_files` but let's keep that conversation in #708.

Issue #987, opened 2019-05-29
Capture stderr/stdout/outputs in monitoring


yadudoc:
Capture output file paths specified to `stderr`, `stdout`, `inputs`, `outputs` in monitoring. 
We should only capture `File` types from inputs/outputs to avoid overrunning the DB with python objects.

Split out from #870

yadudoc:
@ZhuozhaoLi I thought this was done ?

ZhuozhaoLi:
> @ZhuozhaoLi I thought this was done ?

This was partially fixed in PR #1026, since there is still an open discussion #1016 about what we should capture in monitoring.

Issue #1016, opened 2019-06-05
Should `inputs`, `outputs` be allowed to contain raw strings?


benclifford:
Sometimes (for example in parts of the tutorial) `inputs` is given a list of `str` typed filenames. Sometimes it is given a list of `File` objects.

What is the meaning of passing in `str` typed filenames? For example, does that mean no staging should happen and every component involved should assume that the file is accessible at this path without further action? If so, is that the route we want to introduce files to people in the tutorial?

The immediate reason this issue comes up is because since #880, `File` objects are no longer `str` objects and so some code which was able to assume `inputs` were always `str` is now invalid - 
for example, code fragments like `" ".join(inputs)` to generate a space separated list of filenames (as long as those filenames don't contain spaces) would work for `str` inputs and for `File` inputs.

cc @danielskatz

danielskatz:
given that the tutorial doesn't do any staging, I think we just did things with strings to demonstrate parsl features that were not really related to files at all, but used files as a convenient (local) tool.  I don't know that it would make sense to think of these as Parsl Files.

annawoodard:
If we want to make good on the 'write once, run anywhere' dream, I think that we would need to move towards always passing `File` objects. Making a simplifying assumption (strings means everything is local) adds mental load for users trying to understand what's going on-- they need to learn both the special (local execution) and the general (anywhere execution) case.

benclifford:
The phrasing in #987 suggests that it is OK to pass in other non-`File` objects through `inputs` and `outputs`. My preference is that this also should be prohibited: python already has a non-parsl specific way to pass arbitrary objects to and from function calls.

benclifford:
crossref #1125 which is about behaviour when an input is a `DataFuture`, not a `File`.

benclifford:
As an example of this, the test `parsl/tests/test_python_apps/test_mapred.py` passes AppFutures in the `inputs=` parameter - see https://github.com/Parsl/parsl/blob/fd1992fb569d68bcfd05dd0055c73843c087f0ae/parsl/tests/test_python_apps/test_mapred.py#L36

This does not fit the style of usage I am expecting for `inputs` - instead those futures should somehow be passed in as regular function parameters... although the presence of those future in a list rather than as individual values makes things awkward.

annawoodard:
> although the presence of those future in a list rather than as individual values makes things awkward.

[...]

PR #1026, opened 2019-06-09
fix stderr for monitoring

ZhuozhaoLi:
Minor fixes for stderr for monitoring
partially for #987

[Discussion?] Issue #987 mentions only `File` type inputs/outputs should be logged. But currently, the monitoring logs any types of inputs, and only the outputs that are already specified when the app is submitted (actually only `str` type as a file path I think).
Do you think the monitoring should be kept in current version right now?

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -188,7 +188,7 @@ class DataFlowKernel(object):
         task_log_info['tasks_completed_count'] = self.tasks_completed_count
         task_log_info['task_inputs'] = str(self.tasks[task_id]['kwargs'].get('inputs', None))
         task_log_info['task_outputs'] = str(self.tasks[task_id]['kwargs'].get('outputs', None))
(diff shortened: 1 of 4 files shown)


benclifford:
If we're logging stdout/stderr paths, we should also log stdin paths as before - that is, it is OK for this PR to add in a capture for stderr path, but it is not OK for this PR to remove the capture of the stdin path.

> [Discussion?] Issue #987 mentions only File type inputs/outputs should be logged. But currently, the monitoring logs any types of inputs, and only the outputs that are already specified when the app is submitted (actually only str type as a file path I think).

The input/output file types are a little informal. If a user passes in a `File` object, staging behaviour can happen. However, they can pass in arbitrary other python objects without error (I think), and staging will not happen (and is not even defined). #1016 contains some discussion on this.

ZhuozhaoLi:
> don't remove stdin

Fixed now.

Issue #992, opened 2019-05-30
Remove keyword substitution from bash apps


benclifford:
The value returned from a bash app passes through some keyword substitution for being evaluated.

This is rarely used, and means that the string returned by a bash app has slightly different syntax from raw bash - `{` cannot be used in the same way in the two contexts.

Remove this implicit formatting.

Any substitutions that can be done this way can be done more directly in the user supplied user code (although with increased complexity) - as well as any other style of dynamic generation of bash code.

annawoodard:
:+1: * 1000000

benclifford:
Fix in #1034

PR #1034, opened 2019-06-10
Remove implicit {}-formatting of bash app command lines.

benclifford:
This will be a breaking change for users with { symbols in their
existing bash_app command lines - either for getting literal {
symbols in the command line (in which case, a layer of escaping
should be removed) or to perform arg substition (the functionality)
just removed, in which case users should perform that substitution
explicitly in their bash_app code.

See issue #992


benclifford:
There's a test failure here that doesn't fail on my laptop - I think it's because that particular test case looks for a file to come into existence but doesn't ensure it doesn't exist before the test starts - so in a fresh run directory, the test fails, but in a directory where the test has passed in the past, it will now pass forever. I'm distracted on something higher priority now so leaving this as a note for posterity.

PR #1047, opened 2019-06-11
0.8.0 doc updates

yadudoc:
Updates configs for:
* Comet
* Cori
* Stampede2
* Theta
* CC-IN2P3
* AWS

### DIFF
--- a/parsl/configs/ec2_htex_single_node.py
+++ b/parsl/configs/ec2_htex_single_node.py
@@ -1,0 +1,36 @@
+"""Config for EC2.
+
+Block {Min:0, init:1, Max:1}
+==================
+| ++++++++++++++ |
+| |    Node    | |
+| |            | |
+| | Task  Task | |
+| |            | |
+| ++++++++++++++ |
+==================
+
+"""
+from parsl.config import Config
+from parsl.providers import AWSProvider
+from parsl.executors import HighThroughputExecutor
+from parsl.addresses import address_by_query
+
+config = Config(
+    executors=[
+        HighThroughputExecutor(
+            label='ec2_single_node',
+            address=address_by_query(),
+            provider=AWSProvider(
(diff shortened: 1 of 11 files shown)


parsl/configs/ec2_htex_single_node.py
```diff
@@ -0,0 +1,36 @@
...
+from parsl.executors import HighThroughputExecutor
+from parsl.addresses import address_by_query
+
+config = Config(
+    executors=[
+        HighThroughputExecutor(
+            label='ec2_single_node',
+            address=address_by_query(),
+            provider=AWSProvider(
+                image_id='YOUR_AMI_ID',     # Please replace image_id with your image id, e.g., 'ami-82f4dae7'
```
annawoodard: Remove 'please's, consider removing comments entirely, they aren't adding any info

parsl/configs/theta_local_htex_multinode.py
```diff
@@ -0,0 +1,30 @@
...
+            max_workers=4,
+            address=address_by_hostname(),
+            provider=CobaltProvider(
+                queue="debug-flat-quad",
+                launcher=AprunLauncher(),
+                walltime="00:30:00",
+                nodes_per_block=2,
+                init_blocks=1,
+                max_blocks=1,
+                scheduler_options='',     # Input your scheduler_options if needed
```
annawoodard: This isn't very informative, I favor a description of scheduler_options instead, for example:
```
# string to prepend to #SBATCH blocks in the submit script to the scheduler
```

parsl/configs/ec2_htex_single_node.py
yadudoc: Fixed.

parsl/configs/theta_local_htex_multinode.py
yadudoc: Agreed. I've updated all configs from my test group. Will update the rest once they are merged.

PR #1052, opened 2019-06-11
add cooley htex config

ZhuozhaoLi:


parsl/configs/cooley_htex_multinode.py
```diff
@@ -0,0 +1,31 @@
...
+config = Config(
+    executors=[
+        HighThroughputExecutor(
+            label="cooley_htex",
+            worker_debug=False,
+            cores_per_worker=1,
+            address=address_by_hostname(),
+            provider=CobaltProvider(
+                queue='debug',
+                account='',  # project name to submit the job
```
annawoodard: Can the config run without this? If not, let's rename `YOUR_ACCOUNT`

ZhuozhaoLi: It does not work. Fixed now

PR #1061, opened 2019-06-13
0.8.0 release prep

yadudoc:
Updates version info and documentation in preparation for 0.8.0 upgrade.


docs/reference.rst
```diff
@@ -24,6 +24,7 @@ Reference guide
     parsl.executors.swift_t.TurbineExecutor
     parsl.channels.LocalChannel
     parsl.channels.SSHChannel
+    parsl.channels.OAuthSSHChannel
```
benclifford: this is not exported from `parsl.channels`

benclifford: wait yes it is.

PR #1063, opened 2019-06-14
Fix flake8 F811 violations and remove F811 exemption

benclifford:
Where two functions with the same name are defined differently, the second
one is used as that seems to reflect how python actually does things.

### DIFF
--- a/parsl/tests/integration/test_channels/test_ssh_errors.py
+++ b/parsl/tests/integration/test_channels/test_ssh_errors.py
@@ -29,17 +29,6 @@ def test_error_3():
     ''' This should work
     '''
     try:
-        connect_and_list("login.mcs.anl.gov", "yadunand")
-    except AuthException as e:
-        print("Caught exception : ", e)
-    else:
-        assert type(e) == BadHostKeyException, "Expected SSException, got :{0}".format(e)
-
-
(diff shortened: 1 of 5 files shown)


parsl/tests/integration/test_channels/test_ssh_errors.py
```diff
@@ -25,17 +25,6 @@ def test_error_2():
         raise Exception("Expected SSException, got :{0}".format(e))
 
 
-def test_error_3():
```
yadudoc: Why is this being removed? Can we not just rename it?

benclifford: This test doesn't ever run - it's shadowed by the subsequent definition. This PR was basically removing all such code as being effectively commented out and dead.

The version that runs on `login.mcs.anl.gov` was introduced in  64fa6ab64b on  Oct 12 17:24:25 2017, but then shadowed a few minutes later by the `edison` version in e6190c833f.

I'd be fine with the `login.mcs.anl.gov` test being re-enabled by renaming the method, if that's still a sensible test - these tests are pretty specific to @yadudoc 's accounts so I don't really have the ability to verify re-enabling test_error_3.

PR #1073, opened 2019-06-17
refactor kubernetes providers: each block in kube refers to a pod

### LINKED ISSUE #853: scale out mechanism in HTEX not compatible with kubernetes provider
Currently, the `scale_out` mechanism in HTEX is to scale out block by block, as in https://github.com/Parsl/parsl/blob/a00acdd411c44777a01d79b0f9cde2cb51e27aff/parsl/executors/high_throughput/executor.py#L480-L482

This is conflicted with the mechanism in kubernetes provider. In kube provider, the number of blocks refer to the number of pods, and it creates the blocks (pods) directly, as in 
https://github.com/Parsl/parsl/blob/a00acdd411c44777a01d79b0f9cde2cb51e27aff/parsl/providers/kubernetes/kube.py#L126-L131

Need to fix to make them compatible.

ZhuozhaoLi:
fix #853.

Now kubernetes provider creates non-deployment pods for each executor, rather than deployments. 
The scaling of pods will be managed by Parsl. 

Refer the details to the discussions in #853


parsl/providers/kubernetes/kube.py
```diff
@@ -65,7 +65,7 @@ def __init__(self,
                  max_blocks=10,
                  parallelism=1,
                  worker_init="",
-                 deployment_name=None,
+                 pod_name=None,
```
benclifford: needs docstring

parsl/providers/kubernetes/kube.py
```diff
@@ -161,9 +159,10 @@ def cancel(self, job_ids):
         for job in job_ids:
             logger.debug("Terminating job/proc_id: {0}".format(job))
             # Here we are assuming that for local, the job_ids are the process id's
-            self._delete_deployment(job)
```
benclifford: Maybe remove _delete_deployment from the source file?

parsl/providers/kubernetes/kube.py
```diff
@@ -110,31 +110,29 @@ def submit(self, cmd_string, blocksize, tasks_per_node, job_name="parsl"):
...
-            if not self.deployment_name:
-                deployment_name = '{}-deployment'.format(job_name)
-            else:
-                deployment_name = '{}-{}-deployment'.format(self.deployment_name,
-                                                            cur_timestamp)
-
-            formatted_cmd = template_string.format(command=cmd_string,
-                                                   worker_init=self.worker_init)
-
-            self.deployment_obj = self._create_deployment_object(job_name,
```
benclifford: _create_deployment and _create_deployment_object are dead code that should be removed maybe now.

parsl/providers/kubernetes/kube.py
```diff
@@ -181,6 +180,93 @@ def _status(self):
...
         # do something to get the deployment's status
 
+    def _create_pod(self,
+                    image,
+                    pod_name,
+                    job_name,
+                    port=80,
+                    cmd_string=None,
+                    volumes=[]):
+        """ Create a kubernetes non-deployment pod for the job.
```
benclifford: Refer to this as this a "pod", not a "non-deployment pod".

parsl/providers/kubernetes/kube.py
```diff
@@ -181,6 +180,93 @@ def _status(self):
...
+        environment_vars = client.V1EnvVar(name="TEST", value="SOME DATA")
+
+        launch_args = ["-c", "{0};".format(cmd_string)]
+
+        volume_mounts = []
+        # Create mount paths for the volumes
+        for volume in volumes:
+            volume_mounts.append(client.V1VolumeMount(mount_path=volume[1],
+                                                      name=volume[0]))
+        # Configureate Pod template container
```
benclifford: "configure"?

parsl/providers/kubernetes/kube.py
```diff
@@ -181,6 +180,93 @@ def _status(self):
...
+        container = None
+        if security_context:
```
benclifford: I think the only purpose of this is to supply or not supply the `security_context` parameter. It looks (from the generated python code of `V1Container` in https://github.com/kubernetes-client/python/blob/master/kubernetes/client/models/v1_container.py) that it is fine to pass in `None` for `security_context` and so this `if`-statement could collapse into a single code path.

benclifford:
This looks ok but I'd like someone to confirm they've definitely run this code and seen it scale correctly.

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: fixed

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: fixed

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: removed

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: fixed

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: fixed

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: collapsed

ZhuozhaoLi:
> This looks ok but I'd like someone to confirm they've definitely run this code and seen it scale correctly.

@benclifford I have tested this on petrelkube with DLHub. This includes the following tests.

1. Pods terminated when the main parsl process exits
2. Pods terminated when there is no task
3. Pods scaled out after termination when there are tasks

PR #1075, opened 2019-06-18
Add command timeout for Condor provider

### LINKED ISSUE #891: Condor provider has a hard-coded `cmd_timeout`.
This timeout should be configurable and probably default to something much longer.

https://github.com/Parsl/parsl/blob/1d8c9e35752274af6ae2ce2f07107474ece4561c/parsl/providers/condor/condor.py#L225

cc @ZhuozhaoLi who noted this in a comment to #889

annawoodard:
Fixes #891.

### DIFF
--- a/parsl/providers/condor/condor.py
+++ b/parsl/providers/condor/condor.py
@@ -81,8 +81,6 @@ class CondorProvider(RepresentationMixin, ClusterProvider):
                  cmd_timeout=60):
 
         label = 'condor'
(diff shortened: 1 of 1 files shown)


annawoodard:
Note @lgray that now in the latest Parsl you can use the `cmd_timeout` keyword arg to the `CondorProvider` to increase the command timeout. This was intermittently causing failures for me at Wisconsin.

PR #1083, opened 2019-06-20
Make bash apps use unique log files - fixes #1043

### LINKED ISSUE #1043: remote_side_bash_executor global logger conflates logs from multiple tasks/sources
`remote_side_bash_executor` configures and uses the global `logging` logger, rather than one scoped to that function. Once it has configured a log file for output, all subsequent global logs from any further `remote_side_bash_executor` invocation in that process, as well as other uses of logging, such as htex `process_worker_pool.py`, end up in earlier configured log files.

This results in `/tmp/bashexec` logs containing a confused assortment of logs from different sources, rather than being focuses on a single bash execution.

benclifford:


parsl/app/bash.py
```diff
@@ -21,8 +18,20 @@ def remote_side_bash_executor(func, *args, **kwargs):
     import subprocess
     import logging
     import parsl.app.errors as pe
+    from parsl import set_file_logger
+
+    logbase = "/tmp"
```
annawoodard: Would it be more polite to collect these in a dedicated directory? `/tmp/parsl_logs` or something?

benclifford: opened #1107 for something related

PR #1024, opened 2019-06-07
Add resource efficiency visualization

ZhuozhaoLi:
This PR

1. fix #931 -- monitoring now logs the node info to a NODE table (sent from interchange). Based on the node table, we compute/plot the CPU/memory resource efficiency, and display the figure to viz server

2. some minor changes to the gantt chart layout.

PR #1086, opened 2019-06-20
Task gantt plot improvement

ZhuozhaoLi:
This PR is part of the big PR #1024, which mainly improves the layout of the first `task_gantt` plot.

1. Remove useless `print`
2. Change y axis -- not showing task IDs on y axis in case there are too many tasks and texts overlap, instead task ID is shown in hover text.
3. Minor correction of time.

### DIFF
--- a/parsl/monitoring/visualization/views.py
+++ b/parsl/monitoring/visualization/views.py
@@ -56,9 +56,8 @@ def workflow(workflow_id):
                                 % (workflow_id), db.engine)
     task_summary = db.engine.execute(
         "SELECT task_func_name, count(*) as 'frequency' from task WHERE run_id='%s' group by task_func_name;" % workflow_id)
-    time_completed = None
-    if hasattr(workflow_details, 'time_completed'):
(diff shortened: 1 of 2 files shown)


parsl/monitoring/visualization/views.py
```diff
@@ -56,11 +56,13 @@ def workflow(workflow_id):
                                 % (workflow_id), db.engine)
     task_summary = db.engine.execute(
         "SELECT task_func_name, count(*) as 'frequency' from task WHERE run_id='%s' group by task_func_name;" % workflow_id)
-
+    time_completed = None
+    if hasattr(workflow_details, 'time_completed'):
```
benclifford: What are the circumstances where this `if` statement isn't true?

ZhuozhaoLi: Well, I think I met once before but I was not able to reproduce one for this `if` just now. So I will remove this `if` for now then.

ZhuozhaoLi: Fixed

PR #1087, opened 2019-06-20
add reg_time, cpu and memory, rename hostname in manager registration msg

ZhuozhaoLi:
This PR is a small part of PR #1024 . This includes:

1. rename `hname` to `hostname`

2. Add `reg_time` to manager registration message -- this specifically refers to the time on the manager node, but not the interchange node. I have seen the times on manager node and on interchange node have huge difference (e.g., on Theta). We need the manager-side time to infer the resource efficiency plot, as all the info like resource usage is collected on managers.

3. Add `cpu_count` and `total_memory` to registration message in HTEX.

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -358,8 +358,7 @@ class Interchange(object):
 
                     try:
                         msg = json.loads(message[1].decode('utf-8'))
-                        if 'reg_time' in msg:
-                            msg['reg_time'] = datetime.datetime.strptime(msg['reg_time'], "%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -357,6 +358,8 @@ def start(self, poll_period=None):
 
                     try:
                         msg = json.loads(message[1].decode('utf-8'))
+                        if 'reg_time' in msg:
```
benclifford: Other parts of this code assumes that fields always exist and let python's exceptions deal with the case they don't. I think that would be OK to do here too.

ZhuozhaoLi: Fixed.

Issue #1076, opened 2019-06-18
parsl-visualize should default to monitoring.db from pwd


benclifford:
parsl-visualise should default to using the monitoring db in pwd as that is the usual use case.

annawoodard:
Adding another item:
- [ ] Default to using the monitoring db in pwd
- [ ] Allow relative paths

benclifford:
sqlite already supports relative path uris, like this: `sqlite:///abc.db` with three rather than four slashes. This is the format used by default for a parsl run to record to `monitoring.db` in the current directory.

However this format appeared to not work with `parsl-visualize`. It looks like what is happening is the current directory is being set somewhere else:

```
$ parsl-visualize sqlite:///abc.db
 * Serving Flask app "parsl.monitoring.visualization.app" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:8080/ (Press CTRL+C to quit)

^C
$ find . -name abc.db
./parsl/monitoring/visualization/abc.db
```

PR #1088, opened 2019-06-21
Use monitoring.db in the current directory as visualization default.

benclifford:
Partially addresses #1076

### DIFF
--- a/parsl/monitoring/visualization/app.py
+++ b/parsl/monitoring/visualization/app.py
@@ -1,14 +1,15 @@
 from flask import Flask
 from parsl.monitoring.visualization.models import db
 import argparse
+import os
 
 
 def cli_run():
     """ Instantiates the Monitoring viz server
     """
     parser = argparse.ArgumentParser(description='Parsl visualization tool')
-    parser.add_argument('db_path', type=str,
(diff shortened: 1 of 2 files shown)


parsl/monitoring/visualization/app.py
```diff
@@ -1,14 +1,15 @@
...
 
 
 def cli_run():
     """ Instantiates the Monitoring viz server
     """
     parser = argparse.ArgumentParser(description='Parsl visualization tool')
-    parser.add_argument('db_path', type=str,
-                        help='Database path in the format sqlite:///<absolute_path_to_db>')
+    parser.add_argument('db_path', type=str, default="sqlite:///{cwd}/monitoring.db".format(cwd=os.getcwd()),
+                        nargs="?", help='Database path in the format sqlite:///<absolute_path_to_db>')
```
ZhuozhaoLi: Why do we need `nargs="?"` here?

benclifford: `nargs="?"` makes the `db_path` parameter optional, so that it can be omitted and the default used: without it, that `db_path` parameter defaults to needing exactly one value.

That's different behaviour to the other arguments because the others are options beginning with `-`, but `db_path` is just taken from the main command line text.

PR #1091, opened 2019-06-21
Add typeguard/mypy annotations on user-facing kubernetes initializer

benclifford:


ZhuozhaoLi:
This looks okay. But is it still a draft?

benclifford:
@ZhuozhaoLi mostly I left it as a draft because I haven't actually tried it on any real kubernetes configs as I'm not really set up for that in my own environment at the moment and there aren't any in the test suite.

ZhuozhaoLi:
> @ZhuozhaoLi mostly I left it as a draft because I haven't actually tried it on any real kubernetes configs as I'm not really set up for that in my own environment at the moment and there aren't any in the test suite.

> okay. I can test this if you could tell me how I can test appropriately.

benclifford:
All kubernetes configs which already work should work as before.

Passing in parameters of the wrong type should throw an exception right away. The `@typechecked` decorate adds on checks at __init__ time, so you only need to test as far as constructing/initialising the config object, rather than running a job all the way through kubernetes.

The main thing I want to flush out is kubernetes configs people are already using which don't actually match the type specification but still work (and so would break with this PR).

Issue #1008, opened 2019-06-03
bash_app executable info appears out of order wrt rest of output sometimes


benclifford:
I have observed in one run that bash_app executable information appears at the end of a stderr file, rather than at the start.

I think this is unintentional.

My superficial diagnosis is that it is because there is no `stderr.flush()` after the `executable follows` print statement but before `subprocess.Popen` which directs more stuff into stderr; so the relative ordering of the output of the two users of stderr is undefined.

PR #1092, opened 2019-06-24
Flush stderr after printing executable diagnostic

benclifford:
In some cases, this message wasn't being sent to stderr
before the stderr of the spawned executable, meaning the
executable diagnostic information appeared at the end,
rather than start, of the logged stderr.

This added flush should provide the necessary sequencing
to ensure the message appears before the stderr of the
spawned process.

See issue #1008


annawoodard:
Fixes #1008

PR #1103, opened 2019-06-27
Add node resource monitoring

ZhuozhaoLi:
This PR is a major breaking down of PR #1024 . It tries to add a node table to monitoring.db, which logs the node information including resource capacity, reg_time, python version, etc.

1. The node info is first collected by the interchange.
2. The interchange send data to the Hub (then send to db_manager). The db_manager logs all the data then.

To allow interchange to connect to the Hub, we need dfk to pass the hub address and hub port to executor and interchange.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -86,18 +86,22 @@ class DataFlowKernel(object):
         self.usage_tracker.send_message()
 
         # Monitoring
+        self.run_id = str(uuid4())
         self.tasks_completed_count = 0
         self.tasks_failed_count = 0
 
         self.monitoring = config.monitoring
+        # hub address and port for interchange to connect
+        self.hub_address = None
+        self.hub_interchange_port = None
         if self.monitoring:
             if self.monitoring.logdir is None:
                 self.monitoring.logdir = self.run_dir
-            self.monitoring.start()
+            self.hub_address = self.monitoring.hub_address
+            self.hub_interchange_port = self.monitoring.start(self.run_id)
 
         self.time_began = datetime.datetime.now()
         self.time_completed = None
-        self.run_id = str(uuid4())
 
         # TODO: make configurable
         logger.info("Run id is: " + self.run_id)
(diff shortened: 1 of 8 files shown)


parsl/dataflow/dflow.py
```diff
@@ -773,6 +777,8 @@ def first(t):
     def add_executors(self, executors):
         for executor in executors:
             executor.run_dir = self.run_dir
+            executor.hub_address = self.hub_address
+            executor.hub_port = self.hub_interchange_port
```
benclifford: can you add a note to parsl.providers.base that this property exists?

ZhuozhaoLi: I am not too sure that I understood -- why it goes to `parsl.providers.base` but not executor?
Do you have an example for that? This is actually a property automatically set by DFK, not set by user.

benclifford: you are right: executor base not provider base.

It should be documented there because some other people will want to implement an executor, or understand how executors will be driven by the DFK.

That might be "users" or it might be "parsl developers": workflow users probably won't call any methods on an executor at all.

`run_dir` is an example of a similar property that holds some DFK context that the executor runs with; or the `label` field. They are documented in different ways in `parsl.executor.base`

ZhuozhaoLi: I have added them to `parsl.executor.base`. See if that is what you expect.

parsl/monitoring/db_manager.py
```diff
@@ -141,6 +141,20 @@ class Task(Base):
...
+        run_id = Column('run_id', Text, nullable=False)
+        hostname = Column('hostname', Text, nullable=False)
+        cpu_count = Column('cpu_count', Integer, nullable=False)
+        total_memory = Column('total_memory', Integer, nullable=False)
+        active = Column('active', Boolean, nullable=False)
+        worker_count = Column('worker_count', Integer, nullable=False)
+        python_v = Column('python_v', Text, nullable=False)
+        reg_time = Column('reg_time', DateTime, nullable=False)
+        __table_args__ = (
+            PrimaryKeyConstraint('hostname', 'run_id'),
```
benclifford: I think this constraint is wrong.

A given host might report multiple times inside one run - consider I am allocated a node for some maximum permitted time, eg 24h, and then at the end of that, I request another allocation and am given the same node again (this seems unlikely when thinking of requesting one node on a large machine, but seems very likely when running a machine-sized workflow that runs for a long time)

PR #1104, opened 2019-06-28
fix monitoring for retries, add hostname, task_fail_count to monitoring

ZhuozhaoLi:
This PR is to partially address @TomGlanzman 's suggestion for monitoring when `retries` mechanism is enabled, including

1. Fix bug in monitoring for retries.
2. Add `task_fail_count` column to TASK table. It logs the number of task failures for a task. This is useful when `retries` mechanism is enabled.
3. Add `hostname` column to STATUS table. For each trial of a task, it logs the `hostname` of the node where the task run on.

I will make a separate PR to add the `task_fail_history` to TASK table that logs all history exceptions.

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -270,14 +270,11 @@ class DatabaseManager(object):
                                          messages=[msg])
                     else:                             # TASK_INFO message
                         all_messages.append(msg)
-                        if msg['task_time_returned'] is not None:
+                        if msg['task_id'] in inserted_tasks:
                             update_messages.append(msg)
                         else:
-                            if msg['task_id'] in inserted_tasks:
-                                update_messages.append(msg)
-                            else:
-                                inserted_tasks.add(msg['task_id'])
-                                insert_messages.append(msg)
+                            inserted_tasks.add(msg['task_id'])
+                            insert_messages.append(msg)
 
                             # check if there is an left_message for this task
                             if msg['task_id'] in left_messages:
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -108,6 +108,7 @@ class Database(object):
         task_status_name = Column(Text, nullable=False)
         timestamp = Column(DateTime, nullable=False)
         run_id = Column(Text, sa.ForeignKey('workflow.run_id'), nullable=False)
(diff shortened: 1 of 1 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -271,8 +273,11 @@ def start(self, priority_queue, resource_queue):
                         if msg['task_time_returned'] is not None:
                             update_messages.append(msg)
                         else:
-                            inserted_tasks.add(msg['task_id'])
-                            insert_messages.append(msg)
+                            if msg['task_id'] in inserted_tasks:
```
benclifford: I'm not clear what this change is trying to achieve: is it solving some problem related to duplicate inserts?

ZhuozhaoLi: For each retry of a task, monitoring receives a message of the task and try to insert the message again to db. However, the messages of the same task cannot be inserted again (identified as duplicate inserts due to duplicate primary IDs of task ID and workflow ID). This is to ensure that if the task is inserted before, we just do updates on several columns. Did I make it clear?

ZhuozhaoLi: Actually this is a bug but we did not test `retries` before -- it is not related to the PR title maybe

benclifford: If retries were broken in some cases before and this fixes it, this is worth putting in the PR title and description.

ZhuozhaoLi: Modified

benclifford: OK. So there's now a three way case here that I don't fully understand:

case 1:   `task_time_returned` is not None:    always update, never insert.
case 2:   `task_time_returned` is None, task_id is not in inserted tasks:  insert
case 3:   `task_time_returned` is None, task_id is in inserted tasks: update

I think case 1 here would be handled by the two new cases 2 and 3, and doesn't need to exist as a separate case;

and,

case 1 at the moment means that we must send a task message *without* a task_time_returned first, before reporting a task message *with* task_time_returned set. That is how the DFK seems to call things now (although I haven't verified properly) but it seems an unnecessary complication in this code.

ZhuozhaoLi: Oh you are right, good catch! It should be fixed now.

PR #1110, opened 2019-07-01
Catch exception for retries in monitoring

### LINKED ISSUE #1105: Get the proper exception logs of a task when the task is retried
Looking at the code at https://github.com/Parsl/parsl/blob/185283452bafdd477a638c5882f259a962abf5d3/parsl/dataflow/dflow.py#L254, 

It does not seem to work. It is always `None`.
It would be great to get the exception so that we can log that to monitoring.db.

Probably related to #111.

ZhuozhaoLi:
As @TomGlanzman suggested, we should log the exception for each retry of a task to monitoring.
This PR adds the exceptions to `monitoring.db` as a column, in the format of `exception 1; exception 2; exception 3 ...`

Related to #1104 .
Fix #1105 .

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -193,7 +193,7 @@ class DataFlowKernel(object):
         task_log_info['task_stderr'] = self.tasks[task_id]['kwargs'].get('stderr', None)
         task_log_info['task_fail_history'] = None
         if self.tasks[task_id]['fail_history'] is not None:
-            task_log_info['task_fail_history'] = "; ".join(self.tasks[task_id]['fail_history'])
+            task_log_info['task_fail_history'] = ",".join(self.tasks[task_id]['fail_history'])
         task_log_info['task_depends'] = None
         if self.tasks[task_id]['depends'] is not None:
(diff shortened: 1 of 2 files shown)


parsl/dataflow/dflow.py
```diff
@@ -191,6 +191,9 @@ def _create_task_log_info(self, task_id, fail_mode=None):
         task_log_info['task_stdin'] = self.tasks[task_id]['kwargs'].get('stdin', None)
         task_log_info['task_stdout'] = self.tasks[task_id]['kwargs'].get('stdout', None)
         task_log_info['task_stderr'] = self.tasks[task_id]['kwargs'].get('stderr', None)
+        task_log_info['task_fail_history'] = None
+        if self.tasks[task_id]['fail_history'] is not None:
+            task_log_info['task_fail_history'] = "; ".join(self.tasks[task_id]['fail_history'])
```
benclifford: It would be more consistent with`task_depends` to use `,` as a separator. Maybe fix that in a follow up PR.

ZhuozhaoLi: Okay. I changed that.

PR #1111, opened 2019-07-02
Add a test that passes in pytest but actually fails in two different ways

benclifford:
On my laptop, this test always passes according to pytest, but it behaves
wrongly in two different ways:

1. Sometimes it shows issue #1109 at the end of the test.

2. Sometimes it does not record task rows in the monitoring DB at all.
      - out of around 10 runs, I only get three task rows in my monitoring.db

I suspect for 2. that there is some race condition happening at shutdown
where the DFK is saying it is all closed before monitoring really is
finished.

Invoke like this:

```
pytest parsl/tests/sites/test_local_monitoring_with_failure.py --config local
```

### DIFF
--- a/parsl/tests/sites/test_local_monitoring_with_failure.py
+++ b/parsl/tests/sites/test_local_monitoring_with_failure.py
@@ -1,0 +1,31 @@
+import pytest
+
+import parsl
+from parsl.dataflow.dflow import DataFlowKernel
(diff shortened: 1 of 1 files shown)


benclifford:
closing this - it was for @ZhuozhaoLi to see a problem I was having, but isn't suitable for merge to master

PR #1113, opened 2019-07-03
minor code enhancement in workflow_resource_plot

ZhuozhaoLi:
A small PR breaking down from #1024.
1. Using `dataframe`'s built-in `max()` method
2. Break a loop earlier for speed


benclifford:
1. Using dataframe's built-in max() method
2. Break a loop earlier for speed

PR #1114, opened 2019-07-04
Resource efficiency visualization

ZhuozhaoLi:
The final break down of #1024 

This PR includes changes in visualization

1. removing multiple plots that are not useful (time series of cpu time/percent and memory usage/percent) 
2. Add two plots that show CPU and memory efficiency -- defined as `resource usage/requested resources`


parsl/monitoring/visualization/plots/default/workflow_resource_plots.py
```diff
@@ -91,3 +91,67 @@ def resource_time_series(tasks, type='psutil_process_time_user', label='CPU user
...
+                for i in range(int(row['last_timestamp']), int(row['relative_time'])):
+                    if label == 'CPU':
+                        diff = (row['psutil_process_time_user'] - row['last_cputime']) / (row['relative_time'] - row['last_timestamp'])
+                    elif label == 'mem':
+                        diff = row['psutil_process_memory_resident'] / 1024 / 1024 / 1024
+                    task_plot[i] += diff
+
+        if label == 'CPU':
+            name1 = 'Used CPU cores'
+            name2 = 'Requested CPU cores'
```
benclifford: I dislike the phrase `requested` here and in the memory usage. In both cases, it is the total amount of CPU or memory being reported by the systems - not the amount requested.

parsl/monitoring/visualization/plots/default/workflow_resource_plots.py
```diff
@@ -91,3 +91,67 @@ def resource_time_series(tasks, type='psutil_process_time_user', label='CPU user
...
+            tmp = resource[resource['task_id'] == task_id]
+            tmp['last_timestamp'] = tmp['relative_time'].shift(1)
+            if label == 'CPU':
+                tmp['last_cputime'] = tmp['psutil_process_time_user'].shift(1)
+            for index, row in tmp.iterrows():
+                if np.isnan(row['last_timestamp']):
+                    continue
+                for i in range(int(row['last_timestamp']), int(row['relative_time'])):
+                    if label == 'CPU':
+                        diff = (row['psutil_process_time_user'] - row['last_cputime']) / (row['relative_time'] - row['last_timestamp'])
```
benclifford: This should probably also include `psutil_process_time_system` - see https://stackoverflow.com/questions/556405/what-do-real-user-and-sys-mean-in-the-output-of-time1

ZhuozhaoLi: fixed. Now it is the sum of `psutil_process_time_system` and `psutil_process_time_user`

Issue #1118, opened 2019-07-05
Remove unused walltime keyword arg


annawoodard:
Currently there is a `walltime` keyword arg to both the python and bash app decorators and classes-- all of which do nothing. The only thing that does anything is an undocumented reserved keyword arg `walltime` for bash apps. We should remove these unused keyword args.

benclifford:
This was implemented via proc.wait quite early on in parsl, although that implementation is, I think buggy - see issue https://github.com/Parsl/parsl/issues/2824

The python implementation is differently awkward - it cannot raise an exception in a thread exception between python VM instructions, so a long lived instruction (such as a call out to heavy C code) will block the timeout.

PR #1119, opened 2019-07-05
Add Python app time limits.

annawoodard:
This PR adds python app time limits. As noted in #1118, there is currently a 'walltime' keyword arg to both the python and bash app decorators and classes-- all of which do nothing. I chose to match what is currently implemented for bash apps-- a reserved `walltime` keyword arg to the app. I'll remove the unused args in a separate PR.

### DIFF
--- a/docs/userguide/apps.rst
+++ b/docs/userguide/apps.rst
@@ -53,7 +53,7 @@ Any Parsl app (a Python function decorated with the ``@python_app`` or ``@bash_a
 2. outputs: (list) This keyword argument defines a list of output :ref:`label-futures` that
    will be produced by this app. Parsl will track these files and ensure they are correctly created.
    They can then be passed to other apps as input arguments.
-3. walltime: (int) If the app runs longer than ``walltime`` seconds, a ``TimeoutError`` will be raised.
+3. walltime: (int) If the app runs longer than ``walltime`` seconds, a ``parsl.app.errors.AppTimeout`` will be raised.
 
 Returns
 ^^^^^^^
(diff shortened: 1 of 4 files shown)


parsl/app/python.py
```diff
@@ -45,6 +45,29 @@ def __call__(self, *args, **kwargs):
         else:
             dfk = self.data_flow_kernel
 
+        walltime = self.kwargs.get('walltime')
+        if walltime is not None:
+            def timeout(f, walltime):
```
benclifford: I think this code would be more readable if `def timeout` was not inline but was a separate toplevel function -  I don't think it takes in anything from scope in which it is defined in.

docs/userguide/apps.rst
```diff
@@ -53,6 +53,7 @@ Any Parsl app (a Python function decorated with the ``@python_app`` or ``@bash_a
 2. outputs: (list) This keyword argument defines a list of output :ref:`label-futures` that
    will be produced by this app. Parsl will track these files and ensure they are correctly created.
    They can then be passed to other apps as input arguments.
+3. walltime: (int) If the app runs longer than ``walltime`` seconds, a ``TimeoutError`` will be raised.
```
benclifford: The code raises `parsl.app.errors.AppTimeout`, not `TimeoutError`

parsl/app/python.py
```diff
@@ -45,6 +45,29 @@ def __call__(self, *args, **kwargs):
...
+                def wrapper(*args, **kwargs):
+                    import threading
+
+                    def stop(thread):
+                        import ctypes
+                        import parsl.app.errors
```
benclifford: i have this niggling doubt that doing these import here means that the `stop` routine might take a while. it's probably desirable for the stop routine to get the exception set as fast as it can, as I think there's a race between i) the Timer expiring and the exception being set; and ii) the function successfully completing, so that the async exception gets thrown into some later part of the execution. Making `stop` be as fast as possible mitigates that.

parsl/app/python.py
annawoodard: will do

docs/userguide/apps.rst
annawoodard: good catch

parsl/app/python.py
annawoodard: Great point-- agreed. Thanks for catching that.

Issue #1094, opened 2019-06-25
aws ubuntu ami hangs on launch


macintoshpie:
The apt-get install command in the aws template.py results in a hang due to an interactive prompt:
```
whiptail--backtitlePackage configuration--titleConfiguring libssl1.1:amd64--output-fd11--defaultno--yesno--There are services installed on your system which need to be restarted 
when certain libraries, such as libpam, libc, and libssl, are upgraded. 
Since these restarts may cause interruptions of service for the system, 
you will normally be prompted on each upgrade for the list of services 
you wish to restart.  You can choose this option to avoid being 
prompted; instead, all necessary restarts will be done for you 
automatically so you can avoid being asked questions on each library 
upgrade.
```

## Replicate
Python: 3.6
Parsl: 0.8.0
```python
import parsl
from parsl.config import Config
from parsl.executors import HighThroughputExecutor
from parsl.providers import AWSProvider
from parsl.app.app import python_app

parsl.set_stream_logger()

parsl_config = Config(
  executors=[
    HighThroughputExecutor(
      label='htex_local',
      address=myip,
      cores_per_worker=1,
      max_workers=1,
      provider=AWSProvider(
        image_id='ami-0c55b159cbfafe1f0', # ubuntu 18.04
        instance_type='t2.micro',
        nodes_per_block=1,
        init_blocks=1,
        max_blocks=1,
        min_blocks=0,
        region='us-east-2',
        key_name='myKeyName'
      ),
    )
  ],
  strategy=None,
)

parsl.load(parsl_config)

@python_app
def hello ():
  return 'Hello World!'

print(hello().result())
```

benclifford:
Yesterday on slack we talked about adding a line

```
export DEBIAN_FRONTEND=noninteractive
```

near the start of `template.py`.

Did that work?

benclifford:
fixed by #1098

PR #1121, opened 2019-07-07
Resurrecting Azure Execution provider

benhg:
This pull request re-adds the Azure execution provider.


parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+        Returns
+        -------
+        None or str
+            If at capacity, None will be returned. Otherwise, the job identifier will be returned.
+        """
+        self.resource_client.resource_groups.create_or_update(
+            self.group_name, {'location': self.location})
+        self.resources["group"] = self.group_name
+
+        logger.info('\nCreating NIC')
```
benclifford: `\n` at the start of log message is weird

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+        Minimum number of blocks to maintain. Default is 0.
+    max_blocks : int
+        Maximum number of blocks to maintain. Default is 10.
+    region : str
+        Azure region to launch machines. Default is 'westus'.
+    key_name : str
+        Name of the Azure private key (.pem file) that is usually generated on the console
+        to allow SSH access to the Azure instances. This is mostly used for debugging.
+    walltime : str
+        Walltime requested per block in HH:MM:SS. This option is not currently honored by this provider.
```
benclifford: I think it would be better to remove this parameter rather than document it as being ignored.

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+        Nodes to provision per block. Default is 1.
+    init_blocks : int
+        Number of blocks to provision at the start of the run. Default is 1.
+    min_blocks : int
+        Minimum number of blocks to maintain. Default is 0.
+    max_blocks : int
+        Maximum number of blocks to maintain. Default is 10.
+    region : str
+        Azure region to launch machines. Default is 'westus'.
+    key_name : str
```
benclifford: does this get used anywhere?

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+          'version': VM OS version
+          'vm_size': VM Size, analogous to instance type for AWS
+          'disk_size_gb': (int) size of VM disk in gb
+          "admin_username": (str) admin username of VM instances,
+          "password": (str) admin password for VM instances
+        }
+    worker_init : str
+        String to append to the Userdata script executed in the cloudinit phase of
+        instance initialization.
+    walltime : str
```
benclifford: duplicate comment with line 90

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+        Walltime requested per block in HH:MM:SS.
+    key_file : str
+        Path to json file that contains 'Azure keys'
+    nodes_per_block : int
```
benclifford: remove this param if it is unused?

parsl/providers/azure/template.py
```diff
@@ -0,0 +1,14 @@
+template_string = """#!/bin/bash
+cd ~
```
benclifford: In the AWS version of this, commit 604d610d7d7e5d883c643daeba09f5b65db943ce added in
``
export DEBIAN_FRONTEND=noninteractive
``
before running any `apt` commands, to stop occasional interactive prompts.

See #1094, #1098 

That is probably a useful setting here too.

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,489 @@
...
+                        'location': self.location,
+                        'address_space': {
+                            'address_prefixes': ['10.0.0.0/16']
+                        }
+                    })
+            vnet_info = async_vnet_creation.result()
+            self.resources["vnet"] = vnet_info
+
+        except Exception:
+            logger.info('Found Existing Vnet. Proceeding.')
```
benclifford: That exception text seems suspicious - what happens if the create fails for other reasons?

benclifford:
I've made a bunch of comments - I think they're all usability related rather than anything to do with functionality.

I'm not set up for azure so I can't run this code so I won't click approve.

parsl/providers/azure/azure.py
benhg: I mirrored this from the AWS provider. Is this planned to be a feature in the future? And if so, is it better to leave it in so that the constructor will be invoked the same way in future releases?

parsl/providers/azure/azure.py
benhg: Azure requires an "admin password" to be set for Linux VM instances (which I really don't like) but optionally allows for SSH keys to be used as well. I am open to opinions on whether it would be useful to have keys implemented, knowing that passwords are required anyways.

parsl/providers/azure/azure.py
benhg: Whoops - I meant to switch that over to a specific type of msrazure.CloudError. Thanks for catching that.

benhg:
I have asked questions on a couple of them - will fix the rest later today. Thanks for the review !

parsl/providers/azure/azure.py
yadudoc: Cloud providers don't generally work with walltime constraints like batch systems. Let's remove this argument.

parsl/providers/azure/azure.py
benhg: Same question as with walltime param - will this break compatibility in future releases? Should it be changed in the AWS provider as well?

parsl/providers/azure/azure.py
benclifford: even if we do support it in future, my feeling is that compatibility is better by removing it now and adding it back in with a sensible default when there is meaning.

the present layout encourages people to specify walltimes using the wrong syntax, which will break if it is ever implemented; to puzzle over why it is not respected; and in potential-future to puzzle over why their jobs are suddenly terminated after the wrong walltime after upgrading parsl.

as long as the keyword has a default value if ever added back in, then calls to the initializer without `walltime` will still behave correctly in that potential-future version.

parsl/providers/azure/azure.py
benclifford: this line is really only about the docstring

parsl/providers/azure/azure.py
benclifford: I'm a big fan of SSH keys, so I'm not going to say no.

parsl/providers/azure/azure.py
benhg: Got it - sounds good.

parsl/providers/azure/azure.py
benhg: Sorry, commented on the wrong review here. meant to put that on the 'nodes_per_block' one

parsl/providers/azure/azure.py
benhg: Same question as with walltime param - will this break compatibility in future releases? Should it be changed in the AWS provider as well?

benclifford: The compatibility argument is the same as for walltime, I think: it is better to remove it now.

A difference between this code and the AWS code is that azure isn't in the last parsl release so we shouldn't be expecting users to already have azure constructors in their workflow, but there might be (in theory) people with AWS constructors that set nodes_per_block.

I expect there aren't though and it's probably fine to remove from both. If you're going to modify AWS, do it in a different PR though.

parsl/providers/azure/azure.py
benhg: Because of the way this is handled on the Azure end, my fix for this is very gross. If you have a better idea, please let me know and I will change it. 

The issue is that these errors don't actually appear to be subtypes of msrazure.CloudError - they seem to just be error messages in the string representations of the errors which are stylized as if they are actual python exceptions, eg `AzureError: InUseSubnetCannotBeDeleted`

test-requirements.txt
```diff
@@ -9,4 +9,3 @@ nbsphinx
 sphinx_rtd_theme
 mypy
 sqlalchemy-stubs
```
benclifford: spurious whitespace change

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+    'VM stopped': 'COMPLETED',  # We shouldn't really see this state
+}
+
+
+class AzureProvider(ExecutionProvider, RepresentationMixin):
+    """
+    A Provider for using Microsoft Azure Resources
+
+    One of 2 methods are required to authenticate: keyfile, or environment
+    variables. If  keyfile is not set, the following environment
```
benclifford: where can a user find out what `keyfile` is meant to look like?

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+        self.key_file = key_file
+        self.location = location
+        self.group_name = group_name
+
+        self.launcher = launcher
+        self.linger = linger
+        self.resources = {}
+        self.instances = []
+
+        env_specified = os.getenv("AZURE_CLIENT_ID") is not None and os.getenv(
```
benclifford: should this check for all four of the `AZURE_*` environment variables ?

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+        self.init_blocks = init_blocks
+        self.min_blocks = min_blocks
+        self.max_blocks = max_blocks
+        self.max_nodes = max_blocks
+        self.parallelism = parallelism
+        self.nodes_per_block = 1
+
+        self.worker_init = worker_init
+        self.vm_reference = vm_reference
+        self.vm_disk_size = self.vm_reference["disk_size_gb"]
```
benclifford: Does this `self.vm_disk_size` need to be set? I can't see anywhere that it is used

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+                 location='westus',
+                 group_name='parsl.auto',
+                 key_name=None,
+                 key_file=None,
+                 vnet_name="parsl.auto",
+                 linger=False,
+                 launcher=SingleNodeLauncher()):
+        if not _api_enabled:
+            raise OptionalModuleMissing(
+                ['azure', 'msrestazure'], "Azure Provider requires the azure module.")
```
benclifford: Look in `setup.py`  at `extra_requires` - add a block in there and a user will be able to say `pip install parsl[azure]` to get all the relevant azure dependencies, in the same way that they can say `pip install parsl[aws]` to get the relevant AWS dependencies

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+        down the nodes.
+    """
+
+    def __init__(self,
+                 init_blocks=1,
+                 min_blocks=0,
+                 max_blocks=10,
+                 parallelism=1,
+                 worker_init='',
+                 vm_reference=None,
```
benclifford: Probably don't assign a default value here unless it actually  makes sense to use that default - I think it never makes senese for `vm_reference` to be None.

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+        Required structure:
+        {
+          'publisher': VM OS publisher
+          'offer': VM OS offer
+          'sku': VM OS SKU
+          'version': VM OS version
+          'vm_size': VM Size, analogous to instance type for AWS
+          'disk_size_gb': (int) size of VM disk in gb
+          "admin_username": (str) admin username of VM instances,
+          "password": (str) admin password for VM instances
```
benclifford: some of these fields use ', some use " - that's inconsistent

benclifford:
I have fiddled with azure enough to have this start up a VM for me, using the below configuration. However it doesn't seem to successfully start up the htex workers - from `dmesg` on the started VM, it looks like maybe I'm running out of memory so I'll try a larger one.

Killing parsl with ctrl-C doesn't kill the VMs, which is worrying as that leaves a drain on real money behind at the end of the run.

```
config = Config(
    executors=[
        HighThroughputExecutor(
            label='azure_htex_single_node',
            address="fitzroy.cqx.ltd.uk",
            provider=AzureProvider(
                vm_reference=
                  {
                    'publisher': "Canonical",
                    'offer': "UbuntuServer",
                    'sku': "19.04",
                    'version': "latest",

                    'vm_size': "Standard_B1ls",

                    'disk_size_gb': 10,
                    "admin_username": "benc",
                    "password": "ligo123"
                },

                key_name='aws-us-east1',
                key_file='./azurecreds.pem',
                init_blocks=1,
                max_blocks=1,
            ),
        )
    ],
)
```

benclifford:
Using the below config I am able to run some simple test `@python_app`s:

```
￼config = Config(
    executors=[
        HighThroughputExecutor(
            label='azure_htex_single_node',
            address="fitzroy.cqx.ltd.uk",
            provider=AzureProvider(
                vm_reference=
                  {
                    'publisher': "Canonical",
                    'offer': "UbuntuServer",
                    'sku': "19.04",
                    'version': "latest",

                    'vm_size': "Standard_B4ms",

                    'disk_size_gb': 10,
                    "admin_username": "benc",
                    "password": "ligo123"
                },

                key_name='aws-us-east1',
                key_file='./azurecreds.pem',
                init_blocks=1,
                max_blocks=1,
            ),
        )
    ],
)

File: parsl/providers/azure/azure.py [hunk 15]
```diff
@@ -0,0 +1,478 @@
...
+    AZURE_SUBSCRIPTION_ID.
+
+    Parameters
+    ----------
+    vm_reference : dict
+        Dictionary describing the parameters of the Azure VM.
+
+        Required structure:
+        {
+          'publisher': VM OS publisher
```
benclifford: I found it quite awkward to find a userful publisher/offer/sku/version combo - it involved a lot of messing round in the Azure Marketplace. I don't know if there is a better way. But make some note about where to get these values from.

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+
+        # state = translate_table.get(instance.state['Name'], "PENDING")
+
+        self.resources[vm_info.id] = {
+            "job_id": vm_info.id,
+            "instance": vm_info,
+            "status": "Test State"
+        }
+
+        virtual_machine = async_vm_creation.result()
```
benclifford: is this a different value to `vm_info_ on line 239?

parsl/providers/azure/azure.py
```diff
@@ -0,0 +1,478 @@
...
+        logger.info('Creating Linux Virtual Machine')
+        vm_parameters = self.create_vm_parameters(nic.id,
+                                                  self.vm_reference)
+
+        # Uniqueness strategy from AWS provider
+        job_name = "parsl.auto.{0}".format(time.time())
+
+        async_vm_creation = self.compute_client.\
+            virtual_machines.create_or_update(
+                self.vnet_name, job_name, vm_parameters)
```
benclifford: Using job name in this format of `parsl.auto.{}` is awkwardly here: inside linux, the full hostname is set to `parsl.auto.WHATEVER`, which is generally interpreted as the base hostname being `parsl` and the domain name being `auto.WHATEVER`.

This means that tooling which reports a short hostname always reports `parsl` and it is then hard to tell multiple VMs in the same run apart.

Contrast this with AWS which uses a hostname based on the instance's internal IP address, so the short hostname is reported differently for each VM, something like this: `ip-10-0-4-242`


Put some kind of unique ID into the first bit of the computer name, before the first `.` symbol. time() would do for now, but see #1122

benhg:
> Killing parsl with ctrl-C doesn't kill the VMs, which is worrying as that leaves a drain on real money behind at the end of the run.

Killing parsl with ctrl-C should send cancel() requests and thus should close all of the VMs. In my testing, this worked as it should. Perhaps your VMs running out of memory caused this to not occur. Did the VMs go away with the updated, working config?


With regards to the rest of these changes, I will make them this evening.

benhg:
I've done some more testing on the issue of not deleting on ctrl-C. If a submit() call is not still running, the VMs will indeed delete themselves. Unfortunately, Azure's API for running scripts on the VM (required for connecting back to Parsl controller) is both synchronous and very slow (30 sec min), so if you ctrl-C within 30 sec or so of a submit() request, Parsl doesn't know that an instance has been started, and so can't clean it up on shutdown. Any ideas on how to deal with this?

benclifford:
There's enough information stored at this point, before the script is run, to be able to kill the VM, even though parsl doesn't know about it.

submit, line 255-256

```
        vm_info = async_vm_creation.result()
        self.instances.append(vm_info.name)
```

Once you have that info, maybe you could put a `try` block around the whole of the rest of the code in `submit` that deals with killing the VM if an exception occurs. I think you might be able to see a `KeyboardInterrupt` exception there when ctrl-C is pressed (I'm not sure).

benhg:
> There's enough information stored at this point, before the script is run, to be able to kill the VM, even though parsl doesn't know about it.
> 
> submit, line 255-256
> 
> ```
>         vm_info = async_vm_creation.result()
>         self.instances.append(vm_info.name)
> ```
> 
> Once you have that info, maybe you could put a `try` block around the whole of the rest of the code in `submit` that deals with killing the VM if an exception occurs. I think you might be able to see a `KeyboardInterrupt` exception there when ctrl-C is pressed (I'm not sure).
> 
> The idea being that until submit returns, it is the responsibility of that `try` block to tidy up; and then as soon as it ends and passes the ID back to the rest of parsl, it is the responsibility of the rest of parsl to call `cancel`.
This makes sense to me.

benclifford:
ok this is looking good. I will give it another try out probably tomorrow and hopefully will merge it then

PR #1123, opened 2019-07-08
[WIP] data manager staging provider prototype

benclifford:
This PR contains all of my work in progress related to data manager.

The main interesting thing is the replacement of hard-coded staging mechanisms with a `Staging` interface - something I've had in my head for a while but wanted to actually code up to see what it looked like in real life.

Executors now have a list of `Staging` providers (a more elaborate form of the earlier globus staging configuration - so still supplied in a `storage_access` attribute on the executor).

For each file to be staged in, each staging provider is offered the file in turn to see if it knows how to perform the staging.
There are 4 staging providers which replicate the existing behavior: FileNoOpStaging - which handles `file://` URLs by doing nothing; `HTTPSeparateTaskStaging` and `FTPSeparateTaskStaging` which perform HTTP, HTTPS and FTP staging as separate parsl tasks; and `GlobusScheme`  which performs transfers using globus.

There are 2 further staging providers which attempt some more interesting functionality:
- `FileArgsStaging` provides an alternative method for staging-in (small) `file://` URLs by reading the content and storing it in a wrapper that goes around the app function
- `HTTPInTaskStaging` which stages-in files in a wrapper around the app, so that the stage-in is guaranteed to happen on the same node as the app execution. This is a prototype for running on executors with no shared file system.

The main staging provider interface is in parsl/data_provider/staging.py and provides three hooks: one for replacing stage-in Files with DataFutures (as happens with the existing staging mechanism) and one for wrapping the called app with a wrapper function. This API deliberately does not try to constrain how a staging provider implements those two hooks. A further hook allows a staging provider to indicate if it can or cannot stage a particular URL.

The implementations of the above staging providers are found in the various files in parsl/data-provider/*.py

There is a bunch of other stuff in this PR as fairly clean separate commits that I would like to get into master at some point.

I would like to understand how this interacts with the workqueue (PR #771) method of staging files, which looks like it can only deal with files already on the submit-side filesystem.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -14,6 +14,7 @@ Reference guide
     parsl.dataflow.futures.AppFuture
     parsl.dataflow.dflow.DataFlowKernelLoader
     parsl.data_provider.data_manager.DataManager
+    parsl.data_provider.data_manager.Staging
     parsl.data_provider.files.File
     parsl.executors.base.ParslExecutor
     parsl.executors.ThreadPoolExecutor
--- a/docs/stubs/parsl.data_provider.staging.Staging.rst
+++ b/docs/stubs/parsl.data_provider.staging.Staging.rst
@@ -1,0 +1,8 @@
+parsl.data\_provider.staging.Staging
+==============================================
+
+.. currentmodule:: parsl.data_provider.staging
+
+.. autoclass:: Staging
+   :members:
+   :undoc-members:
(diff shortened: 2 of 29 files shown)


yadudoc:
@benclifford  Could you mark this ready for review? It is really silly that this being a draft, github doesn't let me leave any comments. I've skimmed most of the changes here.

benclifford:
marked as review to enable comments

benclifford:
This has pretty much been split into other PRs now. I still have ongoing development in my private patch stack but I'll put up changes from there as new PRs as appropriate.

PR #1126, opened 2019-07-09
Minor cleanup for stress test

annawoodard:
This PR updates the stress test to use the latest config paradigm and adds some minor cleanup.

### DIFF
--- a/parsl/tests/integration/test_stress/test_python_simple.py
+++ b/parsl/tests/integration/test_stress/test_python_simple.py
@@ -6,6 +6,7 @@ from parsl.configs.htex_local import config
 import time
 import argparse
 
+
 @python_app
 def increment(x):
     return x + 1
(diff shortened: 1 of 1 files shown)


parsl/tests/integration/test_stress/test_python_simple.py
```diff
@@ -1,16 +1,14 @@
...
+from parsl import python_app
+from parsl.configs.htex_local import config
+# from parsl.configs.local_threads import config
 
 import time
 import argparse
 
-workers = ThreadPoolExecutor(max_workers=100)
-dfk = DataFlowKernel(executors=[workers])
+parsl.load(config)
```
benclifford: In some cases, test modules get imported in workers (I think something to do with pytest adding on annotations, so probably not a problem with this test). In those cases, parsl.load() needs to happen in the protected `__main__` block, rather than everywhere that this file is imported, and it's probably stylistically better to try to stick to that for all tests. See c8c451b45c1b9e9163a2bf2a94ede74d9621fa91

benclifford:
This test only takes a couple of seconds - could add it onto travis with other integration tests if it is useful

annawoodard:
I'm not sure how useful it is in integration tests, as it doesn't assert anything. I think its main use is for 'make change, test time, make other change, test time'

PR #1144, opened 2019-07-17
Move ftp and http staging into own files and out of DM

benclifford:
The intention here is to move the particular staging
mechanism out of the DM and into something configurable
along the lines of GlobusScheme

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -1,43 +1,14 @@
 import os
 import logging
-import requests
-import ftplib
 import concurrent.futures as cf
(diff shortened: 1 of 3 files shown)


benclifford:
@yadudoc yeah the aim is to get the existing per-method staging code into separate files in this PR - I have another more complicated one for globus that comes after this.

I wanted the un-tangling of different methods to happen first, so that a later PR putting the API in place isn't more focused and doesn't move code round so much.

Most of the individual commits in #1123 should turn into individual PRs, so that list of commits should give you a feel for what else I want do make PRs for in this sequence.

PR #1150, opened 2019-07-19
Add control of logging level to the Config, stream_ and file_logger

Lnaden:
This adds new settings in the `Config` class which enables setting
the logging levels to the stream_logger and the file_logger.

The main use case here is for those directly invoking the
`DataFlowKernel` with a `Config` who don't want the output loggers to
be automatically set to `DEBUG` level.

Default behavior should be identical to the current behavior and this
*should* be an extension of the existing API without breaking backwards
compatibility.

It is a rather simplistic change, so I'm not sure I made the changes
everywhere that is needed nor have I figured out how to properly test
this, so any comments or suggestions would be helpful.

Change is related to a conversation @yadudoc and myself had previously. There is no issue directly tied to this, but I can open one if requested and if this feature needs more discussion before a PR.

### DIFF
--- a/parsl/__init__.py
+++ b/parsl/__init__.py
@@ -79,12 +79,14 @@ wait_for_current_tasks = DataFlowKernelLoader.wait_for_current_tasks
 
 
 @typeguard.typechecked
-def set_stream_logger(name: str = 'parsl', level: int = logging.DEBUG, format_string: Optional[str] = None):
+def set_stream_logger(name: str = 'parsl', level: int = logging.DEBUG, stream_level: Optional[int] = None,
+                      format_string: Optional[str] = None):
     """Add a stream log handler.
 
     Args:
          - name (string) : Set the logger name.
          - level (logging.LEVEL) : Set to logging.DEBUG by default.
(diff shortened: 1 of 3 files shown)


benclifford:
I don't really understand from the PR description what this brings me as a user - can you elaborate?

Lnaden:
Reduces the amount of logger information shown to the end user if they choose as a quality of life aspect.

Specifically, from our use case, we directly instance the `DataFlowKernel` and pass jobs into it for distribution on our hardware. Within our software stack, we also have our own logging.

When we have our users start up the code, the fact that the `DataFlowKernel` calls `set_file_logger` at `level=logging.DEBUG`, which means there is going to be a bunch of information which fills a user screen. For our users who don't know and don't need to see all the logging information for the entire software stack, it just tidies up the output a bit. 

In the end, its a rather small thing, but one which can help users not feel overwhelmed at the cost of less information if/when something goes awry.

dgasmith:
@yadudoc Pinging you here.

Lnaden:
Superseded by #1335

PR #1152, opened 2019-07-22
Added WQ Executor Functionality to Increase Throughput

tjdasso:
This pull request makes modifications to the WorkQueue Executor to provide better throughput. Instead of using the ipyparallel.serialize library to serialize Python functions, a switch has been implemented for the Executor that allows the user to choose whether to transport the function as source code. This greatly diminishes the overhead for sending and receiving task information that was due to the expensive serialization. It also uses the tblib library to serialize traceback objects for tasks that failed. 

To test full functionality, you can use the "source" option in the parsl/tests/workqueue_tests/wqex_local.py to switch between transferring function data as a serialized object or through source code.


benclifford:
I've really been pushing in parsl to have feature PRs not contain tidyups. There are a lot of useful looking tidy-ups in this PR and having the extra comments in there will be useful, but it makes it more time consuming to evaluate the PR in the context of what the description claims it does - my preference would be for a tidyup PR (which claims to make no changes to behaviour) and a feature PR (which doesn't have stuff that doesn't change behaviour, such as tidying up comments far from the meat of the changes).

Most of the CI related failures are to do with code formatting so should be straightforward to fix: run `flake8 parsl/` to recreate that test on your local machine.

There is also a `mypy` failure that comes from using the variable name `source` for two different purposes in the same function - I think that should be straightforward to fix too. Run `mypy ./parsl/executors/workqueue/workqueue_worker.py` to run that mypy test locally.

tjdasso:
Sorry about incorporating the tidyups, the commenting was mostly for my own sake as I tried to became better acquainted with the project after Andrew finished his work with it. I can definitely see why that makes it more difficult to evaluate the PR.

The meat of the changes has to do with the serialization of the function and traceback objects when tasks are created and completed. The main changes occur between lines `350-355` and `587-610` of the `executors/workqueue/executor.py` script, as well as between lines `53-70` and `129-143` of `executors/workqueue/workqueue_worker.py` where the function/traceback objects are serialized or the function code is extracted and put into the message queue.

I also examined the scripts using flake8 and mypy and made the appropriate changes, so hopefully there should be no CI related failures now.

tjdasso:
Work Queue has also been working on implementing a way to "transport" Python environments and dependencies to worker machines using Conda environments. More information can be found in the CCTools pull request #2109 (https://github.com/cooperative-computing-lab/cctools/pull/2109), which includes the scripts and documentation on how to use them, but the essence is that a Python script/function is analyzed for its library dependencies, a Conda environment is created that contains all these necessary dependencies, and the Conda environment is then sent to the worker machine where the task is then run within that environment. This ensures that Parsl tasks using Work Queue will not fail due to the absence of the required libraries or Python interpreter on the remote machine. 

I have been working on the implementation of these tools with the Work Queue Executor for Parsl. However, this work is dependent upon the changes made in this pull request. As such, I do not want to submit a separate PR to Parsl as they will contain overlapping changes. With that being said, the changes and implementation to the executor can be found under the condaEnvSetup branch of my forked Parsl repo (https://github.com/tjdasso/parsl/tree/condaEnvSetup). I can submit a PR for this branch in the future if this (#1152) PR gets merged, as it would be then easier to see the specific changes that were made to the executor's implementation.

benclifford:
I usually submit my sequenced work as separate PRs, one by one as the previous one gets merged I submit the next.

I'll review this PR soonish but to be honest I haven't yet because it is a lot of screenfuls of noise wrt what the title says it should do that I need to pay attention to.

benclifford:
What we've done for other PRs in that situation is leave the big PR, and pull out individual changes into separate PRs that are easily reviewable, and then eventually the original big PR evaporates away to nothing as the separate PRs are merged in.

benclifford:
(which I'm happy to help you with, as it's basically what needs to happen for reviewing anyway...)

tjdasso:
That would be great. How would the PRs be best organized? The code can generally be separated into the three separate categories: 1) changing the serialization/deserialization techniques for the Python functions/exceptions, 2) incorporating Conda environment wrapping when sending tasks to workers, and 3) general tidyups that include added comments and more verbose logging statements. 

(1) and (3) are generally included in this PR #1152, while (2) is on a separate branch that includes all the previous work.

benclifford:
If you were explaining this PR to me, you'd probably say "here are some fixes that are so trivial you don't need to think about them" - fixing log messages, adding explanatory comments for existing code, for example. Start by getting all of those into a separate PR so that whoever reviews them doesn't need to engage brain much to review.

When that gets merged, then what is left in this #1152 PR should be the (de)serialisation and nothing else distracting.

benclifford:
@tjdasso - now #1161 is merged, merge master into this branch. It looks like there are a couple of merge conflicts from the changes but I think they should be easy for you to resolve. Then have a read through the diff again, check it looks beautiful, check there is no fluff in there.

tjdasso:
I merged master into this branch, so now the pull request should only contain feature changes for the executor, notably in `executors/workqueue/executor.py` and `executors/workqueue/workqueue_worker.py`. I used `tests/workqueue_tests/test_scale.py` to test its functionality, and you can use the `source` switch in the `tests/workqueue_tests/wqex_local.py` configuration file to compare functionality when the function is serialized versus when the source code is transferred.

parsl/executors/workqueue/executor.py
```diff
@@ -352,12 +352,17 @@ def WorkQueueCollectorThread(collector_queue=multiprocessing.Queue(),
...
             result = item["result"]
-            future_update, _ = deserialize_object(result["result"])
+            future_update = result["result"]
             logger.debug("Updating Future for Parsl Task {}".format(parsl_tid))
             if result["failure"] is False:
                 future.set_result(future_update)
             else:
-                future.set_exception(RemoteExceptionWrapper(*future_update))
+                future_fail = pickle.loads(future_update)
+                exc = RemoteExceptionWrapper(*future_fail)
```
benclifford: This bit of code is maybe more convoluted than it needs to be. RemoteExceptionWrapper is intended to transport exceptions through serialisation (which has been a problem for us in the past). But, this code creates that wrapper and then immediately unwraps it by calling `reraise`.

If you just want to set that exception you could call `reraise(t, v, tb)` (where those are the three parameters you're probably getting from `future_fail`) on line 363 without the RemoteExceptionWrapper

parsl/executors/workqueue/executor.py
```diff
@@ -402,6 +407,7 @@ def __init__(self,
                  port=WORK_QUEUE_DEFAULT_PORT,
                  env=None,
                  shared_fs=False,
+                 source=False,
```
benclifford: I think this default means that the new code path has not been tested in CI?

benclifford: I tried the workqueue pytest stuff from `.travis.yml` on my laptop with source=True and I get this error pretty much straight away:

```
E           _pickle.PicklingError: Can't pickle <function echo_to_file at 0x7f387eb802f0>: it's not the same object as parsl.tests.test_bash_apps.test_apptimeout.echo_to_file
```

tjdasso: Where was the error traced back to? The goal of this PR is to avoid pickling a function object altogether, instead obtaining the source code and transferring that as a string to the worker.

tjdasso: Also, I can update the code to change source=True for the build

benclifford: Try in CI - I won't be able to look at this for a few more hours

benclifford:
waiting to hear from @tjdasso about pytest / potential CI problems

tjdasso:
@benclifford Sorry for the delay, as I was away for several weeks with limited access to my computer, but I will have access and time for the next several months. I ran the `test_scale.py` script locally from this branch but I did not have issues with pickling functions. How do I test it in CI? I will also change the `source` switch to True for the build, for better testing.

benclifford:
Change the source switch and push to your PR branch... the CI will run automatically, and part of that is a command near the end of .travis.yml which runs tests with parsl/test/configs/workqueue_ex.py, which is the workqueue test configuration. Then you'll get a green or red symbol next to your commit in this PR.

benclifford:
@tjdasso ok, so now you have a commit that is failing in CI. It's failing/hanging with this command, which is what runs the test suite against workqueue:

```
pytest parsl -k "not cleannet" --config parsl/tests/configs/workqueue_ex.py --cov=parsl --cov-append --cov-report= --random-order --bodge-dfk-per-test
```

Hopefully that also fails for you on your own machine.

benclifford:
The way to read the output is: green dot means a test was run and passed; red F means a test was run and failed; yellow s means a test was skipped because it has been marked by a developer as to be skipped - ignore those ones. It looks like there is a mixture of passes and fails.

tjdasso:
Great, I looked at the tests that are failing in CI and I know where to look. Is there any easy way to run each test locally with the WQ executor configuration file to see where the error occurs? Or would I have to make individual changes and have each test run in CI?

benclifford:
If you have the right stuff installed (which is requirements.txt and test-requirements.txt and workqueue) you should be able to type that `pytest` commandline directly into your own machine.

You can specify an individual test file like this, by replacing the `parsl` directory name with a path to an individual test file:

```
pytest parsl/tests/test_python_apps/test_basic.py --config parsl/tests/configs/workqueue_ex.py 
```

benclifford:
Also, if you want to run the same test using the local threads executor, to compare behaviour with workqueue, use this config file (which runs near the start of CI): `parsl/tests/configs/local_threads.py`

tjdasso:
After running a few of the tests locally, it seems like a main problem is that the WQ executor cannot pickle the Parsl app functions into files, which causes a Pickling error. This PR uses an alternative method to pickling by converting the Parsl app to source code and executing that directly. However, after looking at some of the tracebacks for the tests, it seems as though some Parsl apps pass in other Parsl apps as arguments, in the form of a function. Thus, when this PR implementation tries to pickle the args, it receives a Pickling error because it cannot pickle the function object.

For example, in `parsl/tests/test_docs/test_from_slides.py`, the `echo()` Bash app lists the `echo()` function itself as an argument to the app. Thus, when the args are pickled into a file for execution by the WQ worker, Pickle has a problem with pickling the function. Are function objects passed as arguments into its own function for all Bash apps?

tjdasso:
After running a few of the tests locally, it seems like a main problem is that the WQ executor cannot pickle the Parsl app functions into files, which causes a Pickling error. This PR uses an alternative method to pickling by converting the Parsl app to source code and executing that directly. However, after looking at some of the tracebacks for the tests, it seems as though some Parsl apps pass in other Parsl apps as arguments, in the form of a function. Thus, when this PR implementation tries to pickle the args, it receives a Pickling error because it cannot pickle the function object.

For example, in parsl/tests/test_docs/test_from_slides.py, the echo() Bash app lists the echo() function itself as an argument to the app. Thus, when the args are pickled into a file for execution by the WQ worker, Pickle has a problem with pickling the function. Are function objects passed as arguments into its own function for all Bash apps?

benclifford:
A bash app gets launched like this -- see parsl/app/bash.py. What actually gets submitted to the DFK for execution (and so what will be passed onto an executor for execution) is the result of a few layers of wrapping.

I think maybe (though I'm not entire sure) that you are seeing the remote_side_bash_executor function (which is then suitably renamed by that update_wrapper to "look like the wrapped function"), and which is passed the original @bash_app function as a parameter - see the second line of the code fragment below which passes the original function (self.func) as the first parameter.

That remote_side_bash_executor is the function that: i) calls the actual user defined function to generate the commandline to run, and then ii) executes that command line; `@python_app`s do not need this wrapping structure because they are already real Python code.

```
        app_fut = dfk.submit(wrap_error(update_wrapper(remote_side_bash_executor, self.func)),
                             self.func, *args,
                             executors=self.executors,
                             fn_hash=self.func_hash,
                             cache=self.cache,
                             **self.kwargs)
```

tjdasso:
Yes, after running a few tests locally, it seems that this error is specific to Bash apps in particular. Is it possible to determine whether a Parsl app is a `bash_app` or `python_app` from the function object itself? If so, the source code could be extracted when the executor is given a `python_app`, otherwise it would perform the normal serialization methods.

benclifford:
> Is it possible to determine whether a Parsl app is a bash_app or python_app from the function object itself?

Not in a clean way at the moment. Perhaps check for a `__wrapped__` attribute on the function that is being passed to executor.submit(), which I think will be defined for bash_apps because of the wrapping, and usually not for python apps.

I think this is a more general problem though - that this mode can't cope with function parameters anywhere, when the original mode can, and it just happens that we've discovered this through testing bash apps.

If that is the case, then maybe it is fine to say "this mode doesn't work for bash apps or if you try to pass functions as a parameter to a python app" in the docstring? (the case where that doesn't work is where you want to have the fast behaviour for python apps and slow behaviour for bash apps in the same program)

benclifford:
The API is already loose enough for that, I think -- for example, if you pass an object in through the local thread executor, your @python_app runs with that original object, and no serialisation happens - changes to the object in an app are visible everywhere; if you pass objects to the htex executor, that object must be suitably serializable, and you end up with a copy of the object - changes to the object in an app are not visible outside of that app.

tjdasso:
It seems as though the `__wrapped__ attribute is set for both `python_app`s and `bash_app`s. I could perhaps parse the decorator from the source code to determine what type of Parsl app it is, but that does not seem very clean, and might get convoluted if multiple decorators are stacked. If I find a way to distinguish between the two different Parsl apps, would it still be useful to transfer the function information as source code for `python_app`s, given that it would not work for `bash_app`s and situations where a function object is passed into another app as a parameter? 

When I originally made this PR, my testing had only been on "pure" `python_app`s, that is, apps that have all the source code completely maintained within the function definition, but I did not test for `bash_app`s. However, in situations where a user wants to perform a lot of short Python tasks within WQ, this is a situation where the implementation might still be useful, as the overhead of communication between the WQ master and worker is minimized. Otherwise, the time saved by not serializing the function might not outweigh the limits that this implementation puts on the user in terms of not allowing the user to write `bash_app`s.

tjdasso:
It seems as though the` __wrapped__` attribute is set for both `python_app`s and `bash_app`s. I could perhaps parse the decorator from the source code to determine what type of Parsl app it is, but that does not seem very clean, and might become convoluted if multiple decorators are stacked. If I find a way to distinguish between the two different Parsl apps, would it still be useful to transfer the function information as source code for `python_app`s, given that it would not work for `bash_app`s and situations where a function object is passed into another app as a parameter?

When I originally made this PR, my testing had only been on "pure" `python_app`s, that is, apps that have all the source code completely maintained within the function definition, but I did not test for `bash_app`s. However, in situations where a user wants to perform a lot of short Python tasks within WQ, this is a situation where the implementation might still be useful, as the overhead of communication between the WQ master and worker is minimized. Otherwise, the time saved by not serializing the function might not outweigh the limits that this implementation puts on the user in terms of not allowing the user to write `bash_app`s.

benclifford:
> However, in situations where a user wants to perform a lot of short Python tasks within WQ, this is a situation where the implementation might still be useful, as the overhead of communication between the WQ master and worker is minimized.

I think this is a decent use-case.

benclifford:
so if you want to proceed that way - make it clear in the docstring for this executor that this option will work in fewer situations but give better throughput; run `git revert 7ea381b71897f0898b9fd4d77d0e2eb290c752d2` to change the config back to what it was before.

Then this PR can probably be merged.

tjdasso:
I updated the docstring to reflect the functionality of the `source` option. Would it make sense to have the `source` switch default to `False`, since the current implementation works in all known cases while this PR only works when it meets certain criteria?

benclifford:
> Would it make sense to have the source switch default to False, since the current implementation works in all known cases while this PR only works when it meets certain criteria?

I think so.

benclifford:
The most recent commit is failing flake8 in CI:

```
$ flake8 parsl/
parsl/executors/workqueue/executor.py:393:64: W291 trailing whitespace
parsl/executors/workqueue/executor.py:394:62: W291 trailing whitespace
The command "flake8 parsl/" exited with 1.
```

THat should be an easy fix.

parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -105,32 +117,44 @@ def check_file(parsl_file_obj, mapping, file_type_string):
...
                     resultname: resultname})
 
-    code = "{0} = {1}(*{2}, **{3})".format(resultname, fname,
-                                           argname, kwargname)
+    # Import function source code and create function call
+    if source:
+        source_list = source_code.split('\n')[1:]
+        full_source = ""
+        for line in source_list:
+            full_source = full_source + line + "\n"
```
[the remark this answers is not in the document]
tjdasso: The goal is to remove the `@python_app` decorator from the source code, as calling `exec()` on the function definition with the decorator will call an error. Perhaps using `'\n'.join(source_code.split('\n')[1:])` is a more straightforward way to write the code?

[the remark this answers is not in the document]
benclifford: aha! OK. Maybe add a comment to that effect?

PR #1155, opened 2019-07-22
Re-order extreme scale and workqueue for better dependency validation

benclifford:
This attempts to reduce the optional dependencies installed before the majority of tests, and installs those dependencies later on only for mechanism-specific testing.

This addresses a problem encountered a few times in CI, where missing dependencies  in the core `requirements.txt` were not discovered because of the more exciting dependency options installed in CI.


benclifford:
@yadu please re-review

PR #1156, opened 2019-07-22
Make monitoring dependencies optional #1139

### LINKED ISSUE #1139: Monitoring should be an optional extra
We should push this out as 0.8.1. This is needed by the nersc DESC stack.

ZhuozhaoLi:
fix #1139


.travis.yml
```diff
@@ -27,7 +27,7 @@ before_install:
 # command to install dependencies
 install:
     - pip install -r requirements.txt
-    - pip install .[extreme_scale]
+    - pip install .[extreme_scale] .[monitoring]
```
benclifford: This should maybe be phrased as `.[extreme_scale,monitoring]`

ZhuozhaoLi: fixed

Issue #785, opened 2019-02-24
serialised globus exception does not deserialise correctly: `__init__() missing 1 required positional argument: 'exc'`


benclifford:
```
2019-02-23 00:25:41 parsl.data_provider.globus:149 [WARNING]  Non-critical Globus Transfer error event for globus://005fe85c-36da-11e9-9838-0262a1f2f698/home/benc/parsl/workdir2/sorted.txt: "file not found" at 2019-02-23 00:25:04+00:00. Retrying...
2019-02-23 00:25:41 parsl.data_provider.globus:149 [WARNING]  Non-critical Globus Transfer error event for globus://005fe85c-36da-11e9-9838-0262a1f2f698/home/benc/parsl/workdir2/sorted.txt: "file not found" at 2019-02-23 00:25:04+00:00. Retrying...
2019-02-23 00:25:41 parsl.data_provider.globus:150 [DEBUG]  Globus Transfer error details: {
  "context": [
    {
      "operation": "Directory List / File Scan",
      "path": "/home/benc/parsl/workdir2/sorted.txt"
    }
  ],
  "error": {
    "body": "550-GlobusError: v=1 c=PATH_NOT_FOUND\\r\\n550-GridFTP-Errno: 2\\r\\n550-GridFTP-Reason: System error in stat\\r\\n550-GridFTP-Error-String: No such file or directory\\r\\n550 End.\\r\\n",
    "code": 550,
    "endpoint": "my laptop 2nd work location (005fe85c-36da-11e9-9838-0262a1f2f698)",
    "server": "gsiftp://172.20.44.170:37028",
    "type": "FTPServerError"
  }
}
2019-02-23 08:54:02 parsl.dataflow.dflow:261 [ERROR]  Task 2 failed
Traceback (most recent call last):
  File "/home/benc/parsl/src/parsl/parsl/dataflow/dflow.py", line 258, in handle_exec_update
    res.reraise()
  File "/home/benc/parsl/src/parsl/parsl/app/errors.py", line 149, in reraise
    reraise(dill.loads(self.e_type), dill.loads(self.e_value), self.e_traceback.as_traceback())
  File "/home/benc/parsl/virtualenv/lib/python3.5/site-packages/dill/_dill.py", line 317, in loads
    return load(file, ignore)
  File "/home/benc/parsl/virtualenv/lib/python3.5/site-packages/dill/_dill.py", line 305, in load
    obj = pik.load()
[...]

PR #1160, opened 2019-07-25
Remove explicit pickling in RemoteExceptionWrapper

benclifford:
cc @annawoodard

Backstory:

I was investigating exception handling in the context of globus errors thrown from the existing globus staging apps.

GlobusConnectionErrors cannot be serialised/deserialised properly (see #785) - which is an error that has happened with a bunch of parsl exception classes, and been fixed in those cases by modifying the source code of the exceptions (although in a way I am not entirely comfortable with, on review).

However, in the globus staging case, these errors are being thrown from apps running in an executor running locally - no serialisation is necessary.

I started investigating only doing the `dill` calls when a RemoteExceptionWrapper is itself pickled, so that RemoteExceptionWrappers staying within the same process do not do any pickling.

I removed the `dill` calls entirely from RemoteExceptionWrapper, as shown in this PR, and was expecting at this point that local exceptions would work, but that htex and IPP would fail to pass back exceptions correctly.

However, it seems that they do - so maybe RemoteExceptionWrapper doesn't need these explicit `dill` calls.

But I implemented #662 specially because there were problems with RemoteExceptionWrapper being serialised.

I'm confused.

Update:
I'm going to keep this in my dev stack for now and see if I encounter problems in my general use of parsl


benclifford:
This PR, although not causing trouble before, has a test failure in a recently restored test that checks bash error codes.

PR #1161, opened 2019-07-25
Test that exceptions from apps are delivered to caller

benclifford:
I can't find where this is explicitly tested in the test suite at the moment, but maybe it is already?


annawoodard:
I think we should try reviving this: https://github.com/Parsl/parsl/blob/master/parsl/tests/test_python_apps/test_basic.py#L94

That tests custom exceptions which aren't in the standard library.

PR #1162, opened 2019-07-25
Work Queue Executor Tidy Up Changes

tjdasso:
This pull request makes many small changes to the `parsl/executors/workqueue/executor.py` and `parsl/executors/workqueue/workqueue_worker.py` script. These changes changes primarily deal with adding additional commenting and more verbose logging messages to help better describe the script. It additionally adds error messages to handle more Work Queue system failures, which were previously  not handled.

This PR contains all the tidyup changes that were originally submitted in PR #1152. If these changes are merged, then PR #1152 should only contain feature changes, i.e., changes that affect the serialization and deserialization of the Python functions and exceptions.

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -28,6 +28,7 @@ from work_queue import cctools_debug_config_file
 
 logger = logging.getLogger(__name__)
 
+
 def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
                           queue_lock=threading.Lock(),
                           launch_cmd=None,
@@ -61,7 +62,7 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
         logger.error("Unable to create WorkQueue object: {}".format(e))
         raise e
 
-    # Specify WorkQueue queue attributes 
+    # Specify WorkQueue queue attributes
     if project_name:
         q.specify_name(project_name)
     if project_password:
@@ -105,7 +106,7 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
                 continue
             parsl_id = item["task_id"]
 
-            # Extract information about the task 
+            # Extract information about the task
             function_data_loc = item["data_loc"]
             function_data_loc_remote = function_data_loc.split("/")[-1]
             function_result_loc = item["result_loc"]
@@ -159,7 +160,7 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
             if env is not None:
                 for var in env:
                     t.specify_environment_variable(var, env[var])
-            
+
             # Specify script, and data/result files for task
             t.specify_file(full_script_name, script_name, WORK_QUEUE_INPUT, cache=True)
             t.specify_file(function_data_loc, function_data_loc_remote, WORK_QUEUE_INPUT, cache=False)
@@ -217,7 +218,7 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
                     task_result = t.result
                     msg = None
 
-                    # Task failure 
+                    # Task failure
                     if status != 0 or (task_result != WORK_QUEUE_RESULT_SUCCESS and task_result != WORK_QUEUE_RESULT_OUTPUT_MISSING):
                         logger.debug("Wrapper Script status: {}\nWorkQueue Status: {}".format(status, task_result))
                         # Wrapper script failure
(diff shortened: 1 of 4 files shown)


parsl/executors/workqueue/executor.py
```diff
@@ -46,69 +46,71 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
...
     if project_name:
         q.specify_name(project_name)
-
     if project_password:
         q.specify_password(project_password)
     elif project_password_file:
         q.specify_password_file(project_password_file)
 
-    # Only write Logs when the log_dir is specified, which is most likely always will be
+    # Only write logs when the wq_log_dir is specified, which it most likely will be
```
benclifford: if that's a parsl-user-configured parameter, it's quite common to not want to specify an explicit log directory (or explicit anything, as much as you can get away with)

parsl/executors/workqueue/executor.py
```diff
@@ -136,40 +140,44 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
...
                 for var in env:
                     t.specify_environment_variable(var, env[var])
 
+            # Specify script, and data/result files for task
             t.specify_file(full_script_name, script_name, WORK_QUEUE_INPUT, cache=True)
-            t.specify_file(function_result_loc, function_result_loc_remote, WORK_QUEUE_OUTPUT, cache=False)
             t.specify_file(function_data_loc, function_data_loc_remote, WORK_QUEUE_INPUT, cache=False)
+            t.specify_file(function_result_loc, function_result_loc_remote, WORK_QUEUE_OUTPUT, cache=False)
             t.specify_tag(str(parsl_id))
+            logger.debug(t.id)
```
benclifford: it can be frustrating to get a value appearing in the log message without a description like "here is the task id: {}"

parsl/executors/workqueue/executor.py
```diff
@@ -184,51 +192,74 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
                 collector_queue.put_nowait(msg)
                 continue
 
-            logger.debug("Task {} submitted workqueue with id {}".format(parsl_id, wq_id))
+            logger.debug("Task {} submitted WorkQueue with id {}".format(parsl_id, wq_id))
```
benclifford: submitted *to* WorkQueue ?

parsl/executors/workqueue/executor.py
```diff
@@ -184,51 +192,74 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
...
-                                reason += "command line parsing"
-                            if status == 2:
+                                reason += "problem parsing command line options"
+                            elif status == 2:
                                 reason += "problem loading function data"
-                            if status == 3:
+                            elif status == 3:
                                 reason += "problem remapping file names"
-                            if status == 4:
+                            elif status == 4:
```
benclifford: it would be a bit defensive to stick a final "else" on the end here that says something like "i don't know what the reason was" that will deal with future changes adding on wrapper status numbers rather than leaving the reason blank.

parsl/executors/workqueue/executor.py
```diff
@@ -184,51 +192,74 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
...
+                            elif task_result == 4 << 3:
+                                reason += "result could not be classified"
+                            elif task_result == 5 << 3:
+                                reason += "task failed, but not a task error"
+                            elif task_result == 6 << 3:
+                                reason += "unable to complete after specified number of retries"
+                            elif task_result == 7 << 3:
+                                reason += "task ran for more than the specified time"
+                            elif task_result == 8 << 3:
+                                reason += "task needed more space to complete task"
```
benclifford: same final `else` comment as above.

parsl/executors/workqueue/executor.py
```diff
@@ -375,9 +414,10 @@ def __init__(self,
         self.shared_files = set()
         self.registered_files = set()
         self.worker_output = see_worker_output
-        self.full = True
+        self.full = False
```
benclifford: is this a fix for a bug or is it a behaviour change?

tjdasso: I apologize, I missed this line when reading through the tidy-up changes. But what this does is enables full Work Queue monitoring. It was changed to false because CCTools monitoring was a bottleneck for our throughput tests, and removing it helped get higher throughput values for the tests. Perhaps, the monitoring could be added as an option for the executor? It is currently only enabled/disabled manually through the script right now.

benclifford: My gut feeling for new stuff in parsl is to default to lots of debugging when the code is new and then as time passes and we care more about throughput rather than debugging, reduce that level.

Other places in parsl have user controllable worker debugging - for example, see the `worker_debug` parameter to `HighThroughputExecutor` in parsl/executors/high_throughput/executor.py

parsl/executors/workqueue/executor.py
tjdasso: The wq_log_dir is automatically specified according to the current directory. The only time it is would not be specified is if `os.path.join()` returns None, but otherwise the path is autoamatically created.

parsl/executors/workqueue/executor.py
tjdasso: Ok, it sounds like it is best to set it as a parameter for the Work Queue Executor, but have it default to enable debugging.

PR #1163, opened 2019-07-25
Set optional port as proper Channel attribute.

jhover:
This avoids this error when custom port is set:

AttributeError: class OAuthSSHChannel uses port in the constructor, but does not define it as an attribute


benclifford:
I feel like we should be addressing this with a call to super().__init__. but this PR should be fine

PR #1165, opened 2019-07-26
Move globus code into one file, ready to become a staging provider

benclifford:
There will be one Globus object per GlobusScheme now, rather than one per data manager - this is
to move globus-related state out of the data manager, in preparation for future PRs where
staging providers manage their own configuration, with the data manager knowing nothing other than
a reference to the staging provider.

I think this shouldn't cause significantly-noticeable change in behaviour.

Tested on my laptop with stage-in and stage-out testing.


parsl/data_provider/data_manager.py
```diff
@@ -156,22 +122,8 @@ def stage_out(self, file, executor):
         elif file.scheme == 'ftp':
             raise Exception('FTP file staging out is not supported')
         elif file.scheme == 'globus':
-            globus_ep = self._get_globus_endpoint(executor)
-            stage_out_app = self._globus_stage_out_app()
-            return stage_out_app(globus_ep, inputs=[file])
+            globus_scheme = _get_globus_scheme(self.dfk, executor)
```
yadudoc: In the case of multi-site with globus enabled sites, we may not know what the stage_out destination is. In that situation, we just need to pass a pointer to where the file is and have the next stage_in handle the actual transfer.

benclifford:
We always know the stage-out endpoint, and the stage-in endpoint:  it's the endpoint listed in the `globus:` URL.

This PR isn't trying to implement anything that causes a subsequent stage-in to pull data from a some different endpoint, not the endpoint that was specified in the `File` description.

PR #1169, opened 2019-07-30
Replace stage_out inhibiting with a more general test

benclifford:
When an app finishes, stage out of that app's outputs is inhibited
in certain circumstances: when that app is a staging app, so that
staging is not re-invoked.

Prior to this patch, the list of checks was fairly arbitrary.

This patch removes some of those checks: we can assume that the app
future is stored in the relevant DFK task slot, and that at the time
the done callback is invoked, it really is done.

The remaining checks attempt to determine if this task is a staging
task.

Previously this happened either by checking if the task was run
with the data_manager executor or if it was one of two specific app
names. Those tests are tied closely to the implementation of
staging.

This patch replaces that with an app keyword argument that sits
alongside output=[...], called staging_inhibit_output, which
inhibits staging of outputs if set to true.

That keyword is now set on the existing app invocations which
need staging inhibited.

This is intended to allow arbitrary other apps to be used for staging,
in a future PR, with stageout inhibited for those apps.


benclifford:
`staging_inhibit_output = False` should be treated the same as it not being there at all -- so mostly I think in those situations it would be left undeclared. I can't off the top of my head imagine a situation where some calls to an app would want staging on, and other calls to the same app would want staging off.

PR #1174, opened 2019-08-01
Add options for the user to specify cpu and memory for each kube pod

ZhuozhaoLi:
adding options for the user to specify cpu and memory for each pod

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -34,6 +34,18 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         Minimum number of blocks to maintain.
     max_blocks : int
         Maximum number of blocks to maintain.
+    max_cpu : float
+        CPU limits of the blocks (pods).
+        In cpu units, check kubernetes docs for more details. Default is 2.
+    max_mem : str
+        Memory limits of the blocks (pods).
+        In Mi or Gi, check kubernetes docs for more details. Default is 500MiB.
+    init_cpu : float
+        CPU initial requests of the blocks (pods).
+        In cpu units, check kubernetes docs for more details. Default is 1.
+    init_mem : str
+        Memory initial requests of the blocks (pods).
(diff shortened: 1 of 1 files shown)


parsl/providers/kubernetes/kube.py
```diff
@@ -63,6 +75,10 @@ def __init__(self,
                  init_blocks: int = 4,
                  min_blocks: int = 0,
                  max_blocks: int = 10,
+                 max_cpu: float = 2,  # in cpu units
+                 max_mem: str = "500Mi",  # in Mi, Gi
```
benclifford: These units do not need to be documented here - use the style used elsewhere, where the units are in the main docstring

parsl/providers/kubernetes/kube.py
```diff
@@ -34,6 +34,18 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         Minimum number of blocks to maintain.
     max_blocks : int
         Maximum number of blocks to maintain.
+    max_cpu : float
+        CPU limits of the blocks (pods).
+        In cpu units, check kubernetes docs for more details. Default is 2.
+    max_mem : str
+        Memory limits of the blocks (pods).
+        In Mi or Gi, check kubernetes docs for more details. Default is 500Mi.
```
benclifford: For existing kubernetes users, it would be useful to make it clear that there fields correspond to kubernetes side `limit` and `request`. That is obvious looking at this PR but maybe won't be later.

benclifford:
The original description of this was a better title than the original title so I copied it over with minor edits

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: fixed the comments and docstrings.

parsl/providers/kubernetes/kube.py
ZhuozhaoLi: Added this to docstrings.

Issue #1176, opened 2019-08-05
Tidy up unused parameters in provider and launcher API


benclifford:
Remove:

- [x] block_size from provider.submit   -- this is  mostly ignored and always(?) 1 when invoked
- [x] walltime from launcher.__call__ -- this is mostly ignored and always(?) not supplied on invocation

This came from review of PR #1175.

PR #1177, opened 2019-08-05
Remove unused walltime parameter from launcher API

benclifford:
This will affect parsl users who have implemented their own launcher.

Addresses part 2/2 of #1176


benclifford:
This will affect parsl users who have implemented their own launcher.

Addresses part 2/2 of #1176

Issue #1173, opened 2019-07-31
Globus stage_out does not work if executor.working_dir is set


benclifford:
During stage in of a file `f`, `f.local_path` is set to a path under the executor `working_dir`, and globus stages in to that same path. This works.

When instead an app is annotated to stage out a File `f`, `f.local_path` is not set during app execution - so the file path for `f` defaults to the current directory, which may different from the executor `working_dir`; however globus staging attempts to stage out from the `working_dir`. When that differs from the current directory, that stage out fails.

I have encountered this when trying to test staging with apps running on local executors with different working_dirs to keep their files separate (otherwise, working files from one app mayu appear to another app, even if staging failed)

I think this may be the behaviour I was originally experiencing with #784 that I couldn't reproduce.

If the staging code is to be allowed to choose where `f.local_path` points to, it should probably be given the opportunity to make such annotations before task execution, even though stage out itself happens after task execution. This will also be the case for future stage out work where non-globus providers are used.

benclifford:
fixed in #1191

Issue #778, opened 2019-02-19
Failing stageout does not cause any failure, and apps complete even with failed stageout


benclifford:
AppFutures and DataFutures are marked as finished before output stage-out happens.

DataFutures marked as successfully should probably mean that stage-out has happened successfully.

Possibly AppFutures should not be marked as successfully complete until stage out has happened either.

I've tested this with the below code, running to my own Globus Connect Server installation, to demonstrate this - the code reports that the test has completed successfully, and then: in the case that stageout doesn't work due to permissions, hangs rather than exiting (see #762); or in the case of a successful transfer, performs the transfer after the test has reported successfully.

Stage out jobs are submitted as separate tasks at the point that an app completes. Probably all of the stage out jobs should be joined on, and that join point is when the App and DataFutures report themselves as complete.

If stageout fails, then this is not reported except in debug logs. In particular, the test suite does not detect failing stageouts.

yadudoc:
Just for context, I think the thought process at the time was that, we planned two distinct types of data movement:
1. Wide-area transfers with globus, usually onto a persistent-node.
2. Data transfer between transient nodes (worker-login)

Globus and such transfer modes run from the parsl-client side and can be restarted/retried and data loss is unlikely. Here decoupling AppFutures and DataFutures made some sense.

In the second mode (which we don't support yet), staging in/out is packaged together with the task, and the task completely only when the data has been moved out. We assume data is not yet available until it is moved to client-side.

PR #1180, opened 2019-08-09
Move output staging task submission to before task launch

benclifford:
This facilitates a few things which will appear in future PRs:

 * pluggable app replacement stage-out, where the submitted app
   task needs to know about stage outs at the time that it is submitted,
   rather than the DFK managing this separately after the task
   has completed.

 * a fix for issue #778, so that DataFutures completion can
   indicate successful stageout

 * a fix for issue #1173, where the stage-out code must be given a chance
   to set the localpath on a File before the main app task runs

Stage out tasks cannot run immediately upon submission now, as they are
submitted before their corresponding main task has completed. This PR
adds an extra parameter to stage out tasks to allow them to depend on the
main task.

That extra dependency means that stageout tasks depend on a task (the main
task) that does not fully exist at stageout submission time, although we
know the main task ID. `check_dep` is modified to take that into account,
by treating task IDs which have no DFK task structure as being
not-yet-complete.

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -69,7 +69,7 @@ class DataManager(object):
         else:
             raise Exception('Staging in with unknown file scheme {} is not supported'.format(file.scheme))
 
-    def stage_out(self, file, executor):
+    def stage_out(self, file, executor, app_fu):
         """Transport the file from the local filesystem to the remote Globus endpoint.
 
         This function returns a DataFuture.
@@ -87,6 +87,6 @@ class DataManager(object):
         elif file.scheme == 'globus':
             globus_scheme = _get_globus_scheme(self.dfk, executor)
             stage_out_app = globus_scheme._globus_stage_out_app(executor=executor, dfk=self.dfk)
-            return stage_out_app(inputs=[file])
+            return stage_out_app(app_fu, inputs=[file])
         else:
             raise Exception('Staging out with unknown file scheme {} is not supported'.format(file.scheme))
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -335,21 +335,6 @@ class DataFlowKernel(object):
             if self.checkpoint_mode == 'task_exit':
                 self.checkpoint(tasks=[task_id])
 
-        # Submit _*_stage_out tasks for output data futures that have output files,
-        # that do not have stageing inhibited.
-
-        logger.debug("Submitting stage out jobs")
-        app_fu = self.tasks[task_id]['app_fu']
-
-        if app_fu.exception() is None and not self.check_staging_inhibited(self.tasks[task_id]['kwargs']):
-            for dfu in app_fu.outputs:
-                f = dfu.file_obj
-                if isinstance(f, File) and f.is_remote():
-                    logger.debug("Submitting stage out job for output file {}".format(f))
-                    self.data_manager.stage_out(f, self.tasks[task_id]['executor'])
-                else:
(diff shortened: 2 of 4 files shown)


parsl/dataflow/dflow.py
```diff
@@ -646,8 +643,17 @@ def submit(self, func, *args, executors='all', fn_hash=None, cache=False, **kwar
             raise ValueError("Task {} supplied invalid type for executors: {}".format(task_id, type(executors)))
         executor = random.choice(choices)
 
+        task_stdout = kwargs.get('stdout')
```
annawoodard: Will putting this before [this block](https://github.com/Parsl/parsl/blob/master/parsl/dataflow/dflow.py#L652-L664) break the functionality of `parsl.AUTO_LOGNAME`?

benclifford: Probably yes

benclifford: The test introduced in #1182 shows this breakage

benclifford: #1183 is an attempt to make AppFuture's property handling a bit cleaner, driven by this problem. This PR (#1180) will need updating to accomodate that if it is merged.

parsl/dataflow/dflow.py
```diff
@@ -510,6 +495,18 @@ def _add_input_deps(self, executor, args, kwargs):
 
         return tuple(newargs), kwargs
 
+    def _add_output_deps(self, executor, args, kwargs, app_fu):
+        logger.debug("Adding output dependencies")
+
+        if not self.check_staging_inhibited(kwargs):
+            outputs = kwargs.get('outputs', [])
+            for f in outputs:
+                if isinstance(f, File) and f.is_remote():
```
annawoodard: Is `is_remote` the right description (or does it not matter?) I'm thinking what we really mean here is `is_remote_writeable` (which wouldn't include `http` for example).

parsl/data_provider/data_manager.py
```diff
@@ -87,6 +87,6 @@ def stage_out(self, file, executor):
         elif file.scheme == 'globus':
             globus_scheme = _get_globus_scheme(self.dfk, executor)
             stage_out_app = globus_scheme._globus_stage_out_app(executor=executor, dfk=self.dfk)
-            return stage_out_app(inputs=[file])
+            return stage_out_app(app_fu, inputs=[file])
```
yadudoc: It doesn't look like the signature for _globus_stage_out matches what it's being called with here.

parsl/dataflow/dflow.py
benclifford: That's a bit vague. That remote concept goes away in a few PRs time I hope - see big data flow PR - so I haven't been too concerned. As of this PR, I think it means that an http output file would be handed to the stage out code and break there with a reasonable error message.

parsl/data_provider/data_manager.py
benclifford: @yadu:

It works when I run it!

stage_out_app is a partially applied decorated form of `_globus_stage_out`. 

The signature of _globus_stage_out is:

```
def _globus_stage_out(scheme, executor, app_fu, inputs=[]):
```

and then `_globus_stage_out_app` partially applies the first two parameters, so that the signature of `stage_out_app` is what remains: `app_fu` and `inputs`.

I think probably when all the staging API changes have settled down, the globus staging provider could be revisited to simplify the code: it's got more complicated as part of adjustments to fit into the upcoming API, but those complications maybe aren't necessary.

Issue #1125, opened 2019-07-09
staging-in does not happen for apps waiting on a file from an earlier app wrapped in a DataFuture


benclifford:
When an app takes an input that is a `DataFuture` from a previous app, rather than an explicitly created `File` object, staging-in does not happen - because staging only happens for `File`s, not for `DataFuture`s wrapping `Files`.

This occurs in code that looks like `parsl/tests/test_data/test_file_apps.py` where the output of each step is a DataFuture that will eventually resolve to a File that should then be used by the next step.

In that test, `parsl/tests/test_data/test_file_apps.py`, the lack of staging-in is hidden because only local files that do not need staging are used.

benclifford:
This probably interacts with #778 - staging-in for a subsequent task probably needs to be aware of when staging-out for a preceding task has completed (or has failed)

benclifford:
fixed by #1181

PR #1181, opened 2019-08-09
Treat DataFuture inputs as dependencies

benclifford:
Prior to this PR, DataFutures were treated, dependency-wise, like
any other input value, and did not cause any deferral of execution.

This meant that output DataFutures from one app passed as inputs
to a second app would not cause the second app to wait for the
first app to complete.

Fixes issue #1125

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -54,12 +54,12 @@ class DataManager(object):
         if file.scheme == 'ftp':
             working_dir = self.dfk.executors[executor].working_dir
             stage_in_app = _ftp_stage_in_app(self, executor=executor)
-            app_fut = stage_in_app(working_dir, outputs=[file], staging_inhibit_output=True)
+            app_fut = stage_in_app(working_dir, parent_fut=parent_fut, outputs=[file], staging_inhibit_output=True)
             return app_fut._outputs[0]
         elif file.scheme == 'http' or file.scheme == 'https':
             working_dir = self.dfk.executors[executor].working_dir
             stage_in_app = _http_stage_in_app(self, executor=executor)
-            app_fut = stage_in_app(working_dir, outputs=[file], staging_inhibit_output=True)
+            app_fut = stage_in_app(working_dir, parent_fut=parent_fut, outputs=[file], staging_inhibit_output=True)
             return app_fut._outputs[0]
         elif file.scheme == 'globus':
             # what should happen here is...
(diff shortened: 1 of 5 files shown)


parsl/dataflow/dflow.py
```diff
@@ -496,17 +497,25 @@ def _add_input_deps(self, executor, args, kwargs):
...
             if isinstance(f, File) and f.is_remote():
-                inputs[idx] = self.data_manager.stage_in(f, executor)
+                inputs[idx] = self.data_manager.stage_in(f, executor, None)
+            # if neither of the above, then local File, or local DataFuture, or non-File-related
+            # so do no stage-in
 
         for kwarg, f in kwargs.items():
-            if isinstance(f, File) and f.is_remote():
-                kwargs[kwarg] = self.data_manager.stage_in(f, executor)
+            if isinstance(f, DataFuture) and f.file_obj.is_remote():
```
annawoodard: Consider making `parent_fut` required and performing the `isinstance(f, DataFuture)`/`isinstance(f, File)` check inside `data_manager.stage_in`-- then we avoid the repeated checks below

parsl/data_provider/data_manager.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, dfk):
         self.dfk = dfk
         self.globus = None
 
-    def stage_in(self, file, executor):
+    def stage_in(self, file, executor, parent_fut):
```
annawoodard: If this remains optional, I favor `parent_fut=None` to make explicit that it is optional.

benclifford:
I can't see how this PR ever worked except for accidentally... I'll try to get it more comprehensively tested.

benclifford:
recent commits on this branch pass the prior future to apps as I originally intended. this should be ready for review now.

parsl/data_provider/data_manager.py
```diff
@@ -24,26 +27,39 @@ def __init__(self, dfk):
...
             - executor (str) : an executor the file is going to be staged in to.
         """
 
+        if isinstance(input, DataFuture) and input.file_obj.is_remote():
+            file = input.file_obj
+            parent_fut = input
+        elif isinstance(input, File) and input.is_remote():
+            file = input
+            parent_fut = None
+        else:
```
annawoodard: Should we have some kind of warning to the user in case they've passed a string path and are expecting a file with that name to be staged?

PR #1182, opened 2019-08-11
Add CI test for parsl.AUTO_LOGNAME [NEEDS stdout/stderr BEHAVIOUR TO BE BETTER DEFINED]

benclifford:


benclifford:
This new test is failing in the workqueue executor. I can't immediately see the cause. @dthain?

benclifford:
Further playing suggests that the workqueue executor does not like complicated pathnames: those with directory components in them, at least for stdout. See PR #1185. This might be the failure for this PR too, as parsl.AUTO_LOGNAME paths are absolute paths on the filesystem.

benclifford:
I think assumptions about shared filesystems need to be tightened up before this will work with workqueue - which is where it is failing at the moment. I'm leaving this PR open because it is something that definitely needs fixing and testing.

benclifford:
closing this because it's unlikely stderr/out handling will get tidied up as needed for this

PR #1183, opened 2019-08-11
Make AppFuture task-related properties be view of task_def struct

benclifford:
Prior to this, AppFuture properties were copies provided at construction time,
which was fragile in the presence of these properties being modified - specifically,
the handling of parsl.AUTO_LOGNAME.

This PR makes the task_def structure the only place where these values
live, and makes the corresponding AppFuture properties a view into that
task_def structure.

### DIFF
--- a/parsl/dataflow/futures.py
+++ b/parsl/dataflow/futures.py
@@ -126,11 +126,11 @@ class AppFuture(Future):
 
     @property
     def stdout(self):
-        return self.task_def['kwargs']['stdout']
+        return self.task_def['kwargs'].get('stdout')
 
     @property
     def stderr(self):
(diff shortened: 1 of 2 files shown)


parsl/dataflow/futures.py
```diff
@@ -131,15 +126,15 @@ def parent_callback(self, executor_fu):
 
     @property
     def stdout(self):
-        return self._stdout
+        return self.task_def['kwargs']['stdout']
```
annawoodard: I think it is ok for these to be undefined, so this should return `self.task_def['kwargs'].get('stdout')` (and similarly for stderr below).

PR #1184, opened 2019-08-11
Improve wrapper script error handling for workqueue

benclifford:
In some situations that I've encountered (working on PR#1181),
t.output is None and so error reporting breaks with:

```
...
  File "/home/benc/parsl/src/parsl/parsl/executors/workqueue/executor.py", line 244, in WorkQueueSubmitThread
    reason += "\nTrace:\n" + t.output
TypeError: Can't convert 'NoneType' object to str implicitly
```

This PR also logs the numerical value of WQ task return status when it does not have a corresponding text description.

When the above error happens, apps hang rather than return an error. With this PR applied, those apps fail rather than hang.


benclifford:
@dthain etc - does this look ok?

dthain:
Right, the output of a task could potentially be blank, in which case `t.output` may be null.

Issue #363, opened 2018-07-01
Automatically transfer 'stdout' and 'stderr'


annawoodard:
We allow a user to pass `stdout` and `stderr` keywords to bash apps, but these have no effect if the app is run remotely. We should automatically transfer these files back. Related to #276 and #361.

annawoodard:
This issue depends on #219

kylechard:
Or support streaming of stdout/err. This would be useful for development/testing. E.g., 

Example from Francois: 

``
@bash_app
def run_process(job_dir, job_script, stdout=sys.stdout, stderr=sys.stderr):
    return f"cd {job_dir} && bash {job_script}
``

benclifford:
The test suite contains at least one test for stdout/stderr which makes assumptions about std file appending/location: `test_stdout_append` in `parsl/tests/test_bash_apps/test_stdout.py`.

This test is likely to not work in the presence of any kind of non-shared filesystem, which this issue is related to.

benclifford:
I think staging of these parameters should be made to be the same as for any other I put and output file, using a `File` object. Things like streaming should be implemented as regular staging mechanisms that can work for all files.

PR #1185, opened 2019-08-11
Add more complicated filepath testing

benclifford:
These tests fail for me with workqueue. cc @dthain

This is motivated by a problem with PR #1182 which has exposed our lack of interesting file path testing in CI.


benclifford:
closing because stdout/stderr handling is not cleanly defined - see issue #363

PR #1186, opened 2019-08-13
Add stageout test for globus

benclifford:


parsl/tests/test_staging/test_implicit_staging_globus.py
```diff
@@ -32,7 +33,34 @@ def test_implicit_staging_globus():
     sorted_file = File('sorted.txt')
 
     f = sort_strings(inputs=[unsorted_file], outputs=[sorted_file])
+
+    f.result()
+
+
+@pytest.mark.local
+def test_stage_in_out_globus():
+    """Test stage-in then stage-out to/from Globus
```
annawoodard: > Test stage-in then stage-out to/from Globus
> "sometimes [this test will] pass even though stageout is not working."

😆 😢 

I think we can tighten this test up (or add a new one) when #778 is addressed.

PR #1200, opened 2019-08-20
Remove parent member from AppFuture implementation

benclifford:
This put all knowledge of the parent(s) of an AppFuture
into a single place - the DFK - rather than having it
split between the DFK and the AppFuture.

The AppFuture parent_callback is now invoked by the
DFK in the callback that the DFK has registered with
parent futures, rather than the parent futures having
multiple registrations. This makes the result handling
codepath driven through that one single callback.

Code relating to the parent attribute of AppFuture
is removed. This includes removing the overridden
running() method, which was not implemented in any
existing parent future implementations; and removing
app future parent references from DataFuture.

This will possibly change the order in which callbacks
happen - and subtle races there might be exposed. This
would most likely be exposed to an end-user if they have
registered callbacks with an AppFuture which expect the
DFK task state to be in a particular state; the order
of state update vs AppFuture callback invocation may
now be different.


parsl/dataflow/dflow.py
```diff
@@ -247,6 +247,8 @@ def handle_exec_update(self, task_id, future):
              makes this callback
         """
 
+        self.tasks[task_id]['app_fu'].parent_callback(future)
```
yadudoc: Why do we need this callback registration here ?

benclifford: It's not a registration - it's calling the actual callback.

The whole mechanism of update_parent registering an executor future level callback to call this app future callback is removed by this PR, and replaced by this call here (which also ultimately triggers when the exec future completes... But all the exec future completion code is now called from handle_exec_update instead of being split across two callbacks that are separately registered)

I think that AppFuture callback should go away entirely in a future PR.

PR #1201, opened 2019-08-20
Implement file staging provider API

benclifford:
Instead of dispatching inside the DFK explicitly based on a
hard-coded map between URI schemes and staging implementations,
go through each staging scheme in sequence, giving that scheme the chance
to handle staging

The concept of 'is_remote' being a property of files defined
by the URL scheme is also superceded by this, and so
is_remote is removed. For behaviour of "file:" URLs to mean
'no staging', then a no-op staging scheme is implemented
in this PR. Other  staging providers can implement different
semantics for `file:` URLs.

There's a user-facing API change here: if a configuration specifies
`storage_access=[]` rather than omitting it or specifying `storage_access=None`,
then the default staging mechanisms will be removed and
even no-op file access won't work; and the way to configure Globus has changed.

Start reading at parsl/data_provider/staging.py, which defines the new API.
Existing staging code is modified to implement that API, and existing code that calls into staging does so via that API - defaulting to a set of staging providers giving equivalent behaviour to that  immediately before this PR.

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -1,12 +1,22 @@
 import logging
+from concurrent.futures import Future
+from typing import Any, Callable, List, Optional, TYPE_CHECKING
 
 from parsl.app.futures import DataFuture
 from parsl.data_provider.files import File
-from parsl.data_provider.ftp import _ftp_stage_in_app
-from parsl.data_provider.globus import _get_globus_scheme
-from parsl.data_provider.http import _http_stage_in_app
+from parsl.data_provider.file_noop import NoOpFileStaging
+from parsl.data_provider.ftp import FTPSeparateTaskStaging
+from parsl.data_provider.http import HTTPSeparateTaskStaging
+from parsl.data_provider.staging import Staging
+
+if TYPE_CHECKING:
+    from parsl.dataflow.dflow import DataFlowKernel
 
 logger = logging.getLogger(__name__)
+
+# these will be shared between all executors that do not explicitly
+# override, so should not contain executor-specific state
+defaultStaging = [NoOpFileStaging(), FTPSeparateTaskStaging(), HTTPSeparateTaskStaging()]  # type: List[Staging]
 
 
 class DataManager(object):
@@ -14,7 +24,7 @@ class DataManager(object):
 
     """
 
-    def __init__(self, dfk):
+    def __init__(self, dfk: "DataFlowKernel") -> None:
         """Initialize the DataManager.
 
         Args:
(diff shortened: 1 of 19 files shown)


annawoodard:
> even no-op file access won't work.
How does this look to the user?

benclifford:
> > even no-op file access won't work.
> 
> How does this look to the user?

The user gets this exception:

> raise ValueError("Executor {} cannot stage file {}".format(executor, repr(file)))

if they try to stage any `File`s.

This is the exception that happens when no staging provider claims a particular File, and having no staging providers at all is a specific case of that.

parsl/executors/threads.py
```diff
@@ -21,7 +20,7 @@ class ThreadPoolExecutor(ParslExecutor, RepresentationMixin):
         Number of threads. Default is 2.
     thread_name_prefix : string
         Thread name prefix (only supported in python v3.6+).
-    storage_access : list of :class:`~parsl.data_provider.scheme.Scheme`
+    storage_access : list of :class:`~parsl.data_provider.staging.Staging`
```
annawoodard: I think we need to flesh out the docstring a bit to help people that aren't reading the source. What is the default? What happens if you pass `None`? What happens if you pass `[]`?

Also, consider copy-pasting this to the other executors, which don't appear to have any entries for `storage_access`. (I see this problem existed before your changes, but this seems like as good a time as any to add it. I'm also happy to defer this to a 'make it possible for users to understand data management by updating all docs and docstrings' issue.)

parsl/data_provider/data_manager.py
```diff
@@ -1,20 +1,30 @@
...
+from parsl.data_provider.staging import Staging
+
+if TYPE_CHECKING:
+    from parsl.dataflow.dflow import DataFlowKernel
 
 logger = logging.getLogger(__name__)
 
+# these will be shared between all executors that do not explicitly
+# override, so should not contain executor-specific state
+defaultStaging = [NoOpFileStaging(), FTPSeparateTaskStaging(), HTTPSeparateTaskStaging()]  # type: List[Staging]
```
annawoodard: i favor `defaultStaging` -> `default_staging`

parsl/data_provider/staging.py
```diff
@@ -0,0 +1,68 @@
...
+    will call both `stage_in` and `replace_task` to give the provider the
+    opportunity to perform staging.
+
+    For each file to be staged out, the data manager will follow the same
+    pattern using the corresponding stage out methods of this class.
+
+    The default implementation of this class rejects all files, and
+    performs no staging actions.
+
+    To implement a concrete provider, one or both of the `can_stage_*`
```
annawoodard: Should we enforce the interface with ABCs?

parsl/data_provider/staging.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Optional, Callable
+from parsl.app.futures import DataFuture
+from parsl.data_provider.files import File
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from parsl.data_provider.data_manager import DataManager
+
+
+class Staging:
```
annawoodard: It seems we mix `Staging` and `Scheme` (i.e. `GlobusScheme` but `FTPSeparateTaskStaging`, and use 'staging provider/staging scheme' interchangeably) throughout. Do they mean different things? If not, can we pick one and remove the other?

parsl/data_provider/staging.py
```diff
@@ -0,0 +1,68 @@
...
+    For each file to be staged out, the data manager will follow the same
+    pattern using the corresponding stage out methods of this class.
+
+    The default implementation of this class rejects all files, and
+    performs no staging actions.
+
+    To implement a concrete provider, one or both of the `can_stage_*`
+    methods should be overridden to match the appropriate files, and then
+    the corresponding `stage_*` and/or `replace_task*` methods should be
+    implemented.
```
annawoodard: I don't see `defaultStaging` documented anywhere user-facing. Maybe this is a good place to mention it?

parsl/executors/threads.py
```diff
@@ -21,7 +20,7 @@ class ThreadPoolExecutor(ParslExecutor, RepresentationMixin):
...
         Specifications for accessing data this executor remotely. Multiple `Scheme`s are not yet supported.
```
annawoodard: Needs update?

parsl/data_provider/staging.py
benclifford: There isn't anything to enforce here, I think -- it's fine to leave any of the methods non-overridden because they have "sensible" do-nothing defaults, and it is not likely that a particular provider would want to override all of the methods.

annawoodard: :+1:

parsl/executors/threads.py
benclifford: I would like to write some user-guide chapter on "What is file handling/staging" which this can refer to. Not in this PR, though.

parsl/data_provider/staging.py
benclifford: The word "scheme" in file staging has had two meanings: one for the URL scheme (the bit before the colon, which is implicitly `file:` if not specified), and the other for the code that started off as GlobusScheme and has subsequently evolved into what I'm calling staging providers. I've added a commit that tries to distinguish the two uses of the term, keeping "scheme" for URLs, but using "provider" for the second case.

parsl/executors/threads.py
```diff
@@ -31,15 +30,20 @@ class ThreadPoolExecutor(ParslExecutor, RepresentationMixin):
     @typeguard.typechecked
     def __init__(self, label: str = 'threads', max_threads: int = 2,
                  thread_name_prefix: str = '', storage_access: List[Any] = None,
+                 # storage_access should be a list of Staging, but Lists are by default
+                 # invariant, not co-variant, and it looks like 'typeguard' author actually
+                 # prefers to fix stuff in 'pytypes' not 'typeguard' - so it's a list of Any for now.
+                 # maybe should port to pytypes, our third live typechecker?
```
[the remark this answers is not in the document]
benclifford: I've fiddled with this text and moved it into the `ParslExecutor` docstring.

parsl/data_provider/staging.py
benclifford: I think the right place is in an updated userguide data section which I'd like to write in a different PR

parsl/executors/threads.py
annawoodard: :+1:

parsl/data_provider/staging.py
annawoodard: I see an example has been added to the user guide. I still favor putting important bits in docstrings for users that like to stay in their terminal or notebook and do `help(foo)`. But I'm also ok with pressing forward as-is.

Issue #942, opened 2019-05-14
Support for node resource hints in provider


yadudoc:
With IPP we had to explicitly specify the number of workers to be launched per node, since the launch system doesn't necessarily run on the compute node and lacked information about the available resource before hand. As we developed HTEX, we moved to an architecture where we might use a launch system that starts 1 manager per node, and this manager after probing available cpu/mem/ resources on the node launches workers based on per-worker resource limits set by the user. This only helps once we have a node provisioned and doesn't help in the strategy planning phase where resource information is still unavailable and we assume the worst case, only 1 worker can run per node, leading to gross overestimation of resources required. One potential automatic method would be to launch a single block and wait until resource information is available before attempting to scale further, but this might work poorly with slow queues. This would definitely take more effort, and is better done with the plan to revamp the strategy component.

When all else fails, we could fallback to the user to specify resource hints to the provider : 

```python
HighThroughputExecutor(
            label="htex",
            worker_debug=True,
            cores_per_worker=2,
            mem_per_worker=2,        # 2GB
            provider=SlurmProvider(
                'debug',
                nodes_per_block=2,
                cpus_per_node=64,         # <-- New
                mem_per_node=128,         # <-- New
            )
)
```

This is a MolSSI/QCArchive requirement (@Lnaden)

annawoodard:
Note already have `cpus_per_node` in the `PBSProProvider`. Also note we probably want to keep #943 in mind as we implement this because they are closely related.

yadudoc:
This is high priority for QCArchive.

yadudoc:
The quickest way to go about this would be to specify exactly the resource slice you need from the scheduler, via Parsl's `provider.scheduler_options` kwarg. This would be followed by the `provider.worker_init` kwarg which would export a few global variables that Parsl can use. We need to add support for using these exported options in the Parsl worker code. I'll add a separate issue for that.

Here's a sample config that would work *after* we've got the bits implemented : 

```python
HighThroughputExecutor(
    label="htex",
    cores_per_worker=2,
    mem_per_worker=2,        # 2GB
    provider=SlurmProvider(
          'debug',
          nodes_per_block=2,
          scheduler_options = '#SBATCH --cpus-per-task=2 --mem-per-cpu=1g --ntasks=1',
          worker_init='export PARSL_MAX_MEMORY=2G; export PARSL_MAX_CPUS=2',
     )
)
```
@Lnaden, @dgasmith Could you take a look please ?

[...]

PR #1207, opened 2019-08-22
Support cpu and mem specification via env

annawoodard:
This commit adds support for respecting specification of how much CPU
and memory should be used by workers via the PARSL_CORES and
PARSL_GB_MEMORY environmental variables (as opposed to inferring it
to be the full node, once a job has started on the node.) After this commit, the config
below will result in 4 workers being started, using a total of 4 cores
and 12 GB memory.

This is a pre-requisite to a full solution to #942.

```

config = Config(
    executors=[
        HighThroughputExecutor(
            cores_per_worker=1,
            mem_per_worker=3,
            provider=SlurmProvider(
                'broadwl',
                cmd_timeout=60,
                nodes_per_block=1,
                init_blocks=1,
                min_blocks=1,
                max_blocks=1,
                scheduler_options='#SBATCH --cpus-per-task=4 --mem-per-cpu=3g,
                worker_init='export PARSL_GB_MEMORY=$(expr $SLURM_CPUS_ON_NODE \* $SLURM_MEM_PER_CPU); export PARSL_CORES=$SLURM_CPUS_ON_NODE',
                exclusive=False
            ),
        )
    ],
)
```


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -123,8 +123,15 @@ def __init__(self,
...
 
-        cores_on_node = multiprocessing.cpu_count()
-        available_mem_on_node = round(psutil.virtual_memory().available / (2**30), 1)
+        if os.environ.get('PARSL_CORES'):
+            cores_on_node = int(os.environ['PARSL_CORES'])
+        else:
+            cores_on_node = multiprocessing.cpu_count()
+
+        if os.environ.get('PARSL_GB_MEMORY'):
+            available_mem_on_node = float(os.environ['PARSL_GB_MEMORY'])
```
benclifford: PARSL_MEMORY_GB?

benclifford:
I'm happy with the `process_worker_pool` changes.

I'm less sure about the changes to `launchers.py` - I'm not really familiar with all of the calculations there, so it's not immediately obviously incorrect or correct. One thing I'll note is that even in the default case, the way that process_worker_pool detects the default number of cores now changes. Previoulsy it would use `multiprocessing.cpu_count()` in python; but now it will use `getconf _NPROCESSORS_ONLN`. I'm unsure if that matters.

PR #1210, opened 2019-08-23
One pytest to rule them all

benclifford:
This is the beginning of some work to massively speed up pytest by running a single pytest rather than once-per-file - I spent one hour timeboxed to see how far I could get.

It aggressively removes bad test files and manual tests, and DFK loads in `local` tests - each of these should be revisited properly.

```
$ time pytest parsl --config parsl/tests/configs/local_threads.py --maxfail=1
...
real    1m9.365s
```

This is encouraging enough that I think this work should be pursued.


benclifford:
Status:

Some `local` tests relied on mutating a global config. This can't be done in a shared environment. The config mutations are commented out where I have encountered them, and can be seen in the diff.

Some files were deleted entirely because they were too hard for me to deal with in initial prototype. I have reviewed some and put in separate PRs to delete them entirely. Others should be restored and fixed to work with the new way of running tests.

benclifford:
working through this still.

main problem at the moment is that configs can't be (in general) used more than once in a session. This is even more restrictive than the "configs can't be mutated" restriction that I previously thought applied. However... my approach for solving that latter (with a `fresh_config`f function instead of a `config` variable exported from a test config) will also work for that.

benclifford:
there's already a pytest.ini which ignores a bunch of tests by pathname, it turns out...

benclifford:
It looks like a `-k` specified in that file is entirely overridden by a `-k` specified on the commandline - for example to specify `-k not cleannet` for use on network-restricted hosts.

PR #1221, opened 2019-08-24
Make and test that sites/ tests directory passes mypy

benclifford:
This directory contains currently-maintained test files so we should
care for it as much as the other test directories.


benclifford:
Test failed with an OS update problem, not a parsl problem

PR #1222, opened 2019-08-24
Add whole-run DFK pytest fixture. Remove many-config capability.

benclifford:
This change will (in non-local mode) load a single DFK for a specified
config, which will be used for all tests in the run.

For local mode, new ways of initialising tests are added: local_config,
local_setup and local_teardown, that can be used to write 'local'-mode
tests which do not interfere with other tests - the parsl test suite
local mode tests, at present, do lots of global stuff, and these new
parameters and methods allow that globus stuff to be more constrained
to individual test runtime.

This conflicts with the implementation of multiple test configurations
in one run - that feature causes the DFK to be loaded/unloaded for each
test module as pytest sweeps over the list of configs, even if the list
only has one entry.

Multiple test configuration handling is removed, and --configs is renamed
to --config.

The behaviour can be demonstrated by running:

pytest parsl/tests/test_bash_apps/test_apptimeout.py \
  parsl/tests/test_bash_apps/test_basic.py \
  --config parsl/tests/configs/local_threads.py --setup-show

before and after, where --setup-show will show when each fixture is executed

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -24,7 +24,7 @@ def pytest_addoption(parser):
         '--config',
         action='store',
         metavar='CONFIG',
-        nargs='1',
+        nargs=1,
         help="only run parsl CONFIG; use 'local' to run locally-defined config"
     )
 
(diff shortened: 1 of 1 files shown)


benclifford:
oops missing a bit

benclifford:
This is failing tests because the rearranged DFK initialisation now means two DFKs are present when a test module loads a DFK globally. I have fixed most of this in my other branch, so I'll probably abandon this PR in favour of merging all of that.

PR #1224, opened 2019-08-25
Generally beautify the user experience for running cctools installer

benclifford:
* throw an error if required environment variable isn't set.
 * make script executable
 * remove bash from travis invocation because script has a shebang


benclifford:
This was merged as part of #1210

Issue #1197, opened 2019-08-20
File `localpath` staging conflict


benclifford:
~~I haven't tried to make this happen, but noting as I think it is likely.~~ I have encountered this.

When multiple sites are involved and staging sets `localpath`, then when `File` objects are shared between sites - for example, Globus staging the same file to two different sites, or staging a file out from one site and then in to the next - then I think that the `File.localpath` attribute may be improperly shared between those sites.

`File.localpath` represents the local path of the file *on a particular site* (perhaps, eventually, even for a particular task, if per-task workdirs come into existence #89) - but staging code mutates `File.localpath` on the submit side without regard for any other conflcting staging action that would want a different values in that attribute.

Probably the correct way to deal with this is to replace no-local-path `File` objects with modified copies, rather than mutating them.

benclifford:
This also happens even in the single site case in the case that the staging provider runs in such a way that it modifies the submit side `File` object - so that the conflict occurs between the `local_path` set for the remote site, and the submit side request for the submit-side path.

I've encountered this for real in a (post-#1201, in development) locally-running staging provider, and will put in a PR to test for this happening.

benclifford:
This was fixed by #1272

PR #1231, opened 2019-08-27
Remove vestigial scheme enumeration

benclifford:
It should be up to staging providers to specify the appropriate
local_path for all schemes, except for file: which has an
implicit meaning on the submit side POSIX filesystem.

File.filepath will now throw an exception in many cases where
this does not make sense. As a consequence, some log messages
throughout the codebase are adjusted to not call filepath,
and instead use the full File repr.

Consequent changes are needed for the existing staging provider
to make them work.

Staging providers that attempted to implement a differently
named scheme would fail.#

This is related to the localpath problems in #1197

### DIFF
--- a/parsl/app/futures.py
+++ b/parsl/app/futures.py
@@ -71,8 +71,7 @@ class DataFuture(Future):
             else:
                 raise NotFutureError("DataFuture can be created only with a FunctionFuture on None")
 
-        logger.debug("Creating DataFuture with parent: %s", self.parent)
-        logger.debug("Filepath: %s", self.filepath)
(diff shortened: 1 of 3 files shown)


benclifford:
This fails where the staging task runs in a different python process to the main process - or specifically, where the `File` objects that the staging process mutates are (serialized...) copies of the `File` objects in the main submitting process, rather than the original `File` objects themselves.

This needs a bit more thinking about the flow of information between different tasks about how files have been staged.

benclifford:
closing this - it needs to be part of the broader #1197 fixing project

PR #1242, opened 2019-08-29
Catch and log errors in flow control callback

benclifford:
Prior to this, exceptions raised by the flow control
callback caused the flow control thread to die.

After this, exceptions raised by the flow control
callback are logged with a stack trace, but flow
control continues.

### DIFF
--- a/parsl/dataflow/flow_control.py
+++ b/parsl/dataflow/flow_control.py
@@ -136,7 +136,10 @@ class FlowControl(object):
                  triggered the callback
         """
         self._wake_up_time = time.time() + self.interval
-        self.callback(tasks=self._event_buffer, kind=kind)
+        try:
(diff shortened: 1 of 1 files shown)


benclifford:
failed in CI with an OS-level MPI install error:

```
0.09s$ bash parsl/executors/extreme_scale/install-mpi.sh $MPI
1577+sudo apt-get install -y -q mpich libmpich-dev
1578Reading package lists...
1579Building dependency tree...
1580Reading state information...
1581Package libmpich-dev is not available, but is referred to by another package.
1582This may mean that the package is missing, has been obsoleted, or
1583is only available from another source
1584
1585E: Unable to locate package mpich
1586E: Package 'libmpich-dev' has no installation candidate
1587The command "bash parsl/executors/extreme_scale/install-mpi.sh $MPI" exited with 100.
```

Issue #308, opened 2018-06-01
Fail if user tries to run multisite + data management


annawoodard:
This is not currently supported.

kylechard:
I don't think we can currently tell if it will or won't work. It will work in multi-site if there is a shared FS (e.g., Mira/Theta) but not otherwise.

benclifford:
Many bugs related to this have been fixed recently. #1197 is the only one outstanding at present.

PR #1253, opened 2019-09-07
Add a lot of debugging around File.local_path

benclifford:
This PR converts local_path into a `@property`.

It then adds log messages around related to use of local_path.

It also makes repeated assignment to local_path an invalid operation which
will now raise an exception. This is to help detect/debug improper sharing
of File objects, as might be encountered in situations related to issues #1197 and #308 
although this PR does not attempt to fix any of those issues.


benclifford:
This is failing in CI with some interesting behaviour around workqueue doing file mapping.

benclifford:
closing this as out-dated

PR #1254, opened 2019-09-07
Rearrange stage-in to facilitate upcoming duplication of Files

benclifford:
This is intended to support a later PR which will attempt to address #1197
by duplicating File objects inside the newly refactored function.

This present PR, however, should not change behaviour.

### DIFF
--- a/parsl/data_provider/data_manager.py
+++ b/parsl/data_provider/data_manager.py
@@ -55,15 +55,20 @@ class DataManager(object):
         # if we reach here, we haven't found a suitable staging mechanism
         raise ValueError("Executor {} cannot stage file {}".format(executor, repr(file)))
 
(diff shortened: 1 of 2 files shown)


parsl/data_provider/data_manager.py
```diff
@@ -55,15 +55,20 @@ def replace_task_stage_out(self, file: File, func: Callable, executor: str) -> C
         # if we reach here, we haven't found a suitable staging mechanism
         raise ValueError("Executor {} cannot stage file {}".format(executor, repr(file)))
 
-    def replace_task(self, input: Any, func: Callable, executor: str) -> Callable:
-        """This will give staging providers the chance to wrap (or replace entirely!) the task function."""
-
+    def optionally_stage_in(self, input, func, executor):
```
annawoodard: would prefer docstring here, data management is hard to grok

Issue #1257, opened 2019-09-08
PR1200 breaks DepFail handling


benclifford:
The below script hangs with master >= PR1200. It should instead exit with a dependency failure exception.

- [x] fix this - see PR #1259 
- [x] fix the test suite to catch it - see PR #1258

```

# test failure of datafutures when grandparent fails

import parsl

@parsl.python_app
def fails():
    raise ValueError("Deliberate failure")

@parsl.python_app
def depends(parent):
    return 1

parsl.load()
parsl.set_stream_logger()
f1 = fails()
f2 = depends(f1)

# f1.result()  # raises a Value error

f2.result()
```

PR #1258, opened 2019-09-08
Restore previously disabled dependency behaviour checks

benclifford:
These would have detected issue #1257 introduced in PR #1200.


benclifford:
This failed in one CI run with:

```
parsl/tests/test_python_apps/test_rand_fail.py .F.

E           RuntimeError: Expected DependencyError, but got no exception
parsl/tests/test_python_apps/test_rand_fail.py:115: RuntimeError
```

benclifford:
... I think this test has some chance of *not* causing dependency errors, because failure is deliberately random.

So actually test_rand_fail probably has no place in the test suite at all, and I will amend this PR to remove it rather than re-enable it.

PR #1259, opened 2019-09-08
Fix depfail handling introduced by PR #1200, issue #1257

benclifford:
In the codebase, dependency failures are handled by creating an executor
future that is populatd with an exception immediately rather than waiting
for some execution to happen.

Both before and after PR #1200, the callback setup differs between
executor futures that come from an actual executor vs for executor
futures that contain a dependency failure.

Prior to #1200, an regular executor future would get two callbacks
registered: one via AppFuture.update_parent to inform the AppFuture
and one to inform the DFK of the result. However, a DependencyFailure
executor future did not have both of those registered - it only had
the AppFuture.update_parent callback registered.

would now happen via the DFK callback; in the dependency failure case
this was not true (just as it had not been true before PR #1200).

This commit makes the handling of regular executor futures vs
dependency failure executor futures more consistent, by reducing the
amount of code in the no-failure vs failure paths and using the
same callback registration code no matter which of these two ways the
executor future is created.


parsl/dataflow/dflow.py
```diff
@@ -388,12 +380,20 @@ def launch_if_ready(self, task_id):
...
+                exec_fu = Future()
+                exec_fu.retries_left = 0
+                exec_fu.set_exception(DependencyError(exceptions,
+                                                      task_id,
+                                                      None))
+
+            if exec_fu:
+
+                try:
+                    exec_fu.add_done_callback(partial(self.handle_exec_update, task_id))
```
yadudoc: If we had the following sequence of deps C --> B ---> A, and A were to fail, were we only triggering a dep_fail on B and not C before this fix ?

benclifford: between PR #1200 and this PR #1259, depfails didn't show up at all in any of the tests I did.

before PR #1200, depfails worked to arbitrary depth - because the results propagated through the AppFuture callback set by AppFuture.update_parent (which #1200 removed)

benclifford: before PR #1200, (and before #1257) the handle_exec_update behaviour for depfailed tasks didn't happen, though -- that's mostly irrelevant in the depfail case except for monitoring updates.

Looking at that code now, the retry logic is in that callback too, so I should test if there is a bad interaction between that and depfails.

benclifford: Related to @yadudoc 's C->B->A question, I've added a pytest case into this PR that checks 4 tasks in a chain get propagated dependency failures.

benclifford: I'm still trying to convince myself of the retries vs depfail behaviour, but I did at least generate PR #1266 on the way.

benclifford: OK, with PR #1261 merged in, which runs pytests with (amongst other excitements) retries > 0, it looks like the depfail tests introduced in 7ddd676 fail, which aligns with the suspicion I had above. I'll dig into this more.

benclifford: I am comfortable with the dep_fail handling introduced in 4a84064 which passes the tests now

benclifford:
Exception coming out of callbacks fall out non-deterministically in various places, killing whoever was unlucky enough to invoke them. The behaviour needs to be that callbacks do not leak exceptions upward to their mystery caller who cannot handle them.

hategan:
You are probably referring to add_callback() occasionally invoking the callback directly. Callback exceptions should generally not be thrown, but, if they are, they should be handled in add_callback(), not whoever calls add_callback(), whose contract should not change nondeterministically.

benclifford:
They aren't handled in add_callback() - they're propagated outwards to the caller. In Python 3.7, the exception from the f callback propagates into a log message saying the exception is being logged and discarded (1 below). the g callback propagates up into the main thread (2 below)

```
from concurrent.futures import Future

def raiser(x):
  raise RuntimeError("foo")


f = Future()
f.add_done_callback(raiser)
f.set_result(7)


g = Future()
g.set_result(7)
g.add_done_callback(raiser)
```

(1):
```
exception calling callback for <Future at 0x7f3da87ecd68 state=finished returned int>
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 324, in _invoke_callbacks
    callback(self)
  File "a.py", line 4, in raiser
    raise RuntimeError("foo")
RuntimeError: foo
```

```
  File "a.py", line 14, in <module>
    g.add_done_callback(raiser)
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 403, in add_done_callback
    fn(self)
  File "a.py", line 4, in raiser
    raise RuntimeError("foo")
RuntimeError: foo
```

Mostly where add_done_callback is wrapped with a catch-and-log is to roughly emulate that first behaviour.

If we're going to do more interesting stuff when those exceptions come out, we also need to accept that sometimes we won't get those exceptions and they'll disappear out of our control in the style of (1) above.

benclifford:
> This was pretty annoying to reproduce, I remember it needed a fast app on a slow machine ?

I think this will happen on any use of memoized results where the future in the memo table is already completed.

benclifford:
My code demonstrates behaviour contrary to a simple reading of https://docs.python.org/3/library/concurrent.futures.html that:

> If the callable raises an Exception subclass, it will be logged and ignored.

hategan:
> My code demonstrates behaviour contrary to a simple reading of https://docs.python.org/3/library/concurrent.futures.html that:
> 
> > If the callable raises an Exception subclass, it will be logged and ignored.

That's because the Python future does the same thing as the parsl one. It ignores it if the callback is invoked later, and throws it if the callback is invoked immediately. So the code and docs are inconsistent. Here's the code (concurrent.futures._base.py):

```python
    def add_done_callback(self, fn):
        # [doctstring]
        with self._condition:
            if self._state not in [CANCELLED, CANCELLED_AND_NOTIFIED, FINISHED]:
                self._done_callbacks.append(fn)
                return
        fn(self)

    def _invoke_callbacks(self):
        for callback in self._done_callbacks:
            try:
                callback(self)
            except Exception:
                LOGGER.exception('exception calling callback for %r', self)
```

It does not seem to be very well thought out (or well documented; or both).

So maybe we shouldn't emulate that behavior.

benclifford:
Executors return Python futures. This is the behaviour we get and have to deal with, unless we change the executor API to require some other implementation of futures.

hategan:
great!

PR #1261, opened 2019-09-09
Make CI run htex tests twice, once with a more interesting config

benclifford:
This commit adds an htex_local_alternate file, and enables some different features in there in order to attempt to cover a wider set of parsl features in CI with more interesting apps than individual feature local tests.

This does not assert anything new explicitly - but checks that all of the existing tests hold in a reasonably different environment.

I initially configured monitoring in this alternate config, but it is unable to pass CI - instead it hangs. I have commented out the monitoring config clause for future reference.


benclifford:
The added test invocation appears to hang at exit time.

Issue #1274, opened 2019-09-12
test_docs/test_workflow1 hangs with monitoring enabled


hategan:
Running the test suite with configs/local_threads_monitoring hangs at test_workflow1.

This is possibly related to #1267

Issue #1267, opened 2019-09-10
Monitoring does not survive full test suite


benclifford:
When running the pytest suite with local executor with this clause added, I get a hang at the end of tests (after the test summary is printed) in Travis, although the same test works for me on my laptop. This clause is also the commented out clause added in PR #1261 

```  monitoring=MonitoringHub(
￼                    hub_address="localhost",
                    hub_port=55055,
                    logging_level=logging.INFO,
                    resource_monitoring_interval=1,
                )
```

My guess is a race condition in monitoring shutdown.

annawoodard:
One requirement for considering this issue fixed is that the original monitoring clause in PR #1261 is re-introduced and CI passes.

benclifford:
@TomGlanzman has seen a monitoring related hang at shutdown, with PR #1367. Here's the stack trace after pressing ctrl-C:

```
^CError in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_19-dev/python/miniconda3-4.5.12/lib/python3.7/multiprocessing/popen_fork.py", line 28, in poll
Process Monitoring-DBM-Process:
    pid, sts = os.waitpid(self.pid, flag)
KeyboardInterrupt
Traceback (most recent call last):
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_19-dev/python/miniconda3-4.5.12/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_19-dev/python/miniconda3-4.5.12/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/global/homes/d/descdm/.local/lib/python3.7/site-packages/parsl/monitoring/db_manager.py", line 451, in dbm_starter
    dbm.start(priority_msgs, node_msgs, resource_msgs)
  File "/global/homes/d/descdm/.local/lib/python3.7/site-packages/parsl/monitoring/db_manager.py", line 327, in start
    threshold=self.batching_threshold)
  File "/global/homes/d/descdm/.local/lib/python3.7/site-packages/parsl/monitoring/db_manager.py", line 395, in _get_messages_in_batch
    x = msg_queue.get(timeout=0.1)
[...]

Issue #282, opened 2018-05-17
Consider renaming `lazy_errors` and re-think behavior


annawoodard:
In implementing the class-based configuration, I am writing docstring entries for things that we currently just have a longer description in the docs for. One of these is `lazyErrors`. In the [docs](http://parsl.readthedocs.io/en/latest/userguide/exceptions.html?highlight=lazyErrors#lazy-fail), we have this description titled `Lazy Fail` (even though the option is called 'lazyErrors'):
> While Retries address resiliency at the level of Apps, lazy failure adds resiliency at the workflow level. When lazy failures are enabled, the workflow does not halt as soon as it encounters a failure, but continues execution of every app that is unaffected. Lazy failures is the default behavior in parsl, with the expectation that when running production workflows, individual app failures can be deferred until the end of the workflow. During the development and testing of workflows, failing immediately on any failure is often preferred and this behavior is possible by setting lazyErrors=False.

I do not favor the name `lazy_fail` or `lazy_errors`. I'm not sure the lazy part is a perfect description of what we're doing, and I think it relies on background knowledge unnecessarily. I propose `halt_on_error` or `halt_on_failure`. In that way, the name would describe what is being done without assuming prior knowledge or a mental mapping between concepts. The docstring entry could be (I'm open to help tweaking this):
> Halt workflow processing and exit immediately if any apps fails.

We've been discussing batching all of the user-facing changes together, which is why I wanted to address this now.

benclifford:
More about the description in general rather than specifically the name:

i) better describe the trade-off between the two modes: non-lazy mode is useful in production *if* you can make use of the other stuff that is being run - for example, if you have sufficient checkpointing in place to usefully resume in a way that uses that later stuff, or because you have some other thing in your app which makes that useful. It's not clear to me at the moment that that is always (or even often) the case (even though we'd like it to be).

ii) The pasted text talks about "exit immediately" - is that really what happens?  or does a Python exception get thrown to the user's workflow code? and if so, where/by whom?

iii) I think my preference for name would be something along the lines of "continue as much as possible" - describing the unusual behaviour that parsl will provide *despite* failure.

annawoodard:
[...]

PR #1275, opened 2019-09-12
Catch errors during launch and set the task future exception appropriately

hategan:
This addresses one of the issues that causes #1274 and possibly #1267.

The gist of it is that, with monitoring enabled, launch_if_ready calls _create_task_log_info which accesses an nonexistent property on AppFutures. That causes an exception to be thrown. Now, launch_if_ready is called from submit(), in particular this snippet:

            def callback_adapter(dep_fut):
                self.launch_if_ready(task_id)

            try:
                d.add_done_callback(callback_adapter)
            except Exception as e:
                logger.error("add_done_callback got an exception {} which will be ignored".format(e))

The issue is that Future.add_done_callback, if the future is closed, directly invokes the callback. Long story short, such errors are explicitly ignored. Perhaps the right solution is to assume that other code paths handle this properly, but, to be safe, I wrapped launch_if_ready in a try/catch which sets an exception on the AppFuture that represents the task currently launched.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -354,6 +354,18 @@ class DataFlowKernel(object):
         launch_if_ready is thread safe, so may be called from any thread
         or callback.
         """
+        try:
+            self._launch_if_ready(task_id)
+        except Exception as ex:
+            # catch errors properly, otherwise it's possible for code to
+            # throw an exception before the task is launched, which causes
+            # things to just hang.
+            #
+            # may want to wrap this exception into something that clearly
+            # indicates that this is an internal error
+            self.tasks[task_id]['app_fu'].set_exception(ex)
+
+    def _launch_if_ready(self, task_id):
         if self._count_deps(self.tasks[task_id]['depends']) == 0:
 
             # We can now launch *task*


benclifford:
i) A bunch of stuff needs to happen at task completion (even failed task completion). A more consistent way to do this would be in the same way as memoization or dependency failure does it: use a Future from somewhere other than `submit` to be the exec future with the appropriate value/exception, and say "this task ran and here is its result".

In the case of memoization, that's a previously used executor future coming via `check_memo` (line 423 of dflow.py); in the case of dep fail handling - see unmerged PR #1258 around line 385.

I think the dep fail handling would roughly the right thing to do here, but using `States.failed` instead of `States.dep_fail`

ii) There is also a potential race condition/fragility here that I'm not entirely sure of: once the task has been launched with `executor.submit` inside `self.launch_task`, but then after that an exception is raised, I'm unclear if it is safe to catch that exception and be able to recover safely: we'd potentially end up potentially with two concurrent things (the real one, and the error handling fake-one) both claiming to be the definitive provider of the tasks result.

For exceptions happening before the submit, I think this is straightforward to deal with: catch and create a failure `Future` deeper in the launch code rather than around the outside of 
launch_if_ready.

For exceptions happening after the submit - I'm not sure what the right thing to do is... catch-and-log at the least.

hategan:
Let me see if I get this straight. We're implementing
seq(
  etc1(), exec_fu = seq(etc2(), future(exec())), etc3(), wait(exec_fu)
)

- errors are maybe handled asynchronously in exec(), but it's not clear what happens with errors in etc2().
- there is no error propagation defined for the etcs() in general (i.e. it gets ignored).
- you are suggesting that a failure in etc1() or etc2() should fake an exec() and create a failed exec_fu, which must go through a particular path in the call stack. Let me rephrase that: if etc1() or etc2() fail, I need to still call exec() (or repeat some of its steps) and put some fake exec_fu where the real one was supposed to be.
- etc3() races with the wait(), so the proper way of handling etc3() failure depends on when that failure happens.

I mean we're obviously not implementing the above kseudocode, but do we want to? Because if we do, the first error in the flattened sequence (etc1(), etc2(), etc3(), exec()) propagates to the app_fu and there is no race because the wait is after etc3().

benclifford:
I don't understand how any of that pseudocode is really meant to map to the existing parsl code so my answers are going to be a bit vague.

> you are suggesting that a failure in etc1() or etc2() should fake an exec() and create a failed exec_fu, which must go through a particular path in the call stack. Let me rephrase that: if etc1() or etc2() fail, I need to still call exec() (or repeat some of its steps) and put some fake exec_fu where the real one was supposed to be.

Approximately.

> etc3() races with the wait(), so the proper way of handling etc3() failure depends on when that failure happens.

There is no wait in the submission code path for exec_fu - there's a callback added for handle_exec_update though, which handles post-completion (for both success and failure completion).

I'm unclear why you argue that:

etc1(), exec_fu = seq(etc2(), future(exec())), etc3(), wait(exec_fu)

can flatten to:


So I'm unclear what you're expecting here:

benclifford:
A related thing to this, maybe, is that there is no reasonable "everything-is-broken-time-to-abandon-everything" behaviour: this is a library sitting inside other code, not a workflow system where we can exit the top level unix process. Something that I think is related arises with (non-)lazy failures - issues #282 has some discussion there.

hategan:
> I don't understand how any of that pseudocode is really meant to map to the existing parsl code so my answers are going to be a bit vague.

This is going to be a mess without a blackboard, but let's try.

> 
> > you are suggesting that a failure in etc1() or etc2() should fake an exec() and create a failed exec_fu, which must go through a particular path in the call stack. Let me rephrase that: if etc1() or etc2() fail, I need to still call exec() (or repeat some of its steps) and put some fake exec_fu where the real one was supposed to be.
> 
> Approximately.

Maybe I misunderstand things. The app_fu is done when a bunch of constituent fus and/or non-fus are done. Presumably any errors in the constituents should cause the app_fu to fail. I see in the code that there is handling on the failure of the exec_fu, which triggers retries and whatnot. Let's call those things, your stuff that needs to happen, exec_cb. What you suggest says that we, in fact, want exec_cb invoked not just when exec_fu fails, but also when other things fail. So it seems to me that we are not clear on what's handled when and where. If exec_cb needs to be called when other app_fu constituents fail, then it should really be an app_cb. Or at least we should be clear and list, perhaps in a table, what an app_fu is made of, and at what level failures on each of those components should be handled.

> 
> > etc3() races with the wait(), so the proper way of handling etc3() failure depends on when that failure happens.
> 
> There is no wait in the submission code path for exec_fu - there's a callback added for handle_exec_update though, which handles post-completion (for both success and failure completion).

pseudocode. It means that the app_fu is only done when the exec_fu is done.

> 
> I'm unclear why you argue that:
> 
> `etc1(), exec_fu = seq(etc2(), future(exec())), etc3(), wait(exec_fu)`
> 
> can flatten to:
> 
> `etc1(), etc2(), etc3(), exec()`
> 
> These are all side-effectful things, and exec() could plausibly run entirely before etc3 even starts, which would be a different behaviour to `etc1(), etc2(), etc3(), exec()`.

I should have said "for reasonable observable purposes". Let's try with fork/join:

exec_fu = fork(exec()), etc3(), join(exec_fu)

It doesn't matter how things run if exec() and etc3() don't change shared state.

So the point is to add the callback handler to the exec_fu after etc3() completes, i.e., as the very last step in the whole launch process. That guarantees that the exec_cb cannot be called nondeterministically, not that exec() completes deterministically. Adding a callback is equivalent to wait() or join() on the future, which, for the purpose of our discussion, do not mean anything at the Python thread level but describe the behavior at an abstract level.

> 
> If etc3() is code that has to happen after execution submission has happened - for example code that logs in monitoring that the task has been submitted - that's wrong to be doing before the task been submitted.

In the pseudocode, future(x) launches x, so that happens before etc3(). x runs in parallel with etc3().

> 
> I'm unclear if that's an artefact of the not-Python language you are using for pseudocode - I would be more comfortable discussing technical details of execution ordering and error propagation using Python-style pseudocode based more around what the actual DFK code does.

Using a language whose semantics are fundamentally ill-equipped to deal with concurrency leads to a lot of unsaid things about what happens when. But let's try:

def submit(app_fu, task):
  try:
    etc1() # there is no point in retrying this since exceptions here are not related to how some executable runs on some remote resource
    exec_fu = launch(task)
    exec_fu.add_callback(...)
  except:
    app_fu.fail()
    (optionally)
    if exec_fu: exec_fu.cancel()

def launch(task):
  etc2() # same as etc1()
  exec_fu = executor.submit(task) # this call guarantees that either an exception is thrown or that task is launched and an exec_fu is returned. All exceptions related to task after this method returns are set on exec_fu
  etc3() # unless etc3() does nothing meaningful, which I don't think can be guaranteed easily, failures of etc3() should be consequential.
  return exec_fu

hategan:
> A related thing to this, maybe, is that there is no reasonable "everything-is-broken-time-to-abandon-everything" behaviour: this is a library sitting inside other code, not a workflow system where we can exit the top level unix process. Something that I think is related arises with (non-)lazy failures - issues #282 has some discussion there.

I'm assuming we're discussing this in the context of how errors in doing all the things that an app is made of are handled. So I'm not assuming some workflow context, but everything is in the context of that app_fu.

benclifford:
> So the point is to add the callback handler to the exec_fu after etc3() completes, i.e., as the very last step in the whole launch process. That guarantees that the exec_cb cannot be called nondeterministically,

It's not just trying to stop the exec_fu callback from happening non-deterministically that I care about: it is that once you've launched the task execution via `executor.submit()` (the potentially large piece of work that is what the user wants done), then the world is pretty much committed to that happening (or happening and failing itself) - the task can't be unlaunched in the case of a later exception.

After that's happened, it's not clear to me that it is reasonable to say "well that core computation is going to progress, but you don't get the result because we hit an exception in some logging code". It certainly is different than the current behaviour but I don't see that it is better.

Once the commitment has been made that the core task code is going to run inside an executor, (roughly `executor.submit` returning) then error handling for anything after that has to take into account that that commitment has been made.

benclifford:
> Incidentally, I'm assuming that there is some head scratching on how to handle global context, like shutting down various central components or what exactly constitutes a "run" that at the moment is handled in an ad-hoc way, like shutting things down when the last known app completes or when the Python VM shuts down, there being cases when only one of these strategies is ideal, but not necessarily the same.

Broadly the state is: a "run" is a DataFlowKernel instance. You can shut it down explicitly, (using `dfk.cleanup` or it will get shut down at process termination.

You can many DFKs in a python process at once, although there sometimes a primary one which app invocations will default to using - this happens less in script style user code, but more in notebook style code.

It's legal to have a DFK instance with no work to do - more work can appear later after some other non-parsl computation makes futher parsl app invocations - and that is a common way to do "dependencies", rather than using parsl's dependency handling.

hategan:
[...]
> 
> After that's happened, it's not clear to me that it is reasonable to say "well that core computation is going to progress, but you don't get the result because we hit an exception in some logging code". It certainly is different than the current behaviour but I don't see that it is better.
> 
> Once the commitment has been made that the core task code is going to run inside an executor, (roughly `executor.submit` returning) then error handling for anything after that has to take into account that that commitment has been made.

If it can be guaranteed that the "after" code is only stuff that we don't care about, sure (though we should be careful not to project what we don't care about as something that shouldn't be cared about). But you can't make this guarantee in general. It could be critical code. So I would say it's better to lose some CPU time than have unpredictable behavior. Fail for all uncaught errors in "after" code and, if there's code that we can explicitly identify as non-critical, for some definition of "non-critical", then wrap that portion into a handler that ignores/logs the error.

benclifford:
> Maybe I misunderstand things. The app_fu is done when a bunch of constituent fus and/or non-fus are done. Presumably any errors in the constituents should cause the app_fu to fail. I see in the code that there is handling on the failure of the exec_fu, which triggers retries and whatnot. Let's call those things, your stuff that needs to happen, exec_cb. What you suggest says that we, in fact, want exec_cb invoked not just when exec_fu fails, but also when other things fail. 

The basic app execution is:
call executor.submit and when the returned executor future completes, whatever comes out of it is used to set the value of the AppFuture.

(Some of) the future related things around that are:

- the memoization table already has a result - in which case instead of getting the executor future from `executor.submit`, an executor future already exists in that memoization table, happens to be completed already, and is used as the executor future for this invocation.

- retries are enabled, in which case an exception in a completed executor future does not turn into a exception in the AppFuture, but instead a new executor future is made by calling `executor.submit` and replaces the old executor future.

- dependencies on earlier apps are handled: if a parameter to the current app invocation is a Future, wait for it to complete before doing the `executor.submit`. So at that point, there is an AppFuture with no executor future to provide the result.
It might be that a dependency completes with an exception rather than a regular result; in that case we cannot proceed and the current app has to fail. To do that, instead of acquiring the executor future with `executor.submit`, the executor future  is a "constant" future that describes the dependency failure. Otherwise, the executor future will come from the usual `executor.submit` invocation.

- File stage-in can make new app invocations and make the current app depend on those completing. As far as future handling is concern, that should be handled by the above dependency mechanism rather than something separate: a file stage-in error is a dependency error that propagates to this app like any other dependency error.

- File stageout can happen as new app invocations which depend on the current exec_fu. The AppFuture does not wait for stage out to complete before it gets a value (for historical implementation reasons, rather than any particularly deliberate choice - see #1279). So as far as the current code is concerned, this isn't a thing that needs to have happened for an AppFuture to complete.

Other sources of breakage are, I think, entirely "parsl internal code" breakages rather than externally caused breakages - the most common at the moment being monitoring code breaking.

benclifford:
Another thing to do with especially monitoring there is that monitoring calls throwing exceptions probably shouldn't be regarded as application failures: they're attempts to describe what is happening, but if they didn't work, that's a failure of monitoring, not a failure of app execution.

hategan:
> Another thing to do with especially monitoring there is that monitoring calls throwing exceptions probably shouldn't be regarded as application failures: they're attempts to describe what is happening, but if they didn't work, that's a failure of monitoring, not a failure of app execution.

We're going a bit backwards here. As I said before, there is no guarantee that errors in etcs() are non-critical. So in deciding how to handle unexpected things, we cannot make the assumption that the right thing to do is ignore errors. It so happens that this discussion was triggered by an error in monitoring, but that has little to do with this PR.

But to get back to your point, yes, they are not application failures. They are failures of the system in running the application with all the bells requested by the user. So when I say parsl.run_my_application_with_bells() and some bell fails, then run_my_application_with_bells() fails.

Perhaps we see things differently because we see monitoring differently: some addon that shouldn't be cared about vs. an integral part of the system. The fact that I jumped on board when monitoring was already there makes me see it more as the latter. But I'd argue that's also how new users would see it.

benclifford:
Monitoring is in no way an integral part of the system, despite its presence laced through out the DFK.

:(

benclifford:
What I think this PR *should* do:

Audit each part of the task launch code path and decide if it failing is the kind of failure which means the app cannot be launched.

If it is such a failure and it is before executor.submit, use the Future() based method described above to give an Exception result.

If it is not that kind of terminal failure, decide who (if anyone) cares about this kind of failure and send the failure that way instead, and continue with submission. Possibly that is only monitoring, and for that, logging it at log level ERROR is probably the right thing to do for now.

If it is that kind of terminal failure and it is *after* provider.submit, then this is probably an irrecoverable parsl internal failure and you are welcome to propose some suitable method of exploding the whole of parsl.

hategan:
> Monitoring is in no way an integral part of the system, despite its presence laced through out the DFK.
> 
> :(

Precisely. However, to me, and everybody using it, and to future users who see it documented and want it, it is an integral part of the system.

But again, we need to address the fact that errors, in integral or non-integral parts of the system, however subjectively defined those are,, are presently being ignored. This can lead to hangs and is not acceptable.

hategan:
> If it is not that kind of terminal failure, decide who (if anyone) cares about this kind of failure and send the failure that way instead, and continue with submission. Possibly that is only monitoring, and for that, logging it at log level ERROR is probably the right thing to do for now.

That is an impossible decision to make when handling things at the top level.

```python
def f():
  a()
  b()
  c()

def g():
  try:
    f()
  except:
    # cannot generally decide here where the error is coming from and whether the error should cause f() to continue to execute after the error was thrown; that is part of the inner workings of f().
```

What could happen is this:
```python
def f():
  try:
   a()
  except:
   # somebody decided that this is not relevant, ignore
  b()
  try:
   c()
  except:
   # idem

def g():
  try:
    f()
  except:
    # this still needs to be here and we still need to do something about it irrespective of what the internals of f() are. We cannot, at this point, assume that the error is benign, nor do we have the means to do much about it without some serious abstraction inversion/spaghetti control flow.
```

So no. This PR is meant to fix g(), not f(). f() needs to be dealt with separately. That said, this particular PR in its current form does not properly fix g() for reasons discussed above, so it needs to be redone.

What I suggest is the following:
```python
def launch_if_ready(task_id):
  assert self.tasks[task_id] is not None
  assert self.tasks[task_id]['app_fu'] is not None
  try:
    exec_fu = self._launch_if_ready(task_id)
    assert exec_fu is not None
    exec_fu.add_done_callback(partial(self.handle_exec_update, task_id))
  except Exception as ex:
    self._fail(task_id, ex)
   # alternatively, self._fail_with_retries(task_id, ex)

def _fail(task_id, ex):
  ...
  fake_exec_fu = Future()
  fake_exec_fu.retries_left = 0
  fake_exec_fu.set_exception(ex)
  self.tasks[task_id]['exec_fu'] = fake_exec_fu
  self.tasks[task_id]['app_fu'].parent_callback(fake_exec_fu)
```

Additionally, but separately, identify code that is broken in the call path of _launch_if_ready and fix it or remove it. Only ignore errors if code, under the assumption that it is stable, should ignore errors.

hategan:
> A broad catch like this makes sense, but we need to be careful about what we do with the caught exception. 

A catch like this would not be there to catch exceptions that we expect to happen during normal usage, but to catch exceptions that otherwise cause silent deadlocks and report them somewhere. So I would recommend not seeing this as a tool that has anything to do with users, except in extreme cases, but something that has to do with us catching a particularly nasty class of problems before the user gets to use the code.

> Setting the exception onto the app_future makes sense in fail early mode, but we've so far taken a lazy approach where we let the workflow proceeed as far as possible. Secondly, app_future exceptions in my mental model are for communicating the reason for the failure of user code. I worry about setting an exception when the result was already computed.

It can be difficult to intuitively think about what to do with errors that occur in one of two parallel sub-units of a parent unit. For concrete cases, it may seem like they should be handled differently based on who finishes first or maybe on how mature the code in each sub-unit is. I would urge you to resist that thinking. 

> 
> ```
> If a failure prevents the computation of a task:
>     report that via the the app_future.exception.
> 
> If a runtime failure potentially blocks the workflow:
>     report a critical error and trigger DFK shutdown.
> else: 
>     report to the log.
> ```
> 
> Do we know what exceptions can bubble up here? Monitoring exceptions for now should be in the non-critical category (report to log, and go on).

That's the problem. Anything can bubble up here and you won't know where it bubbled from, so you must assume the worst.

benclifford:
Closing this as it isn't going to be merged as is. I think work on #1420 and related PRs supercedes this.

hategan:
I don't think #1420 is related to this. This addresses an unrelated unimplemented logical path in Parsl. Ignore at your own peril.

But yeah, the initial version of the code in this PR should not be merged.

Issue #1278, opened 2019-09-12
Add worker plot to monitoring


annawoodard:
Our current CPU utilization plots reflect the CPU utilization of a given App. They don't tell the user, given that I want to run App A (which may be executing some inefficient, crappy software), am I doing it as efficiently as possible?

To mitigate this, we should add a plot showing how many workers are currently connected, and of those workers, how many are actually running tasks.

PR #1284, opened 2019-09-16
Add worker plot monitoring #1278

ZhuozhaoLi:
This PR adds a worker efficiency plot to the resource page of workflow.

The worker efficiency is computed by `total_running_tasks/total_workers`.

`total_running_tasks` is to count the total number of running tasks every one second (by checking if the `task_time_running` and `task_time_returned` include that timestamp).

`total_workers` is computed by adding up the `worker_count` in the `node` table (current assumption is that there is no scaling. If we want to have scaling, we need to add more info to the `node` table)

### DIFF
--- a/parsl/monitoring/visualization/plots/default/workflow_resource_plots.py
+++ b/parsl/monitoring/visualization/plots/default/workflow_resource_plots.py
@@ -129,7 +129,7 @@ def worker_efficiency(task, node):
         return plot(fig, show_link=False, output_type="div", include_plotlyjs=False)
     except Exception as e:
         print(e)
-        return "The worker efficiency plot cannot be generated due to missing data.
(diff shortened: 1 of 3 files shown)


ZhuozhaoLi:
Fixes #1278

yadudoc:
I'm seeing this error now : 
`The worker efficiency plot cannot be generated because of exception can't multiply sequence by non-int of type 'numpy.float64'.`

yadudoc:
Issue #1286 will be addressed separately where we'll handle the case where we have events besides worker registration logged for monitoring.

ZhuozhaoLi:
> I'm seeing this error now :
> `The worker efficiency plot cannot be generated because of exception can't multiply sequence by non-int of type 'numpy.float64'.`

Now it returns an exception saying that the plot cannot be produced due to missing data.

PR #1287, opened 2019-09-16
custom max_idletime in simple strategy

### LINKED ISSUE #1147: Configurable max_idletime on simple strategy
currently it is hard coded to 120s. It would be great to make this parameter configurable.

ZhuozhaoLi:
fixes #1147

### DIFF
--- a/parsl/config.py
+++ b/parsl/config.py
@@ -48,7 +48,7 @@ class Config(RepresentationMixin):
         Strategy to use for scaling resources according to workflow needs. Can be 'simple' or `None`. If `None`, dynamic
         scaling will be disabled. Default is 'simple'.
     max_idletime : float, optional
-        The maximum idle time for an executor in the 'simple' strategy. Default is 120.0.
+        The maximum idle time for an executor in the 'simple' strategy. Default is 120.0 seconds.
     usage_tracking : bool, optional
         Set this field to True to opt-in to Parsl's usage tracking system. Parsl only collects minimal, non personally-identifiable,
         information used for reporting to our funding agencies. Default is False.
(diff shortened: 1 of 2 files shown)


parsl/config.py
```diff
@@ -47,6 +47,8 @@ class Config(RepresentationMixin):
     strategy : str, optional
         Strategy to use for scaling resources according to workflow needs. Can be 'simple' or `None`. If `None`, dynamic
         scaling will be disabled. Default is 'simple'.
+    max_idletime : float, optional
+        The maximum idle time for an executor in the 'simple' strategy. Default is 120.0.
```
benclifford: specify units (seconds)

benclifford:
whoops, I accidentally closed this.

parsl/config.py
```diff
@@ -64,6 +66,7 @@ def __init__(self,
                  retries: int = 0,
                  run_dir: str = 'runinfo',
                  strategy: Optional[str] = 'simple',
+                 max_idletime: Optional[float] = 120.0,
```
benclifford: This type should be `float`, not `Optional[float]`.

`Optional[float]` means "this value may be `None` or `float`" - but `None` does not work with the simple strategy.

This is different from using the word "optional" to mean "there is a default float value so you do not need to explicitly specify one" - for example, look how the `usage_tracking` parameter is not `Optional[bool]`.

parsl/config.py
ZhuozhaoLi: added

parsl/config.py
ZhuozhaoLi: Good point. Fixed this.

PR #1290, opened 2019-09-18
Handle case where `stdout=None` when calling `qstat` in TorqueProvider._status

### LINKED ISSUE #1289: Handle case where `stdout=None` when calling `qstat` in TorqueProvider._status
When shell command `qstat <list of job ids>` fails due to timeout, the returned stdout, stderr will be None causing a `AttributeError: 'NoneType' ` Exception while trying to parse the stdout.

Solution: NoneType case should be handled by first checking whether `stdout is not None` and `stderr is not None`

```
2019-09-17 03:57:23 parsl.dataflow.strategy:204 [DEBUG]  Executor htex has 1 active tasks, 0/0/1 running/submitted/pending blocks, and 8 connected workers
Caught exception: Command 'qstat 8972700.wlm01 8973036.wlm01 8973047.wlm01 8973173.wlm01 8973640.wlm01 8973649.wlm01 8973674.wlm01 8973692.wlm01 8973750.wlm01 8973774.wlm01' timed out after 120 seconds
2019-09-17 03:59:23 parsl.channels.local.local:77 [WARNING]  Execution of command [qstat 8972700.wlm01 8973036.wlm01 8973047.wlm01 8973173.wlm01 8973640.wlm01 8973649.wlm01 8973674.wlm01 8973692.wlm01 8973750.wl
m01 8973774.wlm01] failed due to
 Command 'qstat 8972700.wlm01 8973036.wlm01 8973047.wlm01 8973173.wlm01 8973640.wlm01 8973649.wlm01 8973674.wlm01 8973692.wlm01 8973750.wlm01 8973774.wlm01' timed out after 120 seconds
Exception in thread Thread-2:
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/dataflow/flow_control.py", line 119, in _wake_up_timer
    self.make_callback(kind='timer')
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/dataflow/flow_control.py", line 139, in make_callback
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/dataflow/strategy.py", line 177, in _strategy_simple
    status = executor.status()
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/executors/high_throughput/executor.py", line 600, in status
    status = self.provider.status(self.blocks.values())
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/providers/cluster_provider.py", line 167, in status
    self._status()
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/providers/torque/torque.py", line 118, in _status
    for line in stdout.split('\n'): 
AttributeError: 'NoneType' object has no attribute 'split'
```

ravihansa3000:
Fixes #1289

- Check return code of shell cmd `qstat <job_ids>` before reading stdout
- In case of non-zero exit code, log an error message
- Check whether `stdout` and `stderr` is `None` before parsing the output


benclifford:
Does this end up marking every job as completed when qstat fails, because they are still all listed in the `missing_job` table and then marked as commpleted around line 137 just after your changes?

benclifford:
If so, I'm not sure what the right thing to do is - there isn't handling in the rest of parsl for `status` to fail, rather than return an answer.

Reporting a job as completed is probably going to cause a replacement to be launched by the scaling strategy.

So then maybe the old behaviour was: exceptions logged but scaling strategy still worked on the next round
and the new behaviour is: exceptions logged by this PRs code, but existing jobs are regarded as killed and new replacement jobs are launched by the scaling strategy even though the old jobs are really still running.

And if so, then maybe better behaviour is for this PR to throw an exception in the case of exit code `!= 0` in addition to the log message you have added, and to add some code in `_strategy_simple` around the call to `executor.status()` that catches and logs an exception there, something like "exception getting status - abandoning this strategy round"

That would also make error reporting for all providers look nicer - not just Torque.

@yadudoc @annawoodard might have comments?

ravihansa3000:
To clarify: old behavior was: uncaught exception raised, Parsl hanged

I'm trying to understand the flow here. So in case of `retcode != 0` existing jobs are marked as 'E' for completed and they will be killed and replaced. But how and why?

benclifford:
They'll be marked as completed. I don't think anything will try to kill them - parsl will regard them as completed so wont' attempt to manage them any more.

In your original stack trace in #1289, the status call got up to this place:

```
  File "/home/users/nus/csipav/anaconda3/envs/gdc/lib/python3.7/site-packages/parsl/dataflow/strategy.py", line 177, in _strategy_simple
    status = executor.status()
```

If that status call actually works, then the code following that call in `_strategy_simple` will do some calculations and decide if it wants to `executor.scale_in()` or `executor.scale_out()` or do nothing. Those calls will results in new blocks being launched or cancelled.

If provider.status says there are no blocks (= batch system jobs), then this bit of code (a page or so later) will call `executor.scale_out()` to get some blocks. If provider.status says there are no blocks, even when there really are blocks running in the batch system, then `_strategy_simple` will still try to get more blocks, in addition to the ones it doesn't know about any more.

The end effect of that is maybe something like: "if qstat is often slow, parsl keeps submitting more and more batch jobs".

benclifford:
(this is all based on me reading code, not trying out your PR -- you could experiment by setting `job_status_cmd = "qstat {0}".format(job_id_list)` to `"exit 1"` to get a deliberate fail on status, for example)

annawoodard:
I agree with Ben's recommendation to throw an exception and catch it in `strategy`. We've seen weird behavior with runaway job submission and I think that might kill two birds with one stone, which would be great!

ravihansa3000:
I'm afraid it's not simple as checking the return code. When parsl attempts to check statuses of jobs which were finished, it will return with exit code 35. See output below.

```
2019-09-18 17:12:30.070 parsl.providers.torque.torque:131 [ERROR]  Command 'qstat 8980829.wlm01 8981859.wlm01 8981860.wlm01 8981862.wlm01' failed with return code 35
stderr:qstat: 8980829.wlm01 Job has finished, use -x or -H to obtain historical job information
stdout8981859.wlm01     parsl.156879789  csipav                   0 Q medium          
8981860.wlm01     parsl.156879791  csipav                   0 Q medium          
8981862.wlm01     parsl.156879793  csipav                   0 Q medium
```

Is it safe to just check `if stdout is not None:` and parse the output? Also if `stdout is None` then we log an error msg then raise an Exception which will be handled by `strategy`. 

Is it ok to raise a generic `Exception` or should I refer to a guideline?

benclifford:
> Is it safe to just check if stdout is not None: and parse the output? Also if stdout is None then we log an error msg then raise an Exception which will be handled by strategy.

I think this will work better than the behaviour in `master` now.

We don't seem to have very strict rules on which Exception type to raise - I would probably raise a `RuntimeError`.

benclifford:
@ravihansa3000 actually - what are the conditions that `None` is returned here? Is it only in an error case, or does it also happen in a valid case - for example, if all jobs have finished?

ravihansa3000:
This is due to the version I was using (v0.8.0)
https://github.com/Parsl/parsl/blob/0.8.0/parsl/channels/local/local.py#L81

stdout is `None` should not occur in master codebase. However, this could be an issue for other channels (eg. ssh). I've updated the PR with TimeoutExpired exception handler which is re-raised as a RuntimeError and handled in strategy. If you think that is unnecessary I'll close the PR.

annawoodard:
If jobs have finished, qstat returns this:
```
[nscc04] ~ >qstat 8994335.wlm01
qstat: 8994335.wlm01 Job has finished, use -x or -H to obtain historical job information
```

ravihansa3000:
I think my previous comment might be confusing. To answer @benclifford 's question: the only condition that `stdout` is returned as `None` is in case of a timeout. I forgot to cross-check the code with the Parsl version I was using earlier. But after searching through the v0.8.0 code-base (version which I was using) I noticed that `stdout` is explicitly set to `None` at [1]. This behavior has been changed in a later commit to re-rase the exception, thus `stdout` is `None` condition should not occur in master code-base in case of a timeout.
However, IMO a job status timeout should not cause a strategy round to fail. Either we can explicitly handle the `TimeoutExpired` exception, log an error msg and continue (as I've done in this PR) AND/OR retry the job status command until a certain no of retries.

Also, as I see it parsing the `qstat` command output could use some improvement. Since querying for completed jobs would return non-zero exit code, we can't rely on return code to guarantee the safe execution of parsing its stdout. I'm not sure whether `qstat -H` would be an option here. I'm happy to assist if you think this is worth pursuing.

[1] https://github.com/Parsl/parsl/blob/0.8.0/parsl/channels/local/local.py#L81

PR #1294, opened 2019-09-19
Add a worker watchdog thread

### LINKED ISSUE #1245: process_worker_pool can hang if a worker dies mid-task
I've encountered a situation where a process_worker worker process is getting killed during task execution/result serialisation.

That's pretty obscure; however process_worker_pool as a whole does not continue working properly - that task is never seen as failed by the main parsl process (until the whole pool dies for example by running out of batch job walltime)

annawoodard:
This commit makes two main changes:
1) A shared dictionary is added which the workers can access. They now
   put their in-progress tasks there.
2) A worker watchdog thread is added which monitors the worker
   processes. If it finds a dead worker, it checks for any in-progress
  tasks in the shared dictionary and sends it back to the DFK with a
  WorkerLost error, and restarts the worker.
Fixes #1245.

### DIFF
--- a/parsl/tests/test_error_handling/test_htex_worker_failure.py
+++ b/parsl/tests/test_error_handling/test_htex_worker_failure.py
@@ -4,6 +4,8 @@ import parsl
 from parsl.app.app import python_app
 from parsl.tests.configs.htex_local import fresh_config
 
+from parsl.executors.high_throughput.errors import WorkerLost
+
 
 def local_setup():
     config = fresh_config()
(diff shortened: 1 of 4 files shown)


parsl/tests/test_error_handling/test_htex_worker_failure.py
```diff
@@ -0,0 +1,33 @@
...
+import time
+
+import pytest
+
+import parsl
+from parsl.app.app import python_app
+from parsl.executors.high_throughput.process_worker_pool import WorkerLost
+
+def local_setup():
+    from parsl.tests.configs.htex_local import config
```
benclifford: Don't mutate global config in test - use fresh_config pattern so these changes don't affect other tests

annawoodard: Fixed

annawoodard:
Note this is a requirement for running coffea at Notre Dame, where a certain class of tasks occasionally tasks get stuck in a bad way. For these tasks, the first working solution I've found is to set a timeout which kills the worker. cc @lgray @klannon

PR #1295, opened 2019-09-19
Fix monitoring logging level

### LINKED ISSUE #1085: Make logging in monitoring wrapper parameterisable
The logging is set to DEBUG level in current state. We should make it parameterisable in the future.

ZhuozhaoLi:
This PR enables users to turn on/off monitoring debug logging.

Fix #1085

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -135,7 +135,7 @@ class MonitoringHub(RepresentationMixin):
                  workflow_version=None,
                  logging_endpoint='sqlite:///monitoring.db',
                  logdir=None,
-                 logging_level=logging.INFO,
+                 monitoring_debug=False,
                  resource_monitoring_enabled=True,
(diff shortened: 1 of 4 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -305,7 +317,7 @@ def __init__(self,
                  monitoring_hub_address="127.0.0.1",
                  logdir=".",
                  run_id=None,
-                 logging_level=logging.DEBUG,
+                 logging_level=logging.INFO,
```
yadudoc: Shouldn't this be `logging_level=logging_level`

ZhuozhaoLi: I think this should be `logging.INFO` -- it is the default logging option for `Hub` class and it is set on line 232.

yadudoc: Oh, sorry. I misread this.

PR #1297, opened 2019-09-20
Add ad-hoc provider #941

### LINKED ISSUE #941: Support for Ad-hoc clusters
Ad-hoc clusters are a bunch of machines that can be reached via ssh but otherwise do not have a batch scheduler or container orchestrator. Our approach in general for these situations is to create an executor for each node, with the workers launched via an SSH channel. Creating separate executors means that load-balancing across nodes will not work. This sort of situation is common in clouds and private clusters and the driving use-case is from DESC (@jchiang87).

We could prototype an ad-hoc provider that takes a list channels pointing to each of the available nodes like this : 
```python

node_addresses = ['node1', 'node2', 'node3', ...]
config = Config(
    executors=[
        HighThroughputExecutor(
            label="Ad-Hoc",
            worker_debug=True,
            cores_per_worker=1,
            provider=AdHocProvider(
                channels=[SSHChannel(hostname=n) for n in node_addresses],
            ),
        )
    ],
    strategy=None,
)

```

yadudoc:
Fixes #941


parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,234 @@
...
+    """ Ad-hoc execution provider
+
+    This provider is used to provision execution resources over one or more ad hoc nodes
+    that are accessible over a Channel (say, ssh) but otherwise lack a cluster scheduler.
+
+    Each submit call invoked will go through the list channels in a round-robin fashion
+
+    Parameters
+    ----------
+
```
ZhuozhaoLi: We need docstrings for each parameter?

parsl/dataflow/dflow.py
```diff
@@ -789,15 +788,21 @@ def add_executors(self, executors):
...
-                        if not executor.provider.channel.isdir(self.run_dir):
-                            parent, child = pathlib.Path(self.run_dir).parts[-2:]
-                            remote_run_dir = os.path.join(parent, child)
-                            executor.provider.channel.script_dir = os.path.join(remote_run_dir, 'remote_submit_scripts')
-                            executor.provider.script_dir = os.path.join(self.run_dir, 'local_submit_scripts')
-                    executor.provider.channel.makedirs(executor.provider.channel.script_dir, exist_ok=True)
                     os.makedirs(executor.provider.script_dir, exist_ok=True)
+
+                    if hasattr(executor.provider, 'channels'):
+                        logger.debug("Creating script_dir across ad-hoc cluster")
```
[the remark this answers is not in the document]
yadudoc: I think this functionality is a bit broken here, let me fix that.

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,234 @@
...
+
+def _roundrobin(items):
+    """ Returns items one at a time in an infinite loop
+    """
+    while True:
+        try:
+            for item in items:
+                yield item
+        except Exception:
```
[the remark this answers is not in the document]
yadudoc: That's from an earlier attempt. Fixed.

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,234 @@
...
+      be invoked.
+
+    """
+
+    def __init__(self,
+                 channels=[],
+                 worker_init='',
+                 cmd_timeout=30,
+                 parallelism=1,
+                 init_blocks=0,
```
ZhuozhaoLi: `init_blocks` should default to 1?

yadudoc: In this case it should be set by the user to # of channels.

benclifford: Should *always* ?

If so, then maybe the three `*_blocks` parameters shouldn't be parameters at all and they should all be set to the length of the channels attribute.

ZhuozhaoLi: A bit confused here. Maybe they should not be available to configure?

And it seems there should not be any scaling for AdHocProvider

ZhuozhaoLi: Or it should scale in terms of channels/nodes.

yadudoc: I'll have this be set internally to `init_blocks = min_blocks = max_blocks = #of channels`

parsl/dataflow/dflow.py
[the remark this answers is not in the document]
yadudoc: @benclifford Can you take a look at this bit again ?

benclifford:
Formatting string literals, PEP 498, are Python >=3.6 so not permitted for parsl use

```
E     File "/home/travis/build/Parsl/parsl/parsl/providers/ad_hoc/ad_hoc.py", line 216
E       logger.debug(f"Cancelling jobs: {job_ids}")
E                                                ^
E   SyntaxError: invalid syntax
```

benclifford:
I was able to run `pytest` against two local containers running ssh servers on my laptop.

parsl/dataflow/dflow.py
```diff
@@ -796,15 +795,26 @@ def add_executors(self, executors):
...
+                        counter = 0
```
benclifford: what is `counter` for?

parsl/dataflow/dflow.py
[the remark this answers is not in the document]
benclifford: The logic for multiple channels and one channels looks (unnecessarily?) different: for example, a single channel provider can omit a script dir but a multi-channel provider cannot omit a script dir. Maybe pull the original block into a helper function so that it can either be invoked once on `channel` or invoked multiple times on each element of `channels`?

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,235 @@
...
+from parsl.channels import LocalChannel
+from parsl.launchers import SimpleLauncher
+from parsl.providers.provider_base import ExecutionProvider
+from parsl.providers.error import SchedulerMissingArgs, ScriptPathError
+from parsl.utils import RepresentationMixin
+
+logger = logging.getLogger(__name__)
+
+
+translate_table = {
```
benclifford: I think this dict is unused?

parsl/dataflow/dflow.py
```diff
@@ -10,7 +10,6 @@
 import threading
 import sys
 import datetime
-
```
benclifford: unnecessary whitespace change

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,235 @@
...
+        Raises:
+              SchedulerMissingArgs : If template is missing args
+              ScriptPathError : Unable to write submit script out
+        '''
+
+        try:
+            with open(script_filename, 'w') as f:
+                f.write(script_string)
+
+        except KeyError as e:
```
benclifford: when does this `KeyError` get raised?

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,235 @@
...
+
+        return None
+
+    def submit(self, command, tasks_per_node, job_name="parsl.auto"):
+        ''' Submits the command onto a channel from a round-robin arrangement of channels
+
+        Submit returns an ID that corresponds to the task that was just submitted.
+
+        Args:
+             - command  :(String) Commandline invocation to be made on the remote side.
```
benclifford: omgz the spaces

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,235 @@
...
+            if retcode == 0:
+                rets.append(True)
+            else:
+                rets.append(False)
+            self.resources[job_id]['status'] == 'COMPLETED'
+        return rets
+
+    @property
+    def scaling_enabled(self):
+        return True
```
benclifford: does this being true or false change anything with init_blocks=max_blocks=min_blocks? My initial reaction was to say "this should be false" but maybe it doesn't change anything?

parsl/tests/manual_tests/test_ad_hoc_htex.py
```diff
@@ -0,0 +1,49 @@
+import parsl
```
benclifford: It would be better to put this config in `parsl/tests/configs/` and check that `pytest` can run against it. (I have verified that pytest runs against my own adhoc cluster config... so this should be fine).

Even better would be to parameterize all of `remotes`, `worker_logdir_root`, `address`, `worker_init` `username` `script_dir` into user_opts.py - `address` already has a `user_opts` key of `public_ip`.

benclifford: I tidied up my config into this branch https://github.com/Parsl/parsl/tree/benc-adhoc-test-conf which you could merge if useful.

parsl/dataflow/dflow.py
```diff
@@ -796,15 +795,26 @@ def add_executors(self, executors):
...
-                        if not executor.provider.channel.isdir(self.run_dir):
-                            parent, child = pathlib.Path(self.run_dir).parts[-2:]
-                            remote_run_dir = os.path.join(parent, child)
-                            executor.provider.channel.script_dir = os.path.join(remote_run_dir, 'remote_submit_scripts')
-                            executor.provider.script_dir = os.path.join(self.run_dir, 'local_submit_scripts')
-                    executor.provider.channel.makedirs(executor.provider.channel.script_dir, exist_ok=True)
                     os.makedirs(executor.provider.script_dir, exist_ok=True)
+
+                    if hasattr(executor.provider, 'channels'):
+                        logger.debug("Creating script_dir across ad-hoc cluster")
```
benclifford: This log shouldn't refer to ad-hoc clusters unless this code lives in the ad-hoc provider - maybe more like "creating script_dir across multiple channels" and "creating script_dir on one channel" in the else clause

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: Removed.

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: Looks like we don't do any command string formatting here anymore.

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: if `scaling_enabled==True` then parsl will do scaling actions, otherwise it is assumed that the user is doing the scaling actions separately. So min/max/init blocks controls how it can scale, and this piece enables the scaling actions.

parsl/dataflow/dflow.py
yadudoc: But, it was unnecessary to begin with.

benclifford: it's not to do with ad hoc cluster provision.

parsl/dataflow/dflow.py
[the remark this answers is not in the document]
yadudoc: Done.

parsl/tests/manual_tests/test_ad_hoc_htex.py
yadudoc: Just merged it.

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: I'll wait for #1310, update ad-hoc, and then re-review.

benclifford: this is gone in #1310

benclifford: ... but it needs deleting from this provider as it is now dead code

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,267 @@
...
+from parsl.channels import LocalChannel
+from parsl.launchers import SimpleLauncher
+from parsl.providers.provider_base import ExecutionProvider
+from parsl.providers.error import ScriptPathError
+from parsl.utils import RepresentationMixin
+
+logger = logging.getLogger(__name__)
+
+
+def _roundrobin(items):
```
benclifford: this function can be removed now

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Returns
+        -------
+        list of confirmation bools: [True, False...]
+        """
+        logger.debug("Cancelling jobs: {}".format(job_ids))
+        rets = []
+        for job_id in job_ids:
+            channel = self.resources[job_id]['channel']
+            retcode, stdout, stderr = channel.execute_wait("kill -TERM -{}".format(
```
benclifford: I think this has an extra `-`. I don't see the top process being killed, but with this change I do.
```
-            retcode, stdout, stderr = channel.execute_wait("kill -TERM -{}".format(
+            retcode, stdout, stderr = channel.execute_wait("kill -TERM {}".format(
```

See commentary in other comment about how this doesn't actually kill the block, though.

benclifford:
I tried the latest version of this on my two-worker-container setup.

```
pytest parsl --config parsl/tests/configs/htex_ad_hoc_cluster.py 
```
passes, and I see process worker pools in both worker containers.

I made another script which launches 1000 sleep jobs and then exits immediately (rather than waiting for completion). Although I see this in the log line:

> 2019-09-28 09:21:32 parsl.providers.ad_hoc.ad_hoc:248 [DEBUG]  Cancelling jobs: ['7412', '4845']

my two containers still have work pools on them (pids 7412 and 4845). They eventually terminate (after a couple of minutes, as the timeout parameters would lead me to expect). As part of their exiting, I see a 30 second delay in the worker watchdog thread:

```
2019-09-28 09:27:36.547 parsl:304 Rank:0 [CRITICAL]  [RESULT_PUSH_THREAD] Exiting
2019-09-28 09:28:05.583 parsl:345 Rank:0 [CRITICAL]  [WORKER_WATCHDOG_THREAD] Exiting
```

which might be related to #1307 somehow.

I've also noticed a potential bug in the AdHoc provider's kill command (with a `-`) which I have added as a comment on the relevant line.
However after fixing that, I see the top level shell script that launched the block being killed (for example `/home/benc/parsl/src/parsl/runinfo/036/submit_scripts/parsl.auto.1569663348.398287.sh
`) which is the process that shares the job ID with ad_hoc provider block IDs.
However this doesn't kill the process worker pool itself.

It looks like the `-` will killl there is attempting to kill a process group: however, the block ID is not a process group ID in my environment: for example, I'll get a block ID of `7674`, but the process group ID shared by both the launching shell script (pid `7674`) and the process worker pool processes is `7673`.

parsl/providers/ad_hoc/ad_hoc.py
benclifford: actually this is process group related which is more complicated - see more in https://github.com/Parsl/parsl/pull/1297#issuecomment-536170987

benclifford:
This behaviour (which is based on the (I think broken) local provider behaviour) seems to kill things for me: It examines the block id process to find the pgid of that, sets the output format with '-o pgid=' to mean "only output pgid, and no header"), trims non-numerics (whitespace in this case) with grep and then uses *that* as the process group ID to kill.

-            retcode, stdout, stderr = channel.execute_wait("kill -TERM -{}".format(
-                self.resources[job_id]['job_id']))
+            cmd = "kill -TERM -$(ps -o pgid= {} | grep -o '[0-9]*')".format(self.resources[job_id]['job_id'])
+            retcode, stdout, stderr = channel.execute_wait(cmd)

This process group kill approach still makes the assumption that the channel launched the top level command into a new process group. Maybe that assumption should be written into the Channel docstring.

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,267 @@
...
+        -------
+        list of status strings ['PENDING', 'COMPLETED', 'FAILED']
+        """
+        for job_id in job_ids:
+            channel = self.resources[job_id]['channel']
+            status_command = "ps --pid {} | grep {}".format(self.resources[job_id]['job_id'],
+                                                            self.resources[job_id]['cmd'].split()[0])
+            retcode, stdout, stderr = channel.execute_wait(status_command)
+            if retcode != 0 and self.resources[job_id]['status'] == 'RUNNING':
+                self.resources[job_id]['status'] == 'FAILED'
```
benclifford: This should be `=` not `==`. With `==`, we never notice that a block has died, and scaling up to fill in broken blocks doesn't happen.

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: I've put in your kill cmd string instead of this.

parsl/providers/ad_hoc/ad_hoc.py
yadudoc: There were two `=` vs `==` bugs. Fixed.

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,259 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AdHocProvider(ExecutionProvider, RepresentationMixin):
+    """ Ad-hoc execution provider
+
+    This provider is used to provision execution resources over one or more ad hoc nodes
+    that are each accessible over a Channel (say, ssh) but otherwise lack a cluster scheduler.
+
+    Each submit call invoked will go through the list of channels in a round-robin fashion
```
ZhuozhaoLi: Maybe just a quick fix: it is no longer round-robin fashion here.

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -0,0 +1,259 @@
...
+            logger.debug("Channel_counts : {}".format(channel_counts))
+            if 0 not in channel_counts.values():
+                yield None
+
+            for channel in channel_counts:
+                if channel_counts[channel] == 0:
+                    yield channel
+
+    def submit(self, command, tasks_per_node, job_name="parsl.auto"):
+        ''' Submits the command onto a channel from a round-robin arrangement of channels
```
ZhuozhaoLi: Not round-robin any more?

PR #1300, opened 2019-09-23
Python Dependency Packaging for WQ Executor

tjdasso:
This PR attempts to perform Python dependency analysis on Parsl apps to ensure correct execution at runtime. Using the Cooperative Computing Lab's python packaging tools, each Parsl app is analyzed for its import dependencies. For each unique function, an appropriate Conda environment is created, the necessary modules are installed into the virtual environment, and the environment is then packed into a tarball at the master. At the worker, the tarball is unpacked, paths are changed to activate the virtual environment, and the task is run within the environment, with all dependencies present. 

The current implementation includes the three CCL packaging scripts (`python_package_analyze`, `python_package_create`, and `python_package_run`). The first two scripts are called via subprocesses with the appropriate arguments to analyze the function and create the environment. The last script wraps the function call on the worker side, ensuring that the function is running within the environment. For repeated function calls, the same environment is sent to the worker, which prevents the need to create identical environments.

This PR utilizes several libraries that might not be available in the current Parsl set of software. Particularly, it uses the `stdlib_list` and `conda_pack` modules to analyze and create the packages, respectively. Of course, the PR also requires that Conda be installed on the master machine (the worker machine does not need Conda installed to activate the virtual environment). For more information about the CCL Python packaging tools, please see https://github.com/cooperative-computing-lab/cctools/tree/master/python_packaging for documentation and current/alternate implementations.


benclifford:
Are those dependencies all installable with usual packaging tools?

If so, then tell the CI (and everyone else) to install them like this:

In setup.py, extras_require, there is already a (broken) `workqueue` entry. Get rid of `work_queue` from that entry and put in the requirements which can be installed with usual packaging tools. (when/if work_queue is installable via pip/conda by name, `work_queue` can be added back in there too).

In `.travis.yml`, there is a line:

```
    - pip install .[extreme_scale,monitoring]
```

which installs dependencies for more interesting parsl features (on top of the original deliberately basic install near the start of .travis.yml).

Add in workqueue to that list, something like

```
    - pip install .[extreme_scale,monitoring,workqueue]
```

and CI will install the dependencies specified in the relevant extras_require entry.

Is there anything that can't be installed this way that is needed to run?

benclifford:
I think this PR makes it so that environment packing *always* happens, and there is no option to turn it off and use workqueue in the mode it has been used in so far. I guess I'm ok with that but just checking that's a deliberate decision...

tjdasso:
Yes, as of right now, the packaging _always_ occurs. However, like the `source` option, this can definitely be made into an option of the executor instead of being a default feature of the executor

setup.py
```diff
@@ -26,7 +26,7 @@
     'google_cloud' : ['google-auth', 'google-api-python-client'],
     'gssapi' : ['python-gssapi'],
     'azure' : ['azure', 'msrestazure'],
-    'workqueue': ['work_queue'],
+    'workqueue': ['conda_pack, stdlib_list'],
```
benclifford: this should be a python list of python strings, not a single string with a non-python list inside it.

benclifford:
Does `conda` need installing too? At least on my machine I can `pip install conda`

tjdasso:
You're right, I'll add that now

benclifford:
Almost everything is failing in CI with the present setup... but interestingly, not everything. There are a few tests that are passing.

Does the workqueue pytest work in your dev environment (`pytest parsl -k "not cleannet" --config parsl/tests/configs/workqueue_ex.py --cov=parsl --cov-append --cov-report= --random-order --bodge-dfk-per-test`) ?

benclifford:
I've encountered this similarly elsewhere, I've just realised. When running in pytest, the PARSL_TESTING environment variable must be set on the remote side too, because some test files will throw errors if imported without it. That environment variable should probably be removed from parsl in general.

This might be what is happening here.

tjdasso:
How can I set that environment variable for the remote side within Travis?

benclifford:
Setting up the remote environment is something that is different per executor - it's not really travis related.

When the other executors run inside travis, they inherit the environment variables of the top level environment, so they don't need setting explicitly; on some platforms, I set the environment for HighThroughputExecutor through the `worker_init` parameter.

I don't know what that looks like with the present workqueue configuration and I also don't know if this really is the problem here - I'm going to see if I can get rid of `PARSL_TESTING` because it mostly annoys me, and I think its original purpose is gone.

benclifford:
If/when #1303 gets merged, this should eliminate problems related to `PARSL_TESTING`. This might make this PR work; or it might eliminate this suspicion and we have to find another problem.

tjdasso:
Looking at the output of the tests within CI, it seems as though one problem is that the output JSON file from `python_package_analyze` is either not being written, or cannot be read when it is passed to `python_package_create`. This might be due to the naming conventions that work within my dev environment, but not generalized for every environment tested in CI

benclifford:
#1303 is merged to master, so you will be able to eliminate PARSL_TESTING from the enquiry by merging the latest `master` to your PR branch.

I had a brief look at how `python_package_analyze` is located - it looks a bit different to how other parsl executables are found. Maybe it should be deployed in the same way as `process-worker-pool.py` is from parsl/executor/high_throughput for consistency - installed by an `entry-point` at the end of `setup.py` and invoked assuming that it is on the `$PATH` because it was installed that way by pip.

tjdasso:
I merged master into my branch, but it seems as though the tests are still failing. For the executables in `entry-point` in `setup.py`, how would I add the scripts? Do the executables have to be Python programs, or can they be other forms of executables as well? This implementation also utilizes a shell script to wrap the task at the worker -- while this isnt needed to run at the master, the file is still required by work queue as an input file to be sent to the worker, and thus the executable must be included within the package.

benclifford:
I'm not really sure of the specifics of entry points in python packaging.

I think you should be able to get an environment set up on a machine you have more access to that looks like the CI parsl/workqueue install, rather than your dev env, and that might make it easier for you to explore what is going on (rather than pushing up to CI which isn't really giving much feedback beyond `no`)

tjdasso:
After running some more local tests in my dev env, I've realized that one of the python analysis executables, `python_package_create`, cannot find a path to an input file because the `runinfo` directories are being renamed. I will reorganize the scripts such that they are not hardcoding this name, and hopefully that should fix one of the problems

dthain:
@trshaffer please update this PR and let us know how this stands relative to the latest WQ integration.

trshaffer:
I think this PR has been largely superseded. Ben and I updated the `python_package_{analyze,create,run}` scripts, so the current ones from CCTools should be used. It also looks like some of the testing and WQ executor code have diverged since this PR.

I wrote an updated implementation based off `master` of automatic packaging (on my branch [here](https://github.com/trshaffer/parsl/commits/package), ignore the more recent commits rebased into #1677) that provides equivalent functionality. The new version doesn't have non-stdlib dependencies (aside from Conda + conda-pack on the master, for obvious reasons), and keeps all the new features off by default. I think the new version should avoid the issues raised above, so it might be merge-able fairly quickly.

yadudoc:
#1714 supersedes this PR and is now merged to master.

PR #1302, opened 2019-09-24
convert adhoc cluster test config to use AdHoc provider

benclifford:


parsl/tests/configs/htex_ad_hoc_cluster.py
```diff
@@ -1,29 +1,27 @@
...
-                worker_init="source /scratch/midway2/yadunand/parsl_env_setup.sh",
-                channel=SSHChannel(hostname=m,
-                                   username=username,
-                                   script_dir="/scratch/midway2/{}/parsl_tests/".format(username)
-                )
+            address=user_opts['public_ip'],
+            provider=AdHocProvider(
+                move_files=False,
+                parallelism=1,
+                worker_init=user_opts['adhoc']['worker_init'],
```
yadudoc: Since we use the same `worker_init` we'll be testing against nodes on a similar filesystems. I test against login nodes on the same shared fs, so having the same worker_init is not a problem.

PR #1316, opened 2019-09-27
Fix a dependency bug that only shows with certain stageout patterns

benclifford:
When a staging provider wishes to edit its File object in stage_in
(for example, to set the local file path) but does not wish to insert
anything into the task graph (for example, because staging will
happen in a task wrapper), and when that File object was copied from
a DataFuture that came from a previous task's output list, then
either:

 * the stageout provider could return None to keep the dependency
   on the earlier stageout in place
or

 * it could return the modified File object so that the stagein wrapper
   sees the modifications.

but it could not do both.

This patch changes the File copying mechanism so that returning None
will both keep the dependency on the earlier stageout, and pass on
the modified file rather than the original application-level File.

It does this by creating a new DataFuture which completes when the
original DataFuture completes, but contains the new copy of the
File rather than the original File.

The bug fixed here was introduced as part of PR#1272.
I discovered it while developing a provider which would do file: scheme
staging in task wrappers, as described above, which is not currently
tested in the test suite.


benclifford:
I've added some test cases in #1417 to demo the specific bugs.

There are two staging providers in there, SP1 and SP2, that do not do any actual staging, but do enough setup to demonstrate the bug that this PR tries to fix.

SP1 stage_in returns a File object, and so apps will correctly see the local_path set by the staging provider. (as tested by the first test, `test_1316_local_path_on_execution_side` which I expect to pass against `master`). However, tasks will not be correctly ordered by files in this case (the second test, `test_1316_local_path_setting_preserves_dependency` demonstrates this - I expect it to fail against `master`).

SP2 stage_in returns None, and so DataFuture-based apps will correctly be ordered, but they will not see the local_path set by stage_in. The third and fourth tests replicate the first two tests but with the SP2 provider; they will fail against `master` because of the missing local_path, even though ordering works correctly.

If you merge #1316 and #1417, then the two SP2 tests will pass. The SP1 tests are a code path that is incorrect after #1316 (and #1316 changes providers to use the SP2-style code path)

Issue #751, opened 2019-01-28
Lookup Worker Information from within Workers


WardLT:
Being able to determine information about a worker (e.g., its rank) would be a nice new feature. 

My need for this is because I am running on a machine with multiple GPUs per node. I need a way to pin each worker to a specific GPU, and currently do so by a convoluted system that makes HTTP requests back to the host process to get a task ID. Something more elegant (@yadudoc has already suggested another hack involving directories) would be much appreciated. Example code below.

-------------------------

```python 

# Make a server for assigning threads to a GPU
@cherrypy.expose
class GPUManager:
    def __init__(self):
        self.lookup = {}

    def POST(self, hostname='', pid=1):
        if hostname not in self.lookup:
            self.lookup[hostname] = {}
        if pid not in self.lookup[hostname]:
            self.lookup[hostname][pid] = len(self.lookup[hostname]) % gpus_per_node
        return str(self.lookup[hostname][pid])

# Start the server on a random port
conf = {
    '/': {
        'request.dispatch': cherrypy.dispatch.MethodDispatcher()
   }
}
cherrypy.config.update({
    'server.socket_port': 5738,
    'server.socket_host': '0.0.0.0'
})
cherrypy.tree.mount(GPUManager(), '/', conf)
cherrypy.engine.start()
manager_hostname = 'http://{}:{}/'.format(hostname, 5738)

# Make the task
@bash_app
def run_training(name, size, hostname):
    # Load libraries (needed for parsl)
    import requests
    import socket
    import os

    # Get the hostname and pid for this process,
    #   use it to determine the GPU
    pid = os.getpid()
    nodename = socket.gethostname()
    gpu_id = requests.post(hostname, params={'hostname': nodename, 'pid': pid}).text

    # Print the GPU id to disk
    return "python3 train_models.py --num-workers=4 -d cuda:{} --copy /dev/shm {} {} >> train.log".format(gpu_id, name, size)
```

yadudoc:
This is related to issue #107 and probably should be addressed in the context of workers being aware of resources and affinities.

WardLT:
I am seeing a need for this issue again when running Parsl applications that use Tensorflow. As discussed in [a Tensorflow issue](https://github.com/tensorflow/tensorflow/issues/8602), multiple processes that use Tensorflow will attempt to use the same resources (e.g., same GPUs, same CPU cores) leading to resource contention. 

[...]

PR #1327, opened 2019-10-03
Added a worker ID environment variable

WardLT:
Allows workers to know their rank within a process pool for HTEX. 

Addresses #751, but might not fix it completely. For example, we might want:

- [x] To access information besides local rank
- [ ] To support other executors


parsl/tests/integration/test_worker_info.py
```diff
@@ -0,0 +1,44 @@
+"""Tests related to Parsl workers being able to access their worker ID"""
```
benclifford: I think this test won't run in pytest in CI because for unpleasant historical parsl reasons, the `/integration/` directory should be ignored - can you move it into a different directory?

WardLT: Any recommendations on a directory to move it to? I was confused about where this would fit. `/sites/`?

benclifford: I've put stuff like this into `/sites/` even though it isn't really a site, so sure.

WardLT:
I'm having trouble running pytest locally. I tried copying the configuration used on Travis, but still get an import error when attempting to run the tests:

```
ImportError while loading conftest '/mnt/c/Users/Logan/Documents/UC/ExaLearn/software/parsl/parsl/tests/conftest.py'.
py._path.local.LocalPath.ImportMismatchError: ('parsl.tests.conftest', '/mnt/c/Users/ward6/Documents/UC/ExaLearn/software/parsl/parsl/tests/conftest.py', local('/mnt/c/Users/Logan/Documents/UC/ExaLearn/software/parsl/parsl/tests/conftest.py'))
```

Any thoughts?

WardLT:
Alright, the worker info here is enough to fix #751 for my needs. Any other features I should add while at it?

benclifford:
@WardLT did you get pytest running?

WardLT:
No, I haven't yet. I'll try to dig in to it later this week

benclifford:
@WardLT I don't fully understand that error message - but I see that it is referring to paths both under  `/mnt/c/Users/ward6` and `/mnt/c/Users/Logan`, and googling suggests that this error happens when .pyc files refer to one place when they are in another place. That maybe suggest something to do with path aliasing or copying parsl checkout directories around after running python code inside them or something along those lines.

parsl/executors/high_throughput/executor.py
```diff
@@ -60,6 +60,13 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
                      +----update_fut-----+
 
 
+    Each of the workers in each process_worker_pool has access to its local rank through
+    an environmental variable, ``PARSL_WORKER_RANK``. The local rank is unique for each process
```
yadudoc: I think RANK is a bit overloaded and can be a bit confusing.
What do you think about `PARSL_WORKER_ID`, `PARSL_WORKER_POOL_ID` (as-is) and `PARSL_WORKER_POOL_SIZE` ?

parsl/tests/sites/test_worker_info.py
```diff
@@ -0,0 +1,44 @@
...
+    ],
+    strategy=None,
+)
+
+
+@python_app
+def get_worker_info():
+    from time import sleep
+    import os
+    sleep(2)
```
yadudoc: We should avoid sleeps in tests if possible.

parsl/executors/high_throughput/executor.py
yadudoc: Or even, `PARSL_WORKER_RANK` and `PARSL_WORKER_COUNT` might be clearer.

WardLT: This is a good point. I did struggle with figuring out what to name the environmental variables.

I would also vote to stick with `PARSL_WORKER_RANK` and `PARSL_WORKER_COUNT`.

Rank is indeed overload, but I was considering that a benefit for people recognizing the purpose of this environmental variable. The issue is that the rank is local to a single executor and some users might expect it to be global for the global pool of executors. However, I like it better than `ID` because I know to expect `RANK` to start at 0 and increase sequentially.

PR #1330, opened 2019-10-03
Adds Python 3.7 testing

dgasmith:
Adds Python 3.7 to the Travis matrix (Python 3.8 is due out soon). Unlikely to cause feature conflicts due to your upgrade cycle, but they do occasional make tweaks to `concurrent.futures`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "3.5"
   - "3.6"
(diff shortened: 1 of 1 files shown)


benclifford:
This fails on 3.6 (not the added 3.7) with an MPI install error that seems unrelated to the PR... this happens sometimes because we install so much stuff each time.

PR #1335, opened 2019-10-04
Make DFK optionally not initialize any logging

benclifford:
The use case for this is when parsl is used as a library
as part of a bigger system which wants to configure logging
in a way that makes sense for that bigger system as a whole.

In such a situation, parsl should step out of the way
and let that system do what it wants.

### DIFF
--- a/parsl/config.py
+++ b/parsl/config.py
@@ -52,17 +52,15 @@ class Config(RepresentationMixin):
     usage_tracking : bool, optional
         Set this field to True to opt-in to Parsl's usage tracking system. Parsl only collects minimal, non personally-identifiable,
         information used for reporting to our funding agencies. Default is False.
-
     initialize_logging : bool, optional
         Make DFK optionally not initialize any logging. Log messages
-        will still be passed into the python logging system, but the
-        logging system will not by default perform any further log
-        system configuration. Most noticeably, it will not create a
-        parsl.log logfile.  The use case for this is when parsl is
-        used as a library in a bigger system which wants to configure
-        logging in a way that makes sense for that bigger system as
-        a whole.  In such a situation, parsl should step out of the way
-        and let that system do what it wants.
+        will still be passed into the python logging system under the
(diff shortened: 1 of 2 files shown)


benclifford:
cc @dgasmith

dgasmith:
Do you want to allow api access to the logger or should we always assume it will be ‘Parsl’?

@lnaden this will supersede you’re PR.

benclifford:
I think it's fine to assume it will always be `parsl`. I'll update the doc for that.

Lnaden:
I'm okay with a different way to adjust logging, but can we have something other than DEBUG or None? The PR I had made (#150) was because I think having Parsl handle its own logging is really helpful, but I wanted to see if we could have a slightly less noisy output, or even have the terminal output be a different level than the written output.

Lnaden:
> I think it's fine to assume it will always be parsl

Keeping this consistent helps us who use Parsl as a lib as well

benclifford:
Really I'm trying to avoid recreating a complex log configuration language that is already expressible using existing python logging configuration language. There is at least one other user that wants to do similar app-specific power-user log configuration, and I'm trying to avoid putting any of this in parsl.

Lnaden:
> Really I'm trying to avoid recreating a complex log configuration language that is already expressible using existing python logging configuration language.

Thats fair. Would you settle for a middle ground then of instead of:

```python
`initialize_logging: bool = True` 
...

    if config.initialize_logging:
            parsl.set_file_logger("{}/parsl.log".format(self.run_dir), level=logging.DEBUG)
```

```python
logging_level: int = logging.DEBUG
...

        parsl.set_file_logger("{}/parsl.log".format(self.run_dir), level=logging_level)
```

Where it just exposes the level Parsl starts at, leaving the default how it is now?

Lnaden:
I realize this still makes Parsl configure its logger, so it dashes one of the uses cases there. I am okay with the original proposed idea of this PR as well, I just wanted to make suggestions.

Closes #1150

Issue #1326, opened 2019-10-03
Add WorkQueueExecutor to parsl docs


annawoodard:

annawoodard:
cc @tjdasso @dthain

dthain:
@tjdasso please take a crack at this, perhaps you can adapt the README that is in the source code directory.  (Users need to know that it exists in order to use it!)

tjdasso:
@annawoodard I'm going through the WQExecutor right now and adding docstrings to the appropriate functions, but are there other places I should look as well to add the WQExecutor to the Parsl docs?

benclifford:
The big user-facing source code documentation place is the docstring for `class WorkQueueExecutor: __init__()` because that the method that users are going to invoke to configure WorkQueueExecutor.

There is also a configuration section in the user guide, `docs/userguide/configuring.rst`, which documents both machine-specific and technology-specific configuration. You could add a section there too, and what I would care about is: "ok, I've initialized a WorkQueueExecutor class and passed it into parsl... but what about the rest of work queue? what else do I have to do to make a job run?". That question is already answered very basically in code in .travis.yml (and it happens on every CI run so we know it works) but could be written in some more human friendly way. Bonus points if it links to relevant sections of workqueue documentation at CCL.

benclifford:
(the target audience for that might be "here's someone who has used parsl a bit before and knows how to make it run jobs with the ThreadPoolExecutor - now you have to point them at all the things they need to know to get going with workqueue")

annawoodard:
@tjdasso fantastic job in #1337. I missed the boat on reviewing that, so here's thoughts on a few minor pieces I think are still missing. I definitely think we want to add to [configuring.rst](https://github.com/Parsl/parsl/blob/master/docs/userguide/configuring.rst) as Ben mentioned. It would also be great to add an example configuration (could be one that would run at ND-- a Condor Provider + WQExecutor config) to that list of example configs we have on that page. Finally, we need to add it to the [reference guide](https://github.com/Parsl/parsl/blob/master/docs/reference.rst) which appears [here](https://parsl.readthedocs.io/en/latest/reference.html). There is some documentation on the documentation [here](https://github.com/Parsl/parsl/tree/master/docs) but in summary, in case you want to build the docs yourself locally, this should do it:

```
cd parsl # wherever the top-level parsl directory is
pip install -r test-requirements.txt # our doc requirements and test requirements are in the same file :(. Also: you may need to add `--user` if you're running in userspace
[...]

PR #1337, opened 2019-10-07
Update Documentation for WorkQueueExecutor

tjdasso:
This PR organizes and adds docstrings to the WQExecutor to better specify the documentation for its use. This attempts to fulfill issue #1326, and if there are other files that need to be updated with the Work Queue documentation, they can be included in this PR.

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -1,3 +1,6 @@
+"""WorkQueueExecutor utilizes the Work Queue distributed framework developed by the Cooperative Computing Lab (CCL) at Notre Dame to provide a fault-tolerant, high-throughput system for delegating Parsl tasks to thousands of remote machines
+"""
+
 import threading
 import multiprocessing
 import logging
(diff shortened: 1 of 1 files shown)


parsl/executors/workqueue/executor.py
```diff
@@ -369,36 +382,61 @@ def WorkQueueCollectorThread(collector_queue=multiprocessing.Queue(),
...
-            password for the work queue project
+            Optional password for the Work Queue project
+
         project_password_file: str
-            password file for the work queue project
+            Optional password file for the work queue project
+
         port: int
-            port to connect to
+            TCP port for Work Queue workers to connect to
```
benclifford: Where is this port if I'm running a parsl submission program on machine A and some workers on machines B and C. I think it is on machine A?

parsl/executors/workqueue/executor.py
```diff
@@ -369,36 +382,61 @@ def WorkQueueCollectorThread(collector_queue=multiprocessing.Queue(),
...
+            TCP port for Work Queue workers to connect to 
+
         env: dict{str}
-            environmental variables
+            Dictionary that contains the environmental variables that
+            need to be set on the Work Queue worker machine
+
         shared_fs: bool
-            define if working in a shared file system or not
+            Define if working in a shared file system or not
```
benclifford: explain why I care / what changes?

parsl/executors/workqueue/executor.py
tjdasso: Yes, on A

PR #1355, opened 2019-10-14
Cleaner reporting for Serialization Errors

yadudoc:
Error message now includes a link to the FAQ.


yadudoc:
Commit history on this PR looks ugly, I will clean it up when this gets squashed.

docs/faq.rst
```diff
@@ -306,6 +306,32 @@ Run::
 Now all available conda environments (for example, one created by following the instructions `here <quickstart.rst#installation-using-conda>`_) will automatically be added to the list of kernels.
 
 
+Addressing SerializationError
```
benclifford: Maybe this text is user guide material not FAQ material

yadudoc: We already have a little note here https://parsl.readthedocs.io/en/latest/userguide/apps.html#limitations. Most users don't run into this issue, so I didn't think it made sense to duplicate this content over on the limitations section. I think it makes it easier to search `SerializationError` and have a note available that explains it that we can point to from the exception message.

benclifford: Those two sections should at least cross-reference each other then, so that someone reading one of them gets to also read the other.

docs/faq.rst
```diff
@@ -309,7 +309,7 @@ Now all available conda environments (for example, one created by following the
 Addressing SerializationError
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-As of `0.9.0` Parsl will raise a `SerializationError` when it encounters a data object passed as an argument to an app.
+As of `1.0.0` Parsl will raise a `SerializationError` when it encounters a data object passed as an argument to an app.
```
benclifford, line 312: I am unclear what a 'data object' is

yadudoc: I've updated text to just say object.

yadudoc:
Will cover in #1489

Issue #1324, opened 2019-10-02
Running tasks with varying "core" requirements in same batch job


TomGlanzman:
It would be beneficial to run tasks with different core_per_worker requirements within a single batch job.  The motivation is to run a heterogeneous set of tasks (tasks with differing core requirements) on the same compute node at NERSC.   I have a workflow that generates many tasks (identical code, different data) to run under the same (htex) executor so that the tasks run in the same batch job.  Each task, in general, needs a different number of cores.  The Cori machine has batch "Haswell" nodes with either 32 cores (and 64 hw threads) or 68 cores (and 272 hw threads).  To efficiently utilize the node, one must be able to keep as many core busy as possible.

This request is to support the ability for the user to specify the number of needed cores at task creation time and to have the appropriate bookkeeping performed to avoid oversubscribing a node.

For "SimpleLauncher", this would mean Parsl would have to handle bookkeeping (i.e., #cores available vs in-use) For "SrunLauncher", srun would presumably do the bookkeeping (potentially across multiple nodes).

annawoodard:
Work Queue supports this functionality. This would be a great use case for the `WorkQueueExecutor`, but there may be a few tweaks needed in to fully exploit those features.  cc @dthain, @btovar, @tjdasso

annawoodard:
Cross ref #1326

TomGlanzman:
@annawoodard does the WorkQueueExecutor have the same basic functionality as the htex?  That is, is wqex a superset of htex or, if not, what functionality will be given up?

tjdasso:
In regards to WorkQueue, each task can specify the number of cores required. With that being said, the WorkQueue executor does not currently specify the number of cores per task, but it shouldn't be a hard feature to implement. With that being said, how will the Parsl app specify the number of cores it needs to run the task, when it is submitted to the WQExecutor?

annawoodard:
One possibility is to add it as a keyword argument to the decorator, for example:

```python
@python_app(cores=2)
def foo():
    return 'Hello, world!'
```

[...]

Issue #873, opened 2019-04-13
Add type annotations for all config-like functions


benclifford:
PR #860 introduces runtime checking of top level parameters to `Config` using enforce.py and PEP484 type annotations.

This should be expanded to all functions likely to be called by a user as part of configuration: at least all of the `__init__` methods for executors, channels, providers, and the address helper functions.

PR #1358, opened 2019-10-14
Specify Resources in App Decorator

tjdasso:
Refer #1324 for issue, but this PR allows users to optionally specify resource requirements within the Parsl app decorator, which enables Work Queue to specify tasks to run with specific resource usage.

This can be done at the function level, such as:
```
@python_app(cores=4, mem=100, disk=100)
def example():
    ...
```
which would specify each function call to example() to require 4 cores, 100MB of memory, and 100MB of disk. Not all resources need to be specified: any combination of cores, memory, and disk can be specified, and if they are not listed, it will default to Work Queue's default settings (where one task consumes an entire worker). 

Additionally, users have an option to specify resource requirements _per task_, by passing in resource requirements as a keyword argument to the function. More specifically, it uses the keyword argument `parsl_resource_specification` to map to a dictionary of resource requirements. For example:
```
@python_app
def example(parsl_resource_specification={cores=2, mem=100, disk=100}):
    ...
```
This allows for the use case where a user might want to specify different resource requirements for different iterations of the same function call, such as if one function call has a much bigger input which would require more computational resources. If resources are specified in both the decorator and as an argument to the function using the `parsl_resource_specification` keyword, it defaults to using the argument values, allowing the values within the decorator to act as "default" resource specifications for the app.

### DIFF
--- a/parsl/app/app.py
+++ b/parsl/app/app.py
@@ -21,7 +21,7 @@ class AppBase(metaclass=ABCMeta):
 
     """
 
-    def __init__(self, func, data_flow_kernel=None, walltime=60, executors='all', cache=False, cores=1):
+    def __init__(self, func, data_flow_kernel=None, walltime=60, executors='all', cache=False, cores=None, disk=None, mem=None):
         """Construct the App object.
 
         Args:
@@ -45,6 +45,8 @@ class AppBase(metaclass=ABCMeta):
         self.status = 'created'
         self.executors = executors
         self.cores = cores
+        self.disk = disk
+        self.mem = mem
         self.cache = cache
         if not (isinstance(executors, list) or isinstance(executors, str)):
             logger.error("App {} specifies invalid executor option, expects string or list".format(
@@ -78,7 +80,7 @@ class AppBase(metaclass=ABCMeta):
         pass
 
 
-def App(apptype, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
+def App(apptype, data_flow_kernel=None, walltime=60, cache=False, cores=None, mem=None, disk=None, executors='all'):
     """The App decorator function.
 
     Args:
@@ -116,11 +118,13 @@ def App(apptype, data_flow_kernel=None, walltime=60, cache=False, cores=1, execu
                          walltime=walltime,
                          cache=cache,
                          cores=cores,
+                         mem=mem,
+                         disk=disk,
                          executors=executors)
     return wrapper
 
 
-def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
+def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=None, mem=None, disk=None, executors='all'):
     """Decorator function for making python apps.
 
     Parameters
@@ -149,6 +153,8 @@ def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, c
                              walltime=walltime,
                              cache=cache,
                              cores=cores,
+                             mem=mem,
+                             disk=disk,
                              executors=executors)
         return wrapper(func)
     if function is not None:
@@ -156,7 +162,7 @@ def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, c
     return decorator
 
 
-def bash_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
+def bash_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=None, mem=None, disk=None, executors='all'):
     """Decorator function for making bash apps.
 
     Parameters
@@ -185,6 +191,8 @@ def bash_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cor
                            walltime=walltime,
                            cache=cache,
                            cores=cores,
+                           mem=mem,
+                           disk=disk,
                            executors=executors)
         return wrapper(func)
     if function is not None:
--- a/parsl/app/app.py
+++ b/parsl/app/app.py
@@ -21,7 +21,7 @@ class AppBase(metaclass=ABCMeta):
 
     """
 
-    def __init__(self, func, data_flow_kernel=None, walltime=60, executors='all', cache=False):
+    def __init__(self, func, data_flow_kernel=None, walltime=60, executors='all', cache=False, cores=1):
         """Construct the App object.
 
         Args:
@@ -44,6 +44,7 @@ class AppBase(metaclass=ABCMeta):
         self.data_flow_kernel = data_flow_kernel
         self.status = 'created'
         self.executors = executors
+        self.cores = cores
         self.cache = cache
         if not (isinstance(executors, list) or isinstance(executors, str)):
             logger.error("App {} specifies invalid executor option, expects string or list".format(
@@ -77,7 +78,7 @@ class AppBase(metaclass=ABCMeta):
         pass
 
 
-def App(apptype, data_flow_kernel=None, walltime=60, cache=False, executors='all'):
+def App(apptype, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
     """The App decorator function.
 
     Args:
@@ -114,11 +115,12 @@ def App(apptype, data_flow_kernel=None, walltime=60, cache=False, executors='all
                          data_flow_kernel=data_flow_kernel,
                          walltime=walltime,
                          cache=cache,
+                         cores=cores,
                          executors=executors)
     return wrapper
 
 
-def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, executors='all'):
+def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
     """Decorator function for making python apps.
 
     Parameters
@@ -146,6 +148,7 @@ def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, e
                              data_flow_kernel=data_flow_kernel,
                              walltime=walltime,
                              cache=cache,
+                             cores=cores,
                              executors=executors)
         return wrapper(func)
     if function is not None:
@@ -153,7 +156,7 @@ def python_app(function=None, data_flow_kernel=None, walltime=60, cache=False, e
     return decorator
 
 
-def bash_app(function=None, data_flow_kernel=None, walltime=60, cache=False, executors='all'):
+def bash_app(function=None, data_flow_kernel=None, walltime=60, cache=False, cores=1, executors='all'):
     """Decorator function for making bash apps.
 
     Parameters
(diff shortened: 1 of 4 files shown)


tjdasso:
It looks like this PR is failing in CI because the other executors aren't expecting the additional kwarg that this PR creates ('parsl_resource_specification`), and thus fails when it passes in the kwargs to the function during execution. Is there any way to only add the kwargs if we know what executor is being used (ie, if we know that the executor is WQ Executor, which has built-in handling for this additional argument)?

annawoodard:
@tjdasso: Yikes, I was not thinking carefully-- I'm sorry for giving you bad advice! Since inserting into kwargs was a hack anyways, I think the best way forward is to do this by changing the `submit` signature in the executor base class and allowing each executor implementation to decide what to do with it. All the existing executors ignore it except wqex. In this way, if someone implements a new executor, they'll know that they have access to the resource specification, if they want to use it. I haven't tested it with wqex, but here's a sketch of how this might look:

```diff
diff --git a/parsl/app/bash.py b/parsl/app/bash.py
index 9ab467c..5c1cb95 100644
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -147,12 +147,6 @@ class BashApp(AppBase):
         """
         # Update kwargs in the app definition with ones passed in at calltime
         self.kwargs.update(kwargs)
-        if 'parsl_resource_specification' not in self.kwargs:
-            self.kwargs['parsl_resource_specification'] = {
-                'cores': self.cores,
-                'disk': self.disk,
-                'mem': self.mem
-            }

         if self.data_flow_kernel is None:
             dfk = DataFlowKernelLoader.dfk()
@@ -164,6 +158,9 @@ class BashApp(AppBase):
                              executors=self.executors,
                              fn_hash=self.func_hash,
                              cache=self.cache,
+                             cores=self.cores,
+                             disk=self.disk,
+                             mem=self.mem,
                              **self.kwargs)

         return app_fut
diff --git a/parsl/app/python.py b/parsl/app/python.py
index 5671d2d..e86cbb3 100644
--- a/parsl/app/python.py
+++ b/parsl/app/python.py
@@ -65,13 +65,6 @@ class PythonApp(AppBase):
         else:
             dfk = self.data_flow_kernel

-        if 'parsl_resource_specification' not in kwargs:
-            kwargs['parsl_resource_specification'] = {
-                'cores': self.cores,
-                'mem': self.mem,
-                'disk': self.disk
-            }
-
         walltime = self.kwargs.get('walltime')
         if walltime is not None:
             self.func = timeout(self.func, walltime)
@@ -79,6 +72,9 @@ class PythonApp(AppBase):
                              executors=self.executors,
                              fn_hash=self.func_hash,
                              cache=self.cache,
+                             cores=self.cores,
+                             disk=self.disk,
+                             mem=self.mem,
                              **kwargs)

         return app_fut
diff --git a/parsl/dataflow/dflow.py b/parsl/dataflow/dflow.py
index 302225b..4ae3cfa 100644
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -439,7 +439,7 @@ class DataFlowKernel(object):
                                                          self.monitoring.resource_monitoring_interval)

         with self.submitter_lock:
-            exec_fu = executor.submit(executable, *args, **kwargs)
+            exec_fu = executor.submit(executable, self.tasks[task_id]['resource_specification'], *args, **kwargs)
         self.tasks[task_id]['status'] = States.launched
         if self.monitoring is not None:
             task_log_info = self._create_task_log_info(task_id, 'lazy')
@@ -600,7 +600,7 @@ class DataFlowKernel(object):

         return new_args, kwargs, dep_failures

-    def submit(self, func, *args, executors='all', fn_hash=None, cache=False, **kwargs):
+    def submit(self, func, *args, executors='all', fn_hash=None, cache=False, cores=None, mem=None, disk=None, **kwargs):
         """Add task to the dataflow system.

         If the app task has the executors attributes not set (default=='all')
@@ -660,6 +660,12 @@ class DataFlowKernel(object):
                                 kw)
                     )

+        resource_specification = {
+            'cores': cores,
+            'mem': mem,
+            'disk': disk
+        }
+
         task_def = {'depends': None,
                     'executor': executor,
                     'func_name': func.__name__,
@@ -674,7 +680,8 @@ class DataFlowKernel(object):
                     'status': States.unsched,
                     'id': task_id,
                     'time_submitted': None,
-                    'time_returned': None}
+                    'time_returned': None,
+                    'resource_specification': resource_specification}

         app_fu = AppFuture(task_def)

diff --git a/parsl/executors/base.py b/parsl/executors/base.py
index 294a2f7..543dbba 100644
--- a/parsl/executors/base.py
+++ b/parsl/executors/base.py
@@ -37,7 +37,7 @@ class ParslExecutor(metaclass=ABCMeta):
         pass

     @abstractmethod
-    def submit(self, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """Submit.

         We haven't yet decided on what the args to this can be,
diff --git a/parsl/executors/high_throughput/executor.py b/parsl/executors/high_throughput/executor.py
index 21f4918..11649fe 100644
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -504,7 +504,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
                 logger.debug("[HOLD_BLOCK]: Sending hold to manager: {}".format(manager['manager']))
                 self.hold_worker(manager['manager'])

-    def submit(self, func, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """Submits work to the the outgoing_q.

         The outgoing_q is an external process listens on this
diff --git a/parsl/executors/ipp.py b/parsl/executors/ipp.py
index 3faa412..307721f 100644
--- a/parsl/executors/ipp.py
+++ b/parsl/executors/ipp.py
@@ -212,7 +212,7 @@ sleep infinity
     def scaling_enabled(self):
         return self._scaling_enabled

-    def submit(self, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """Submits work to the thread pool.

         This method is simply pass through and behaves like a submit call as described
@@ -221,7 +221,7 @@ sleep infinity
         Returns:
               Future
         """
-        return self.lb_view.apply_async(*args, **kwargs)
+        return self.lb_view.apply_async(func, *args, **kwargs)

     def scale_out(self, blocks=1):
         """Scales out the number of active workers by 1.
diff --git a/parsl/executors/low_latency/executor.py b/parsl/executors/low_latency/executor.py
index d24737b..3a9b092 100644
--- a/parsl/executors/low_latency/executor.py
+++ b/parsl/executors/low_latency/executor.py
@@ -189,7 +189,7 @@ class LowLatencyExecutor(ParslExecutor, RepresentationMixin):

         logger.info("[MTHREAD] queue management worker finished")

-    def submit(self, func, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """ TODO: docstring """
         self._task_counter += 1
         task_id = self._task_counter
diff --git a/parsl/executors/threads.py b/parsl/executors/threads.py
index c981b0b..df8663b 100644
--- a/parsl/executors/threads.py
+++ b/parsl/executors/threads.py
@@ -54,14 +54,14 @@ class ThreadPoolExecutor(ParslExecutor, RepresentationMixin):
     def scaling_enabled(self):
         return self._scaling_enabled

-    def submit(self, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """Submits work to the thread pool.

         This method is simply pass through and behaves like a submit call as described
         here `Python docs: <https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor>`_

         """
-        return self.executor.submit(*args, **kwargs)
+        return self.executor.submit(func, *args, **kwargs)

     def scale_out(self, workers=1):
         """Scales out the number of active workers by 1.
diff --git a/parsl/executors/workqueue/executor.py b/parsl/executors/workqueue/executor.py
index 1ac492c..c14f819 100644
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -547,7 +547,7 @@ class WorkQueueExecutor(ParslExecutor):
             index += 1
         return new_name

-    def submit(self, func, *args, **kwargs):
+    def submit(self, func, resource_specification, *args, **kwargs):
         """Submit.

         We haven't yet decided on what the args to this can be,
@@ -560,12 +560,6 @@ class WorkQueueExecutor(ParslExecutor):
         input_files = []
         output_files = []
         std_files = []
-        resources = {}
-
-        # Receive resource specifications from the kwargs, then remove it
-        if 'parsl_resource_specification' in kwargs:
-            resources = kwargs["parsl_resource_specification"]
-            kwargs.pop("parsl_resource_specification")

         # Add input files from the "inputs" keyword argument
         func_inputs = kwargs.get("inputs", [])
@@ -649,7 +643,7 @@ class WorkQueueExecutor(ParslExecutor):
         # Create message to put into the message queue
         logger.debug("Placing task {} on message queue".format(task_id))
         msg = {"task_id": task_id,
-               "resources": resources,
+               "resources": resource_specification,
                "data_loc": function_data_file,
                "result_loc": function_result_file,
                "input_files": input_files,
```

annawoodard:
Unfortunately, specifying resources per task is now a bit awkward. Here is an example of how to do it. In my view, we can decide later if we want to add 'magic' keyword args (analogous to `stderr` and `stdout` currently) and not hold this PR up.

**per task**
```python
import parsl

parsl.set_stream_logger()
parsl.load()

def foo(name):
    print('Hello, {}!'.format(name))

python_app(foo, cores=4)('small per-task world')
python_app(foo, cores=8)('big per-task world')
```
**per app**
```python
import parsl

parsl.set_stream_logger()
parsl.load()

@python_app(cores=2)
def bar(name):
    print('Hello, {}!'.format(name))

bar('per-app world')

benclifford:
@annawoodard if resource_specification is an object rather than a dict, it can be documented and type-checked (in the same way as Configs). This means it can't become a quagmire of random options so easily. Which might be a feature, or might be a downside - depending on whether you intend each executor implementation to be able to have its own arbitrary options not specified by parsl core, or whether you want parsl core to specify what options mean even if they aren't implemented everywhere. (I think the latter, because there's a hard-coded dict initializer in the DFK in your patch)

benclifford:
Maybe for usability parsl should be emitting warning when it ignores fields in resource_specification.

dgasmith:
One note is that `memory` might be better than `mem` from a user point of view. A simple parser for `100MB` or `100GiB` is usually advised as well. Best to double check the `cores` definition complies with the language used in the partial node PR.

@annawoodard Any harm in having thousand of python apps to comply with the new definition?

@benclifford Cannot say enough good things about [Pydantic](https://pydantic-docs.helpmanual.io) for automatic type checked and serialized objects.

benclifford:
@dgasmith we're already using `mypy` for some static and `typeguard` for some dynamic checking using type hints. Issue #873 is a slow ongoing project that would welcome contributions.

dthain:
Adding @tjuedema to this thread.

dthain:
@ZhuozhaoLi is going to move towards a JSON resource spec here.

dthain:
superseded by #1675

PR #1362, opened 2019-10-16
Add a note about summit paths

yadudoc:
Renaming summit config for consistency, and a note about paths on summit.


benclifford:
Is this meant to be a rename of an existing file? This PR adds a new file without removing any.

benclifford:
@yadudoc this PR doesn't rename anything.

yadudoc:
Fixing this in a separate PR. I had done a git mv, but that didn't do a delete apparently.

PR #1363, opened 2019-10-16
Fix and test against Python 3.8

benclifford:
This updates codeutil.py with corresponding changes from ipyparallel to deal with a changed API in Python 3.8.

It also adds python 3.8 to CI


parsl/executors/serialize/codeutil.py
```diff
@@ -23,12 +23,14 @@ def code_ctor(*args):
 
 
 def reduce_code(co):
-    args = [co.co_argcount, co.co_nlocals, co.co_stacksize,
+    args =  [co.co_argcount, co.co_nlocals, co.co_stacksize,
```
yadudoc: Is this extra whitespace intentional ?

parsl/executors/serialize/codeutil.py
```diff
@@ -23,12 +23,14 @@ def code_ctor(*args):
...
             co.co_flags, co.co_code, co.co_consts, co.co_names,
             co.co_varnames, co.co_filename, co.co_name, co.co_firstlineno,
             co.co_lnotab, co.co_freevars, co.co_cellvars]
     if sys.version_info[0] >= 3:
         args.insert(1, co.co_kwonlyargcount)
+    if sys.version_info > (3, 8):
+        args.insert(1, co.co_posonlyargcount)
```
yadudoc: I checked on the sys pages and couldn't find what changed in the api on 3.8. Can you clarify, please ?

parsl/executors/serialize/codeutil.py
benclifford: It comes from re-copying the original at ipyparallel/serialize/codeutil.py where it is indented that way. i've changed it back.

parsl/executors/serialize/codeutil.py
benclifford: This comes from the below two ippyparallel commits which modify the ipyparallel version of codeutil.py to work with Python 3.8, https://github.com/ipython/ipyparallel/blob/master/ipyparallel/serialize/codeutil.py

```
commit 6017b1aa645331b29b5171596f9245b02395dc94
Author: Min RK <benjaminrk@gmail.com>
Date:   Thu Sep 19 19:24:25 2019 +0200

    specify 3.8 version_info
    
    remove awareness of which alpha release changed this
    
    no reason to support 3.8 alpha < 3

commit f88551d8a89510c225f40efe20bf4b9f7fd0a1fb
Author: Mattias Ellert <mattias.ellert@physics.uu.se>
Date:   Mon Aug 12 16:05:13 2019 +0200

    adapt to Python 3.8 with PEP 570
```

It's a change to the codeobject internal type (the internal types section of https://docs.python.org/3/reference/datamodel.html describes the (current, 3.8) fields without describing the change). code objects (like python 3.8) now have positional only arguments (not accessible with keywords), and the code constructor now has a parameter for those. (PEP 570 - https://www.python.org/dev/peps/pep-0570/)

This was introduced in the cpython commit between 3.8a3 and 3.8.0a4: https://github.com/python/cpython/commit/843bf42aa65aaa25b356e7b3d8733a117b8f01a4
```
commit 8c77b8cb9188165a123f2512026e3629bf03dc9b
Author: Pablo Galindo <Pablogsal@gmail.com>
Date:   Mon Apr 29 13:36:57 2019 +0100

    bpo-36540: PEP 570 -- Implementation (GH-12701)
```

PR #1368, opened 2019-10-16
Update stampede2 config to use SrunLauncher

yadudoc:


ZhuozhaoLi:
There is a very similar config of stampede2 on `parsl/parsl/configs/stampede2_htex_multinode.py`. Why not just change that or remove that?

PR #1369, opened 2019-10-17
change the theta config path and add an option to AprunLauncher

ZhuozhaoLi:
This PR
(1) renames the Theta config path
(2) adds `min_blocks` and `max_blocks` to show how users can configure the scaling on Theta
(3) add `overrides="-d 64"` to AprunLauncher, which can deploy the workers across all the 64 cores on compute nodes.

### DIFF
--- a/parsl/configs/theta.py
+++ b/parsl/configs/theta.py
@@ -14,10 +14,12 @@ config = Config(
             provider=CobaltProvider(
                 queue='YOUR_QUEUE',
                 account='YOUR_ACCOUNT',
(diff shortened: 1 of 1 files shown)


benclifford:
Make the PR describe why this change makes parsl better

ZhuozhaoLi:
> Make the PR describe why this change makes parsl better

Edited

benclifford:
See #1370 for resolution of that CI error which is unrelated to this PR.

PR #1370, opened 2019-10-18
Fix formatting parameters that mypy has started analyzing harder

benclifford:
Recent mypy has started checking format parameters harder, resulting in the
following CI failure:

parsl/launchers/launchers.py:118: error: Not all arguments converted during string formatting
parsl/launchers/launchers.py:162: error: Not all arguments converted during string formatting
parsl/launchers/launchers.py:201: error: Not all arguments converted during string formatting

This commit changes those launchers to use a style that mypy and flake8 are both
happy with.

This PR should not change any user facing behaviour.


parsl/launchers/launchers.py
```diff
@@ -81,11 +81,10 @@ def __call__(self, command, tasks_per_node, nodes_per_block):
         - task_block (string) : bash evaluated string.
 
         """
-        task_blocks = tasks_per_node * nodes_per_block
 
         x = '''export CORES=$(getconf _NPROCESSORS_ONLN)
 echo "Found cores : $CORES"
-WORKERCOUNT={3}
+WORKERCOUNT={nodes_per_block}
```
ZhuozhaoLi: If I understand correctly, this changes the original values from `total workers` to `total nodes`, correct?

parsl/launchers/launchers.py
```diff
@@ -137,11 +136,10 @@ def __call__(self, command, tasks_per_node, nodes_per_block):
...
```
ZhuozhaoLi: Same here

parsl/launchers/launchers.py
```diff
@@ -152,14 +150,14 @@ def __call__(self, command, tasks_per_node, nodes_per_block):
 fi
 
 cat << MPIEXEC_EOF > cmd_$JOBNAME.sh
-{0}
+{command}
 MPIEXEC_EOF
 chmod u+x cmd_$JOBNAME.sh
 
 mpiexec --bind-to none -n $WORKERCOUNT --hostfile $HOSTFILE /usr/bin/sh cmd_$JOBNAME.sh
```
ZhuozhaoLi: The change in `WORKERCOUNT` would change the behavior here.

parsl/launchers/launchers.py
```diff
@@ -188,17 +186,17 @@ def __call__(self, command, tasks_per_node, nodes_per_block):
 
         x = '''export CORES=$(getconf _NPROCESSORS_ONLN)
 echo "Found cores : $CORES"
-WORKERCOUNT={3}
+WORKERCOUNT={nodes_per_block}
```
ZhuozhaoLi: Same here.

parsl/launchers/launchers.py
```diff
@@ -188,17 +186,17 @@ def __call__(self, command, tasks_per_node, nodes_per_block):
...
-{0}
+{command}
 MPIRUN_EOF
 chmod u+x cmd_$JOBNAME.sh
 
-mpirun -np $WORKERCOUNT {4} cmd_$JOBNAME.sh
+mpirun -np $WORKERCOUNT {task_blocks} cmd_$JOBNAME.sh
 
 echo "All workers done"
-'''.format(command, tasks_per_node, nodes_per_block, task_blocks, self.bash_location)
```
ZhuozhaoLi: This `bash_location` is completely removed.

ZhuozhaoLi:
The CI failed with similar errors `parsl/tests/test_checkpointing/test_periodic.py:63: error: Not all arguments converted during string formatting`

parsl/launchers/launchers.py
benclifford: Oh, this is an off-by-one error in my patch. I was starting counting at 1.

benclifford:
If the 0.9.0 release is coming from master (rather than eg a fork of 0.9.0-a0) I think this PR should not be merged yet, and instead maybe something that pins an older mypy version should be added - #1370 invalidates a bunch of testing that involves launchers, I expect.

yadudoc:
@benclifford We'll be releasing from master. Let's hold this till 0.9.0 is out.

PR #1375, opened 2019-10-21
Fix sites/test_ec2 failure when aws user opts specified

benclifford:
Previously, if no AWS user opts were specified, then
the tests in sites/test_ec2.py were skipped (due to a
key error on importing the ec2 config); but if AWS
user opts *were* specified, then the tests would run
in non-local mode.

This commit changes the test so that it runs with a
locally defined config.

This test runs in local mode if suitable AWS configuration
is made - which at the moment is a more awkward than
setting the user opts for AWS.

This looks like it was messed up in a combination of
PR #1210 (which replaced the local mark with a noci mark)
and PR #1230 (which removed the noci mark)


benclifford:
Looks like this fails when there isn't a public_ip user option specified... in other places such a condition causes the test to be skipped, rather than causing an exception.

PR #1376, opened 2019-10-21
Fix non-relative import in AWS test config

benclifford:


benclifford:
The previous behaviour raises an exception:

```
pytest parsl --config  parsl/tests/configs/ec2_single_node.py 
...
E   ImportError: attempted relative import with no known parent package
```

It might have taken an import from the current directory at some point but it hasn't for a while at least in some circumstances.

PR #1378, opened 2019-10-21
Add an rsync file staging provider

benclifford:
This supports cloud-based execution where the cloud nodes are able
to connect using rsync back to the submitting system to access the
submit-side filesystem.


parsl/data_provider/rsync.py
```diff
@@ -0,0 +1,108 @@
...
+def in_task_stage_out_wrapper(func, file, working_dir, hostname):
+    def wrapper(*args, **kwargs):
+        import logging
+        logger = logging.getLogger(__name__)
+        logger.debug("rsync in_task_stage_out_wrapper start")
+
+        logger.debug("rsync in_task_stage_out_wrapper calling wrapped function")
+        result = func(*args, **kwargs)
+        logger.debug("rsync in_task_stage_out_wrapper returned from wrapped function, calling rsync")
+        r = os.system("rsync {worker_filepath} {hostname}:{permanent_filepath}".format(hostname=hostname,
```
yadudoc: Can this hang ?

yadudoc:
This is fairly separate from the existing code paths tested so far, so I'm okay with this going into 0.9.0. What do you think @benclifford ?

benclifford:
I don't think it will hurt anything going into master

parsl/data_provider/rsync.py
benclifford: I've never seen rsync hang forever in real life.

PR #1389, opened 2019-10-23
Install test-requirements in the same way as other optional dependencies

benclifford:
This should make installing test requirements from any pip source easier -
for example, when installing parsl on a remote worker for pytest.

Without this change, when installing using pip, the test-requirements.txt
file is not easily available to allow test dependencies to be installed. With
this change, the test dependencies can be installed using the same
optional dependency sytax as with other dependencies, using `[testing]`.


yadudoc:
@benclifford  This PR probably needs another look given the recent switch to using Makefiles.

yadudoc:
@benclifford I think there's some minimal redundancy with this and the the makefile. I think I'm okay with that. Could you check the conflicts in `.travis.yml` please ?

yadudoc:
Closing.

Issue #1382, opened 2019-10-22
KeyError anywhere causes pytest+our fixtures to skip test rather than report failure


benclifford:
**Describe the bug**
A KeyError raised by any code causes a pytest test to be marked as skipped rather than as failed. This means that some failing tests will instead be reported as skipped, and can cause the test suite to complete without error when it should fail.

This happens because the parsl test fixture (`conftest.py`) that translates missing test `user_opts` into skipped tests is overly broad in the kind of failure it translates: any KeyError from anywhere in a test run.

**To Reproduce**
Raise a KeyError somewhere in the parsl source, and run a pytest that touches that bit of source.

**Expected behavior**
The KeyError should be reported as a failure.

**Environment**
My laptop, python 3.7, master 51cd498cb17c908135b9409b21e718f43308a400

PR #1392, opened 2019-10-23
Tighten KeyError handling that was falsely skipping failing tests (#1382)

benclifford:
This pulls out all KeyError handling except at the collect stage.

At the collect stage, keyerrors only cause a skip if the top of the stack
trace is in parsl/tests/configs/ - so you can only reference user_opts there,
not anywhere else.

KeyErrors happening outside of the collect stage, even if they are in a
config file (for example, if a test local_setup() does an import of a
config), will result in test failures.

Broadly this means that KeyErrors happening in parsl/tests/configs/*
imported at the top level into a test will cause a skip; otherwise KeyErrors
will be treated as errors.


yadudoc:
@benclifford Do you think this draft is ready to be reviewed and merged ?

benclifford:
This has been partly (but I think not entirely) superceded by PR #1803 so should be updated in that context

benclifford:
unlikely i'll get this mergeable any time soon - closing until such time as I do

PR #1400, opened 2019-10-25
Add typeguard to MonitoringHub initializer

benclifford:
This is part of issue #873

This converts at least one hang (when hub_port is specified as a string) into a config-time exception.


parsl/monitoring/monitoring.py
```diff
@@ -122,22 +123,23 @@ def send(self, message_type, task_id, message):
...
+                 client_address: str = "127.0.0.1",
+                 client_port_range: Tuple[int, int] = (55000, 56000),
+
+                 workflow_name: Optional[str] = None,
+                 workflow_version: Optional[str] = None,
+                 logging_endpoint: str = 'sqlite:///monitoring.db',
+                 logdir: Optional[str] = None,
+                 monitoring_debug: bool = False,
+                 resource_monitoring_enabled: bool = True,
+                 resource_monitoring_interval: int = 30):  # in seconds
```
ZhuozhaoLi: I think this `resource_monitoring_interval` should be float (the docstrings were wrong before this PR)

benclifford: I've change it to float both here and in the docstring.

PR #1404, opened 2019-10-25
Clean up configs and match comment style

yadudoc:
Updating configs to match in description style.


parsl/configs/kubernetes.py
```diff
@@ -0,0 +1,41 @@
...
+            cores_per_worker=1,
+            max_workers=1,
+            worker_logdir_root='YOUR_WORK_DIR',
+
+            # Address for the pod worker to connect back
+            address=address_by_route(),
+            provider=KubernetesProvider(
+                namespace="default",
+
+                # Docker image url to use for deployment
```
ZhuozhaoLi: Better to say `for pods` rather than `for deployment`?

yadudoc: Fixed.

benclifford:
need to update `userguide/configuring.rst` which includes some of these by path reference?

yadudoc:
@benclifford Just updated the config references

benclifford:
```
make html
```
...

```
/home/benc/parsl/src/parsl/docs/userguide/configuring.rst:318: WARNING: Include file '/home/benc/parsl/src/parsl/parsl/configs/ec2_htex_single_node.py' not found or reading it failed
```

Issue #1205, opened 2019-08-21
Remove DataFuture special-case string handling


benclifford:
When a DataFuture is constructed, it handles strings as a special case; this is used when passing a string as one of an app's `outputs=[]`. The special-casing treats that string as a `file:` URL for the purposes of the data future; this works when `file:` URLs are unstaged, but post-#1201, this breaks: passing a string as an output will not cause file staging, when passing the previously equivalent `file:` URL will.

This special-casing should be removed.

PR #1412, opened 2019-10-29
Remove str special-casing in DataFutures as they should wrap only Files

benclifford:
A number of tests still used the old style of output handling where strings
could be passed in and were implicitly treated as files. Those tests have
been changed to use the mandatory-File form of outputs.

Fixes issue #1205


yadudoc:
@benclifford Looks like the travis build is failing at the check for doc stubs (https://travis-ci.com/github/Parsl/parsl/jobs/322756321#L2320). I am not able to reproduce this on my laptop. Can you take a look please?

benclifford:
most recent fail in ff51e21 is due to flake8 not doc stubs

Issue #1058, opened 2019-06-12
bash_app keyword arg defaults are overwritten when a bash_app is called multiple times


benclifford:
given
```
@bash_app
def f(z=2):
return "echo {z}"
```

then the app call sequence:
```
f()
f(z=3)
f(z=4)
f()
```
will output 2,3,4,4, when it should output 2,3,4,2

I think this comes from the use of a `kwargs` structure that is scoped to the definition of the `f` app and mutated on every invocation, in `parsl/app/bash.py`

This branch adds a more fleshed out test case:
https://github.com/Parsl/parsl/tree/benc-1058-bash-app-kwarg-defaults

benclifford:
fixed by #1413

PR #1413, opened 2019-10-29
Duplicate kwargs in bash_app call rather than modifying reusable dict

benclifford:
The xfailing test for this needed updating a bit to reflect
recent changes in parsl.

Fixes issue #1058


benclifford:
What goes wrong in the case that the default value is a dict? It can be modified by an app running in a local thread executor and I think that modification is seen by all invocations, which is correct python semantics. In the case of a different-process executor like HTEX those modifications don't change the submit side values, which is parsl's behaviour for any objects passed in as a parameter.

yadudoc:
I tested this out a bit, and i don't think the situation where one of the args is a dict itself, is a concern for us.

Issue #648, opened 2018-11-14
Document host address heuristics


annawoodard:
We need to document the changes implemented in #646.

PR #1414, opened 2019-10-29
Add parsl.addresses to reference guide

benclifford:
Partly addresses #648


parsl/addresses.py
```diff
@@ -9,36 +9,52 @@
 logger = logging.getLogger(__name__)
 
 
-def address_by_route():
+def address_by_route() -> str:
+    """Finds an address for the local host by querying the local routing table
+       for the route to Google DNS.
+
+       This will return an incorrect value when the internet-facing address is
```
yadudoc: This may not return a correct value when the network doesn't give an externally accessible address ?

parsl/addresses.py
```diff
@@ -9,36 +9,52 @@
...
     logger.debug("Address found: {}".format(addr))
     return addr
 
 
-def address_by_hostname():
+def address_by_hostname() -> str:
+    """Returns the hostname of the local host.
+
+       This will return an incorrect value when the hostname cannot be
+       resolved correctly from workers.
```
yadudoc: It is returning a correct value, but a value that may not be resolvable on the workers.

benclifford:
I think both your comments relate you having a different interpretation of what "correct" means... I mean correct as in "it will work when passed into htex as the public IP address" rather than "an address was successfully discovered using the described mechanism". Let me pick a better wording to reflect this.

yadudoc:
Fixes #648

benclifford:
I've changed the term '(in)correct' to '(un)usable' which I think better reflects what I intended.

I've also added a docstring for the whole module.

PR #1415, opened 2019-10-29
Update pytest >=4.6

benclifford:
This commit requires a much newer version of pytest.

It changes `conftest.py` to use the API of that newer pytest.

This is needed as part of running CI with python 3.8


benclifford:
This failed once in CI for the python 3.5 test, but not for the other two version, with a ConnectionRefusedError on an ftp test. This seems unrelated to the changes introduced by the PR. I've restarted that test to see if it passes.

benclifford:
looks like our pytest marks are broken by this PR on 3.5 but not on other versions(!)

benclifford:
This now passes in CI: on my laptop, with no `pytest` installed, the most recent valid pytest was install (4.6.6). In CI, pytest 4.1.0 is pre-installed. This was also valid according to the requirements I added, but didn't work. I've updated the requirements to >=4.6 and this now passes in CI.

benclifford:
I'm seeing a bunch of
```
--- Logging error ---
  File "/opt/python/3.5.6/lib/python3.5/logging/__init__.py", line 988, in emit
    stream.write(msg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/_pytest/capture.py", line 441, in write
    self.buffer.write(obj)
ValueError: I/O operation on closed file
Call stack:
  File "/home/travis/build/Parsl/parsl/parsl/dataflow/dflow.py", line 855, in atexit_cleanup
    self.cleanup()
  File "/home/travis/build/Parsl/parsl/parsl/dataflow/dflow.py", line 882, in cleanup
    logger.info("DFK cleanup initiated")
Message: 'DFK cleanup initiated'
Arguments: ()
```

at the end of the `--config local` test run - although this does not cause pytest to fail with an exit code.

It looks related to this pytest issue to do with module imports setting up logging: https://github.com/pytest-dev/pytest/issues/5502

benclifford:
These `I/O operation on closed file` errors also happen on my dev laptop.

benclifford:
It looks like the error messages that I reported were caused by test cases calling `parsl.set_stream_logger()` when they were imported, resulting in log configuration to stderr/stdout that this newer version of pytest does not like.

Those calls shouldn't have been made anyway, so I've removed them / moved to __main__ blocks in 4838b8a097e0dd1c92cc88d0f186cea788e4caab and the error messages no longer happen.

benclifford:
@yadudoc please can you look at this again?

PR #1417, opened 2019-10-29
Add test for file dependency handling bugfix in #1316

benclifford:


benclifford:
With #1316 merged, I'd like to turn the SP2 case of this PR into a pytest case.

Issue #1035, opened 2019-06-10
htex/exex worker startup failure results in silent hangs


benclifford:
Often I find that parsl cannot start the workers for htex or exex, with some error that is extremely obvious and diagnosable from the stderr of the worker pool program if run manually (for example, missing MPI or programmer error in syntax/imports).

These errors aren't straightfowardly presented to the user; instead often parsl seems to hang silently.

hategan:
Outline of a partial solution that addresses the most common scenario: if no workers are running and three failures starting workers accumulate, fail all tasks queued on this executor.

yadudoc:
If the executor.scale_out fails thrice consecutively trigger a failure over all tasks on the executor. We need to do this to let the layer above make decisions.

hategan:
a small update: executor.scale_out rarely fails (e.g., it doesn't fail if process_worker_pool.py is missing), so this needs to be handled somewhere else

hategan:
The "minimum of three failures" doesn't work when only two blocks are requested. This is what happens with my test case, which has init_blocks=2.

One could move the error checking code into the executor and have the executor decide that it should be a "minimum of init_blocks" failures instead of 3.

Thoughts?

PR #1420, opened 2019-10-31
Worker error handling

hategan:
This is a first step in addressing issues like #1035 

It tries to formalize worker error handling a bit and move it outside of Strategy for now. The basic idea is that, for efficiency reasons, executors and their provider are polled regularly, based on custom poll intervals, to get the status of all workers managed by the executor. This information is then forwarded to an error handler and, possibly, to the strategy. The error handler checks for problematic situations and informs the executor when such situations are encountered. It is also possible for the executor to implement a custom error handling check.

Part of the solution involves a bit of scaffolding to better keep track of and propagate errors from providers. The exact details of how providers implement error propagation is only addressed for the local provider with a local channel. More work is needed on having all providers properly address errors in the jobs they launch.


parsl/providers/provider_base.py
```diff
@@ -1,7 +1,46 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
+from enum import Enum
 from typing import Any, List, Optional
 
 
+class JobState(bytes, Enum):
```
benclifford: I'm unclear why this isn't a plain `(Enum)` and what the `__new__` is for

benclifford: I think switching from string based job states to JobState is a useful thing to merge now, separate from the rest of this PR.

parsl/dataflow/dflow.py
```diff
@@ -167,13 +173,6 @@ def __init__(self, config=Config()):
                 logger.error("invalid checkpoint_period provided: {0} expected HH:MM:SS".format(config.checkpoint_period))
                 self._checkpoint_timer = Timer(self.checkpoint, interval=(30 * 60), name="Checkpoint")
 
-        # if we use the functionality of dynamically adding executors
-        # all executors should be managed.
-        if any([x.managed for x in config.executors]):
-            self.flowcontrol = FlowControl(self)
-        else:
-            self.flowcontrol = FlowNoControl(self)
-
```
benclifford: I think it's fine to get rid of FlowNoControl entirely like this -- the flow control/strategy stuff has to cope with a mix of managed and unmanaged executors in the FlowControl case anyway. It leads to the flow control stuff running even when running with just a local thread executor, but I don't think that is a huge price to pay for this simplification.

Getting rid of FlowNoControl could be merged independently of the rest of this PR.

parsl/executors/base.py
```diff
@@ -87,6 +88,83 @@ def scaling_enabled(self) -> bool:
...
+        should be done
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def error_management_enabled(self) -> bool:
+        """Indicates whether worker error management is supported by this executor. Worker error
+        management is done externally to the executor. However, the executor must implement
+        certain methods that allow this to function. The basic idea of worker error management
```
benclifford: what are the certain methods? Is it the _get_job_ids abstract method on StatusHandlingMixin or more?

parsl/executors/status_handling.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class StatusHandlingMixin(object):
+    tasks = None
+    provider = None
+
+    def __init__(self, provider=None):
```
benclifford: This __init__ signature makes me uncomfortable: classes need to call super().__init__(provider) and assumes that the MRO makes super().__init__() be the StatusHandlingMixin, which feels fragile to me in the presence of subclassing.

Maybe it's more reasonable to require that `self.provider`  exists, so that this code does not store a second copy in `self._provider`.

parsl/dataflow/strategy.py
```diff
@@ -189,9 +194,11 @@ def _strategy_simple(self, tasks, *args, kind=None, **kwargs):
             nodes_per_block = executor.provider.nodes_per_block
             parallelism = executor.provider.parallelism
 
-            running = sum([1 for x in status if x == 'RUNNING'])
-            submitting = sum([1 for x in status if x == 'SUBMITTING'])
-            pending = sum([1 for x in status if x == 'PENDING'])
+            running = sum([1 for x in status.values() if x.state == JobState.RUNNING])
+            # I didn't see any provider setting the status to 'SUBMITTING'
```
benclifford: Me neither. I think the only references of this value are the `active_blocks` calculation and log messages a few lines later.

It would be a reasonable separate PR to remove submitting from here.

parsl/executors/base.py
```diff
@@ -87,6 +88,83 @@ def scaling_enabled(self) -> bool:
...
+        and this method is a delegate to the corresponding method in the provider.
+
+        :return: the number of seconds to wait between calls to status() or zero if no polling
+        should be done
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def error_management_enabled(self) -> bool:
```
benclifford: A different structure for this is to make a new ABC that defines all of the error management abstract methods; classes which implement error management inherit from that, classes which do not, do not. That removes the need for NoStatusMixin and a bunch of (I think?) dummy methods there that shouldn't be called but still need implementing.

parsl/executors/high_throughput/executor.py
```diff
@@ -264,6 +265,7 @@ def initialize_scaling(self):
 
         self._scaling_enabled = True
         logger.debug("Starting HighThroughputExecutor with provider:\n%s", self.provider)
+        # TODO: why is this a provider property?
```
benclifford: prototyping history more than anything, I think

parsl/dataflow/dflow.py
hategan: Code can be added to disable it if there is nothing to poll if executors cannot be added at runtime. But the existence of add_executors() seems to suggest that this is not the intent.

parsl/executors/base.py
hategan: I'll add a note.

parsl/executors/status_handling.py
hategan: I think the reasonable solution is to have StatusHandling and NoStatusHandling be abstract classes rather than mixins. I don't see a good reason any more why they should be mixins, since they are very specific and unlikely to be used anywhere outside of an implementation of a ParslExecutor. I'll update the PR.

parsl/dataflow/strategy.py
hategan: If it's reasonable here, there is no need for additional bureaucracy.

parsl/executors/base.py
hategan: True. Feel free... :)

parsl/executors/status_handling.py
hategan: That said, it's probably wise in general if mixins were conventionally restricted to only have trivial constructors.

parsl/dataflow/strategy.py
yadudoc: state:submitting made sense if we were to use `execute_no_wait` calls on batch systems that take minutes to respond. It's obsolete now.

hategan:
Just FYI, I would like to update this PR with a couple of things mentioned by Ben as well as addressing a number of issues raised by mypy. It will not fundamentally change the structure of the solution, so feel free to review this for now.

hategan:
Well?

parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,67 @@
...
+        self._strategy = Strategy(dfk)
+        self._error_handler = JobErrorHandler()
+
+    def poll(self, tasks=None, kind=None):
+        logging.debug("Polling")
+        self._update_state()
+        self._error_handler.run(self._poll_items)
+        self._strategy.strategize(self._poll_items, tasks)
+
+    def _update_state(self):
```
yadudoc: Could be pushed into the poll method.

parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,67 @@
...
+        return self._status.__repr__()
+
+
+class TaskStatusPoller(object):
+    def __init__(self, dfk: "parsl.dataflow.dflow.DataFlowKernel"):
+        self._poll_items = []  # type: List[PollItem]
+        self._strategy = Strategy(dfk)
+        self._error_handler = JobErrorHandler()
+
+    def poll(self, tasks=None, kind=None):
```
yadudoc: What does kind do ? Doesn't seem to be used ?

parsl/executors/base.py
yadudoc: With this we'd need to touch every executor. It would be much more easier to start with just htex supporting the full status reporting feature set.

parsl/executors/high_throughput/executor.py
```diff
@@ -1,34 +1,37 @@
-from concurrent.futures import Future
-import typeguard
+"""HighThroughputExecutor builds on the Swift/T EMEWS architecture to use MPI for fast task distribution
```
yadudoc: This must've come in from a bad merge.

parsl/dataflow/job_error_handler.py
```diff
@@ -0,0 +1,49 @@
...
+
+class JobErrorHandler(object):
+    def __init__(self):
+        pass
+
+    def run(self, status: List[ExecutorStatus]):
+        for es in status:
+            self._check_irrecoverable_executor(es)
+
+    def _check_irrecoverable_executor(self, es: ExecutorStatus):
```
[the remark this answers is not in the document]
hategan: It may be worth focusing on more important stuff than this. There is no "need" for it to be either separate or not separate. So need won't help here. It's a matter of code design and the idea of keeping methods as short and clear as possible. It's not necessarily an antipattern for methods to be a bit longer, although it is an antipattern for methods to be very long (with various definitions of "very"). So official software engineering isn't providing a definite answer either.

As far as I'm concerned, from a conceptual perspective, the code checks if each executor in a list is in an irrecoverable state, and I do not see a more straightforward way of writing that than for x in list check(x).

parsl/dataflow/strategy.py
```diff
@@ -1,9 +1,12 @@
 import logging
 import time
 import math
+from typing import Any, Dict, List  # noqa F401 (used in type annotation comments)
 
+from parsl.dataflow.task_status_poller import ExecutorStatus
```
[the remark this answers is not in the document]
hategan: How on earth is that even possible? I'll fix it.

parsl/dataflow/task_status_poller.py
hategan: See above.

parsl/dataflow/task_status_poller.py
hategan: You tell me. It was there in flow_control.py.

parsl/executors/high_throughput/executor.py
hategan: There was a conflict with master, which I had to resolve by hand, and I might have added what was removed.

parsl/dataflow/job_error_handler.py
```diff
@@ -0,0 +1,49 @@
...
+        err = ""
+        count = 1
+        for js in status.values():
+            if js.message is not None:
+                err = err + "{}. {}\n".format(count, js.message)
+                count += 1
+        if len(err) == 0:
+            err = "[No error message received]"
+        # wrapping things in an exception here doesn't really help in providing more information
+        # than the string itself
```
benclifford: i think weird stuff happens if you raise the same exception object multiple times, which I think happens to this `Exception` later on (not sure)

parsl/dataflow/job_error_handler.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def run(self, status: List[ExecutorStatus]):
+        for es in status:
+            self._check_irrecoverable_executor(es)
+
+    def _check_irrecoverable_executor(self, es: ExecutorStatus):
+        if not es.executor.error_management_enabled:
+            return
+        custom_handling = es.executor.handle_errors(self, es.status)
+        if not custom_handling:
```
benclifford: maybe there's no need for a `not custom_handling` case here? The StatusHandlingExecutor code has a different default that overrides this default, and (I think?) the NoStatusHandlingExecutor shouldn't hit this code path because error_management_enabled() is false? So "if you say you want error management, you have to handle the errors"?

parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,67 @@
...
+        self._interval = executor.status_polling_interval
+        self._last_poll_time = 0.0  # type: float
+        self._status = {}  # type: Dict[Any, JobStatus]
+
+    def _should_poll(self, now: float):
+        return now >= self._last_poll_time + self._interval
+
+    def poll(self, now: float):
+        if self._should_poll(now):
+            logging.debug("Polling {}".format(self._executor))
```
benclifford: s/logging/logger/

parsl/executors/base.py
```diff
@@ -87,6 +90,98 @@ def scaling_enabled(self) -> bool:
...
+    @property
+    @abstractmethod
+    def status_polling_interval(self) -> int:
+        """Returns the interval, in seconds, at which the status method should be called. The
+        assumption here is that, once initialized, an executor's polling interval is fixed.
+        In practice, at least given the current situation, the executor uses a single task provider
+        and this method is a delegate to the corresponding method in the provider.
+
+        :return: the number of seconds to wait between calls to status() or zero if no polling
+        should be done
```
benclifford: is there a situation where no polling should be done but error_management_enabled is True?

parsl/executors/status_handling.py
```diff
@@ -0,0 +1,155 @@
...
+    def _get_job_ids(self) -> List[Any]:
+        raise NotImplementedError("Classes inheriting from StatusHandlingMixin must implement"
+                                  "_get_job_ids()")
+
+    def _fail_job_async(self, job_id: Any, message: str):
+        """Marks a job that has failed to start but would not otherwise be included in status()
+        as failed and report it in status()
+        """
+        if job_id is None:
+            job_id = "block-{}".format(self._generated_job_id_counter)
```
benclifford: put "failed-block" or something in this string so that people wondering why they get two different block ID formats later get more of a clue.

parsl/dataflow/job_error_handler.py
hategan: The original idea was that errors can be handled by a common handler with "if 3/3 blocks fail, the executor is bad". That until you realize that if init_blocks=max_blocks=2, you'll never hit 3 failed blocks.  So for executors like that, some special handling is needed. It turns out that all executors that submit "blocks" have init_blocks and max_blocks, so, for the current set of executors what you say is true. But init_blocks and max_blocks is not standardized in any way, so it would be unwise to pretend that all executors inheriting from StatusHandlingExecutors will magically have this working when the error handing in StatusHandlingExecutors works only if hasattr("init_blocks").

parsl/executors/base.py
hategan: polling is also used by strategies, so the two are separate, at least logically.

benclifford:
One thing I'm not clear of at the moment is how this will capture (or if it is even intended to capture, at the moment) batch jobs failing to launch workers: with unix process launches, there is a useful exit code which is used to set JobStatus.FAILED. With other more asynchronous stuff, I'm less clear what should happen.

parsl/dataflow/task_status_poller.py
hategan: There's inline editing. Might be quicker than comment, wait for me to see, me edit, etc.

parsl/dataflow/task_status_poller.py
benclifford: remove it then

parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,67 @@
...
+    def _update_state(self):
+        now = time.time()
+        for item in self._poll_items:
+            item.poll(now)
+
+    def add_executors(self, executors: Sequence[ParslExecutor]):
+        for executor in executors:
+            if executor.error_management_enabled and executor.status_polling_interval > 0:
+                logging.debug("Adding executor {}".format(executor))
+                self._poll_items.append(PollItem(executor))
```
benclifford: Does polling need to happen even if `not error_management_enabled` so that the strategy gets status updates? I think that strategies used to ask executors for status themselves, but it looks like now they're passed the subset of executors/statuses that match this condition.

benclifford: (specifically strategy could look at *all* executors, but now it only sees error_management_enabled+polling executors)

parsl/dataflow/task_status_poller.py
hategan: But perhaps not here.

parsl/dataflow/task_status_poller.py
hategan: Ah, true. The check is inconsistent with what I said. I can fix that.

parsl/dataflow/job_error_handler.py
benclifford: I think what I'm saying is basically in line with that: don't assume there is any kind of known-to-core-parsl sensible error behaviour and make each executor provide the sensible-for-that-executor behaviour (which it can inherit or delegate if that happens to make sense - but that is up to the executor implementation)

hategan:
> One thing I'm not clear of at the moment is how this will capture (or if it is even intended to capture, at the moment) batch jobs failing to launch workers: with unix process launches, there is a useful exit code which is used to set JobStatus.FAILED. With other more asynchronous stuff, I'm less clear what should happen.

Your comment is hard to parse. I have made a comment when I submitted this PR that this is the infrastructure and only local/localchannel implements all the necessary bits. Batch jobs failing when launched from whatnot would need all the whatnots to have some means of propagating errors. This PR has some changes to the bash launcher used by local/localchannel, so there's a little bit of updates in one whatnot. 

In other words, there is no magic. The hard work needs to be done. Or not, and we can just let it hang.

parsl/dataflow/job_error_handler.py
hategan: Except the initial design, and we can talk separately about the details of how that design was arrived at, called for a known-to-core-parsl error behavior, more in line with the way strategies work. Until it turned out that it's not quite sensible with the current executors. So a compromise solution is there. I don't particularly care, but I also don't see much of a downside in having some not-particularly-not-sensible default behavior and inheritance+delegation, as in this PR.

parsl/executors/base.py
hategan: The easiest is to start with what's there, which is this PR. Going back and changing all the executors that needed to be touched and untouching them (as well as figuring out a clean way for untouched executors to coexist with touched executors) is, as far as I'm concerned, quite clearly not easier.

parsl/dataflow/task_status_poller.py
hategan: see 3371373

parsl/dataflow/job_error_handler.py
```diff
@@ -0,0 +1,49 @@
...
+            self.simple_error_handler(es.executor, es.status, 3)
+
+    def simple_error_handler(self, executor: ParslExecutor, status: Dict[Any, JobStatus], threshold: int):
+        (total_jobs, failed_jobs) = self.count_jobs(status)
+        if total_jobs >= threshold and failed_jobs == total_jobs:
```
yadudoc: I worry that this test might fail if there are pending jobs in the queue. Since we can differentiate failed from completed with no errors, we could fail based on just `failed_jobs >= threshold`.

parsl/providers/provider_base.py
```diff
@@ -48,7 +87,7 @@ def submit(self, command: str, tasks_per_node: int, job_name: str = "parsl.auto"
         pass
 
     @abstractmethod
-    def status(self, job_ids: List[Any]) -> List[str]:
+    def status(self, job_ids: List[Any]) -> List[JobStatus]:
```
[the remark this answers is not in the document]
hategan: Who's "we"?

ParslExector.status() does that.

It's fine here as a List because we can guarantee that it matches job_ids. Alternatively, we could have JobStatus have a job_id property.

The reason the executor is different is because, given the following hypothetical:

ids:List[str] = executor.get_job_ids()
states: List[JobStatus] = executor.status(),

it cannot be guaranteed that states matches ids since the job list internal to the executor may change. There is the additional issue that the job list may be constructed by the executor from a dict, which may or may not have predictable iteration order. Very recent versions of Python guarantee insertion order and slightly older versions of Python use insertion order, but do not officially guarantee it.

parsl/providers/lsf/lsf.py
```diff
@@ -120,15 +121,15 @@ def _status(self):
...
+                state = translate_table.get(parts[2], JobState.UNKNOWN)
+                self.resources[job_id]['status'] = JobStatus(state)
                 jobs_missing.remove(job_id)
 
         # squeue does not report on jobs that are not running. So we are filling in the
         # blanks for missing jobs, we might lose some information about why the jobs failed.
         for missing_job in jobs_missing:
-            if self.resources[missing_job]['status'] in ['PENDING', 'RUNNING']:
-                self.resources[missing_job]['status'] = 'COMPLETED'
+            # if self.resources[missing_job]['status'] in ['PENDING', 'RUNNING']:
```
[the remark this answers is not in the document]
hategan: That opens a can of worms, since you can never guarantee that a FAILED job will not eventually complete. So you're setting it up for hanging.

You might want to consider having a separate job, which is effectively what happens if "job" means an attempt to run something, or having a FAILED_BUT_WILL_RETRY state.

parsl/providers/local/local.py
```diff
@@ -90,20 +91,34 @@ def status(self, job_ids):
...
+                poll_code = job_dict['proc'].poll()
+
+                if poll_code is not None:
+                    out, err = job_dict['proc'].communicate(input=None)
+                    if out:
+                        job_dict['out'] += out.decode()
+                    if err:
+                        job_dict['err'] += err.decode()
+
+                print("Poll code: %s" % poll_code)
```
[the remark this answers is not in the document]
hategan: You might want to look at the latest version of that file.

parsl/providers/local/local.py
```diff
@@ -90,20 +91,34 @@ def status(self, job_ids):
...
+                result.append(status)
+                continue
+
+            if job_dict['proc']:
+                poll_code = job_dict['proc'].poll()
+
+                if poll_code is not None:
+                    out, err = job_dict['proc'].communicate(input=None)
+                    if out:
+                        job_dict['out'] += out.decode()
```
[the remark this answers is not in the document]
hategan: Isn't that a problem with workers?

parsl/launchers/launchers.py
```diff
@@ -34,30 +34,68 @@ class SingleNodeLauncher(Launcher):
...
     of the command to be launched can be controlled.
     """
-    def __call__(self, command, tasks_per_node, nodes_per_block):
+    def __call__(self, command, tasks_per_node, nodes_per_block, fail_on_any=False):
         """
         Args:
         - command (string): The command string to be launched
         - task_block (string) : bash evaluated string.
+        - fail_on_any: If True, return a nonzero exit code if any worker failed, otherwise zero;
+                       if False, return a nonzero exit code if all workers failed, otherwise zero.
```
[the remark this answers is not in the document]
hategan: "Usually", things don't hang. We're here because, sometimes, they do.

parsl/executors/status_handling.py
```diff
@@ -0,0 +1,137 @@
...
+    tasks = None
+    provider = None
+
+    def __init__(self, provider=None):
+        super().__init__()
+        self._provider = provider
+        # errors can happen during the sumbit call to the provider; this is used
+        # to keep track of such errors so that they can be handled in one place
+        # together with errors reported by status()
+        self._simulated_status = {}
```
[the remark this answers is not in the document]
hategan: _job_fail_status would be misleading, no? It would suggest that it contains the status of all failed jobs. Perhaps _simulated_failures. Or maybe it doesn't matter, it being clearly documented in the comment above.

parsl/dataflow/job_error_handler.py
hategan: This will not match if there are pending jobs in the queue. Do we want it to?

failed_jobs >= threshold would match eventually when you have mostly successful jobs and rare occasional failures. I don't think we want that. So maybe failed_job >= threshold and completed_jobs == 0 and running_jobs == 0.

parsl/dataflow/job_error_handler.py
hategan: Right. That was sketchy even if bad stuff doesn't happen. See 2a6add8.

parsl/dataflow/job_error_handler.py
yadudoc: Since we don't know when pending jobs in the queue will be filled, we shouldn't have to wait when we know that `failed_jobs >= threshold`. We shouldn't need `complete_jobs == 0` and limit to cases where no progress is made, when it could be that say the allocation / quotes have been exceeded.

parsl/dataflow/strategy.py
yadudoc: Marking this as resolved.

parsl/executors/high_throughput/executor.py
yadudoc: This comment could be removed at this point.

parsl/launchers/launchers.py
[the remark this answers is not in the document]
yadudoc: If at least 1 launched command ie manager is running, the job should not exit. The decision to terminate a job because the not all managers are running is upto strategy/executor, not the provider.

parsl/dataflow/job_error_handler.py
hategan: See above: "failed_jobs >= threshold would match eventually when you have mostly successful jobs and rare occasional failures."

In other words, it is possible for failures to accumulate over time, even if the executor is mostly successful at running things.

parsl/launchers/launchers.py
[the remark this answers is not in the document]
hategan: Neither of those options cause the launcher to exit before all the launched commands complete. It is simply a question of what constitutes a  failure when multiple commands are involved.

hategan:
@benclifford, do you mind going through conversations and resolving if needed for merging and addressed?

benclifford:
> @benclifford, do you mind going through conversations and resolving if needed for merging and addressed?

sure - I've already indicated which pieces should be merged already as separate PRs though.

hategan:
> > @benclifford, do you mind going through conversations and resolving if needed for merging and addressed?
> 
> sure - I've already indicated which pieces should be merged already as separate PRs though.

I always appreciated the direct nature of your statements. What you are doing here is the opposite. If you think this PR is unmergeable, please state so.

benclifford:
It's got too much going on for me to coherently review without splitting it up, so you're going to have to wait a lot longer for an approving review if you wait for me to split it up.

hategan:
> It's got too much going on for me to coherently review without splitting it up, so you're going to have to wait a lot longer for an approving review if you wait for me to split it up.

Ok, so you mentioned two separate PRs: JobState and ControlNoFlow. You are saying I should manually extract all the changes related to those issues and create two separate branches and PRs. Presumably those would be merged quickly, after which I should manually resolve conflicts in this branch. And that you believe this work would significantly reduce the time it would take you to review the remaining changes. Is this correct?

benclifford:
yes

hategan:
> yes

OK

hategan:
After banging my head against stuff, it looks like what you suggest would involve significantly more work than mentally separating things and would be quite error prone. It would essentially require rewriting the whole code in this PR such that the above commits end up in a particular order. For example, all executors use JobStatus in methods that have been moved to an abstract class. Going back requires plugging the JobStatus stuff into every executor. It would, if I started entirely from scratch, require more work than I spent on this PR because many of those things got abstracted concurrently, without me changing them first to use the new class.

In addition, both the suggested separate PRs are conceptually trivial and do not make much sense by themselves.

So I would rather not invest significant time into something that is likely to be broken and mostly useless. If you would like, let me know when you have some time and we can skype/etc. and go through each change and discuss. I can than identify for you whether a particular change would have belonged in one of these hypothetical separate PRs.

benclifford:
I think parsl/providers/ad_hoc/ad_hoc.py:AdHocProvider use JobState/JobStatus?

benclifford:
I started unpicking this into a sequence of commits that I'm more happy reviewing with but that ends up with bitwise identical to this source tree. That will probably lead to more review comments as I pay closer attention. I'll push up the commit series to a branch in a day or so when I've made my full first pass.

hategan:
> I think parsl/providers/ad_hoc/ad_hoc.py:AdHocProvider use JobState/JobStatus?

True, true; 95e0159

parsl/providers/slurm/slurm.py
```diff
@@ -237,17 +238,21 @@ def cancel(self, job_ids):
         rets = None
         if retcode == 0:
             for jid in job_ids:
-                self.resources[jid]['status'] = translate_table['CA']  # Setting state to cancelled
+                self.resources[jid]['status'] = JobStatus(JobState.CANCELED)  # Setting state to cancelled
```
benclifford: should be two Ls in this CANCELED constant?

hategan: Right. Fixed by e117128. I'm a bit surprised that typechecking doesn't catch this. Kinda defeats the purpose of enums.

benclifford: At least in parsl CI, mypy is configured to only typecheck functions which have type signatures (that is mypy default behaviour, and the parsl codebase isn't ready for anything more aggressive).

I added this mostly-no-op type signature to cancel in the slurm provider:

```
-    def cancel(self, job_ids):
+    def cancel(self, job_ids) -> Any:
```

on the commit right before e117128 and mypy does detect it then:

> parsl/providers/slurm/slurm.py:241: error: "Type[JobState]" has no attribute "CANCELED"

I guess the moral of the story is to press on with making type checking deployed more throughout the codebase?

hategan: Right. I'm not quite sure why adding type annotations is such a piecewise effort. It seems a bit wasteful to me.

yadudoc:
@hategan How much of the PR is yet to be sliced and merged to master ? Just checking if there's value in keeping this around.

hategan:
Not all of it is merged. I don't think it matters how much beyond that.

PR #1422, opened 2019-11-01
Add address heuristics

yadudoc:
This PR will ignore the existing `HighThroughputExecutor(address=X)` option.

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -62,15 +62,17 @@ def get_all_addresses():
         try:
             s_addresses.append(address_by_interface(interface))
         except Exception:
+            logger.exception("Ignoring failure to fetch address from interface {}".format(interface))
             pass
 
     s_addresses = set(s_addresses)
 
     try:
+        s_addresses.add(address_by_hostname())
         s_addresses.add(address_by_route())
         s_addresses.add(address_by_query())
-        s_addresses.add(address_by_hostname())
     except Exception:
+        logger.exception("Ignoring one or more address finder method failure")
         pass
 
     return s_addresses
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -5,6 +5,7 @@ import requests
 import socket
 import fcntl
 import struct
+import psutil
 
 logger = logging.getLogger(__name__)
 
@@ -46,3 +47,30 @@ def address_by_interface(ifname):
         0x8915,  # SIOCGIFADDR
         struct.pack('256s', bytes(ifname[:15], 'utf-8'))
     )[20:24])
+
+
+def get_all_addresses():
+    """ Uses a combination of methods to determine possible addresses.
+
+    Returns:
+         list of addresses as strings
+    """
+    net_interfaces = psutil.net_if_addrs()
+
+    s_addresses = []
+    for interface in net_interfaces:
+        try:
+            s_addresses.append(address_by_interface(interface))
+        except Exception:
+            pass
+
+    s_addresses = set(s_addresses)
+
+    try:
+        s_addresses.add(address_by_route())
+        s_addresses.add(address_by_query())
+        s_addresses.add(address_by_hostname())
+    except Exception:
+        pass
+
+    return s_addresses
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -21,6 +21,7 @@ from parsl.executors.errors import BadMessage, ScalingFailed, DeserializationErr
 from parsl.executors.base import ParslExecutor
 from parsl.providers.provider_base import ExecutionProvider
 from parsl.data_provider.staging import Staging
+from parsl.addresses import get_all_addresses
 
 from parsl.utils import RepresentationMixin
 from parsl.providers import LocalProvider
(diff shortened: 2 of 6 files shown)


parsl/addresses.py
```diff
@@ -46,3 +47,30 @@ def address_by_interface(ifname):
...
+            pass
+
+    s_addresses = set(s_addresses)
+
+    try:
+        s_addresses.add(address_by_route())
+        s_addresses.add(address_by_query())
+        s_addresses.add(address_by_hostname())
+    except Exception:
+        pass
```
benclifford: at least emit a warning here

parsl/addresses.py
```diff
@@ -46,3 +47,30 @@ def address_by_interface(ifname):
...
+         list of addresses as strings
+    """
+    net_interfaces = psutil.net_if_addrs()
+
+    s_addresses = []
+    for interface in net_interfaces:
+        try:
+            s_addresses.append(address_by_interface(interface))
+        except Exception:
+            pass
```
benclifford: at least emit a warning here

parsl/executors/high_throughput/probe.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns
+    -------
+    None or string address
+    """
+    context = zmq.Context()
+    addr_map = {}
+    for addr in addresses:
+        socket = context.socket(zmq.DEALER)
+        # socket.setsockopt(zmq.LINGER, 0)
+        url = f"tcp://{addr}:{task_port}"
```
benclifford: no f-strings in python 3.5

parsl/executors/high_throughput/probe.py
```diff
@@ -0,0 +1,93 @@
...
+        print("Trying url: ", url)
+        socket.connect(url)
```
benclifford: does this wait for a tcp connection to be established/rejected before returning? if so, when the traffic is dropped rather than rejected (which is a fairly common firewall configuration on the internet) then this will maybe hang a-while?

parsl/executors/high_throughput/probe.py
```diff
@@ -0,0 +1,93 @@
...
+    -------
+    None or string address
+    """
+    context = zmq.Context()
+    addr_map = {}
+    for addr in addresses:
+        socket = context.socket(zmq.DEALER)
+        # socket.setsockopt(zmq.LINGER, 0)
+        url = f"tcp://{addr}:{task_port}"
+        print("Trying url: ", url)
```
benclifford: use logging

parsl/executors/high_throughput/probe.py
yadudoc: This is a zmq connect, and will return almost immediately. In fact it will buffer sends until a server is available remotely to receive.

parsl/executors/high_throughput/executor.py
```diff
@@ -188,6 +189,7 @@ def __init__(self,
         self.mem_per_worker = mem_per_worker
         self.max_workers = max_workers
         self.prefetch_capacity = prefetch_capacity
+        self.all_addresses = ','.join(get_all_addresses())
```
benclifford: respect the address parameter still - almost definitely som people are going to need to override this behavior.

if address is specified, don't do any get_all_addresses magic at all

yadudoc: This is a bit that I wanted to ask folks for comments. I'm leaning towards `address=None` will be the default and that implies we try auto-detecting vs, `address=X` means we will use only the user's recommendation.

parsl/addresses.py
yadudoc: Fixed.

parsl/addresses.py
yadudoc: Fixed.

parsl/executors/high_throughput/executor.py
benclifford: i think that is fine.

yadudoc: Fixed.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -107,6 +109,20 @@ def __init__(self,
...
+                logger.error("Unable to connect back on any addresses:{}".format(addresses))
+                exit(5)
+            else:
+                logger.info("Connection to Interchange successful on {}".format(ix_address))
+                task_q_url = "tcp://{}:{}".format(ix_address, task_port)
+                result_q_url = "tcp://{}:{}".format(ix_address, result_port)
+                logger.info("Task url : {}".format(task_q_url))
+                logger.info("Result url : {}".format(result_q_url))
+        except Exception:
+            logger.exception("Caught exception while trying to determine viable address to interchange")
```
benclifford: should the process worker terminate now if this exception path happens? (because the iniitalisations haven't necessarily all been run correctly)

yadudoc: Right, I think we should terminate here if the probe fails.

benclifford: ok. other than this, this PR is ready to be approved

Issue #1416, opened 2019-10-29
SlurmProvider mem_per_node passes incorrectly formatted memory string


mattwelborn:
**Describe the bug**
When specifying the mem_per_node option to SlurmProvider, the submission script requests memory with "#SBATCH --mem=120.0g" (rather than "#SBATCH --mem=120g"), which is invalid in SLURM version 17.11.8. The error in the parsl log is:

```
2019-10-29 17:26:10.215 parsl.providers.slurm.slurm:185 [DEBUG]  Requesting one block with 1 nodes
2019-10-29 17:26:10.215 parsl.providers.slurm.slurm:202 [DEBUG]  Writing submit script
2019-10-29 17:26:10.218 parsl.providers.slurm.slurm:206 [DEBUG]  moving files
2019-10-29 17:26:10.244 parsl.providers.slurm.slurm:222 [ERROR]  Retcode:1 STDOUT: STDERR:sbatch: error: invalid memory constraint 120.0g
2019-10-29 17:26:10.244 parsl.executors.high_throughput.executor:569 [DEBUG]  Launched block 0->None
2019-10-29 17:26:10.244 parsl.dataflow.flow_control:142 [ERROR]  Flow control callback threw an exception - logging and proceeding anyway
Traceback (most recent call last):
  File "/home/mwelborn/miniconda3/envs/parsl9_dev/lib/python3.6/site-packages/parsl/dataflow/flow_control.py", line 140, in make_callback
    self.callback(tasks=self._event_buffer, kind=kind)
  File "/home/mwelborn/miniconda3/envs/parsl9_dev/lib/python3.6/site-packages/parsl/dataflow/strategy.py", line 260, in _strategy_simple
    executor.scale_out(excess_blocks)
  File "/home/mwelborn/miniconda3/envs/parsl9_dev/lib/python3.6/site-packages/parsl/executors/high_throughput/executor.py", line 572, in scale_out
    "Attempts to provision nodes via provider has failed"))
parsl.executors.errors.ScalingFailed: ('slurm', 'Attempts to provision nodes via provider has failed'
```

**To Reproduce**
Steps to reproduce the behavior, for e.g:
1. Setup Parsl 0.8.0 with Python 3.6 on a cluster with slurm 17.11.8
2. Create a SlurmProvider with mem_per_node specified
3. Create tasks

**Expected behavior**
Parsl should complete work without incident. 

**Environment**
 - OS: Linux x86_64
 - Python 3.6.7
[...]

PR #1423, opened 2019-11-05
Add type signatures for slurm initializer

benclifford:
Part of #873


benclifford:
This is part of long running issue #873 to add type signatures to user facing code.

This PR also forces mem_per_node to be an integer, because slurm cannot deal with fractions in this field - see issue #1416

PR #1446, opened 2019-11-14
Guarantee monitoring will send out at least one message

### LINKED ISSUE #1011: monitoring does not always record task_time_running and hostname fields (on short jobs?)
I see missing hostname and task_time_running for the first task, `create_empty_repo`, which runs extremely fast.

```
sqlite> select run_id, task_func_name, hostname, task_time_submitted, task_time_running, task_time_returned from task;
37cb5ae7-f4f9-40a7-86ec-9457760344f7|create_empty_repo||2019-06-04 04:13:31.476801||2019-06-04 04:13:31.553611
37cb5ae7-f4f9-40a7-86ec-9457760344f7|install_transmission_curves|cori06|2019-06-04 04:13:31.554481|2019-06-04 04:13:31.644297|2019-06-04 04:13:36.155113
37cb5ae7-f4f9-40a7-86ec-9457760344f7|import_ci_hsc|cori06|2019-06-04 04:13:31.560238|2019-06-04 04:13:31.650255|2019-06-04 04:18:30.889769
37cb5ae7-f4f9-40a7-86ec-9457760344f7|pccd_show|cori06|2019-06-04 04:18:30.891500|2019-06-04 04:18:30.989590|2019-06-04 04:18:49.810345
37cb5ae7-f4f9-40a7-86ec-9457760344f7|pccd_process|cori06|2019-06-04 04:18:30.898264|2019-06-04 04:18:31.031377|2019-06-04 05:26:59.966683
37cb5ae7-f4f9-40a7-86ec-9457760344f7|makeDiscreteSkyMap|cori06|2019-06-04 05:26:59.968017|2019-06-04 05:27:00.054919|2019-06-04 05:27:08.146285
```

ZhuozhaoLi:
Multiple entries of `task` and `status` tables (e.g., `hostname` and `task_running_time`) rely on the first message from monitoring.

However, for extremely short task, monitoring does not have enough time to send out the first message. This PR addresses this issue by guaranteeing monitoring will send out at least one message.

Fix #1011


parsl/monitoring/monitoring.py
```diff
@@ -280,17 +280,22 @@ def monitor_wrapper(f,
...
                               run_id,
+                              command_q,
                               logging_level,
                               sleep_dur),
                         name="Monitor-Wrapper-{}".format(task_id))
             p.start()
             try:
-                return f(*args, **kwargs)
+                res = f(*args, **kwargs)
+                command_q.put("Finished")
```
benclifford: If `f` throws an exception, then this added sequence does not happen, and the `finally p.terminate()` happens.

That means that a function which fails very fast (which is a common failure mode) will still not be reported.

benclifford: I'm also a little concerned about `put` blocking if the monitoring process no longer exists (for example, if ctrl-C has been pressed and killed some of the processes). I haven't investigated this in depth, but https://docs.python.org/3/library/multiprocessing.html talks about some potential deadlocks.

ZhuozhaoLi: > If `f` throws an exception, then this added sequence does not happen, and the `finally p.terminate()` happens.
> 
> That means that a function which fails very fast (which is a common failure mode) will still not be reported.

I tested the situation that `f` fails by adding `return 1 / 0` before `command_q.put("Finished")`. The added sequence did not happen like what you expected. 

So my new change is to add an `except` block that also does the sequence and raises the exception. 

I did not put the sequence in the `finally`  block, because I think the `p.terminate()` is useful to ensure the monitoring process will exit anyway.

benclifford: I think it is fine to put it in the finally block. Maybe even better to.

The code as it is now will try the "except put/join" if f completes successfully but the first put/join sequence raises an exception.

Putting the put/join code in `finally`, with only the invocation of `f` in the try body, will behave differently: it will catch exceptions from invoking `f`, but will not catch exceptions from invoking put/join.

I think it is better to keep the try block as tightly focused around the invocation of `f` as possible.

To keep the p.terminate() in there even if put/join throws an exception, wrap the whole thing in a new `try/finally` block.

ZhuozhaoLi: Fixed now

parsl/monitoring/monitoring.py
```diff
@@ -525,5 +533,13 @@ def monitor(pid,
         except Exception:
             logging.exception("Exception getting the resource usage. Not sending usage to Hub", exc_info=True)
 
+        try:
+            msg = command_q.get(block=False)
+            if msg == "Finished":
+                logging.info("Received task finished message. Ending the monitoring loop now.")
+                break
+        except Queue.empty:
```
benclifford: @ZhuozhaoLi should this be `queue.Empty`? (different capitalization)

ZhuozhaoLi: You are right. I fixed this in the new PR #1482

PR #1450, opened 2019-11-15
set monitoring zmq channel linger options to 0 to prevent shutdown delay

### LINKED ISSUE #1438: 1 second workflow takes 31 seconds with monitoring enabled
**Describe the bug**
The below workflow takes 31 seconds to run with monitoring enabled. With monitoring disabled, it takes 1 second.

**To Reproduce**
Run this code with master = ea8333c3249fdf3b21691b7a66fb4e9c8ed37ec0

```
import parsl

from parsl.monitoring.monitoring import MonitoringHub

config.monitoring=MonitoringHub(
       hub_port=55055,
   )

parsl.load(config)

@parsl.python_app()
def foo():
    return 7

print(foo().result())
```

```
time python md.py 
7

real    0m31.128s
user    0m1.000s
sys     0m0.260s
```

**Expected behavior**
The workflow should run much closer to the runtime with monitoring disabled:

```
time python md.py 
7

real    0m0.829s
user    0m0.692s
sys     0m0.124s
```

**Environment**
my laptop, python 3.7, parsl master ea8333c3249fdf3b21691b7a66fb4e9c8ed37ec0

ZhuozhaoLi:
As reported in #1438 , with monitoring enabled, it took 31s for a very short workflow.
This is because we recently changed the monitoring closing mechanism in PR #1403 .
But it took a long time for the `Hub` process to join because of the long zmq context termination.

This PR sets the zmq lingering option to 0, so the context will terminate immediately.

Fixes #1438

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -370,11 +370,13 @@ class Hub(object):
 
         self._context = zmq.Context()
         self.dfk_channel = self._context.socket(zmq.DEALER)
+        self.dfk_channel.setsockopt(zmq.LINGER, 0)
         self.dfk_channel.set_hwm(0)
         self.dfk_channel.RCVTIMEO = int(self.loop_freq)  # in milliseconds
         self.dfk_channel.connect("tcp://{}:{}".format(client_address, client_port))
 
         self.ic_channel = self._context.socket(zmq.DEALER)
+        self.ic_channel.setsockopt(zmq.LINGER, 0)
         self.ic_channel.set_hwm(0)
         self.ic_channel.RCVTIMEO = int(self.loop_freq)  # in milliseconds
         self.logger.debug("hub_address: {}. hub_port_range {}".format(hub_address, hub_port_range))


benclifford:
From reading about LINGER, I think this means that:
 * there is a message that was taking 30 seconds to deliver that is now being discarded? 
or
 * LINGER was being set to 30 seconds somewhere
?

benclifford:
In my laptop test of the code in #1438, this reduces the run time from 31s to 4s.

ZhuozhaoLi:
> In my laptop test of the code in #1438, this reduces the run time from 31s to 4s.

This is because I put a timeout of 3s at the `Hub`, so that it has enough time receive all messages.

ZhuozhaoLi:
> From reading about LINGER, I think this means that:
> 
> * there is a message that was taking 30 seconds to deliver that is now being discarded?
>   or
> * LINGER was being set to 30 seconds somewhere
>   ?

The `Hub` will never send messages, but will only receive messages. But upon the close, there should not be any messages coming in.

I think the default of linger is `-1`, which means infinite .
http://api.zeromq.org/2-1:zmq-setsockopt

benclifford:
so where does the 30s come from? (rather than any other number)

ZhuozhaoLi:
> so where does the 30s come from? (rather than any other number)

Removing the linger of `ic_channel` still leads to a run time of ~4s. So that is the problem of the `dfk_channel`. I do not see any monitoring message missing with linger set  to 0. 

Was trying to reproduce with a local zmq script, but unfortunately did not succeed. So I have not found out what happens here.

PR #1459, opened 2019-11-22
Change parsl.auto filenames to component based names

benclifford:
This makes it easier to understand where parsl.auto files
are coming from, and what might be expected in them. This
has proved useful in some debugging situations.


ZhuozhaoLi:
It seems `torque` provider is missing.

benclifford:
> It seems `torque` provider is missing.

fixed - now the only occurence should be in provider_base:

```
$ git grep 'parsl\.auto'
parsl/providers/provider_base.py:    def submit(self, command: str, tasks_per_node: int, job_name: str = "parsl.auto") -> Any:
```

Issue #1448, opened 2019-11-15
A plan for getting rid of LocalChannel special-casing


benclifford:
LocalProvider and AdHocProvider have special casing in them for LocalChannel.

This has resulted in broken code paths in LocalChannel (for example, execute_wait is broken, because it is not used in the special case - even though this method is the core method of a channel)

It has resulted in bugs in the non-special-case paths, because that code path is used less - for example #1318 

Based on some quick fixes I made for imsim, I propose eliminating the LocalChannel special-casing in `master`:

- [x] fix LocalProvider to work with remote channels properly:
  - [x] process killing is broken - PR #1447
  - [x] status checking is broken - PR #1449 
- [x] fix LocalProvider to not assume the Channel command line can be `bash` - specifically, LocalChannel execute_wait uses /bin/sh inside python `Popen` - PR #1452 
- [x] add a simple test for AdHocProvider - PR #1460 (this revealed brokenness somewhere which has been fixed)
- [x] fix LocalChannel to use new process groups for `execute_wait` like it does for `execute_no_wait` - PR #1461 
- [x] remove LocalProvider and AdHocProvider special casing for LocalChannel so that they are channel agnostic and the remote provider code-path is used for all channels  - PR #1471 
- [ ] remove now-unused `execute_no_wait` from all Channels - by this point, it is unused - PR #1629

PR #1461, opened 2019-11-22
Make LocalChannel execute_wait set pgid like execute_no_wait

benclifford:
execute_no_wait sets the process group of launched commands so that
local and ad-hoc providers can kill those launched commands later by
killing the whole process group.

Before this commit, execute_wait did not do that, so execute_wait could
not be used in the same way - instead, processes were left in the same
process group as the parent, and local/adhoc providers kill the entire
parent process group (for example, the entire containing workflow).

This commit also amends the documentation for the Channel base class to
clarify this requirement.

This is part of the plan in issue #1448

### DIFF
--- a/parsl/channels/local/local.py
+++ b/parsl/channels/local/local.py
@@ -65,7 +65,8 @@ class LocalChannel(Channel, RepresentationMixin):
                 stderr=subprocess.PIPE,
                 cwd=self.userhome,
                 env=current_env,
-                shell=True
+                shell=True,
+                preexec_fn=os.setpgrp
             )
             proc.wait(timeout=walltime)
             stdout = proc.stdout.read()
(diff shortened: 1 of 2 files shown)


parsl/channels/local/local.py
```diff
@@ -65,7 +65,8 @@ def execute_wait(self, cmd, walltime=None, envs={}):
                 stderr=subprocess.PIPE,
                 cwd=self.userhome,
                 env=current_env,
-                shell=True
+                shell=True,
+                preexec_fn=os.setpgrp
```
yadudoc: The docs indicate that `preexec_fn` mechanism not threadsafe and recommends using the `start_new_session=True` instead. Perhaps that isn't important right now, since making the change also requires updating the kill mechanism to use sessions instead.

benclifford: This should have around the same level (good or bad) of thread safety as using it in execute_no_wait. I'll open an issue noting this.

yadudoc: Agreed. Merging now.

PR #1465, opened 2019-11-22
Remove unused translate_table in pbspro provider

benclifford:
Status translations are done by the superclass TorqueProvider, using
the translations in parsl.providers.torque.torque.translate_table

That table is identical to the table being removed.


benclifford:
@annawoodard can you test this?

This PR comes from code review of #1420

PR #1466, opened 2019-11-22
Make launcher scripts fail on error. This will only be useful if/when…

hategan:
… launcher failures are reported back to the user.

There are two types of errors that this will catch: errors in the assumptions of the launcher script (such as filesystem errors while writing sub-scripts or missing mpirun/srun/parallel/etc.) and application/block errors if the launching command propagates them. Both are useful to be reported. Moreover, from the school of thought of "all errors should be appropriately handled or reported", this is a must.


benclifford:
I think this is the right thing to do.

I'm also interested in what the error handling behaviour is when launchers write out the launched command into a script that is then run via an intermediate command, such as `srun` in the SrunLauncher. Maybe there's opportunity there to propagate failures there easily too: does a failure inside that launched script propagate through srun? I think probably not without explicitly turning it on in the enclosed bash script, but I'm unsure.

hategan:
> I think this is the right thing to do.
> 
> I'm also interested in what the error handling behaviour is when launchers write out the launched command into a script that is then run via an intermediate command, such as `srun` in the SrunLauncher. Maybe there's opportunity there to propagate failures there easily too: does a failure inside that launched script propagate through srun? I think probably not without explicitly turning it on in the enclosed bash script, but I'm unsure.

That's "and application/block errors if the launching command propagates them". I've tested Local + GNU parallel, and that does this propagation. The others need testing. When launched though queuing systems, this will need additional work to ensure that useful error messages are propagated instead of just a boolean success/failure.

Anyway, one step at a time. What is clear is that, without this, errors in the corresponding scripts will go undetected. So, for comprehensive error propagation, this is necessary.

yadudoc:
We want to either fail very early, or ignore partial failures later in production runs. If this is going to cause blocks to exit if any of the node managers launched by the launcher say srun exits, we will need to make sure that this behavior is optional. In production runs, we often to indicate to the launching command to ignore failures if any launched node/program fails. I worry that in cases where we may invoke several launcher commands in sequence to fill the block, failure in one launch might kill the whole block (eg SrunMPILauncher).

hategan:
> We want to either fail very early, or ignore partial failures later in production runs. If this is going to cause blocks to exit if any of the node managers launched by the launcher say srun exits, we will need to make sure that this behavior is optional. In production runs, we often to indicate to the launching command to ignore failures if any launched node/program fails. I worry that in cases where we may invoke several launcher commands in sequence to fill the block, failure in one launch might kill the whole block (eg SrunMPILauncher).

Neither background tasks or wait cause the script to stop when -e is set. I would suggest trying this on your favorite machine:
```
#!/bin/bash
set -e

/bin/true &
/bin/false &
/bin/doesnotexist &

wait

echo "Done"
```

yadudoc:
Okay great. Let's merge this.

PR #1474, opened 2019-11-28
Remove unused zmq_plumbing module

benclifford:


yadudoc:
If only there was a quick way to find dead code.

benclifford:
This came out of me looking at coverage testing results

PR #1476, opened 2019-11-30
Correct links to source code

WardLT:
Getting 404s after clicking "source" in the rtds has been bugging me for awhile but I'd been too lazy to fix it

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,7 +53,7 @@ def linkcode_resolve(domain, info):
     if not info['module']:
         return None
     filename = info['module'].replace('.', '/')
-    return "http://github.com/Parsl/parsl/blob/master/{}.py".format(filename)
+    return "http://github.com/Parsl/parsl/tree/master/{}.py".format(filename)
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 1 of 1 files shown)


benclifford:
Which URLs are you trying to fix?

https://github.com/Parsl/parsl/blob/master/parsl/data_provider/globus.py works as is, for example.

The source link for `set_stream_logger` doesn't work either with `blob` or `tree` because the code generator thinks it should be in this file:

https://github.com/Parsl/parsl/blob/master/parsl.py

Which links are fixed by this PR?

WardLT:
Sorry, I did not put enough description or thought into this PR. 

The kind of links I'm trying to fix are similar to the `[source]` link on [this page](https://parsl.readthedocs.io/en/stable/stubs/parsl.providers.CondorProvider.html). The fix I propose here doesn't fix that problem. 

Should I open an issue about these links?

benclifford:
open an issue.

it looks like the links are broken roughly in the case that the symbols are defined in an __init__.py - in which case the links link to `directoryname.py`, rather than `directoryname/__init__.py`. Although often in those cases they are re-exports from a different location which would  be a more appropriate source location.

PR #1479, opened 2019-12-02
Exit htex probe loop with first working address

benclifford:
Prior to this commit, the probe will run for the entire timeout
period. This results in slow worker startup if the timeout is
manually changed to something much longer.


benclifford:
failing in CI because mypy 0.750 is out and is stricter on the codebase - not because of this particular PR

PR #1486, opened 2019-12-04
Fix stderr stdout format issue in monitoring

ZhuozhaoLi:
This PR is a portion of the fixes mentioned in #1267 , including

(1) Move a function that extract std file name and mode for bash apps to parsl/utils.py.
I also added another check for the types of fname and mode.
(2) In dfk, use the function above to extract the correct fname and mode, and concatenate them with `;`. This info is for monitoring.


parsl/app/bash.py
```diff
@@ -60,15 +61,11 @@ def open_std_fd(fdname):
...
-        elif isinstance(stdfspec, tuple):
-            if len(stdfspec) != 2:
-                raise pe.BadStdStreamFile("std descriptor %s has incorrect tuple length %s" % (fdname, len(stdfspec)), TypeError('Bad Tuple Length'))
-            fname, mode = stdfspec
-        else:
-            raise pe.BadStdStreamFile("std descriptor %s has unexpected type %s" % (fdname, str(type(stdfspec))), TypeError('Bad Tuple Type'))
+
+        try:
+            fname, mode = get_std_fname_mode(fdname, stdfspec)
+        except pe.BadStdStreamFile:
```
benclifford: does this `except` need to be here?

If get_std_fname_mode does not raise an exception, then this block doesn't fire.
If get_std_fname_mode raises a different exception, the exception propagates upwards.
If get_std_fname_mode raises a BadStdStreamFile exception, then this block propagates it upwards, which would be the behaviour anyway

parsl/dataflow/dflow.py
```diff
@@ -197,8 +197,20 @@ def _create_task_log_info(self, task_id, fail_mode=None):
...
         task_log_info['task_outputs'] = str(self.tasks[task_id]['kwargs'].get('outputs', None))
         task_log_info['task_stdin'] = self.tasks[task_id]['kwargs'].get('stdin', None)
-        task_log_info['task_stdout'] = self.tasks[task_id]['kwargs'].get('stdout', None)
-        task_log_info['task_stderr'] = self.tasks[task_id]['kwargs'].get('stderr', None)
+        stdout_spec = self.tasks[task_id]['kwargs'].get('stdout', None)
+        stderr_spec = self.tasks[task_id]['kwargs'].get('stderr', None)
+        try:
+            stdout_name, stdout_mode = get_std_fname_mode('stdout', stdout_spec)
+        except Exception as e:
+            stdout_name, stdout_mode = str(e), None
```
benclifford: this seems like a bad thing to do with an exception. Maybe it should just be allowed to propagate upwards?

ZhuozhaoLi: I just think it might be useful to catch this exception in monitoring db?

ZhuozhaoLi: probably add a logging line here?

parsl/app/bash.py
ZhuozhaoLi: fixed

parsl/dataflow/dflow.py
ZhuozhaoLi: @benclifford Do you expect this exception to be raised at DFK?

PR #1487, opened 2019-12-04
add try catch to monitoring db insert/update

ZhuozhaoLi:
This is another fix mentioned in #1267  -- add a try-catch to catch the exceptions during insertion or update to monitoring db, so the insertion/update errors won't stop the entire loop in `db_manager`.

This PR can pass the tests with monitoring enabled (except one test in `test_implicit_staging_https.py`).


parsl/monitoring/db_manager.py
```diff
@@ -380,10 +383,18 @@ def _migrate_logs_to_internal(self, logs_queue, queue_tag, kill_event):
                     self.pending_node_queue.put(x[-1])
 
     def _update(self, table, columns, messages):
-        self.db.update(table=table, columns=columns, messages=messages)
+        try:
+            self.db.update(table=table, columns=columns, messages=messages)
+        except Exception:
+            self.db.rollback()
```
yadudoc: Can the rollback fail ?

benclifford: almost definitely because it involves talking to the database... that's probably something that should be fatal to monitoring though.

the log line should go right after the `except` statement so that it is clearer what is happening

ZhuozhaoLi: Added another `try except` around rollback

Issue #1480, opened 2019-12-02
Unhandled exception in worker side result serialization


yadudoc:
**Describe the bug**
When a large numpy array is returned by a worker, one of the serialization steps fails due to the array being a `MemoryView` object. This exception is not caught and causes the worker to crash.

The surprising behavior here is that the following app does not fail for size 10 or 100, but fails for 1000 making this harder to catch in development : 
```
@python_app
def compute_descript(smile, size=10):
    import numpy as np
    import pickle

    x = np.array(list(range(0, size)), dtype=complex).astype(np.float32)
    return x

```

**To Reproduce**
Steps to reproduce the behavior, for e.g:
Run https://gist.github.com/yadudoc/690145693f4984edf43c8041652028f9 
with parsl 0.9.0

**Expected behavior**
Catch and report `can't pickle memoryview objects` from pickle module with the right exception object.

PR #1489, opened 2019-12-04
Better exception handling in serialization code

yadudoc:
#1480 describes a case where ipyparallel.serialize will try to serialize a numpy.array followed by pickle trying to serialize this already serialized buffer. Pickle will fail depending on whether the array is large enough to be internally represented as a MemoryView object which are generally non-serializable. Previously we assumed that all serialization issues would've been caught in the ipyparallel.serialize step and no errors are expected in the pickle step. This PR adds additional error handling for the corner cases (eg, in #1480).


ZhuozhaoLi:
This looks good to go. But I wonder if the `test_1480` would still fail with PR #1654

yadudoc:
@ZhuozhaoLi test_1480 returns a numpy array of 10^6 which exceed the buffer size limit set by 1654.
So I've two tests test_1480 that confirms that a typeError is raised, and 1654 testing that parsl still works at smaller numpy sizes.

PR #1490, opened 2019-12-05
Add try catch monitoring process starter

ZhuozhaoLi:
This PR is another fix mentioned in #1267 . This includes:

1. Wrapping the Hub and DBM starter with a `try catch`. If there is exception on Hub or DBM, let `MonitoringHub` know through a multiprocessing queue (any thought of potential problem on that?).
In case of Hub or DBM exception, `MonitoringHub` cleans up all monitoring processes.

2. Change Hub and DBM processes to daemon.

3. Change multiple threads in DBM to daemon.

PS: it passed pytest with monitoring enabled locally.


parsl/monitoring/monitoring.py
```diff
@@ -433,10 +448,17 @@ def start(self, priority_msgs, node_msgs, resource_msgs):
...
+def hub_starter(comm_q, exception_q, priority_msgs, node_msgs, resource_msgs, *args, **kwargs):
     hub = Hub(*args, **kwargs)
     comm_q.put((hub.hub_port, hub.ic_port))
-    hub.start(priority_msgs, node_msgs, resource_msgs)
+    hub.logger.info("Starting Hub in Hub starter")
+    try:
+        hub.start(priority_msgs, node_msgs, resource_msgs)
+    except Exception:
+        hub.logger.exception("hub.start exception")
+        exception_q.put("Hub got exception")
```
yadudoc: Could we add exception message itself so that the main process can report it better ?

ZhuozhaoLi: Fixed this.

parsl/monitoring/db_manager.py
```diff
@@ -471,8 +471,8 @@ def dbm_starter(exception_q, priority_msgs, node_msgs, resource_msgs, *args, **k
     dbm.logger.info("Starting dbm in dbm starter")
     try:
         dbm.start(priority_msgs, node_msgs, resource_msgs)
-    except Exception:
+    except Exception as e:
         dbm.logger.exception("dbm.start exception")
-        exception_q.put("DBM got exception")
+        exception_q.put(("DBM", e))
```
yadudoc: Sending the exception object like expects serializability, which I worry may not always work. Maybe stringify the exception and traceback ?

parsl/monitoring/monitoring.py
```diff
@@ -454,9 +456,9 @@ def hub_starter(comm_q, exception_q, priority_msgs, node_msgs, resource_msgs, *a
     hub.logger.info("Starting Hub in Hub starter")
     try:
         hub.start(priority_msgs, node_msgs, resource_msgs)
-    except Exception:
+    except Exception as e:
         hub.logger.exception("hub.start exception")
-        exception_q.put("Hub got exception")
+        exception_q.put(('Hub', e))
```
yadudoc: Same as above

parsl/monitoring/db_manager.py
ZhuozhaoLi: fixed

parsl/monitoring/monitoring.py
ZhuozhaoLi: fixed

PR #1492, opened 2019-12-05
Removed check that marks a missing job as completed only if it was in…

hategan:
… certain state before. This check is nondeterministic and depends on the rate at which the job state changes as well as the interval at which the queuing system is polled. It is entirely possible for a job to very quickly move from queued -> running -> failed -> removed. If the check above is kept, such a job will eternally be reported as PENDING. For example:

submit():
	status = 'PENDING'
job runs and is removed from queue
status():
	if status in ['RUNNING', 'KILLING', 'EXITING']: # false, since status == 'PENDING'
		status = 'EXITING'
	# status == 'PENDING'


hategan:
> Functionally this is correct. Is there any reason for using the translate table for cobalt and torque, but not the others?

That's probably best answered by whoever wrote that code, but it's unrelated to this PR. You see those lines as changed because the indentation is different, but there are only things being removed here.

PR #1496, opened 2019-12-11
Fix jumbled legends and colors in workflow dags

benclifford:
Prior to this, colours and labels for jobs in the two DAG graphs were
jumbled up.

The code incorrectly assumed that iterating over a dict would iterate in
the order that the entries appear in a hard-coded literal dict in the
source tree.

For example:

The code assumed that the first element added to the node_traces list, and
thus the element referred to by node_traces[0], would be the element in
groups_list with value (0,...)

This incorrect assumption led to adding DAG nodes to the wrong Scatter
object.

This patch changes the way that the node_traces list is constructed by
populating it with marker places and then using index notation to set the
appropriate entry in the list based on hard-coded index.

Further, that wrong scatter object then had its colour mutated. This patch
sets the colour correctly on initial creation and does not change it again.

For the DAG coloured by state plot, the naming if-statement had no final
else: case, leading to some nodes being incorrectly labelled. This patch
adds an 'Unknown' case which is reached, for example, in jobs which are
never launched due to dependency failure.

### DIFF
--- a/parsl/monitoring/visualization/plots/default/workflow_plots.py
+++ b/parsl/monitoring/visualization/plots/default/workflow_plots.py
@@ -165,7 +165,6 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
     G.add_edges_from(edges)
 
     node_positions = nx.nx_pydot.pydot_layout(G, prog='dot')
-    node_traces = []
 
     if group_by_apps:
         groups_list = {app: (i, None) for i, app in enumerate(
--- a/parsl/monitoring/visualization/plots/default/workflow_plots.py
+++ b/parsl/monitoring/visualization/plots/default/workflow_plots.py
@@ -199,7 +199,6 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
         x, y = node_positions[node]
         if group_by_apps:
             name = dic['task_func_name'][node]
-            index, color = groups_list[name]
         else:
             if dic['task_time_returned'][node] is not None:
                 name = 'Completed'
@@ -209,7 +208,7 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
                 name = "Pending"
             else:
                 name = "Unknown"
-            index, color = groups_list[name]
+        index, color = groups_list[name]
         node_traces[index]['x'] += tuple([x])
         node_traces[index]['y'] += tuple([y])
         node_traces[index]['text'] += tuple(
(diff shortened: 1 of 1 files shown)


parsl/monitoring/visualization/plots/default/workflow_plots.py
```diff
@@ -176,12 +176,14 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
     node_traces = []
```
ZhuozhaoLi: Should remove this then

parsl/monitoring/visualization/plots/default/workflow_plots.py
```diff
@@ -197,25 +199,26 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
...
                 line=dict(width=1, color='rgb(0,0,0)')))
-        node_traces.append(node_trace)
+        node_traces[index]=node_trace
 
     for node in node_positions:
         x, y = node_positions[node]
         if group_by_apps:
             name = dic['task_func_name'][node]
-            index = groups_list[name]
+            index, color = groups_list[name]
```
ZhuozhaoLi: We can probably move this line outside of the `if ... else` now. And also, after removing line 218, `color` is no longer used later I think.

parsl/monitoring/visualization/plots/default/workflow_plots.py
```diff
@@ -197,25 +199,26 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
             name=k,          # legend app_name here
             marker=dict(
                 showscale=False,
-                # color='rgb(200,0,0)',
+                color=color,
```
ZhuozhaoLi: I would need to try this, and see what happen when there are multiple `apps` and the color is set to `None`

parsl/monitoring/visualization/plots/default/workflow_plots.py
benclifford: done

parsl/monitoring/visualization/plots/default/workflow_plots.py
benclifford: The documentation here https://plot.ly/python/reference/#scatter says that colour must be either a colour or a list of colour indices.

Actual behaviour with the version of plotly I'm using is that color=None seems to be the same as not specifying color at all - it allocates a new color automatically.

ZhuozhaoLi:
Tried that on a machine. The `color=None` looks fine.

PR #1498, opened 2019-12-11
Add Provider Functionality to WorkQueueExecutor

tjdasso:
This PR is the first, bare-bones attempt at integrating Provider functionality with the WorkQueueExecutor. It adds implementations for the scaling functions within the executor, and adds a provider attribute to the Executor object to be able to submit jobs using the `work_queue_worker` executable.

This has only been tested with the CondorProvider, which successfully scales out and scales in workers according to the scaling functions within the Executor. Using `condor_q`, we can see that jobs have been submitted to Condor, wrapped by a Parsl command. Additionally, during shutdown, the CondorProvider successfully removes the jobs from the queue. However, the workers that have been spun up on the Condor provider are not yet connecting to the master Executor and receiving tasks. 

I can't tell what the actual command is that is being submitted to Condor because I believe the task is being wrapped by another script by the Provider, but I have two hypotheses as to why the connection is not occurring:
1. The wrong port/hostname is being submitted to Condor by the executor, and so the worker on the Provider cannot connect. I don't think this is likely since the debugging of the launch command sent to the provider is the expected command with the appropriate hostname and port.
2. The Provider cannot find the worker executable to send over to Condor to run. Unlike other Executors like the HighThroughputExecutor which uses a Python script to run on the Provider, Work  Queue has the `work_queue_worker` executable which needs to be sent. Perhaps the worker executable is not being found by the Provider since it is not packaged with the rest of the code when installation occurs, and we have to put it somewhere where the Provider can find it.


annawoodard:
@tjdasso: I'd like to test this, but I'm currently getting what looks like some unrelated errors running the tests with wqex from master using `CCTOOLS_VERSION=7.0.11` on a RHEL6+torque system. Can you confirm which version of WQ you are running, so I can try to sync up? I'll open an issue about the errors separately.

benclifford:
It's frustrating that you don't get a clear error message that tells you more information about 1. vs 2. 

It looks like the CondorProvider (in providers/condor/template.py) directs stdout/stderr like this:

```
output                  = ${submit_script_dir}/${job_name}.stdout
error                   = ${submit_script_dir}/${job_name}.stderr
```

Does any useful failure message appear there, or in something like `condor_q -verbose`?

tjdasso:
I looked at the output and error files for Condor, but they are both empty, even after the tasks are killed at shutdown. Looking at Condor documentation, it mentions that these files are created during job submission with zero length, so it's likely that the files are not getting written back to after the job completes/gets killed.

Looking at the wrapper script that gets submitted and the Condor submit script for each worker, however, it seems as though Condor is receiving the right command to run. I was able to run the wrapper script locally, which correctly spun up a worker that connected to the WQExecutor and completed the tasks. However, due to the wrapping of the task, the work_queue_worker executable isn't automatically transferred with Condor, so I used one of the arguments to the CondorProvider constructor to specify the executable as an input file. The new submit scripts show that this file is specified as an input file, but the workers are still not connecting to the master. Hopefully, I can find out how to get the stdout and stderr files to get written back to the submit machine, which will help with the debugging process.

annawoodard:
> However, due to the wrapping of the task, the work_queue_worker executable isn't automatically transferred with Condor, so I used one of the arguments to the CondorProvider constructor to specify the executable as an input file.

Great sleuthing, that sounds like promising progress...

You can also try `scheduler_options='transfer_output_files = <path_to_logs>'` as a keyword arg to the `CondorProvider`. That's not equivalent getting the full stdout, but perhaps you could modify the command starting the worker to direct the output somewhere you can access. One possibility is that the executable is being transferred, but the permissions are not correct and you're running something like `./work_queue_worker` which is failing with a permission denied error.

If all else fails, sometimes `condor_ssh_to_job`, where you can run `ps` to inspect what processes are running, can be helpful.

dthain:
@tjuedema please take over from @tjdasso and push this one forward as well.

dthain:
@btovar @yadudoc please take a look at getting this PR running.

btovar:
I updated to master 4709b305b78047b3f58cd5a943507081f12744bd, and with some changes, I was able to succesfully run both with local and condor providers.

There are two issues with the condor provider:

1) The work_queue_worker executable is not being sent with the condor job. Using:
`provider=CondorProvider(transfer_input_files = [...path_to_worker...] ` as an argument to the workqueue executor works, but we would not like users to specify that every time.
2) The workqueue executor assumes that the work_queue_worker is in PATH. For condor, I made it work by changing the command to './work_queue_worker', but this breaks the local executor. 

For the first issue, do providers have the concept of 'list of support files that may need to be transferred'? Right now the workqueue executor could detect that it is using the condor provider and modify the provider's self.transfer_input_files directly, but that is ugly and far from general. On the other hand, some providers do not send anything. 

For the second, I think we can solve it by copying the work_queue_worker where we needed it, and executing it with ./work_queue_worker. Most likely, we would need to copy it with some temp name so we can safely remove it at the end of execution.

dthain:
Superseded by #1678

yadudoc:
#1678 supersedes this.

PR #1500, opened 2019-12-12
These should have been part of the job_status_class PR, but weren't.

hategan:
Boubou was committed earlier. The job_status_class branch/PR was missing changes to the local and ad_hoc provider. Somehow it passed tests, review, and whatnot. We should merge this.

You can check that there are no more relevant occurrences of 'PENDING', 'RUNNING', 'COMPLETED', 'CANCELLED', 'FAILED' with:
egrep -r "'PENDING'|'RUNNING'|'FAILED'|'COMPLETED'|'CANCELLED'" .

There are a few in comments, which we should fix, but this is more urgent.

### DIFF
--- a/parsl/providers/ad_hoc/ad_hoc.py
+++ b/parsl/providers/ad_hoc/ad_hoc.py
@@ -4,7 +4,7 @@ import time
 
 from parsl.channels import LocalChannel
 from parsl.launchers import SimpleLauncher
-from parsl.providers.provider_base import ExecutionProvider
+from parsl.providers.provider_base import ExecutionProvider, JobStatus, JobState
 from parsl.providers.error import ScriptPathError
 from parsl.utils import RepresentationMixin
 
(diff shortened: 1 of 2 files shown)


hategan:
> This looks good to go in. Does strategy need to be updated to match the new jobstatus data-structure ?

Should already be there. See, e.g., https://github.com/Parsl/parsl/blob/master/parsl/dataflow/strategy.py#L192

benclifford:
That egrep does not catch job statuses wrapped in `"` rather than `'`

I think the only relevant line uncaught is:

```
parsl/executors/ipp.py:        to_kill = [engine for engine in status if status[engine] == "RUNNING"][:blocks]
```

hategan:
> That egrep does not catch job statuses wrapped in `"` rather than `'`
> 
> I think the only relevant line uncaught is:
> 
> ```
> parsl/executors/ipp.py:        to_kill = [engine for engine in status if status[engine] == "RUNNING"][:blocks]
> ```

Right. The reader is supposed to extend that to double-quotes. I was a bit busy yesterday and prioritized the local provider. I'll add commits for comments and IPP shortly.

PR #1503, opened 2019-12-13
PR to figure out why testing did not uncover issues with the JobStatus class PR

hategan:
! Do not merge !
Also, delete both branches after we figured this out.

A PR was merged which updated everything to use the JobStatus class. Except "everything" did not include the local provider. This was caught after the merge, so this PR tries to reconstruct the environment of that PR so that we can figure out how this slipped through the automated tests.

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -3,7 +3,7 @@ import time
 import math
 
 from parsl.executors import IPyParallelExecutor, HighThroughputExecutor, ExtremeScaleExecutor
-
+from parsl.providers.provider_base import JobState
 
 logger = logging.getLogger(__name__)
 
@@ -189,8 +189,8 @@ class Strategy(object):
             nodes_per_block = executor.provider.nodes_per_block
             parallelism = executor.provider.parallelism
 
-            running = sum([1 for x in status if x == 'RUNNING'])
-            pending = sum([1 for x in status if x == 'PENDING'])
+            running = sum([1 for x in status if x.state == JobState.RUNNING])
+            pending = sum([1 for x in status if x.state == JobState.PENDING])
             active_blocks = running + pending
             active_slots = active_blocks * tasks_per_node * nodes_per_block
 
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -16,7 +16,7 @@ from parsl.executors.high_throughput import zmq_pipes
 from parsl.executors.high_throughput import interchange
 from parsl.executors.errors import BadMessage, ScalingFailed, DeserializationError
 from parsl.executors.base import ParslExecutor
-from parsl.providers.provider_base import ExecutionProvider
+from parsl.providers.provider_base import ExecutionProvider, JobStatus
 from parsl.data_provider.staging import Staging
 from parsl.addresses import get_all_addresses
 
@@ -611,7 +611,7 @@ class HighThroughputExecutor(ParslExecutor, RepresentationMixin):
 
         return r
 
-    def status(self):
+    def status(self) -> List[JobStatus]:
         """Return status of all blocks."""
 
         status = self.provider.status(list(self.blocks.values()))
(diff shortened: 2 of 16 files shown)


hategan:
So nothing in the tests seems to detect this. I would have expected that strategy.py line 192 would have failed because str doesn't have a .state attribute, but that line is either not exercised by any of the tests in conjunction with the local provider, or the failure is not reported.

yadudoc:
@hategan change from PR to draft so that it can't be merged ?

hategan:
> @hategan change from PR to draft so that it can't be merged ?

Can you do that without creating a new PR?

In any event, this would merge into an unused branch, not into master.

hategan:
Ok, looks like an exception is thrown by _strategy_simple, but it is caught in flow_control.make_callback, logged and otherwise ignored.

I would, again, suggest that critical errors in core components should cause everything to fail immediately or that, in the least, they do so when the tests are being run. I understand that there's some opposition to this because it could cause wasted CPU-hours.

I'll use a metaphor to justify the suggested behavior. We have a hotel in a skyscraper. If the TV in a room stops working, the appropriate action is to respond with a localized solution: fix the TV problem in that room. We would not want to kick the guest out of the room. Now, say that a structural problem with the building is detected. The appropriate solution would be to close down the hotel and immediately kick everybody out until it can be asserted that the building is safe. Sure, money will be lost, but significantly more money will be lost if the building collapses with everybody in it.

It's not quite like that in our case, but close. Stable code that users use should have a minimal amount of structural problems. If that isn't true, then it shouldn't be called "stable". However, in order for it to have a minimal amount of structural problems, we must take an active role in ensuring that this is the case by, e.g., seeking and detecting such problems. Tests are a way to do that *if* they accurately convey the existence of problems. In this case, they didn't. We seem to have two options here (and then a middle ground which is probably a slippery slope to one or the other of the two):

- ensure that serious problems are reported early and forcefully; this is equivalent to kicking everybody out of the building if the building is unstable. Unless we expect to have users frequently use an unstable core, this should have little impact on users. In other words, we should avoid opening hotels in unstable buildings and, if we are good at doing early inspections and sound structure, that should minimize the risk that we'll ever have to close a hotel due to structural problems with the building. Of course, we cannot guarantee that all problems will always be caught before the code reaches users, so some limited damage is possible, but I would argue that the alternative is far worse.

- continue hiding things but keep user jobs running as far as possible; this is the equivalent of keeping all guests in the hotel until it collapses, and even after, if not all rooms are fully affected. In a sense, this is almost like not having type checking because "even if the code is incorrect, it may still make progress beyond what a typechecked program would". There are a number of problems with this:
   * it may be difficult for a user to recover a consistent result out of a partial workflow, even if we run as much as we can; my personal preference when working on anything that I would want to later publish is that I would rather lose some CPU hours if it helps ensure the integrity of the scientific result
   * it seems to be very costly for us in terms of development. We must spend nontrivial effort on issues that could otherwise be checked quickly and automatically. 
   * it has hidden costs that we are not quantifying. In concurrent software, hiding errors often leads to a lack of progress. It's a fundamental problem in distributed and concurrent systems that (without sufficient access to low-level details) a process that will never complete is indistinguishable from a process that is taking an arbitrarily long time. The way we deal with this is to impose timeouts. But, from a user's perspective, deciding when something has timed-out, requires a lot of effort, such as logging into the remote system and looking at the queue, inspecting logs, etc. The user may very well prefer wasting some CPU hours to their own time (for reference, converted to dollars, CPU time is way cheaper than human-time). Furthermore, structural problems can also cause their own wasted CPU time. If scaling down of workers fails (which would if something were to happen to the strategy), CPU time will be wasted.

benclifford:
I'm fine with there being a more formal parsl-is-broken state/behaviour.

I'm also fine with the test suite failing if there are unexpected ERROR logs.

yadudoc:
This could be left on a branch instead of being a PR. Plus this is most likely stale at this point. 
What do you think @hategan, @benclifford ?

hategan:
@yadudoc see the very first comment

PR #1506, opened 2019-12-16
Make config-changing staging test run in local mode

benclifford:
Prior to this, one of the staging tests added an executor to
the DFK, to check that staging still worked in that mutated
environment.

However, this test leaves the mutated environment in place,
meaning that subsequent tests will now sometimes run on that
added executor instead of on the command-line specified
executor.

This commit changes that test to run in local mode, so that
there is a DFK specific to just this test module for the
additional executor test.


benclifford:
I decided that this test really should use a fresh config, not the session global htex one.

@annawoodard can you look at this again?

annawoodard:
@benclifford Why did you decide we need a fresh config? I want to update my heuristics on when we need vs when we don't.

benclifford:
> Why did you decide we need a fresh config?

I think I'm pretty much of the opinion that config objects should never be used twice: sometimes they get mutated by tests; and it isn't clear to me what it means for components which live inside the config object (especially executors)  to be used by two different DFKs at once. Who is meant to call start()? Can it be called multiple times? Who is meant to call stop()? (it actually surprises me that it works at all at the moment)

A new python process => a new config object even if it comes from the same source code, but within the same python process, that isn't true.

Following on from that, that makes me think that any `--config local` test that loads a DFK should always load a fresh config object, not a config object that was imported from `parsl.tests.configs`, because there is no guarantee that that config object isn't re-used elsewhere - If you want a test case specific config, put it directly in the test case source file.

annawoodard:
bors merge

annawoodard:
bors merge

annawoodard:
bors merge-

annawoodard:
bors merge

annawoodard:
bors merge-

annawoodard:
Trying again after enabled building pushed branches in the travis settings. I think our whitelist in `.travis.yml` only takes effect if that is enabled.

bors merge

annawoodard:
Travis was run this time, but afterwards bors cancelled because there wasn't a PR approval. Approving and trying again.

bors merge

annawoodard:
bors merge

annawoodard:
Note, we're giving up on bors for now because it seems there is no way to have bors enabled while having GitHub require that all PRs have approvals.

PR #1510, opened 2019-12-27
Dump stack traces in pytest on SIGUSR1

hategan:
When testing, dump a stack trace in ~/parsl_stack_dump.txt when SIGUSR1 is received.


parsl/tests/conftest.py
```diff
@@ -16,6 +19,25 @@
...
+            s += ''.join(traceback.format_stack(frame))
+    except Exception:
+        s = traceback.format_exc()
+    with open(os.getenv('HOME') + '/parsl_stack_dump.txt', 'w') as f:
+        f.write(s)
+    print(s)
+
+
+import signal
+signal.signal(signal.SIGUSR1, dumpstacks)
```
benclifford: can this handler installation happen in something that only runs inside pytest rather than on import? I've spent way too much time tracking down "of course this will be safe to do at import" bugs especially in the test suite...

Maybe like:
```
@pytest.fixture(autouse=True, scope='session')
def load_dfk_session(request, pytestconfig):
```

hategan: d1e075d. There's a pytest hook for session initialization, which, IMO, is clearer in intent than the fixture, but if you really want it to be a fixture, let me know or change directly.

Issue #1347, opened 2019-10-10
Add mechanism to flush tasks from DFK


yadudoc:
**Is your feature request related to a problem? Please describe.**
Currently we store a lot of information about tasks in the DFK. Unfortunately, since the DFK has no mechanism to auto-identify what tasks can be garbage collected, it keeps everything. This can bloat up pretty quickly if you are building a long running service on top of Parsl

**Describe the solution you'd like**
Add a mechanism to flush tasks from the DFK : 
```python
dfk = parsl.dfk()
dfk.flush(<list of task_ids>)
```

**Describe alternatives you've considered**
N/A

**Additional context**
Requested by @dgasmith at Parslfest,

dgasmith:
This is becoming a larger issue now that we leave Parsl instance running for weeks to months that handle multiple millions of tasks.

yadudoc:
@dgasmith On it.

Some thoughts on how we could do this properly. 

Python3 uses a combination of reference counting and generational garbage collection internally
for garbage collection. The DataFlowKernel keeps track of every future in it's internal
task graph which leads to live references that prevent Python's GC from reaping futures (and it's data)
even when no references are held in the main context.

1. We need to differentiate between references held inside the DFK vs outside it.
   We can assume that the task graph inside the DFK will not generate additional references, only additional indirections.
   For example `TaskA <---- TaskB` might transform to `TaskA <--- stage_out() <---- stage_in() <---- TaskB`
   
   Currently these data stageing tasks are the only additional operations that the DFK would insert into the task graph.
   
2. Ideally we make minor changes that enable Python's GC to step in and handle dead data.

   * We can assume that if a live reference exists in the main context the future should not be deleted.
   * If there are no live references outside the DFK no additional references will be added.
   * Once a task is completed and *all* it's dependents used their references to it, then it should be safe to GC
   
So, we could mark a future as a weakref upon resolving and delete it's dependencies, such that the reference counts
within the DFK are updated correctly. Therefore, if no references exist in the main context, the future would be
garbage collected.

If we go this route, the user will be able to flush task by calling `del app_future` rather than having to ask the dfk to flush the tasks, which I think is cleaner.

dgasmith:
Overall this sounds good, but does seem fairly involved. I am unsure how transferable our use case is.

It we can simply delete a DFK and rebuild periodically that could work as well.

yadudoc:
[...]

PR #1512, opened 2020-01-03
Support for garbage collection

yadudoc:
Replace app_future with a weakref and wipe depends list upon task completion


benclifford:
does it work for the appfu to be replaced with None rather than a weakref? What are the circumstances that it is useful to maybe be able to get the app future but not definitely so, after completion?

yadudoc:
@benclifford We only need a ref to the appFuture when the execFu callback triggers, once that is complete it should be safe to remove the ref. However it looks like there's some race condition going on in cases where the appFu is set before the dfk.submit() could return the appFu to the user's context.

benclifford:
> However it looks like there's some race condition going on in cases where the appFu is set before the dfk.submit() could return the appFu to the user's context.

I think a weak ref isn't going to stop that race condition happening? I think broadly if the code can't work with the field being set to None then I don't think it will necessarily work with the field being a weak ref, because the weak ref can turn into a null reference at any time.

parsl/dataflow/dflow.py
```diff
@@ -257,7 +257,7 @@ def handle_exec_update(self, task_id, future):
         structure.
 
         Args:
-             task_id (string) : Task id which is a uuid string
+             task_id (string) : Task idwhich is a uuid string
```
benclifford: accidentally removed space?

yadudoc:
Fixes #1347

parsl/tests/test_checkpointing/test_python_checkpoint_3.py
```diff
@@ -8,6 +8,7 @@
 
 
 def local_setup():
+    parsl.clear()
```
benclifford: if this is needed, another test hasn't cleared up after itself properly, I think

parsl/dataflow/dflow.py
yadudoc: Done

parsl/tests/test_checkpointing/test_python_checkpoint_3.py
yadudoc: I've removed this.

docs/userguide/garbage_collection.rst
```diff
@@ -0,0 +1,33 @@
...
+storing `AppFutures` in an active context. For eq:
+
+.. code-block:: python
+
+   def double(x):
+        return x * 2
+
+   # Since the AppFuture returned is not stored in a variable
+   # the task will be garbage collected.
+   print(double(5).result)
```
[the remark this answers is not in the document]
yadudoc: Fixed.

docs/userguide/garbage_collection.rst
```diff
@@ -0,0 +1,33 @@
+.. _garbage_collection:
+
+Garbage Collection
+------------------
+
+Production workflows often run for days processing millions of tasks, and as tasks accumulate keeping
```
benclifford: I dislike this paragraph. It comes from the angle of "this is how parsl used to work internally and this is how it has been changed" - that belongs in the changelog / PR commit message.

There's no change to the user exposed interface by garbage collection, and so no need to change the documentation.

With memoization/checkpointing on or off, the memory usage of parsl previously differed. Now that difference is more. We didn't document that before. If we're going to document that now, document it as an improvement to the checkpointing documentation.

yadudoc: I'm not sure how to improve this. I tried to write this as a feature note, so that users know this exists, maybe this doesn't warrant any attention.

I don't mind removing this entirely and just leaving a note in checkpointing that some modes can interfere with garbage collection.

annawoodard: I vote to remove this and leave a note in checkpointing

annawoodard: (i agree with @benclifford)

yadudoc: Done.

PR #1519, opened 2020-01-04
Formalized the executor interface more by adding certain methods that…

hategan:
… were more or less tacitly used without a formal interface. These are mostly related to error handling.

Also added abstract classes that are used to reduce copied-and-pasted code in executors, although only implementations for the newly formalized methods are present.


parsl/executors/status_handling.py
```diff
@@ -0,0 +1,109 @@
...
+    @property
+    def bad_state_is_set(self):
+        return self._executor_bad_state.is_set()
+
+    @property
+    def executor_exception(self):
+        return self._executor_exception
+
+    @property
+    def tasks(self) -> Dict[str, Future]:
```
benclifford: this type annotation is in conflict with how HighThroughtputExecutor uses _tasks: it uses keys from `int`, not from `str`.

PR #1521, opened 2020-01-08
Correct Scaling with SimpleLauncher+ClusterProvider

WardLT:
Parsl does not correctly determine the number of active slots when using a SimpleLauncher as part of a ClusterProvider. The current code assumes that you launch a single worker per compute node, which is not the case for SimpleLauncher.


benclifford:
I talked with @WardLT on slack about this a few weeks ago and commented that I didn't like the use of isinstance to bring in magic behaviour for certain configurations. I'm going to close this PR. Reopen/make a new one if there's a better way to do this.

PR #1528, opened 2020-01-09
Fix walltime bug #1527

### LINKED ISSUE #1527: Walltime option via python_app decorator not honored
**Describe the bug**
Walltime specified via the python_app decorator as `python_app(walltime=20)` is ignored, while
walltime kwarg to the decorated function works correctly.

**To Reproduce**
1. Setup Parsl 0.8.0 with Python 3.6 locally
2. Run:
```python
@parsl.python_app(walltime=1)
def my_app_2():
    import time
    time.sleep(2)
    return True

my_app_2().result()  # This should raise AppTimeout
```

**Expected behavior**
Raises AppTimeout

**Environment**
 - OS: ubuntu
 - Python version: Py 3.6
 - Parsl version: 0.9.0 and master

Reported by @ericmjonas

yadudoc:
`walltime` kwarg is now removed from the app decorator base and will not be supported by `@bash_app` or  `@python_app`. 
Adds a test that checks whether python_app walltime kwarg works correctly. The walltime works in second level granularity, so this test will take at least a second to trigger. 

Fixes #1527

### DIFF
--- a/parsl/app/app.py
+++ b/parsl/app/app.py
@@ -21,7 +21,7 @@ class AppBase(metaclass=ABCMeta):
 
     """
 
-    def __init__(self, func, data_flow_kernel=None, walltime=60, executors='all', cache=False):
+    def __init__(self, func, data_flow_kernel=None, executors='all', cache=False):
         """Construct the App object.
 
         Args:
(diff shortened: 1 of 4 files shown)


yadudoc:
Our latest docs specify that `inputs`, `outputs` and `walltime` are shared by both bash_apps and python_apps. Please check here : https://parsl.readthedocs.io/en/latest/userguide/apps.html#special-keyword-arguments
I believe some of the restructuring in the docs here is from Ian's edits.

PR #1540, opened 2020-01-15
Enable the bors merge-bot

annawoodard:
This should enable the bors merge-bot. It follows the vanilla recommendations [here](https://bors.tech/documentation/getting-started/), and further specifies that PRs should be squashed before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ branches:
     # This is where pull requests from "bors try" are built.
     - trying
     # Uncomment this to enable building pull requests.
-    #- master
+    - master
 
 # command to run tests
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,15 @@ install:
 os:
     - linux
 
+branches:
+  only:
+    # This is where pull requests from "bors r+" are built.
+    - staging
+    # This is where pull requests from "bors try" are built.
+    - trying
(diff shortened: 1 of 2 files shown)


bors.toml
```diff
@@ -0,0 +1,9 @@
+status = [
+  "continuous-integration/travis-ci/push",
+]
+
+use_squash_merge = false
```
benclifford: is this meant to be true?

annawoodard: good catch-- fixed

.travis.yml
```diff
@@ -29,6 +29,15 @@ install:
...
     - linux
 
+branches:
+  only:
+    # This is where pull requests from "bors r+" are built.
+    - staging
+    # This is where pull requests from "bors try" are built.
+    - trying
+    # Uncomment this to enable building pull requests.
+    #- master
```
benclifford: I think this turns off building the PR/master merge which is a test that we require to pass at the moment... so tests haven't even been run on this.

So maybe the right thing here is all of staging/trying/master. (Or maybe we need to turn off that required test and let bors do the enforcement rather than github. I'm not sure.)

annawoodard: > So maybe the right thing here is all of staging/trying/master.

I think you're right. In this way, we have the option of using our current workflow or bors.

Issue #1279, opened 2019-09-13
Everything-to-do-with-a-task completion vs AppFuture completion


benclifford:
At present, AppFutures complete when the result of an app is available. That does not mean that stage out has completed. Stage out completion is indicated by the relevant DataFutures completing.

This gives more scope for concurrency: I can use a result as soon as it is available, without waiting for stage out, in places that this is meaningful.

This has potential to cause trouble: it is awkward for users to wait for "everything" and probably not entirely obvious that they should do a more complex thing that just wait for the AppFuture to complete.

One option is to change AppFuture completion to happen only when all the output futures have completed, not just the executor future completing.

benclifford:
I'm seeing related behaviour where wait_for_current_tasks completes too early: it completes when all AppFutures are done, which is before some task ending processing is completed. This can lead to process exit before monitoring information has been recorded for recently finished tasks.

benclifford:
If all post-processing can be represented through futures (as happens with file stage-out for example) then the DFK code which will probably used to implement `join_app` (see #1796) can maybe also be used here to make an app wait for all of its staging/post processing to complete.

benclifford:
Issue #2252 is a checkpointing at shutdown race condition tied strongly to this issue.

PR #1541, opened 2020-01-16
Prohibit ERROR log messages in pytest, with an escape clause

benclifford:
This PR causes test failures if parsl code logs an ERROR or higher during a test. The intention is to capture errors that are logged but aren't presented to the test suite in a form that is mechanically detected.

This introduces a couple of fixtures, a log handler and a context manager
which together check that log messages of severity ERROR or higher are
not logged during test runs, except during periods when a test is in a
permit_severe_log context.

If an ERROR is logged during a test, that test will fail.

All the tests that were failing in the test suite because of this (because they deliberately caused errors) have been fixed by the addition of `permit_severe_log`.

This fixture won't detect logs that happen before the fixture comes alive, or after it has shut down - so perhaps in DFK initialization or shutdown.


benclifford:
This PR is mostly failing in CI because of IPP. There is a plan (#1548) to rapidly remove that.

benclifford:
I've converted this back to a draft PR, as `parsl/tests/test_bash_apps/test_apptimeout.py` is non-deterministically failing for me on the htex_local_alternate config.

What is happening is that the first test there is intended to fail, and is annotated that way in this PR, but the final error message for that failing test is not logged until sometimes 2 seconds later than the app future completes - by which time the test suite has moved onto other tests, out of the permit_severe_log context.

It is not clear to me what is causing that delay. In any case, there's a race condition between permit_severe_log contexts ending because an app future has completed, and parsl not having completed its post-completion processing. Crossref issue #1279 which addresses this race condition in other situations.

benclifford:
this needs the parsl task model to change a bit. that might happen but unlikely to, so closing.

PR #1542, opened 2020-01-16
Require PR approval before allowing bors merge

annawoodard:
I believe that if this isn't enabled, someone could merge a PR via bors without it having been approved.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -3,6 +3,7 @@ status = [
 ]
 
 use_squash_merge = true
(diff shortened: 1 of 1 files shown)


annawoodard:
Looks like this isn't needed because bors just cancelled #1506 due to no approval.

annawoodard:
Re-opening. Maybe with this enabled bors will fail before merging, which would be preferable.

PR #1543, opened 2020-01-16
Cleaner support for wiping tasks after completion from DFK

yadudoc:
Added a new `wipe_task` method.

This should fail current tests, but isn't curiously. Needs more testing.


benclifford:
> This should fail current tests, but isn't curiously. Needs more testing.

@yadudoc what do you think should be failing?

parsl/dataflow/dflow.py
```diff
@@ -368,11 +368,18 @@ def handle_app_update(self, task_id, future, memo_cbk=False):
 
         # If checkpointing is turned on, wiping app_fu is left to the checkpointing code
         # else we wipe it here.
-        if self.checkpoint_mode is None:
-            self.tasks[task_id]['app_fu'] = None
```
benclifford: can also remove the blanking out of 'depends' here too I think

benclifford:
This failed on 31e38de in the garbage collection test with workqueue config.

I think there are the following clashing assumptions in the code in this PR at the moment:

i) If checkpoint_mode == None, then checkpoints will not happen.
ii) If checkpoint_mode == Node, then a user may manually call dfk.checkpoint() to make a checkpoint happen.

Assumption i) means that the DFK deletes the task record "early" because it knows that checkpoints will not happen. Assumption ii) means that the checkpoint code assumes a task record will not be being deleted and that checkpointing will happen.

I think there's then a race (which happens to show with WQ and not with others) where the tasks dict  is being modified (by regular task completion - assumption 1) while checkpointing is iterating over it (assumption 2).

benclifford:
Adding this sleep makes a similar but different race appear on my laptop with `local_threads.py`. (Adding a sleep should never make parsl break)

```
diff --git a/parsl/dataflow/dflow.py b/parsl/dataflow/dflow.py
index 499f7732f..ef3f5bcf6 100644
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -7,6 +7,7 @@ import pickle
 import random
 import typeguard
 import inspect
+import time
 import threading
 import sys
 import datetime
@@ -371,6 +372,7 @@ class DataFlowKernel(object):
         # else we wipe it here.
         self.tasks[task_id]['depends'] = []
         if self.checkpoint_mode is None:
+            time.sleep(3)
             self.wipe_task(task_id)
             # self.tasks[task_id]['app_fu'] = None
         return
```

benclifford:
Maybe the behaviour that is the present `None` checkpointing should be two checkpoint modes: none and manual.

parsl/dataflow/dflow.py
```diff
@@ -370,10 +370,16 @@ def handle_app_update(self, task_id, future, memo_cbk=False):
...
-        self.tasks[task_id]['depends'] = []
+            self.wipe_task(task_id)
         return
 
+    def wipe_task(self, task_id):
+        """ Remove task with task_id from the internal tasks table
+        """
+        logger.warning("Wiping {}".format(task_id))
+        # del self.tasks[task_id]
+        self.tasks.pop(task_id, None)
```
benclifford: does `pop` vs `del` have different the behaviour?

parsl/dataflow/dflow.py
```diff
@@ -1015,7 +1030,7 @@ def checkpoint(self, tasks=None):
                         # mode behave like a incremental log.
                         pickle.dump(t, f)
                         count += 1
-                        self.tasks[task_id]['checkpoint'] = True
+                        # self.tasks[task_id]['checkpoint'] = True
```
benclifford: this `checkpoint` field is referenced in the construction of the task record and also passed to monitoring - probably it should disappear from there?

parsl/dataflow/dflow.py
yadudoc: del raises keyerror if the key is missing, while pop does not when a default is provided.

parsl/dataflow/dflow.py
yadudoc: I will remove those.

parsl/dataflow/dflow.py
benclifford: were KeyErrors really happening?

yadudoc: Good point. That shouldn't really happen. Let me try roll it back and see if it still causes issues in CI

benclifford:
The garbage collector test fails with the workqueue provider since 07b08b4 which makes me think there's still some race condition happening here that was being disguised by the exceptionless `pop`

benclifford:
I think that workqueue problem is coming from the assumption in the test that it is ok to call dfk.checkpoint() even when checkpoint mode is none - that same problem I mentioned above.

The way this PR works, that isn't a safe assumption.

see https://github.com/Parsl/parsl/pull/1543#issuecomment-576246319

yadudoc:
@benclifford  You are right, this test is failing because it is calling `dfk().checkpoint()` when the config specifies `checkpoint_mode=None` and these two are conflicting actions. I think the right option here is to do the checkpoint call only if the checkpoint_mode is not None. Think that's fair ?

benclifford:
I'm uncomfortable with the time delay added in the test suite, but I can't see a clear way to specify the right ordering of operations.

parsl/dataflow/dflow.py
```diff
@@ -370,10 +370,15 @@ def handle_app_update(self, task_id, future, memo_cbk=False):
...
         if self.checkpoint_mode is None:
-            self.tasks[task_id]['app_fu'] = None
-        self.tasks[task_id]['depends'] = []
+            self.wipe_task(task_id)
         return
 
+    def wipe_task(self, task_id):
+        """ Remove task with task_id from the internal tasks table
+        """
+        logger.warning("Wiping {}".format(task_id))
```
benclifford: this should be an info or a debug, not a warning.

yadudoc: Sorry, I should've removed that.

parsl/dataflow/dflow.py
benclifford: there are a few places where code is commented out when it should be removed

yadudoc:
> I'm uncomfortable with the time delay added in the test suite, but I can't see a clear way to specify the right ordering of operations.

Not the best solution, but I think this is passable.

PR #1544, opened 2020-01-17
Make IPP executor return scaled out blocks and not log a spurious error

benclifford:
Indentation meant that an else clause was attached to the scaling for
loop, and so ran at the end of the loop every time, rather than to the
if statement checking if a provider is available.

I discovered this as part of #1541, auditing ERROR log messages in the
test suite.


parsl/executors/ipp.py
```diff
@@ -243,9 +243,9 @@ def scale_out(self, blocks=1):
                                         "Attempts to provision nodes via provider has failed"))
                 self.engines.extend([block])
                 r.extend([block])
-        else:
-            logger.error("No execution provider available")
-            r = None
+            else:
```
yadudoc: You'd think python would complain about this!

benclifford: it's valid syntax - `for` loops can have `else` clauses in python. Which I learned because of this PR.

yadudoc: TIL.

PR #1545, opened 2020-01-17
Fix lazy error handling test

benclifford:
Previously this test checked that 'done' was defined on an
AppFuture, rather than checking that the result of invoking done()
was True. This commit adds the necessary '()'.

This commit also removes some unnecessary complexity in the test.


benclifford:
@yadu I've changed this to test that several failing apps are all reported as `.done()`, and that the appropriate failing exception is returned by `.exception()` without disrupting the main test function control flow.

PR #1546, opened 2020-01-17
Allow plain Futures to be used as dependencies

benclifford:
By ensuring that plain Futures can be used as dependencies, we can be more sure that dependency code does not want to know about the parsl-internal state of dependencies, but instead only relies on what it can observe with the Future interface.

This should be useful in garbage collection work.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -218,7 +218,7 @@ class DataFlowKernel(object):
             task_log_info['task_fail_history'] = ",".join(self.tasks[task_id]['fail_history'])
         task_log_info['task_depends'] = None
         if self.tasks[task_id]['depends'] is not None:
-            task_log_info['task_depends'] = ",".join([str(t.tid) for t in self.tasks[task_id]['depends']])
+            task_log_info['task_depends'] = ",".join([str(t.tid) for t in self.tasks[task_id]['depends'] if isinstance(t, AppFuture) or isinstance(t, DataFuture)])
         task_log_info['task_elapsed_time'] = None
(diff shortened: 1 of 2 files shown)


parsl/dataflow/dflow.py
```diff
@@ -218,7 +218,8 @@ def _create_task_log_info(self, task_id, fail_mode):
             task_log_info['task_fail_history'] = ",".join(self.tasks[task_id]['fail_history'])
         task_log_info['task_depends'] = None
         if self.tasks[task_id]['depends'] is not None:
-            task_log_info['task_depends'] = ",".join([str(t.tid) for t in self.tasks[task_id]['depends']])
+            task_log_info['task_depends'] = ",".join([str(t.tid) for t in self.tasks[task_id]['depends']
```
yadudoc: We'll miss futures that aren't AppFuture or DataFuture here. 

What do you think about replacing with something like this : 
```
deps = ",".join([ str(t.tid) if isinstance(t, AppFuture) or isinstance(t, DataFuture) else str(t) for t in self.tasks[task_id]['depends'] ]
```

benclifford: I think that that field is used as a structured list of integers by downstream monitoring code and we should not break that by putting in a bunch of arbitrary text.

In master at the moment, you don't get dependencies recorded if you do them outside of parsl: primarily if you make your own code implement dependency ordering by waiting for `.result()` before invoking a subsequent app.

I'm fine applying that same argument here: if you do stuff outside of parsl, parsl isn't going to record it.

yadudoc: I see. This is fine.

PR #1549, opened 2020-01-23
Use plain object as task ID type in more places

hategan:
…so str is not quite right.

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -17,7 +17,7 @@ class StatusHandlingExecutor(ParslExecutor):
         self._provider = provider
         self._executor_bad_state = threading.Event()
         self._executor_exception = None
-        self._tasks = {}  # type: Dict[str, Future]
+        self._tasks = {}  # type: Dict[object, Future]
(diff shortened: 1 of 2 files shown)


parsl/executors/status_handling.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, provider):
         self._provider = provider
         self._executor_bad_state = threading.Event()
         self._executor_exception = None
-        self._tasks = {}  # type: Dict[str, Future]
+        self._tasks = {}  # type: Dict[object, Future]
```
benclifford: this same change needs to happen on NoStatusHandlingExecutor

hategan: No Ben, it can't, because self._tasks in NoStatusHandlingExecutor has no type annotation.

That said, I added that type annotation, which hopefully addresses what must have been your actual objection.

benclifford: This came from me validating this PR against my more heavily type-annotated branch - that was more aggressive about the _tasks annotation.

PR #1550, opened 2020-01-23
Implement a status polling loop

hategan:
This PR is in the #1420 series. It allows multiple consumers to access executor/provider status by adding an independent status polling loop whose results are then used as needed.


parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,62 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PollItem(ExecutorStatus):
+    def __init__(self, executor: ParslExecutor):
+        self._executor = executor
+        self._interval = executor.status_polling_interval
+        self._last_poll_time = 0.0
+        self._status: Dict[object, JobStatus] = {}
```
benclifford: This type annotation syntax is only in python >=3.6. Use  something like `# type: Dict[object, JobStatus]` on the end of the line instead

hategan: Thanks. Wasn't 3.6 a minimum requirement for running parsl?

hategan: Just asking. It's a straightforward change that I can do once this call I'm in ends (or maybe even before).

yadudoc: 3.5 is the minimum. I just made an issue about pushing for 3.6+ requirement, but that will take time.

yadudoc:
Seems to be failing in mypy checks, I've got a log snippet here: https://gist.github.com/yadudoc/c0e39493f4efb3885d0be0a9d6a4ec37

hategan:
> Seems to be failing in mypy checks, I've got a log snippet here: https://gist.github.com/yadudoc/c0e39493f4efb3885d0be0a9d6a4ec37

Yes. The patch adds type annotations to the signature of _strategy_simple. This turns on type checking on that method, which, in turn uncovers various existing problems.

It is my understanding, from a previous discussion that we had, that it would not be preferable to fix such problems in the same PR where they were incidentally uncovered (aka. unless they were introduced by other commits in that PR).

parsl/providers/cobalt/cobalt.py
```diff
@@ -228,3 +228,7 @@ def cancel(self, job_ids):
             rets = [False for i in job_ids]
 
         return rets
+
+    @property
```
yadudoc: Could be moved to the parent ClusterProvider

parsl/providers/ad_hoc/ad_hoc.py
```diff
@@ -255,3 +255,7 @@ def scaling_enabled(self):
     @property
     def label(self):
         return self._label
+
+    @property
+    def status_polling_interval(self):
```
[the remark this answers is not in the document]
hategan: cmd_timeout: how long before commands used to interact with the scheduler are considered to have hanged when invoked
status_polling_interval: how frequently a specific command should be run in order to query the status of jobs)

Aren't those two completely unrelated things? On what basis would one ever assume that one can be derived from the other?

parsl/providers/cobalt/cobalt.py
hategan: In practice, probably. In theory, it seems like not something that should be abstracted away.

benclifford:
see draft PR #1554 for some work that I've done on more typing for this on the `benc-mypy` branch.

parsl/providers/ad_hoc/ad_hoc.py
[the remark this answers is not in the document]
yadudoc: They are unrelated until the polls hit the timeouts repeatedly. I guess this is fine since the polls are blocking, and we aren't going to add to the stress on a backed up batch-scheduler.

parsl/dataflow/task_status_poller.py
```diff
@@ -0,0 +1,62 @@
...
+
+from parsl.dataflow.executor_status import ExecutorStatus
+from parsl.dataflow.strategy import Strategy
+from parsl.executors.base import ParslExecutor
+from parsl.providers.provider_base import JobStatus
+
+logger = logging.getLogger(__name__)
+
+
+class PollItem(ExecutorStatus):
```
yadudoc: While this is fine and works great, a simpler approach would be to add a timed cache decorator in the provider, since the provider also holds the timeout period.

hategan: So I gave this some considerable thought. But before I write about that, let me see if I understand you correctly. The following would help:
- do you mean discarding the centralized polling loop altogether or keeping the polling loop but having caching done on provider.getStatus()?
- do you have a specific caching decorator in mind, or did you mean that something can be written if needed?

yadudoc: Central polling is better than each provider polling separately. 
Unfortunately `functools.lru_cache` doesn't have a timer for staleness, so we might have to do something like this : 

```python
import functools
import time

class TimedCache():
    def __init__(self, valid_time=2):                
        self.valid_time = valid_time
        print("valid_time = {}".format(valid_time))        

    def __call__(self, func):        
        print("here")        
        def wrapper(*args, **kwargs):
            
            if time.time() > self.cache_time + self.valid_time:
                print("Stale")
                self.cached_result = func(*args, **kwargs)
                self.cache_time = time.time()                
            else:
                print("Fetching cached")
                    
            return self.cached_result
        
        self.cache_time = 0
        self.cached_result = None
        return wrapper

@TimedCache(60)
def status( ...)
```

hategan: by "central polling" I mean one polling loop for all consumers (strategy, error handling, etc.) vs. each of (strategy, error handling, etc.) calling executor.status() -> provider.status() on their own.

Anyway, the scheme in this PR is fundamentally single-threaded. There is one thread that maintains the status of the system and it dispatches that status to various consumers who deal with a snapshot of the status of the system. It's true that it requires some help from the providers in determining how often certain things should be done.

(1) We could keep the polling loop as it is, remove PollItem and get_polling_interval() methods, implement a cache as above, and have the polling loop call status() on all providers at some regular interval that is smaller than the smallest polling interval (this will be a somewhat magical decision, since the smallest polling interval would not be known to the loop). This is almost fine.

(2) We could also abandon the polling loop entirely and have the strategy and whatever other entities call status() on their own, knowing that the caching decorator does the right thing. The above decorator does not do the right thing. It does not deal with being invoked from multiple threads. If invoked in quick succession without having a cached result, it will pass all the calls through. A proper implementation would determine if an invoking thread is the first to trigger an update, stall all other calling threads while the update happens, and finally return the cached result to all waiting threads when it becomes available. That's not that hard to implement, but it's nowhere near the simplicity of not implementing it at all. The cost, I think, is much higher than explicitly exposing the polling interval. We could, of course, also go with a good enough implementation that doesn't do things quite right. I think that's just incurring technical debt. We'll end up with a method (status()) that seems OK to call freely, but it isn't quite so, and the problems arising from it would be subtle and hard to figure out once we forget about this little conversation. It would be like a button that says "press me", but which would, occasionally, make things fail in subtle ways. In other words, not a good idea. So anyway, I think we'd trade explicitness for hidden complexity. I'd rather have the former.

Even if we keep one loop (1), I don't think that's much different than (2). It would still involve a button that says "press me" which either fails in subtle ways or it's not worth making it function correctly.

yadudoc: I understand. I'm working with the assumption that there's one central polling loop, which is why I haven't added any locking for thread safety to the decorator in the stale case. We should note in the docs that the status is not thread safe, and is not intended for the user to call when strategy is present.

I think 1 is reasonable and gets the job done.

To questions in 2:  
* assume one thread
* provider status polls block, so you can't call it again from 1 thread
* thread safety is not a good assumption in python anyway, so let's not add complexity until it's necessary. Adding a lock in the update method isn't hard.

Let's have a quick chat to make sure we are both on the same page.

yadudoc: Here are some conclusions from an offline discussion: 
* decorator approach vs the PollItem are different implementations of the same functionality
* we want to keep the providers as simple and stateless as possible
* state/cache should be held in a centralized place and the PollItems do that.

PR #1554, opened 2020-01-30
Extracts of provider/executor typing from benc-mypy branch for #1550

benclifford:
@hategan for the type errors in #1550, here some extracts from the `benc-mypy` branch laid on top of the current #1550 head.

Those changes apply fairly cleanly on top of `master` too.

These changes use 3.6 syntax for properties (because I target `benc-mypy` at 3.7 minimum) so if you want to make a real PR based on this, those would need porting to 3.5. It's not a high priority for me to introduce stronger type checking into `master`, though.

I basically did this:

```
git diff -r origin/master...benc-mypy parsl/executors/base.py | patch -p1
git diff -r origin/master...benc-mypy parsl/providers/provider_base.py | patch -p1
```

and fixed up merge conflicts and add one ugly `cast` which would maybe be better implemented some other way (eg using isinstance and an appropriate superclass; or moving that property to be mandatory)

I'm making this a PR to see how it checks out in CI.


benclifford:
this PR isn't intended for merge, so closing - it was for showing a code extract from elsewhere

PR #1561, opened 2020-02-06
Fixes walltime format for GridEngineProvider

### LINKED ISSUE #1556: GridEngine provider mis-parses wall times, treating minutes as seconds
**Describe the bug**
The GridEngine system expects walltime in seconds, but parsl provides `10`, which leads to difficult-to-recognize errors relating to walltime errors.

**To Reproduce**
1. Setup Parsl 0.9.0 with Python 3.6 on cluster
2. Run a sleep job with GridEngine provider
3. Watch the `qstat`, where you will see jobs disappear

**Expected behavior**
10 minute default walltime

**Environment**
 - OS: CentOS 7
 - Python version 3.6
 - Parsl version 0.9

**Distributed Environment**
- Where are you running the Parsl script from ? Login node

- Where do you need the workers to run ?Worker nodes on same network as login node

yadudoc:
GridEngine takes an `h_rt` attribute that expects HH:MM:SS format or seconds. 
Previously we incorrectly passed it walltime in minutes resulting in jobs terminating early.

Unfortunately the GridEngine scheduler did not give any warnings on hitting the walltime, and allowed this bug to go on undetected.

@benhg I'd appreciate it if you can test this out on your cluster.

Fixes #1556

### DIFF
--- a/parsl/providers/grid_engine/grid_engine.py
+++ b/parsl/providers/grid_engine/grid_engine.py
@@ -7,7 +7,7 @@ from parsl.providers.cluster_provider import ClusterProvider
 from parsl.providers.grid_engine.template import template_string
 from parsl.launchers import SingleNodeLauncher
 from parsl.providers.provider_base import JobState, JobStatus
-from parsl.utils import RepresentationMixin, wtime_to_minutes
+from parsl.utils import RepresentationMixin
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


benhg:
Thanks for this quick fix !! I will test it out this evening, after my current run is finished

benhg:
I can verify that this works as expected on Grid Engine 8.1.9

PR #1563, opened 2020-02-08
Support for scaling down in HTEX

yadudoc:
Adds support to HTEX for scaling down idle blocks. `Config.max_idletime` will now be applied to each block individually when `Config.strategy='htex'` is set.

The driving use-case for this feature is from ArcticDem (@clairecporter) where there are several large blocks requested, each filled with a task that consumes the whole block (assume MPI). These tasks have wide ranging walltimes, and since in this case tasks map to blocks, it naturally leads to idle blocks without having to worry about task de-fragmentation, towards the end of the campaign.


parsl/dataflow/strategy.py
```diff
@@ -269,3 +271,131 @@ def _strategy_simple(self, tasks, *args, kind=None, **kwargs):
...
+            elif active_slots == 0 and active_tasks > 0:
+                # Case 4
+                # Check if slots are being lost quickly ?
+                logger.debug("Requesting single slot")
+                if active_blocks < max_blocks:
+                    executor.scale_out(1)
+
+            # Case 4
+            # More slots than tasks
+            elif active_slots > 0 and active_slots > active_tasks:
```
yadudoc: Use idle counter here to make sure we do not kill while tasks might be in flight

ZhuozhaoLi:
I have tested several configs for this branch (using the test in this PR with minor changes) and it worked fine. I will continue testing this branch.

It seems `scale_in(force=True)` is never used. But @yadudoc can you remind me what is blocking this from being reviewed?

benclifford:
@ZhuozhaoLi this branch is pretty out of date wrt master - it forked from master in February:

```
commit a33d465b8024d261b1fbe79cafc506053386fcef
Author: Yadu Nand Babuji <yadudoc1729@gmail.com>
Date:   Mon Feb 3 12:10:49 2020 -0600

    Cleaner support for wiping tasks after completion from DFK (#1543)
```

It is probably better testing this against a more up-to-date master, because other stuff has changed in strategy behaviour since that merge point.

ZhuozhaoLi:
> @ZhuozhaoLi this branch is pretty out of date wrt master - it forked from master in February:
> 
> ```
> commit a33d465b8024d261b1fbe79cafc506053386fcef
> Author: Yadu Nand Babuji <yadudoc1729@gmail.com>
> Date:   Mon Feb 3 12:10:49 2020 -0600
> 
>     Cleaner support for wiping tasks after completion from DFK (#1543)
> ```
> 
> It is probably better testing this against a more up-to-date master, because other stuff has changed in strategy behaviour since that merge point.

Agreed. That was my plan too.

parsl/executors/high_throughput/executor.py
```diff
@@ -491,6 +492,11 @@ def connected_managers(self):
         workers = self.command_client.run("MANAGERS")
         return workers
 
+    @property
+    def block_status(self):
```
yadudoc, line 496: Let's just get the `idle_duration` from the `MANAGERS` command itself. That would be simpler.

parsl/executors/high_throughput/interchange.py
```diff
@@ -300,6 +301,21 @@ def _command_server(self, kill_event):
                                 'active': self._ready_manager_queue[manager]['active']}
                         reply.append(resp)
 
+                elif command_req == "BLOCKS":
```
yadudoc, line 304: Some of this logic would be moved to `MANAGERS`, with `idle_duration` reporting 0 when manager is not idle, and a float otherwise, in seconds.

parsl/executors/high_throughput/interchange.py
```diff
@@ -501,6 +519,8 @@ def start(self, poll_period=None):
 
                     self.results_outgoing.send_multipart(b_messages)
                     logger.debug("[MAIN] Current tasks: {}".format(self._ready_manager_queue[manager]['tasks']))
+                    if len(self._ready_manager_queue[manager]['tasks']) == 0:
```
yadudoc, line 522: Ok, this logic makes sense. The only point where a manager can go idle is when a result is returned and #tasks == 0

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,42 @@ def scale_in(self, blocks=None, block_ids=[]):
...
+                if not manager['active']:
+                    continue
+                b_id = manager['block_id']
+                if b_id not in blks:
+                    blks[b_id] = [0, float('inf')]
+                blks[b_id][0] += manager['tasks']
+                blks[b_id][1] = min(blks[b_id][1], manager['idle_duration'])
+
+            sorted_blks = sorted(blks.items(), key=lambda item: (item[1][1], item[1][0]))
+            logger.warning("YADU: sorted_block : {}".format(sorted_blks))
```
yadudoc, line 633: Remove log line.

parsl/dataflow/strategy.py
```diff
@@ -119,7 +119,9 @@ def __init__(self, dfk):
         for e in self.dfk.config.executors:
             self.executors[e.label] = {'idle_since': None, 'config': e.label}
 
-        self.strategies = {None: self._strategy_noop, 'simple': self._strategy_simple}
+        self.strategies = {None: self._strategy_noop,
+                           'simple': self._strategy_simple,
+                           'htex': self._strategy_htex}
```
benclifford, line 124: "htex" isn't a very user friendly name for a strategy

parsl/dataflow/strategy.py
```diff
@@ -275,3 +277,132 @@ def _strategy_simple(self, status_list, tasks, *args, kind=None, **kwargs):
...
+            self.unset_logging()
+
+            # FIXME we need to handle case where provider does not define these
+            # FIXME probably more of this logic should be moved to the provider
+            min_blocks = executor.provider.min_blocks
+            max_blocks = executor.provider.max_blocks
+            if isinstance(executor, HighThroughputExecutor):
+
+                tasks_per_node = executor.workers_per_node
+            elif isinstance(executor, ExtremeScaleExecutor):
```
benclifford, line 313: this if will never fire, because an ExtremeScaleExcecutor is a HighThroughputExecutor and will be captured by the previous case.

parsl/dataflow/strategy.py
```diff
@@ -275,3 +277,132 @@ def _strategy_simple(self, status_list, tasks, *args, kind=None, **kwargs):
...
+                tasks_per_node = executor.ranks_per_node
+
```
benclifford, line 315: add default behaviour (error?) when neither of the above cases fire and task_per_node is left undefined

parsl/dataflow/strategy.py
```diff
@@ -275,3 +277,132 @@ def _strategy_simple(self, status_list, tasks, *args, kind=None, **kwargs):
...
+                    # logger.debug("Strategy: Case.2b")
+                    excess = math.ceil((active_tasks * parallelism) - active_slots)
+                    excess_blocks = math.ceil(float(excess) / (tasks_per_node * nodes_per_block))
+                    excess_blocks = min(excess_blocks, max_blocks - active_blocks)
+                    logger.debug("Requesting {} more blocks".format(excess_blocks))
+                    executor.scale_out(excess_blocks)
+
+            elif active_slots == 0 and active_tasks > 0:
+                # Case 4
+                # Check if slots are being lost quickly ?
```
benclifford, line 390: what does this comment mean?

parsl/dataflow/strategy.py
```diff
@@ -275,3 +277,132 @@ def _strategy_simple(self, status_list, tasks, *args, kind=None, **kwargs):
...
+                logger.debug("Requesting single slot")
+                if active_blocks < max_blocks:
+                    executor.scale_out(1)
+
+            # Case 4
+            # More slots than tasks
+            elif active_slots > 0 and active_slots > active_tasks:
+                logger.warning("YADU: More slots than tasks")
+                if isinstance(executor, HighThroughputExecutor):
+                    blocks = {}
```
benclifford, line 400: `blocks` never gets assigned?

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,42 @@ def scale_in(self, blocks=None, block_ids=[]):
         if block_ids:
             block_ids_to_kill = block_ids
         else:
-            block_ids_to_kill = list(self.blocks.keys())[:blocks]
-
+            managers = self.connected_managers
+            blks = {}
```
benclifford, line 622: use whole words

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,42 @@ def scale_in(self, blocks=None, block_ids=[]):
...
+            for manager in managers:
+                if not manager['active']:
+                    continue
+                b_id = manager['block_id']
+                if b_id not in blks:
+                    blks[b_id] = [0, float('inf')]
```
benclifford, line 628: more readable with a dict than [0], [1]?

parsl/dataflow/strategy.py
yadudoc: I think we are better off raising an exception if the executor isn't of type HTEX. What do you think ?

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,42 @@ def scale_in(self, blocks=None, block_ids=[]):
...
+                blks[b_id][0] += manager['tasks']
+                blks[b_id][1] = min(blks[b_id][1], manager['idle_duration'])
+
+            sorted_blks = sorted(blks.items(), key=lambda item: (item[1][1], item[1][0]))
+            logger.warning("YADU: sorted_block : {}".format(sorted_blks))
+            if force is True:
+                block_ids_to_kill = [x[0] for x in sorted_blks[:blocks]]
+            else:
+                if not max_idletime:
+                    block_ids_to_kill = [x[0] for x in sorted_blks[:blocks] if x[1][0] == 0]
```
benclifford, line 638: I think this can select fewer than `blocks` blocks because it only examines the first `blocks` blocks to see if they are without tasks. (truncate first, then filter).

Maybe it should filter first, then truncate, which is what the `if max_idletime` clause below does.

parsl/dataflow/strategy.py
yadudoc: Stale comment from when the method was copied over. I'll go through and trim old comments.

It meant that if there are >1 tasks and active slots=1, it might be useful to check if the blocks are failing.

parsl/executors/high_throughput/executor.py
```diff
@@ -599,6 +599,15 @@ def scale_in(self, blocks=None, block_ids=[]):
         blocks : int
              Number of blocks to terminate and scale_in by
 
+        force : Bool
+             Used along with blocks to indicate whether blocks should be terminated by force.
+             When force = True, we will kill blocks regardless of the blocks being busy
+             When force = False, we will kill only idle blocks
```
benclifford, line 605: When `force` is `False`, then `scale_in` will sometimes return without killing the requested number of blocks, and sometimes will kill no blocks at all, right?

parsl/dataflow/strategy.py
```diff
@@ -275,3 +277,129 @@ def _strategy_simple(self, status_list, tasks, *args, kind=None, **kwargs):
...
+                if active_blocks < max_blocks:
+                    executor.scale_out(1)
+
+            # Case 4
+            # More slots than tasks
+            elif active_slots > 0 and active_slots > active_tasks:
+                logger.debug("More slots than tasks")
+                if isinstance(executor, HighThroughputExecutor):
+                    if active_blocks > min_blocks:
+                        executor.scale_in(1, force=False, max_idletime=self.max_idletime)
```
benclifford, line 400: self.max_idletime comes from config.max_idletime which is documented in its docstring (not in this diff) as being for the simple strategy, not for this strategy - could fix that

parsl/tests/manual_tests/test_scaling.py
```diff
@@ -0,0 +1,82 @@
+import parsl
```
benclifford, line 1: Run this test in CI - it takes 25s on my laptop which I think is time worth spending.

parsl/tests/manual_tests/test_scaling.py
```diff
@@ -0,0 +1,82 @@
...
+    assert len(dfk.executors['htex_local'].connected_managers) == 2, "Expected 2 managers once tasks are running"
+    print("Between")
+    print("Managers   : ", dfk.executors['htex_local'].connected_managers)
+    print("Outstanding: ", dfk.executors['htex_local'].outstanding)
+
+    fus[2].result()
+    fus[3].result()
+    time.sleep(10)
+    [x.result() for x in fus]
+    # assert len(dfk.executors['htex_local'].connected_managers) == 1 , "Expected 1 manager cleaning up"
```
benclifford, line 70: is this assertion meant to hold here?

benclifford:
The new strategy introduced is so close in implementation to the existing `simple` strategy (a single clause different, I think?) that it might be nice to figure out how to make them the same strategy?

parsl/dataflow/strategy.py
benclifford: I don't have a clear idea of what to do when a strategy goes wrong... strategy exceptions at the moment propagate upwards into flow control and raise this:
```
    logger.error("Flow control callback threw an exception - logging and proceeding anyway", exc_info=True)
```

It bugs me that this doesn't work for WQ, which also has providers and scale_out/in behaviour. That's probably something for another time, though.

parsl/dataflow/strategy.py
yadudoc: Updated config docs.

parsl/executors/high_throughput/executor.py
yadudoc: Yes. That's right, do you think the text needs updating to clarify this further ?

parsl/tests/manual_tests/test_scaling.py
yadudoc: I believe so. Let me add that back. Once the first 4 short tasks are done, one manager should be idle, and after a 10s gap, only one should be still connected.

parsl/tests/manual_tests/test_scaling.py
ZhuozhaoLi: I used this test when I was debugging. This test passed when there were multiple bugs. 

For instance:
1. This test checked the min_blocks at the end of test, but that is actually resulted from case 1 in strategy (when there is no task), not case 4
2. This test does not check if the strategy follows `max_idletime` to kill blocks (from case 4 in strategy).

I had another test in this PR that is a bit more comprehensive, but that one takes ~50s. I could try to change the time to make it shorter.

parsl/executors/high_throughput/executor.py
benclifford: yeah - it wasn't obvious to me

parsl/tests/manual_tests/test_scaling.py
benclifford: I think 50s is worth spending too.

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,41 @@ def scale_in(self, blocks=None, block_ids=[]):
...
+                if not max_idletime:
+                    block_ids_to_kill = [x[0] for x in sorted_blocks[:blocks] if x[1][0] == 0]
+                else:
+                    block_ids_to_kill = []
+                    for x in sorted_blocks:
+                        if x[1][1] > max_idletime and x[1][0] == 0:
+                            block_ids_to_kill.append(x[0])
+                            if len(block_ids_to_kill) == blocks:
+                                break
+                logger.warning("Selecting block ids to kill since they are idle : {}".format(
```
benclifford, line 645: `.debug`, not `.warning` ?

parsl/executors/high_throughput/executor.py
```diff
@@ -609,14 +618,41 @@ def scale_in(self, blocks=None, block_ids=[]):
...
+                    block_ids_to_kill))
+
+        logger.warning("CUrrent blocks : {}".format(self.blocks))
```
benclifford, line 648: debug not warning

benclifford: also, case typo `CUrrent` with a capital U

benclifford:
Suppose I have a job that in its first part has 10 blocks worth of tasks (eg 1000 tasks), and then in its second part has 2 blocks worth of tasks (eg 200 tasks).  One expectation of scaling down is that you'd end up with those 200 tasks on 2 blocks and the other 8 blocks would get scaled down. But potentially those 200 tasks are scattered uniform randomly across all 10 blocks, and so none of those blocks can end. I don't think there's a mechanism to make the first outcome happen, not the second.

benclifford:
If I have 2 blocks running, and 2 blocks in the queue not running, and I have enough tasks to run on 3 blocks, the new strategy tries a scale down; but htex does not pick a block to scale down: the 2 blocks running are not idle, and the two blocks in the queue don't seem to be regarded as "idle". Maybe the strategy should remove one of the queue blocks?

benclifford:
I had a talk with @yadudoc  about this yesterday. The two unsupported cases I wrote above are from test runs of the LSST DRP workflow that I've been working on. I think fundamentally this PR isn't any use in the case of that workflow without other work to allow blocks to empty out (for example, the ability to place new tasks "most full block first"?)

yadudoc:
As @benclifford pointed out, this PR only adds the ability to scale_down if and when some blocks go idle. For this feature to become really useful, we need better task packing so that they are not evenly spread across nodes/blocks but consolidated. For this to work we need some task walltime estimates to pack better along the walltime dimension.

@ZhuozhaoLi is interested in doing this sort of smarter scheduling, but that work is out of scope for this PR.

parsl/executors/high_throughput/executor.py
yadudoc: @benclifford Can you check the docstring, please ?

benclifford: i think you should clarify that with force = false, this call will complete without killing the specified number of blocks and without throwing an error.

PR #1564, opened 2020-02-10
Fix "Task N waiting on" message when task isn't waiting on anything

benclifford:
Before this commit, the message was awkwardly phrases as "waiting on" and then nothing further.

  Task 35 submitted for App echo_to_streams, waiting on

Now the message looks like:

  Task 35 submitted for App echo_to_streams, not waiting on any dependency


yadudoc:
I think this looks good, but I'll leave this one to our log lines expert @annawoodard :)

Issue #1548, opened 2020-01-21
Remove IPyParallel executor


benclifford:
**Is your feature request related to a problem? Please describe.**
The parsl ipp code isn't really maintained any more. As I try to tidy up some stuff in the test suite I find myself spending more time than I'd like to debugging what is happening inside the IPP  parsl executor, while at the same time the folklore is "don't use IPP".

**Describe the solution you'd like**
I'd like to remove all IPP tests from the test suite now. This will inevitably lead to the ipp code rotting fast, and so in addition the IPP executor should probably be removed in its entirely.

**Describe alternatives you've considered**
An alternative is to put developer effort into maintaining IPP as a supported executor.

**Additional context**
The folklore on slack seems to be "don't use IPP". This feature request turns that folklore into reality.

I invite comment in support or against from @kylechard @yadudoc @annawoodard especially as regards how this would affect existing users who have not moved to htex (if any?).

see #1328

annawoodard:
I strongly support removing IPP.

yadudoc:
My vote is on removing IPP from tests and clarifying in docs that it is no longer supported. I would defer removing it entirely from the codebase for later.

annawoodard:
What is the point of having something that isn't being tested (and therefore we can assume isn't working 😆) in the codebase?

yadudoc:
I think it would be nicer to the users to have a release go out that warns anyone using IPP of the impending deprecation before we drop it entirely. Same process as planned for #1553

annawoodard:
Sounds good, but after we depreciate, let's open an issue to remind us to remove it after the depreciation period has ended.

benclifford:
> I think it would be nicer to the users to have a release go out that warns anyone using IPP of the impending deprecation before we drop it entirely.

If we're intending users to use it, we should be testing it. Otherwise, what is the response to broken IPP when a user reports it?

annawoodard:
Thinking about it more, I agree with Ben. I think these are our two options:

1. Add depreciation warnings in next release, continue testing until the following release, when we completely remove IPP and IPP testing.
2. Add retroactive depreciation warnings (if someone tries to use IPP, we tell them to install the previous release) and remove IPP and IPP testing now.

I slightly favor option 2.

kylechard:
I vote for option 2. If people really need to use IPP then they can use an old version of Parsl until its time to update. I can't think of a single situation where IPP would work and HTEX would not.

benclifford:
[...]

PR #1565, opened 2020-02-12
Remove IPyParallel executor

benclifford:
This commit removes most of the IPyParallel executor.

Stub classes are left in place which will raise an error if instantiated, to give users a clue what has happened.

See #1548.


parsl/executors/ipp.py
```diff
@@ -1,303 +1,12 @@
...
-from parsl.executors.ipp_controller import Controller
-from parsl.executors.status_handling import StatusHandlingExecutor
-from parsl.providers import LocalProvider
-from parsl.utils import RepresentationMixin
-from parsl.utils import wait_for_file
 
-logger = logging.getLogger(__name__)
+class IPyParallelExecutor(ParslExecutor):
+    """This stub exists to issue a more helpful warning about the IPyParallel
+    executor being removed from parsl some time afer v0.9.
```
annawoodard: afer -> after

parsl/executors/ipp_controller.py
```diff
@@ -1,166 +1,8 @@
-import logging
-import os
-import random
-import signal
-import subprocess
-import time
+class Controller():
+    """This stub exists to issue a more helpful warning about the IPyParallel
+    executor being removed from parsl some time afer v0.9.
```
annawoodard: afer -> after

PR #1572, opened 2020-02-15
Move retries_left to task record, instead of on the executor Futures

benclifford:
This should not change behaviour in usual operation, but moves
the retries_left information to the task record along with other
task related information.

It also makes type-checking with mypy easier for the Futures returned
by executors as there is no longer an expectation that arbitrary
Futures will have retries_left attributes.


benclifford:
yeah I think with things re-arranged it makes sense to look at what the task record variables should be

Issue #1553, opened 2020-01-27
Make Python3.6 the minimum requirement for Parsl


yadudoc:
**Is your feature request related to a problem? Please describe.**
Supporting Py3.5 has slowly become a hassle. For eg. these have changed for the better:
1. fspath objects are only supported in 3.6+
2. fstrings are supported in 3.6+.
3. Bunch of improvements in the typing module 

**Describe the solution you'd like**
Push requirements to 3.6+

**Additional context**
The daily download proportion plot shows that Python3.6 downloads make up the majority of the downloads. This shows that only a tiny fraction of the user base might be affected by discontinuing support for 3.5.
https://pypistats.org/packages/parsl
This requires feedback from users.

benclifford:
We could make `master` now strongly warn users of 3.5 that this version is deprecated, and then end support for 3.5 in two releases time.

danielskatz:
This is a great idea, and a good policy for use to implement more generally

annawoodard:
I support depreciating 3.5

kylechard:
I agree we should deprecate 3.5. I like Ben's idea of noting now that 3.5 support will end in one-two releases time.

benclifford:
PR #1557 adds a deprecation warning issued at DFK initialization on Python < 3.6, stating that Parsl 0.10 will be the last version supporting earlier versions.

After the 0.10 release, this warning could turn into a RuntimeException for one or more future releases, and perhaps be eventually removed.

yadudoc:
Assuming the next release is 1.0 and the one after will be 1.1, I'm retagging this for 1.1.

benclifford:
Docs also contain at least one reference to python 3.5 specific behaviour (in docs/futures.rst). When 3.5 support is removed, grep for `3.5` in the docs and tidy.

benclifford:
done in  #1753, #1758

PR #1583, opened 2020-03-01
Fix Torque job_id truncation

glentner:
This is a simple naive fix to a critical issue with the Torque provider.

The implementation of the Torque provider keeps track of running jobs using the output it got from the `qsub` command which gives a full hostname. But the status method runs `qstat` whose output typically provides truncated job_id values.

In my case (on Purdue's Brown cluster):

```
$ qsub ...
6058223.brown-adm.rcac.purdue.edu
```
```
$ qstat 6058223.brown-adm.rcac.purdue.edu ...

Job ID                    Name             User            Time Use S Queue
------------------------- ---------------- --------------- -------- - -----
6058223.brown-adm         ...96831.6737266 glentner               0 Q debug
```

This results in an almost immediate `KeyError` being raised on startup and a full panic.

All this fix does to resolve the issue is to check existing "long form" job_id values to see if it _starts with_ the job_id it got from `qstat`. Since Torque job ID values are always NUM.HOSTNAME format I think this is pretty safe. I'm open to suggestions though.

Note: We're almost done migrating all our clusters at Purdue to Slurm. I tried to get this tested before I didn't have Torque available -- a time horizon now measured in days! 🤣

### DIFF
--- a/parsl/providers/torque/torque.py
+++ b/parsl/providers/torque/torque.py
@@ -124,7 +124,7 @@ class TorqueProvider(ClusterProvider, RepresentationMixin):
             job_id = parts[0]  # likely truncated
             for long_job_id in job_ids:
                 if long_job_id.startswith(job_id):
-                    logger.debug(f'coersed jobid {job_id} -> {long_job_id}')
+                    logger.debug('coerced job_id %s -> %s', job_id, long_job_id)
                     job_id = long_job_id
                     break
             state = translate_table.get(parts[4], JobState.UNKNOWN)
--- a/parsl/providers/torque/torque.py
+++ b/parsl/providers/torque/torque.py
@@ -111,6 +111,7 @@ class TorqueProvider(ClusterProvider, RepresentationMixin):
               [status...] : Status list of all jobs
         '''
 
+        job_ids = list(self.resources.keys())
         job_id_list = ' '.join(self.resources.keys())
 
         jobs_missing = list(self.resources.keys())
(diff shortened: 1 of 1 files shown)


parsl/providers/torque/torque.py
```diff
@@ -120,7 +121,12 @@ def _status(self):
             parts = line.split()
             if not parts or parts[0].upper().startswith('JOB') or parts[0].startswith('---'):
                 continue
-            job_id = parts[0]
+            job_id = parts[0]  # likely truncated
+            for long_job_id in job_ids:
+                if long_job_id.startswith(job_id):
+                    logger.debug(f'coersed jobid {job_id} -> {long_job_id}')
```
benclifford: f-strings are not valid in python 3.5 (which Parsl plans to support for the next version or two - see https://github.com/Parsl/parsl/issues/1553). This makes this code fail the test suite (see the big red X next to your commit).

benclifford:
I feel like it would be better if torque could be made to output job IDs in a consistent format between the two commands, rather than encoding some equivalence rules inside parsl. But if there isn't an obvious way to do that (and I didn't find any) then I think the way this PR does it is ok.

glentner:
I agree; that was my initial thought as well.  We talked about it internally here at Purdue and I’m told this is just how Torque behaves.

parsl/providers/torque/torque.py
glentner: Oh okay, that’s an easy fix. Actually, we don’t necessarily even need to have that logging statement. I was just using during tests.

benclifford: spell coerced with a *c*

PR #1601, opened 2020-03-09
Remove unused retries field from AppFailure exception

benclifford:


benclifford:
This is superceded by #1602

PR #1616, opened 2020-03-23
Simplify return type of check_memo

benclifford:
There is no need to separately indicate if the memo
result is present and return that memo result.

A memo result is always a Future, so usual
Future or None protocol can be used.

This should not change the behaviour of memoization.

This comes from work on benc-mypy branch.


benclifford:
> If the result from the app is `None` this code change won't be able to differentiate. That's why this code is structured this way. Maybe we should add a note in comments to clarify.

what comes out in that case is a `Future[None]`, not a `None` I think? (I should test)

yadudoc:
@benclifford  You are right, the memoization table holds futures rather than the PyObj results.

Issue #1444, opened 2019-11-14
LocalProvider+LocalChannel discards stdout/stderr of launched command


benclifford:
**Describe the bug**
LocalProvider+LocalChannel directs stdout/stderr of the submitted command into a pipe that is then not used.

This means that stdout/err from the command is lost.

**To Reproduce**
Run HTEX with LocalProvider+LocalChannel+SingleNodeLauncher. The "Found cores :" message from SingleNodeLauncher does not appear in a log file.

**Expected behavior**
This output should not be discarded. For example, with slurm, the output of the executed batch job goes into a log file alongside the batch job submit file.

**Environment**
parsl master ea8333c3249fdf3b21691b7a66fb4e9c8ed37ec0

benclifford:
the plan in #1448 should resolve this, as the non-LocalChannel code path directs output to a file

WardLT:
Following up on the reference in #1620, not using the buffered stdout/stderr can even lead to stalling if the command produces enough output.

The stalling is particularly common when using Tensorflow-enabled apps, which produce a lot of stderr.

PR #1620, opened 2020-03-24
Route stdout/err to DEVNULL

WardLT:
The PIPE was never read from, leading to stalling if the subprocess creates too much output

### DIFF
--- a/parsl/channels/local/local.py
+++ b/parsl/channels/local/local.py
@@ -77,10 +77,10 @@ class LocalChannel(Channel, RepresentationMixin):
             logger.warning("Execution of command '{}' failed due to \n{}".format(cmd, e))
             raise
 
-        return (retcode, stdout.decode("utf-8"), stderr.decode("utf-8"))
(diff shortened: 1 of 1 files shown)


benclifford:
https://github.com/Parsl/parsl/pull/1471 should remove the execute_no_wait code paths entirely. I don't know if we have enough review bandwidth to merge it ever, so don't hold your breath.

WardLT:
Thanks for the quick reply, Ben.

I'll keep using my fork for now and gladly wait for #1471.

Do you think it's worthwhile to just close this PR to avoid confusion? I can open an issue instead to give other users guidance if their apps are stalling.

benclifford:
There's a related issue here: https://github.com/Parsl/parsl/issues/1444 which talks about *discarding* output into a pipe; but doesn't record the fact that when the pipe gets full, stuff hangs. Maybe that's the place.

benclifford:
closing in favour of #1471

PR #1626, opened 2020-03-25
Rearrange bash app output checking codepath

benclifford:
Previously this code assumed that a output was either:

* a string

or

* something with a .filepath attribute

and would fail if an output was anything else.

Presently, outputs of a bash app are always File objects
(with a .filepath attribute, and never anything else).

This commit tidies that codepath.

### DIFF
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -97,9 +97,7 @@ def remote_side_bash_executor(func, *args, **kwargs):
 
     missing = []
     for outputfile in kwargs.get('outputs', []):
-        fpath = outputfile
-        if type(outputfile) != str:
-            fpath = outputfile.filepath
+        fpath = outputfile.filepath
 
         if not os.path.exists(fpath):
             missing.extend([outputfile])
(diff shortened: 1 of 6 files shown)


benclifford:
This needs tests to use new-style File() outputs, not strings in outputs. which in turn needs test suite to know how to checkpoint File parameters. I have that prototyped somewhere.

parsl/app/bash.py
```diff
@@ -97,9 +97,7 @@ def open_std_fd(fdname):
 
     missing = []
     for outputfile in kwargs.get('outputs', []):
-        fpath = outputfile
-        if type(outputfile) != str:
-            fpath = outputfile.filepath
+        fpath = outputfile.filepath
```
annawoodard, line 100: Now that we're really requiring this not be a string I think we need to adjust the datafutures description [here](https://github.com/Parsl/parsl/blob/master/docs/userguide/futures.rst) where we say:
```
hello = echo('Hello World!', outputs=['hello1.txt'])
```

benclifford: see #1640

PR #1636, opened 2020-04-09
Added developer friendly Makefile; added ssh_keyfile name to SSHChannel 

lhayhurst:
This commit does two things.

First, it adds a developer friendly Makefile at the root of the project. When run with no args, it will show what targets it offers:

```
$ make      
clean                          remove venv and flush out
coverage                       show the coverage report
deploy                         deploy the distribution
deps                           install the dependencies
flake                          run flake
help                           me
htex_local_test                run all tests with htex_local config
lint                           run linter script
local_thread_test              run all tests with local_thread config
package                        package up a distribution.
release                        create a release. To run, do a 'make VERSION="version string"  release'
tag                            create a tag in git. to run, do a 'make VERSION="version string" tag
test                           run all tests
```

The targets has sensible dependencies; ie, a `make test` will build the python environment for you (via `make deps`). 

Right now,`make test` only runs the `local_thread_test`s and the `htex_local_test` configs. More work can be done to add the other config types as make targets (see the .travis.yml file for all the options). 

For the `make release` workflow, the `tag_and_release.sh` script now passes shellcheck. It's usage has changed to support the new make targets; before, you'd just run the script, now you run the script with the function to call as an argument. 

2) `ssh_keyfile` is now being passed into SSHChannel, and passed along to the Paramiko connect. There is an argument to be made for passing this in via `**kwargs`; this is the minimum change needed.

Please note that it appears that the integration scripts to test this sort of thing are run manually; I'll be testing this tomorrow morning in my local uat environment, so please hold off merging until then in case this PR sails through :-) 

Final thing to note, running all the tests dumps all sorts of .out, .err, and .txt files in various places from the current working directory where make is invoked. Probably all the tests should output to just one place, but this is a larger bit of work to cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,5 +77,4 @@ coverage: test ## show the coverage report
 
 .PHONY: clean
 clean: ## remove venv and flush out
-	rm -rf $(VENV) dist *.egg-info .mypy_cache
-	find . -name __pycache__ | xargs --no-run-if-empty rm -rf
+	rm -rf $(VENV) dist *.egg-info .mypy_cache build .pytest_cache .coverage
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,81 @@
+SHELL := $(shell which bash) # Use bash instead of bin/sh as shell
+SYS_PYTHON := $(shell which python3.7 || echo ".python_is_missing")
+GIT := $(shell which git || echo ".git_is_missing")
+VENV = .venv
+VENV_BIN := $(VENV)/bin
+PIP := $(VENV_BIN)/pip
+DEPS := $(VENV)/.deps
+PYTHON := $(VENV_BIN)/python3
+PYTEST := $(VENV_BIN)/pytest
+export PATH := $(VENV_BIN):$(PATH)
+
+.PHONY: help
+help: ## me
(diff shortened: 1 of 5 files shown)


lhayhurst:
build failed on azure 

```
Collecting azure (from parsl==0.9.0)
  Downloading https://files.pythonhosted.org/packages/e8/48/ef73d5e71ecb84cacaeaf533f519be81825ad6198acc41bd821f440e7f7c/azure-5.0.0.zip
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-txihrfjo/azure/setup.py", line 60, in <module>
        raise RuntimeError(message)
    RuntimeError:

    Starting with v5.0.0, the 'azure' meta-package is deprecated and cannot be installed anymore.
    Please install the service specific packages prefixed by `azure` needed for your application.
```

benclifford:
This really needs to be two PRs because it makes two totally separate changes, with separate reviews.

benclifford:
see PR #1637 for azure dependency

lhayhurst:
I'll split this into 2 PRs, thank you for fb.

PR #1638, opened 2020-04-09
Project master makefile

lhayhurst:
This PR adds a developer friendly Makefile at the root of the project. When run with no args, it will show what targets it offers:

```
$ make      
clean                          remove venv and flush out
coverage                       show the coverage report
deploy                         deploy the distribution
deps                           install the dependencies
flake                          run flake
help                           me
htex_local_test                run all tests with htex_local config
lint                           run linter script
local_thread_test              run all tests with local_thread config
package                        package up a distribution.
release                        create a release. To run, do a 'make VERSION="version string"  release'
tag                            create a tag in git. to run, do a 'make VERSION="version string" tag
test                           run all tests
```

The targets has sensible dependencies; ie, a `make test` will build the python environment for you (via `make deps`). 

Right now,`make test` only runs the `local_thread_test`s and the `htex_local_test` configs. More work can be done to add the other config types as make targets (see the .travis.yml file for all the options). 

For the `make release` workflow, the `tag_and_release.sh` script now passes shellcheck. It's usage has changed to support the new make targets; before, you'd just run the script, now you run the script with the function to call as an argument. 

Note: running all the tests dumps all sorts of .out, .err, and .txt files in various places from the current working directory where make is invoked. Probably all the tests should output to just one place, but this is a larger bit of work to cleanup.


lhayhurst:
Note, there is a potential issue here, which is that now there are don't-repeat-yourself violations between the travis.yml file and this Makefile. 

Probably the correct way to do this is to have the travis `script` target call `make test`. the Makefile's `make test` target needs to do everything the travis script target call is doing though!

lhayhurst:
This is in a pretty good state now, in that the D.R.Y. issue is addressed.

note that the build is failing for some dependency reasons that don't seem related to this change. Otherwise the tests seem to be running well.

PR #1639, opened 2020-04-09
Add key_filename parameter to SSHChannel

lhayhurst:
This PR adds the `key_filename` name to the SSHChannel params, allowing callers to take advantage of Paramiko's `key_filename` connect() param.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ extras_require = {
     'docs' : ['nbsphinx', 'sphinx_rtd_theme'],
     'google_cloud' : ['google-auth', 'google-api-python-client'],
     'gssapi' : ['python-gssapi'],
-    'azure': ['azure<=4', 'msrestazure'],
+    'azure' : ['azure<=4', 'msrestazure'],
     'workqueue': ['work_queue'],
 }
 extras_require['all'] = sum(extras_require.values(), [])
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ extras_require = {
     'docs' : ['nbsphinx', 'sphinx_rtd_theme'],
     'google_cloud' : ['google-auth', 'google-api-python-client'],
     'gssapi' : ['python-gssapi'],
-    'azure' : ['azure', 'msrestazure'],
+    'azure': ['azure<=4', 'msrestazure'],
     'workqueue': ['work_queue'],
 }
 extras_require['all'] = sum(extras_require.values(), [])
--- a/docs/userguide/data.rst
+++ b/docs/userguide/data.rst
@@ -122,7 +122,7 @@ Thus, we can write programs such as the following that take a file on one Globus
         f.result()
 
 Configuration
-^^^^^^^^^^^^^
+"""""""""""""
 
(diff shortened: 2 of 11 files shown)


yadudoc:
@lhayhurst Might be easier to scrap that last commit and just merge in changes from master. Currently it looks like there's a merge conflict from a whitespace

setup.py
```diff
@@ -25,7 +25,7 @@
     'docs' : ['nbsphinx', 'sphinx_rtd_theme'],
     'google_cloud' : ['google-auth', 'google-api-python-client'],
     'gssapi' : ['python-gssapi'],
-    'azure' : ['azure', 'msrestazure'],
```
yadudoc, line 28: Here's the whitespace change that's causing a merge conflict.

lhayhurst:
Cool will fix ty

lhayhurst:
Should be good to go now, just one file changed.

Before you do, though, @benclifford  chatted me in slack this afternoon and said:

```
You said in other comment somewhere that it might be useful to pass in **kwargs all the way through and maybe that's a good idea because it is annoying repeatedly going through "this underlying lib has parameter X that isn't exposed one level up"
```

So maybe there is some conversation to be had here.

yadudoc:
Now that I think about this, this is the second time that I can remember having to make updates to this interface.@lhayhurst I think @benclifford 's suggestion to pass through kwargs is the right fix here.  What do you think?

lhayhurst:
For the sake of discussion, here's a few (potential) downsides to consider: 

1) If parsl ever migrates off of Paramiko, you'll have to maintain all of Paramiko's connect() args in order to stay compatible, because they are being **kwarg'd in. Or adapt all those argument's to whatever you move to. 

2) If kwargs is blithely passed into connect(), there's a chance the user provided you some params that actually don't work for Paramiko's interface. This may have been part of the motivation for @benclifford 's[ commit 13 days ago](https://github.com/Parsl/parsl/commit/fe31ac5cface54862f5143c9157820334cc11cdd#diff-3c7a0c6cde9a251077adcaca97e904a6) that removed kwargs from the __init__? 

Given that these are optional arguments that don't effect existing clients, I think the more cautious approach is change the interface one param at a time.

benclifford:
This breaks when trying to print an SSHChannel using RepresentationMixin.

```
> ssh.SSHChannel("172.17.0.1")
[...]
AttributeError: class SSHChannel uses key_filename in the constructor, but does not define it as an attribute
```

I got that at a python interactive prompt, but I'd (probably?) expect it to happen at parsl initialization when it logs out the current full configuration to `parsl.log`.

RepresentationMixin wants every parameter passed into the constructor to be stored as an attribute so that it can be printed out again when calling __repr__.

benclifford:
I've tested this works from a python REPL, but not inside parsl:

```
>>> c = ssh.SSHChannel("172.17.0.1", key_filename=["./tmpkg"])
>>> c.execute_wait("uptime")
(0, ' 13:59:37 up 27 days,  3:50, 22 users,  load average: 1.35, 1.13, 1.80\n', '')
```

lhayhurst:
I've fixed the RepresentationMixin error.

PR #1643, opened 2020-04-15
Added 'pull_file' to the channel interface and to the local channel i…

hategan:
…mplementation

### DIFF
--- a/parsl/channels/base.py
+++ b/parsl/channels/base.py
@@ -71,6 +71,19 @@ class Channel(metaclass=ABCMeta):
         '''
         pass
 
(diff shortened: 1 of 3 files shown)


hategan:
This is needed if we ever want to get back those stdout/stderr files using channels.

parsl/channels/base.py
```diff
@@ -71,6 +71,19 @@ def push_file(self, source, dest_dir):
         '''
         pass
 
+    def pull_file(self, remote_source, local_dir):
```
benclifford, line 74: I think this should be an abstract method - PR #1645 adds that decorator.

PR #1644, opened 2020-04-15
Client side poll item updates

### LINKED ISSUE #1632: Scaling up beyond what is necessary and beyond max_blocks
**Describe the bug**
Once a block has been launched, the simple scaling strategy will not see that that block exists until after one executor poll period. The strategy will continue to launch blocks every 5 seconds until that poll period has been reached and statuses updated.

I have experienced this on cori using slurm. @annawoodard might also have experienced it.

It looks like this was introduced in PR #1550 which reworked status polling.

**To Reproduce**
#1621 provides a pytest case that runs without needing a batch system

**Expected behavior**

**Environment**
#1621 shows on my laptop. I've experienced it trying to upgrade lsst workflow to recent master.

hategan:
Should fix #1632

### DIFF
--- a/parsl/dataflow/task_status_poller.py
+++ b/parsl/dataflow/task_status_poller.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Sequence
 from parsl.dataflow.executor_status import ExecutorStatus
 from parsl.dataflow.strategy import Strategy
 from parsl.executors.base import ParslExecutor
-from parsl.providers.provider_base import JobStatus
+from parsl.providers.provider_base import JobStatus, JobState
 
 logger = logging.getLogger(__name__)
 
@@ -34,6 +34,20 @@ class PollItem(ExecutorStatus):
     @property
     def executor(self) -> ParslExecutor:
         return self._executor
+
+    def scale_in(self, n):
+        ids = self._executor.scale_in(n)
+        if ids is not None:
+            for id in ids:
+                del self._status[id]
+        return ids
+
+    def scale_out(self, n):
+        ids = self._executor.scale_out(n)
+        if ids is not None:
+            for id in ids:
+                self._status[id] = JobStatus(JobState.PENDING)
+        return ids
 
     def __repr__(self):
         return self._status.__repr__()
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -1,5 +1,6 @@
 import logging
 import threading
+from itertools import compress
 from abc import abstractmethod
 from concurrent.futures import Future
 from typing import List, Any, Dict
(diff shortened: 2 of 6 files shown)


yadudoc:
The `parsl/tests/test_flowcontrol/test_one_block.py` from Ben's #1621 is now passing on this branch.

parsl/dataflow/task_status_poller.py
```diff
@@ -35,6 +35,20 @@ def status(self) -> Dict[object, JobStatus]:
...
 
+    def scale_in(self, n):
+        ids = self._executor.scale_in(n)
+        if ids is not None:
+            for id in ids:
+                del self._status[id]
+        return ids
+
+    def scale_out(self, n):
+        ids = self._executor.scale_out(n)
```
yadudoc, line 46: Is None returned if there was a scale_out error ? Just trying to understand if anything needs to happen here if the scale out failed.

parsl/dataflow/task_status_poller.py
```diff
@@ -35,6 +35,20 @@ def status(self) -> Dict[object, JobStatus]:
     def executor(self) -> ParslExecutor:
         return self._executor
 
+    def scale_in(self, n):
+        ids = self._executor.scale_in(n)
+        if ids is not None:
```
yadudoc, line 40: `executor.scale_in` should return a list of ids that were removed, it shouldn't return `None`. If it failed to remove the blocks it is supposed to return an empty list.

parsl/executors/status_handling.py
```diff
@@ -87,6 +87,17 @@ def tasks(self) -> Dict[object, Future]:
     def provider(self):
         return self._provider
 
+    def _filter_scale_in_ids(self, to_kill, killed):
+        assert len(to_kill) == len(killed)
+
+        result = []
+
+        for i in range(len(killed)):
```
yadudoc, line 95: If this is filtering, it might be clearer to use itertools.compress : 
```
from itertools import compress
result = list(compress(to_kill, killed)) # Filters first iterable by bool values in second.
```

hategan: Sure. I started that way, but then I didn't think "compress" was standard FP naming (Ben can correct me if I'm wrong), and thought that being clear is better than being compact. But it looks like a carefully placed comment may do the trick.

hategan: To be clear, by "that way" I meant

 `reduce(lambda l, x: l + [x[0]] if x[1] else l, zip(to_kill, killed), [])`

This is clearly not the same as using `compress`, whose name also doesn't have to match what I thought was akin to a `filter` function, but isn't.

benclifford: I haven't seen `compress` in other FP situations.

I don't have a lot of opinion either way, but I think I slightly prefer the compress style...

yadudoc: @hategan Can we do compress with a comment to clarify?

annawoodard:
Note I have been running with this for a few days and things look good so far.

PR #1645, opened 2020-04-16
Make channel.pull_file an @abstractmethod like other channel methods

benclifford:
This will change channel implementations behaviour, so that subclasses
without a pull_file will error rather than silently skip file transfer.
I think this does not affect any existing channel.


hategan:
My bad. I meant it as abstract.

PR #1665, opened 2020-05-05
Results from zombie managers

yadudoc:
Catching the case of results being received from a manager that is marked dead due to missing heartbeats.

In the new mode, we will ignore results from a zombie manager. This is done since the assumption is that at the point where a manager is called dead, we send back notifications of the tasks being lost. Those tasks will then be relaunched. We should not send conflicting results back to the interchange.


benclifford:
I think it would be nicer to not make a broken manager data structure in response to surprisingly ordered message processing, rather than catch weird exceptions that happen later due to having created a broken internal state. (and I say that as the person that prototyped the catch/ignore approach used in this PR)

benclifford:
(if the manager is gone, don't create a new manager structure for it and then get upset that the new structure is broken)

yadudoc:
@benclifford  I've updated this PR so that we do not add a register managers unless a registration message is received. We'd receive the result messages, but ignore it.

PR #1667, opened 2020-05-06
Adding new configurable option address_probe_timeout to HTEX.

yadudoc:
* Updates the default address probe timeout from 2s to 30s
* Adds a configurable option for htex to specify the timeout.

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -73,7 +73,7 @@ class Manager(object):
              comma separated list of addresses for the interchange
 
         address_probe_timeout : int
-             Timeout is seconds for the address probe to detect viable addresses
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -68,6 +69,13 @@ def __init__(self,
         """
         Parameters
         ----------
+        addresses : str
+             comma separated list of addresses for the interchange
+
+        address_probe_timeout : int
+             Timeout is seconds for the address probe to detect viable addresses
```
benclifford, line 76: *in* seconds, not *is* seconds

benclifford: fixed it myself

benclifford:
In my test environment, I set this to 0 to see if it got an error without waiting for any results. That is how it behaved.

Issue #1669, opened 2020-05-06
Version mismatches are getting suppressed


yadudoc:
**Describe the bug**
#1357 changed the default behavior to suppress failures in the interchange's handling of messages
from the managers. However, this was very broad. While this change was made to broadly ignore garbled messages, and messages from managers marked as dead, this had the unintended effect of also masking version mismatches between the interchange and the workers.

**To Reproduce**

Steps to reproduce the behavior, for e.g:
1. Update worker config to load python3.7
2. Run parsl script with python3.6

**Expected behavior**
Parsl should shutdown with an exception pointing out the version mismatch in the env.

**Environment**
 - OS: [e.g. ubuntu, centos, MacOS, windows]
 - Python version
 - Parsl version

benclifford:
#1537 is a code-of-conduct change, not htex

yadudoc:
Sorry. fixed in original post to https://github.com/Parsl/parsl/pull/1357

PR #1671, opened 2020-05-07
Remove suppress failures #1669

yadudoc:
* `dfk`  does not call `shutdown` or `scale_in`, if the executor is in a bad_state to avoid hangs.
* `suppress_failures` option is now removed, this used to cover the following situations:
* manager registration fails dues to version mismatch -> Now a critical failure
* manager registration fails due to bad message -> logged and ignored
* interchange received a spurious message -> logged and ignored 
* `address_probe_timeout` is not specified to the manager if user does not specify explicitly for better backward compatibility.

The last item can be separated to a different PR.

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -948,7 +948,7 @@ class DataFlowKernel(object):
         self.flowcontrol.close()
 
         for executor in self.executors.values():
-            if executor.managed:
+            if executor.managed and not executor.bad_state_is_set:
                 if executor.scaling_enabled:
                     job_ids = executor.provider.resources.keys()
                     executor.scale_in(len(job_ids))
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -138,9 +138,6 @@ class HighThroughputExecutor(StatusHandlingExecutor, RepresentationMixin):
         Managers attempt connecting over many different addesses to determine a viable address.
         This option sets a time limit in seconds on the connection attempt. Default is 30s.
 
-    suppress_failure : Bool
-        If set, the interchange will suppress failures rather than terminate early. Default: True
-
     heartbeat_threshold : int
         Seconds since the last message from the counterpart in the communication pair:
         (interchange, manager) after which the counterpart is assumed to be un-available. Default: 120s
@@ -177,7 +174,6 @@ class HighThroughputExecutor(StatusHandlingExecutor, RepresentationMixin):
                  heartbeat_period: int = 30,
                  poll_period: int = 10,
                  address_probe_timeout: int = 30,
-                 suppress_failure: bool = True,
                  managed: bool = True,
                  worker_logdir_root: Optional[str] = None):
 
@@ -226,7 +222,6 @@ class HighThroughputExecutor(StatusHandlingExecutor, RepresentationMixin):
         self.heartbeat_threshold = heartbeat_threshold
         self.heartbeat_period = heartbeat_period
         self.poll_period = poll_period
-        self.suppress_failure = suppress_failure
         self.run_dir = '.'
         self.worker_logdir_root = worker_logdir_root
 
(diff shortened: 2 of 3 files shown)


yadudoc:
Fixes #1669

hategan:
Hmm, I clicked that by accident, how do I undo it?

yadudoc:
@hategan It is ready for review, thanks for switching the state.

hategan:
I accidentally did the right thing? What are the odds?!

hategan:
From what I understand, a spurious connection or a version mismatch would still lead to _kill_event being set, which would bring down the whole interchange, no?

yadudoc:
Fixed. I've updated the handling of spurious message / registration failed conditions to not set the kill_event or send an exception notice to the executor. It logs the message and continues.

hategan:
... and the version mismatch?

Also, question: are there replies? Aka, if a manager of questionable version connects and gets booted by this code, will it know about it?

yadudoc:
Version mismatch triggers an exception on the parsl client side. 

There is no reply to the registration message, also there won't be any heartbeats which is how managers will drop off.

hategan:
> Version mismatch triggers an exception on the parsl client side.

And brings down the interchange through _kill_event. I don't think it should, unless all managers are of the wrong version. 

> 
> There is no reply to the registration message, also there won't be any heartbeats which is how managers will drop off.

A twisted way of achieving the right thing by doing the wrong thing. That's a handshake and it fails. The manager should know about it. And it should fail immediately. Then, with proper error handling, the executor should detect if all managers have failed and take appropriate action if necessary.

yadudoc:
@hategan, Since all managers that connect to an interchange are created via the same config, it is reasonable to assume that if the env on one is broken, it is broken on the others as well. Here it makes sense to exit early. Secondly how does the interchange know "all" managers are bad ? It has no idea of how many to expect.

If the workers are not on the same env, then there's the possibility of python serialization itself not being reliable or the worker might not have the same shutdown methods. I think the heartbeat method works reliably and in a general enough way. Don't fix something that isn't broken ?

hategan:
> @hategan, Since all managers that connect to an interchange are created via the same config, it is reasonable to assume that if the env on one is broken, it is broken on the others as well.

In Swift, we've had issues with workers from other runs connecting by accident. So I don't think it's safe to make the assumptions that you are making, although it may be the most common case.


> Here it makes sense to exit early. Secondly how does the interchange know "all" managers are bad ?

I said: I don't think it should [set _kill_event], *unless* all managers are of the wrong version.

If it cannot assert that all managers are of the wrong version, then it should *not* shut down the whole interchange. Point being, don't use _kill_event except when shutting everything down.

> It has no idea of how many to expect.
> 
> If the workers are not on the same env, then there's the possibility of python serialization itself not being reliable or the worker might not have the same shutdown methods. I think the heartbeat method works reliably and in a general enough way. Don't fix something that isn't broken ?

Just because it rolls down a hill, reliably as it may,, it doesn't mean it's not broken. Not sure what you're saying there, but if the manager can't properly communicate with the interchange because of a version mismatch or a serialization issue, state clearly that that's the case and don't send the user on a wild goose chase of figuring out why heartbeats are getting lost.

I'm not saying do the whole handshake in this PR. Eventually. But I think you should remove the _kill_event getting set in response to a message from a single manager. It's not the idea of fault tolerance that I would expect.

yadudoc:
@hategan, Parsl has port selection logic that avoid multiple runs sharing ports. So cross talk between runs will not happen. 
We want to cover for the most common failure mode, which is version mismatch due to botched conda envs. The right behavior is to shutdown early.
To the last point, I don't want to fix something that isn't broken. If we want to kill the managers faster, we can force the provider to scale down, rather than add more code complexity for little value. Either way, that should be covered in a separate PR.

hategan:
I'm really unsure what thought process leads to you chose to have a network client just sit there wasting resources when it fails to connect or call that behavior not broken.

PR #1675, opened 2020-05-10
add resource specification kwarg

ZhuozhaoLi:
This PR is a followup on #1358. It includes:

- exposing an interface to specify a `parsl_resource_specification` kwarg per app.
- logging a warning for executors that do not support resource specification yet and pointing the users to `WorkQueueExecutor`


parsl/dataflow/dflow.py
```diff
@@ -729,6 +729,8 @@ def submit(self, func, app_args, executors='all', fn_hash=None, cache=False, ign
                                     kw)
                     )
 
+        resource_specification = app_kwargs.pop('parsl_resource_specification', {})
```
benclifford, line 732: Other references to parsl-specific keyword args use `get` or `[]`, rather than `pop` which removes `parsl_resource_specification` from the dictionary.

The behaviour should be consistent unless there is a clear reason to make the behaviour different.

parsl/executors/base.py
```diff
@@ -41,7 +41,7 @@ def start(self) -> None:
         pass
 
     @abstractmethod
-    def submit(self, func: Callable, *args: Any, **kwargs: Any) -> Future:
+    def submit(self, func: Callable, resource_specification: Dict, *args: Any, **kwargs: Any) -> Future:
```
benclifford, line 44: Is there more description of what goes in `resource_specification` than a plain Dict? For example, is it a dictionary of strings -> anything? (`Dict[str,Any]`) or strings -> strings? (`Dict[str,str]`)

parsl/executors/high_throughput/executor.py
```diff
@@ -524,6 +524,11 @@ def submit(self, func, *args, **kwargs):
         Returns:
               Future
         """
+        if len(resource_specification) > 0:
+            logger.warning("Ignoring the resource specification. "
+                           "Parsl resource specification is not supported in HighThroughput Executor. "
+                           "Please check WorkQueueExecutor if resource specification is needed.")
```
benclifford, line 530: I would prefer that this raises an error, rather than logs a message: the user is asking the executor to do a thing that it cannot do.

parsl/dataflow/dflow.py
ZhuozhaoLi: It is `pop` here because `parsl_resource_specification` cant be a kwarg passed to the app---it will raise an exception of unknown kwarg

benclifford: > parsl_resource_specification cant be a kwarg passed to the app---otherwise it will raise an exception of unknown kwarg

That is the same behaviour as `stdout`, `stderr` and `label`. What is the story to tell a user that `label` or `stdout` needs to be declared, but `parsl_resource_specification` doesn't?

benclifford:
I'm also uncomfortable with the usability issues that come from this parameter being a dictionary. Some of the arguments around using object based config for parsl configs, rather than dicts, might apply here: For example, how does parsl help me debug that I put in (for example) `walltime` instead of `wall_time` (or some other wrong parameter like that) or that I used a parameter that doesn't exist at all?

parsl/dataflow/dflow.py
ZhuozhaoLi: I see your point now. I changed to `get`

dgasmith:
I like Pydantic in these situations:

```python
from typing import Optional
from pydantic import BaseModel, ByteSize, PositiveInt

class ResourceSpec(BaseModel):
    ncores: PositiveInt
    wall_time: Optional[str] = None
    memory: ByteSize = -1

    class Config:
        extra = "forbid"
```

I never built one out, but you can have an object like `TimeSecond` or whatnot which converts `5hours` to a common unit.

```python
>>> print(ResourceSpec(ncores=1, memory='1gb'))
ncores=1 wall_time=None memory=1000000000

>>> print(ResourceSpec(ncores=1, memory=1e6, walltime=5))
  File "tmp.py", line 14, in <module>
    print(ResourceSpec(ncores=1, walltime=5))
  File "pydantic/main.py", line 338, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for ResourceSpec
walltime
  extra fields not permitted (type=value_error.extra)
```

parsl/executors/high_throughput/executor.py
```diff
@@ -524,6 +529,12 @@ def submit(self, func, *args, **kwargs):
         Returns:
               Future
         """
+        if len(resource_specification) > 0:
```
yadudoc, line 532: Since resource spec is a dict, we can say `if resource_specification` since an empty dict evals to False.

parsl/executors/errors.py
```diff
@@ -30,6 +30,24 @@ def __repr__(self):
                                                                 self.nodes_per_block)
 
 
+class UnsupportedFeatureError(ExecutorError):
+    """Error raised when attemping to use unsupported feature in an Executor"""
+
+    def __init__(self, feature, current_executor, target_executor):
+        self.feature = feature
+        self.current_executor = current_executor
+        self.target_executor = target_executor
```
yadudoc, line 39: If we have a complex feature matrix, maintaining what target_executor(s) are supported might become annoying.
This is definitely useful to the user.

yadudoc:
@ZhuozhaoLi Let's have a typeguard do the basic check on resource_spec with `Dict[str, Any]` for now, since this is not finalized. We should switch to a better system like @dgasmith pointed out using pydantic or a proper object, for the next release.

ZhuozhaoLi:
> @ZhuozhaoLi Let's have typeguard do the basic check on resource_spec with `Dict[str, Any]` for now, since this is not finalized. We should switch to a better system like @dgasmith pointed out using pydantic or a proper object, for the next release.

I change it to `Dict[str, Any]`  for typeguard to do the basic check

parsl/tests/test_error_handling/test_resource_spec.py
```diff
@@ -0,0 +1,30 @@
+import parsl
+from parsl.app.app import python_app
+from parsl.tests.configs.local_threads import config
+from parsl.executors.errors import UnsupportedFeatureError
+from parsl.executors import WorkQueueExecutor
+
+local_config = config
```
benclifford, line 7: is this intended to be a local test or a test run on arbitrary configs? (the second, i think)

ZhuozhaoLi: Yes, the second. I am removing this line.

PR #1677, opened 2020-05-10
Add support for Work Queue auto labeling resources

trshaffer:
This PR adds options to the Work Queue executor to turn on automatic resource labeling for submitted tasks. WQ can collect stats from running tasks to automatically determine resource requirements. Another PR (#1675) is adding explicit resource labels. This feature is useful if the resource requirements aren't known in advance or the user doesn't want to worry about it, but still wants to subdivide large workers.

This PR also introduces a complementary option to automatically assign categories to functions, with each defined function placed in its own category. WQ manages resources separately for different categories, so this complements the auto-labeling feature. If different functions have different resource requirements, WQ will measure and label each function category separately.

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -155,6 +155,7 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
             input_files = item["input_files"]
             output_files = item["output_files"]
             std_files = item["std_files"]
+            category = item["category"]
 
             full_script_name = workqueue_worker.__file__
             script_name = full_script_name.split("/")[-1]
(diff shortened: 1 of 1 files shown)


dthain:
Q that came up in Monday meeting: can this feature be tested robustly in CI?  Discussion was inconclusive.

parsl/executors/workqueue/executor.py
```diff
@@ -101,6 +104,10 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
         q.specify_password(project_password)
     elif project_password_file:
         q.specify_password_file(project_password_file)
+    if autolabel:
+        q.enable_monitoring()
+        q.tune('category-steady-n-tasks', 1)
```
ZhuozhaoLi, line 109: As mentioned in the meeting, maybe make this `1` configurable.

trshaffer:
Rebased off Ben's changes, so this should be good to go

parsl/executors/workqueue/executor.py
```diff
@@ -191,6 +201,13 @@ def WorkQueueSubmitThread(task_queue=multiprocessing.Queue(),
                 logger.error("Unable to create task: {}".format(e))
                 continue
 
+            if not autocategory or category is None:
```
yadudoc, line 204: For a tiny bit more consistency you could say `autocategory is False`

trshaffer: Yeah, that ended up looking awkward as things changed... Just fixed it

Issue #1228, opened 2019-08-27
workqueue tests don't run with session-scoped DFK


benclifford:
The workqueue executor does not always seem to work running multiple tasks with a DFK instance.

This has come up from work on PR #1210 which rearranged tests to use a single DFK for a whole session instead of a new DFK for each `test_*` function; but I have also seen this fail when running a command line parsl test. The present test driver contains a workqueue specific bodge to work around this.

This command usually fails on a test:

```pytest parsl -k "not cleannet and not noci" --config parsl/tests/configs/workqueue_ex.py --cov=parsl --cov-append --cov-report= --random-order```

This command, which is what is used in CI now and which starts up one DFK for each test function, does not:

```
pytest parsl -k "not cleannet and not noci" --config parsl/tests/configs/workqueue_ex.py --cov=parsl --cov-append --cov-report= --random-order --bodge-dfk-per-test
```

I haven't dug much into this but sometimes I've seen the code hit this codepath in the WQ executor:

```
                            else:
                                reason += "unable to process wrapper script failure with status = {}".format(status)
```

with `status = -1`.

dthain:
@tjdasso adding you to this conversation.  Next thing to look at after your current WQ PR gets finalized.

benclifford:
@tjdasso @dthain let me know if you need more info about what is actually going on here on the parsl side

tjdasso:
@benclifford could you give a general rundown of what goes on when the tests run, and what the current WQ workaround does to provide a temporary fix?

benclifford:
@tjdasso what happened before recent rearrangements in #1210 was that each test file got an individual `pytest` invocation, in a for-loop in .travis.yaml.

This incurred a bunch of startup time and the test suite took many minutes to run.

PR #1210 changed things so that pytest created a single DFK at the start of the test run, ran all the tests using that DFK, and then shut down that DFK.

This means that the whole stack of DFK and everything underneath it (including in the case of this issue, the workqueue executor and workqueue worker executable) get a different workload: instead of getting a very small number of tasks (often, only one), the entire set of pytest tasks goes to the same DFK/executor.

This doesn't work for all configurations ... but it should!

Notably it doesn't work for exex (at least for me) which is issue #1229 even though exex is tested for a small number of tasks elsewhere in the test suite.

And it doesn't work for workqueue (at least for me).

[...]

PR #1680, opened 2020-05-11
Cleanup function serialization in workqueue executor

btovar:
Moves the serialization code into its own function. Pickled functions are now identified as source or byte code. This simplifies their execution in the wq worker wrapper.


btovar:
This pr is towards a solution to #1228.

parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -59,15 +57,19 @@ def check_file(parsl_file_obj, mapping, file_type_string):
         input_function = open(input_function_file, "rb")
         function_info = pickle.load(input_function)
```
yadudoc, line 58: Worth adding a try catch here. If the python versions do not match, this load can fail.

parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -59,15 +57,19 @@ def check_file(parsl_file_obj, mapping, file_type_string):
...
             kwargs = function_info["kwargs"]
         # Extract information from function pointer
-        else:
+        elif "byte code" in function_info:
+            fn_from_source = False
             from ipyparallel.serialize import unpack_apply_message
-            func, args, kwargs = unpack_apply_message(function_info, user_ns, copy=False)
+            func, args, kwargs = unpack_apply_message(function_info["byte code"], user_ns, copy=False)
+        else:
+            raise Exception("Function file does not have a valid function representation.")
```
yadudoc, line 72: If the function fails at this point, how does the error notification propagate back to Parsl ?

btovar: I don't know if they do with the current version. I think that all exceptions should be reworked to match what parsl expects. Right now there is generic wq error exception, but as in this code snippet, this errors may be better communicated to parsl. Let me open an issue for that.

parsl/executors/workqueue/workqueue_worker.py
btovar: Sounds good!  I think that we have to rework this wrapper so that all exceptions are sent back to parsl as they should.

btovar: See #1688

parsl/executors/workqueue/workqueue_worker.py
btovar: See #1688

PR #1683, opened 2020-05-12
Adding equality/inequality operators for JobStatus

### LINKED ISSUE #1682: JobStatus objects failing equality checks
**Describe the bug**
We need to be able to check the status returned from the `provider.status()` and that is currently not possible now, for eq:

```
>>> from parsl.providers.provider_base import JobState, JobStatus
>>> JobStatus(JobState.RUNNING) == JobStatus(JobState.RUNNING)
False
```

**Expected behavior**
We need at least equality and inequality to work for JobStatus

yadudoc:
Adding __ne__ and __eq__ magic operators so that JobStatus objects can be compared.

Fixes #1682

### DIFF
--- a/parsl/providers/provider_base.py
+++ b/parsl/providers/provider_base.py
@@ -34,6 +34,12 @@ class JobStatus(object):
     def terminal(self):
         return self.state.terminal
 
+    def __eq__(self, other):
+        return (self.state) == other.state
+
+    def __ne__(self, other):
+        return (self.state) != other.state
+
     def __repr__(self):
         if self.message is not None:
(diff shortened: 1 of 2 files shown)


hategan:
Please don't. If you want to determine whether two jobs are in the same execution stage, use status1.state == status2.state. status1 should, in most cases, not be equal to status2 unless they are precisely the same object. The point is that there is potentially more information in status objects (stdout, stderr, exitcode, timestamps possibly, etc.) and equality is not that well defined.

yadudoc:
It's preferable to compare the state within the JobStatus object than to compare the objects directly. Closing.

PR #1684, opened 2020-05-12
cleanup stop condition of wait and submit threads in wq executor

btovar:
Simplifies stop conditions for wait and submit threads by using a single variable (should_stop) rather than two (continue_running and cancel_value).


btovar:
Given the changes in #1705, and the issue in #1704, it is easier to resubmit these changes in another pr.

PR #1693, opened 2020-05-14
Avoid overwriting launch_cmd in executor 

yadudoc:
Currently we overwrite launch_cmd with a formatted version of itself in the executor. This causes problems when the executor is reused between some tests.


benclifford:
I'm unclear that it is ever safe to re-use active pieces of a `Config` tree between different DFKs and have been working on the assumption that it is not.

That lead to the `fresh_config()` style of test configuration which constructs a new configuration on demand.

ZhuozhaoLi:
The CI tests failed---the one in `ExtremeScaleExecutor` also needs to be changed.

yadudoc:
As @benclifford pointed out, I've switched the configs in #1489 to use fresh config to avoid this issue. This PR is not necessary.

PR #1699, opened 2020-05-14
Replace executor __repr__ with executor label in periodic state report

benclifford:
Prior to this, I was seeing 38 lines of executor configuration sent to the
parsl log file every 5 seconds.

This commit adds python 3.5 compatible code to declare the type of label,
so that `make mypy` will pass. This code is a bit ugly but can be replaced
once we drop support for Python 3.5

### DIFF
--- a/parsl/executors/base.py
+++ b/parsl/executors/base.py
@@ -32,6 +32,16 @@ class ParslExecutor(metaclass=ABCMeta):
               @typeguard the constructor, you'll have to use List[Any] here.
     """
 
+    # This __init__ method exists to provide a python 3.5 compatible type declaration
+    # for the `label` attribute. If/when python 3.5 is deprecated (see parsl issue #1553)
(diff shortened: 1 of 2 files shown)


parsl/executors/base.py
```diff
@@ -32,6 +32,16 @@ class ParslExecutor(metaclass=ABCMeta):
...
     """
 
+    # This __init__ method exists to provide a python 3.5 compatible type declaration
+    # for the `label` attribute. If/when python 3.5 is deprecated (see parsl issue #1553)
+    # this could be replaced with a python 3.6 style attribute type declaration.
+    #   label: str
+    # and __init__ removed.
+
+    @abstractmethod
+    def __init__(self) -> None:
```
yadudoc, line 42: We key the executors on the label, so it would be very nice to make label a required argument once we switch to 3.6.

PR #1700, opened 2020-05-14
Test that parsl-visualize does not return HTTP errors

benclifford:
This wget recursively follows every link starting at the top level page of
a parsl-visualize web server, and checks that each of them returns with
a successful response, rather than a server error.

This commit also fixes some pages so that this test can pass, by replacing
some pages which returned a redirect (to add /) with the target of that redirect.


benclifford:
This fails in CI because the PR assumes that parsl has been installed in the default python environment, but recent Makefile based testing adds an extra layer of virtualisation. PR #1710 removes that extra layer, which should make this PR work.

benclifford:
This PR is currently failing CI because an incompatible version of the `click` package is installed.

This is logged but allowed to continue by an earlier CI step which reports:
> ERROR: oauth-ssh 0.11 has requirement click<7.0,>=6.7, but you'll have click 7.1.2 which is incompatible.

One requirement comes from Flask:

> Requirement already satisfied: click>=5.1 in /home/benc/parsl/virtualenv-3.7/lib/python3.7/site-packages (from Flask>=1.0.2->parsl==0.9.0) (7.0)

The other comes from oauth-ssh.

More broadly, I think this means that the two parsl features: oauth-ssh and monitoring are incompatible and cannot even both be installed together at the moment.

ZhuozhaoLi:
Can we fix click version to `click==0.7.0`? I just checked, it seems that flask only require `click>5.1`.
https://github.com/pallets/flask/blob/master/setup.py

benclifford:
It is probably the wrong thing to do to pin that in the parsl library requirements file - this isn't a requirement of the parsl library.

It might be ok to do that at the start of CI, and hope that `pip` provides a better dependency resolver in the future.

benclifford:
Removing 1.0 tag from this because it isn't urgent and packaging is awkward

Issue #1703, opened 2020-05-14
work queue executor's wrapper does not communicate correctly errors back to parsl


btovar:
**Describe the bug**
For example, if reading the function file fails because of pickling errors, the wrapper does not communicate the exception as such, but transforms it into a wrapper exception.

**To Reproduce**
1. Launch parsl with python 3.6 and wait for a task result using the work queue executor.
2. Launch a work_queue_worker that has python 3.8 in PATH.
3. Task will fail with WorkqueueTaskFailure, rather than PicklingError.

**Expected behavior**
Any exception related to the processing of the task should be communicated back to parsl.  WorkqueueTaskFailure should only be used for when there was no exception, but the task did not complete as expected (e.g., some output file was not generated, resources were exhausted, etc.)

**Environment**
 - OS: linux
 - Python version: For the example above, incompatible versions of python for main and worker processes.
 - Parsl version: At least from master commit 70d6f0f1, but much earlier.

**Distributed Environment**
- Using the work queue executor.

Issue #1694, opened 2020-05-14
work queue executor does not have type annotations


btovar:
**Describe the bug**
work queue executor does not have type annotations. This means that it always passes `make mypy` when testing the code, even when incorrect types are used.

**To Reproduce**
1. Checkout master branch commit 70d6f0f
2. Confirm that parsl/executors/workqueue/executor.py does not have type annotations.

**Expected behavior**
All functions and methods in the executor should have type annotations.

**Environment**
 - OS: all
 - Python version: all
 - Parsl version: at least from commit 70d6f0f, but much earlier.

**Distributed Environment**
It is independent of the distributed environment.

benclifford:
cross-reference #873 which is about putting @typeguard.typechecked annotations on any function call that directly faces a user

Issue #1704, opened 2020-05-14
work queue executor hangs when collector thread dies


btovar:
**Describe the bug**
If the collector thread, which converts parsl tasks into work queue tasks, dies, then the executor hangs. This is because the main thread first waits for the submission thread and then for the collector thread. If the submission thread never exits, then the executor never exits.

**To Reproduce**
Steps to reproduce the behavior, for e.g:
1. Add a syntax error to the collector thread.
2. Run the test suite of work queue.
3. The first task will fail, but the execution hangs.

**Expected behavior**
If the collector thread terminates, then whole executor should also terminate.

**Environment**
 - OS: linux
 - Python version: any
 - Parsl version: at least in master's commit 70d6f0f1, but much earlier.

**Distributed Environment**
- Using the work queue executor.

btovar:
The problem seems to be that the task futures are never updated, and thus the is an indefinite wait.

PR #1705, opened 2020-05-14
work queue executor: rework how file names are handled

btovar:
This pr is to solve issue #1228.

There was some name clashing when the executor was used across the different test sets. This pr changes the way the names of files that need to be staged in/out are handled. Now each tasks gets its own directory, and the remapping of .local_paths is done at the worker from a dictionary. Before, each name had to be tracked, and new names were generated on demand. However this logic was sometimes wrong, and it is much simpler to compute a single distinct name for the directory of the task.

To match this change:
- The staging of stdout and stderr was simplified to match that of the rest of the files.
- Files with url's but no .local_name's are not staged by work queue.
- The wrapper that executes the task at the worker was modified to receive the dictionary of names as pickled file. With this, the wrapper is always called as:  `workqueue_worker map_file function_file result_file`.
- work queue code is tested both with and without shared filesystems. For shared filesystems only tests marked with issue363 are considered.

Future pull requests will fix:
- Correct communication of wrapper exceptions to parsl (issue #1703)
- Add type information to user facing functions (issue #1694)
- When collector thread dies, the executor hangs (issue #1704)


parsl/executors/workqueue/workqueue_worker.py
```diff
@@ -1,51 +1,43 @@
...
-    # Obtain the local_path from the mapping
-    if type_desc == file_type_string:
-        if parsl_file_obj.filepath in mapping:
-            parsl_file_obj.local_path = mapping[parsl_file_obj.filepath]
+def load_pickled_file(filename):
+    try:
+        with open(filename, "rb") as f_in:
+            return pickle.load(f_in)
+    except pickle.PickleError:
+        print("Error while unplickling {}".format(filename))
```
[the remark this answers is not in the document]
btovar: Indeed! This is related to issue #1703.  Since the previous code sometimes missed to send the appropriate errors to parsl, I did not work to include a fix for that in this pr. Do you recommend to fix it in this pr?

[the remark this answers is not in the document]
btovar: Yadu, since the code changed a bit anyway, I am going to push the changes related to serializing the errors here. That way you don't have to re-read code that has changed.

btovar:
@yadudoc : It should fix #1703, #1704, and #1228.

I have a pr ready on top of this one to fix #1694.

btovar:
Working on the conflicts.

parsl/executors/workqueue/exec_parsl_function.py
```diff
@@ -0,0 +1,183 @@
...
+        pickle.dump(result_package, f_out)
+
+
+def remap_location(mapping, parsl_file):
+    if not isinstance(parsl_file, File):
+        return
+    # Below we rewrite .local_path when scheme != file only when the local_name
+    # was given by the main parsl process.  This is the case when scheme !=
+    # 'file' but .local_path (via filepath) is in mapping.
+    if parsl_file.scheme == 'file' or parsl_file.local_path:
```
benclifford, line 44: I'm concerned (though I haven't tested this) that this will break with eg HTTPInTaskStaging, which sets local path ahead of time to mean "the file will appear here on the execute side, and a wrapper will later do that staging" rather than "the file is here on the submit side for workqueue to stage in". Fixing that should probably wait for another round of thinking about data staging architecture with the WQ staging model better incorporated.

parsl/executors/workqueue/exec_parsl_function.py
```diff
@@ -0,0 +1,183 @@
...
+        #              this file.
+        try:
+            (map_file, function_file, result_file) = sys.argv[1:]
+        except ValueError:
+            print("Usage:\n\t{} function result mapping\n".format(sys.argv[0]))
+            raise
+
+        result = execute_function(map_file, function_file)
+    except Exception:
+        print("There was an error while setting up the function.")
```
benclifford, line 174: This exception clause also fires if there is a trigger in *executing* the function, not just in setting it up right?

parsl/executors/workqueue/exec_parsl_function.py
```diff
@@ -0,0 +1,183 @@
...
+        traceback.print_exc()
+        result = RemoteExceptionWrapper(*sys.exc_info())
+
+    # Write out function result to the result file
+    try:
+        dump_result_to_file(result_file, result)
+    except Exception:
+        print("Could not write to result file.")
+        traceback.print_exc()
```
benclifford, line 183: Should this exception clause make the process with a non-zero exit code? I think this `except` clause falls off the end and lets the process exit normally with a zero exit code.

parsl/executors/workqueue/executor.py
```diff
@@ -451,15 +124,21 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
...
             Choose whether to transfer parsl app information as
             source code. (Note: this increases throughput for
             @python_apps, but the implementation does not include
             functionality for @bash_apps, and thus source=False
             must be used for programs utilizing @bash_apps.)
-            Default is False.
+            Default is False. Set to True if pack is True
 
-        pack: bool
+         pack: bool
```
benclifford, line 141: spuriously introduced space?

parsl/executors/workqueue/executor.py
```diff
@@ -808,6 +411,55 @@ def _serialize_function(self, fn_path, parsl_fn, parsl_fn_args, parsl_fn_kwargs)
...
+        self.registered_files.add(parsl_file)
+
+        return ParslFileToWq(parsl_file.filepath, to_stage, to_cache)
+
+    def _std_output_to_wq(self, fdname, stdfspec):
+        """Find the name of the file that will contain stdout or stderr and
+        return a ParslFileToWq with it. These files are never cached"""
+        fname, mode = putils.get_std_fname_mode(fdname, stdfspec)
+        to_stage = not os.path.isabs(fname)
+        print(ParslFileToWq(fname, stage=to_stage, cache=False))
```
benclifford, line 460: use logger or remove?

benclifford:
> work queue code is tested both with and without shared filesystems. 

@btovar  is this a change introduced in this PR? I only see a CI config disabling issue363

parsl/executors/workqueue/exec_parsl_function.py
benclifford: (turns out it doesn't...)

I'm not super happy about the parsl-level story for how this is working though but I'll leave that for future work.

btovar: Got it. Work queue should be only remapping files that it staged. If it is not working, maybe it means that in some cases it is staging the wrong files. For example, if the .local_name set for execution also happens to exists at the main master process, but they refer to different contents. Is there enough information at the main parsl process to implement something like a `File.executor_should_stage()`?

benclifford: > Is there enough information at the main parsl process to implement something like a File.executor_should_stage()?

No. That's why I'm unhappy. it's not a workqueue problem.

btovar:
It was there, but I reverted it in 150a8e9d5d071a1c035e34539575f3539c8cb712 because of https://github.com/Parsl/parsl/issues/1709#issuecomment-629197549.

parsl/executors/workqueue/exec_parsl_function.py
btovar: Fixed in 4098f35c9628f6e78f1f92b3180a50d6b060cb09. Exec and load error messages are now separated.

parsl/executors/workqueue/executor.py
btovar: Fixed in a45233b8775652b5e7f5bf34df2c189a6f22bade

parsl/executors/workqueue/exec_parsl_function.py
btovar: Fixed in e5a1aea50e1a4c2f520e6dc02e86d57e6f4e1db5

parsl/executors/workqueue/executor.py
btovar: Removed in 2a6830c04ffebf34fb7170f9a1857eebc7e4fc62

Issue #1692, opened 2020-05-13
Parsl 1.0.0 Release Checklist


yadudoc:
## Checklist

Please edit the checklist if I've missed any items. 

### Documentation updates :

- [x] Update docs to point at 1.0.0 as the latest
- [x] Make sure docs are not broken on readthedocs, since a broken doc build will stick on as stable till next release.
- [x] Update changelog with summary of changes since 0.9.0
- [x] Update Parsl tutorial repo with a 1.0.0 branch that folds in changes 
- [x] Add `Beta` tags to components/features that are not yet stable.

### Testing :

- [x] All testing should be green on Travis
- [x] Update all configs in `parsl/parsl/configs` to match current best practices
- [x] Update all test configs in `parsl/parsl/test/configs`
- [x] Test notebooks/tutorials and basic tests on a Mac

- [ ] Site testing:
  - [x] Comet (SDSC)
  - [x] Cori (NERSC)
  - [x] Stampede2 (TACC) (* tested with pre-alpha)
  - [x] Frontera (TACC) (* tested with pre-alpha)
  - [x] Theta (ALCF)
  - [ ] ~~Cooley (ALCF)~~
  - [x] Bluewaters (NCSA)
  - [ ] ~~Swan (Cray partner network)~~
  - [x] Summit (ORNL)
  - [x] CC-IN2P3 (French Grid)
  - [x] Midway (RCC, UChicago)
  - [ ] Open Science Grid
  - [x] AWS
  - [ ] ~~Azure~~
  - [ ] Kubernetes
  - [ ] Ad-Hoc clusters

### Release Tagging and pushing to PyPI

The Makefile now supports `make release` that covers the tagging, and push to pypi. We need to push an alpha quickly to smoothen the site specific testing.

An earlier version of this is available [here](https://gist.github.com/yadudoc/168ccdfe93d99d2e62548ce025e2afde) as a draft.

danielskatz:
0.10.0 in a few places rather than 1.0.0?

yadudoc:
@danielskatz I've updated the text to say 1.0.0 instead.

PR #1707, opened 2020-05-14
Site testing

yadudoc:
This PR adds support for running `make site_test` using a site specific config on each of our priority testing sites listed in #1692. 

* `site_config_selector.py` now selects configs based on hostname
* updated per site configs for:

Sites with shared-fs
  - [x] Comet (SDSC)
  - [x] Cori (NERSC)
  - [x] Stampede2 (TACC)
  - [x] Frontera (TACC)
  - [x] Theta (ALCF)
  - [ ] Cooley (ALCF)
  - [x] Bluewaters (NCSA) [NEED HELP]
  - [x] Summit (ORNL)
  - [x] CC-IN2P3 (French Grid)
  - [x] Midway (RCC, UChicago) [NEED HELP]
  - [ ] ~~Swan (Cray partner network)~~

Sites with no shared-FS
  - [ ] Open Science Grid [NEED HELP]
  - [ ] AWS - benc
  - [ ] Azure - benc
  - [x] Kubernetes [NEED HELP]
  - [ ] Ad-Hoc clusters 

* `conda_setup` script add support for creating a conda env against github hash.

My `local_user_opts.py` file that goes in `parsl/parsl/tests/configs` is [here](https://gist.github.com/yadudoc/b71765284d2db0706c4f43605dd8b8d6)


Makefile
```diff
@@ -86,6 +86,11 @@ config_local_test: $(DEPS) ## run all tests with workqueue_ex config
 	pip3 install ".[extreme_scale]"
 	PYTHONPATH=. pytest parsl -k "not cleannet" --config local --cov=parsl --cov-append --cov-report= --random-order
 
+.PHONY: site_test
+site_test:
+	pytest parsl -k "not cleannet" --config parsl/tests/site_tests/site_config_selector.py --cov=parsl --cov-append --cov-report= --random-order
+	pytest parsl/tests/site_tests/ --config local
```
benclifford, line 92: maybe should attempt to run the full set of `local` tests rather than the parsl/tests/site_test/ subdir unless there is a specific reason not to?

parsl/tests/site_tests/conda_setup.sh
```diff
@@ -0,0 +1,124 @@
...
+        else
+            echo "Please install conda to your home dir at ~/anaconda3"
+        fi
+        conda create -p $CONDA_TARGET python=3.7 --yes --force
+        conda activate $PWD/$CONDA_TARGET
+        echo "source ~/anaconda3/bin/activate;"          >  ~/setup_parsl_test_env.sh
+        echo "conda activate $PWD/$CONDA_TARGET"         >> ~/setup_parsl_test_env.sh
+
+    else
+        echo "Treating $(hostname) as Local"
```
benclifford, line 102: This doesn't properly test the unknown system in "test batch system and running on worker nodes" sense.
So I think should give a failure rather than run a local test and give potentially false confidence. It's fine to say "we don't know how to test this system", and it's what we would say on slack when someone shows up with a question about why the site test passes even though parsl doesn't work on their site.

parsl/tests/site_tests/conda_setup.sh
```diff
@@ -0,0 +1,124 @@
...
+fi
+
+
+create_conda() {
+    pushd .
+    cd $PARSL_ROOT
+
+    if [[ "$(hostname)" =~ .*thetalogin.* ]]
+    then
+        echo "On theta"
```
benclifford, line 24: having this stuff hardcoded into an if-statement rather than in separate files makes me cry

parsl/tests/site_tests/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+  1. Checkout the git tag/hash for testing
+     * If you are running tests on a system where the $HOME is not mounted on compute nodes
+       checkout to $WORK / $SCRATCH or the equivalent. For eg, Comet and Stampede2
+
+  2. `cd parsl/tests/site_tests`
+  3. `bash conda_setup.sh`
+  4. Now check if the setup script is available at `~/setup_parsl_test_env.sh`
+  5. Go back to the parsl root dir
+  6. Run tests with `make site_test`
```
benclifford, line 37: Can I run steps 1 and 6, ignoring steps 2-5? If not, please script steps 2-5 so I don't have to do them explicitly.

parsl/tests/site_tests/README.rst
```diff
@@ -0,0 +1,60 @@
...
+   3. User's conda installation
+* Add a new block to `conda_setup.sh` that installs a fresh environment and writes out
+  the activation commands to `~/setup_parsl_test_env.sh`
+* Add a site config to `parsl/tests/configs/<SITE.py>` and add your local user options
+  to `parsl/tests/configs/local_user_opts.py`. For eg, `here's mine<https://gist.github.com/yadudoc/b71765284d2db0706c4f43605dd8b8d6>`_
+  Make sure that the site config uses the `fresh_config` pattern.
+  Please ensure that the site config uses:
+    * max_workers = 1
+    * init_blocks = 1
+    * min_blocks = 1
```
benclifford, line 55: at least the local config in site_selector has min_blocks = 0 - what is the actual constraint on this parameter for these tests?

benclifford:
The cloud based configs don't fit the shared filesystem/conda model. i) There's no shared filesystem, and ii) our example cloud templates use `pip` to install directly into the system-wide python install (because the VM is the virtual environment, so there isn't a generally need for further python level virtualization).

This might be fixed relatively easily by moving any requirement to use conda into the site-specific pieces of code.

parsl/tests/configs/cori.py
```diff
@@ -0,0 +1,40 @@
+from parsl.config import Config
```
benclifford, line 1: There is already a cori_htex.py configuration in this directory. That's a mess. Get rid of one of them.

Makefile
yadudoc: We are running:
* full test suite against the site specific config
* Site specific tests that can't run on a single node that test provider behavior against the site's config.

We can run the full test suite with `--config local`, but I didn't think that added much more value, since we can just do that with `make config_local_test`. There's definitely some overlap here.

ZhuozhaoLi:
I have added the config for midway.

I tried the site test on Kubernetes as well. But as @benclifford mentioned, there are some problems for the tests on sites that do not have a shared file system. Many of the tests need a share fs to work. For example, the `test_5` in `parsl/tests/test_regression/test_69b.py`: it assumes `hello2.txt` result can be read from the client side.

parsl/tests/site_tests/conda_setup.sh
yadudoc: Fixed.

parsl/tests/site_tests/README.rst
yadudoc: That'll be in round 2.

parsl/tests/site_tests/README.rst
yadudoc: typo. fixed.

yadudoc:
I've added a new env variable "SHARED_FS_OPTIONS" to deselect tests that rely on a shared-fs (issue #363). This is required on cc-in2p3. Another issue is that CC-IN2P3 so far, only lets you request single node jobs, which breaks the provider test which assumes a 2 node block. Otherwise everything checks out fine.

ZhuozhaoLi:
> I've added a new env variable "SHARED_FS_OPTIONS" to deselect tests that rely on a shared-fs (issue #363). This is required on cc-in2p3. Another issue is that CC-IN2P3 so far, only lets you request single node jobs, which breaks the provider test which assumes a 2 node block. Otherwise everything checks out fine.

With this, Kubernetes should also survive the test suites now. I will test this on Kubernetes and report. 

One question is that PetrelKube needs to be accessed from Cooley. Is it fine if the site test is on Cooley (i.e., `[[ "$(hostname -f)" =~ .*cooley.* ]]` in `conda_setup.sh`)?

yadudoc:
It's okay for now. Let's get it tested and checked off for now. Testing cooley specifically is less important since we test the cobalt provider over on Theta. Let's treat cooley submit node == kubernetes for now.

benclifford:
This PR conflates issue 363 (which is to do with lack of staging std* files) with lack of shared filesystem. It adds an `issue363` marker onto tests which do not have that issue. Those annotations should be a new marker such as `shared-fs` (so that pytest can be invoked with a `not sharedfs` selector).

benclifford:
Actually, I've changed my mind: the problem is that `file:` staging is not properly configured. Some of those tests require working `file:` staging, not specifically a shared filesystem, so *should* pass if any `file:` provider is used, not specifically `NoOpFileStaging`.

So the pytest mark name should reflect that the tests require file staging, not that they need a shared fs, perhaps.

parsl/tests/configs/user_opts.py
```diff
@@ -21,7 +21,24 @@
...
+    },
+    'summit': {
+        'worker_init': 'source ~/setup_parsl_test_env.sh;',
+    },
+    'midway': {
+        'worker_init': 'source ~/setup_parsl_test_env.sh;',
+    },
+    'petrelkube': {
+        'worker_init': '~/setup_parsl_test_env.sh',
+    },
```
ZhuozhaoLi, line 41: @yadudoc I think you missed the user opts of `in2p3` here

yadudoc: These need to be edited by the user anyway, and I have mine in a local_user_opts.

PR #1708, opened 2020-05-15
Fix source mode in Work Queue executor

trshaffer:
I found a bug in the WQ executor that prevents any tasks from starting when `source=True`.

The steps to actually write out the serialized function were under the default non-source branch, so configuring the executor with `source=True` resulted in no function info being written. Work Queue of course failed with missing inputs before actually running anything.

Also, the error message was kind of confusing (`unable to process wrapper script failure with status = -1`). I think the worker wrapper is in flux, would it be better to say something here to the effect of "this is probably a bug in Parsl" in case a user happens get into an invalid state?

cc @btovar  @yadudoc


btovar:
Ah, beat me to it! 

With regards to the error message, there is the issue #1703. I think we are missing one type of exception for this case, such as WorkQueueWrapperFailure. Then there will be WorkQueueTaskFailure (for tasks that failed according to work queue, such as resource exhaustion), and any other exception that are passed to parsl.

PR #1710, opened 2020-05-18
Rearrange Makefile and CI

benclifford:
This PR rearranges the Makefile to:
 - allow use of the Makefile in an existing python environment rather than forcing a new `venv`. This allows `make test` to be used to test against development environments
 - removes the use of a venv in CI. CI provides a container environment for the entire system, so a python level virtual environment is unnecessary
 - removes a duplicate execution of module linting and flake8
 - makes mypy be executed the same as lint/flake8
 - installs optional dependencies in the targets that need them - prior to this, optional dependencies were sometimes installed in a different target that happens to execute in the correct order in `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,9 +61,7 @@ htex_local_test: ## run all tests with htex_local config
 
 .PHONY: htex_local_alternate_test
 htex_local_alternate_test: ## run all tests with htex_local config
-	echo "$(MPI)}"
-	parsl/executors/extreme_scale/install-mpi.sh $(MPI)
-	pip3 install ".[extreme_scale,monitoring]"
+	pip3 install ".[monitoring]"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -52,18 +52,16 @@ mypy: ## run mypy checks
 	MYPYPATH=$(CWD)/mypy-stubs mypy parsl/app/ parsl/channels/ parsl/dataflow/ parsl/data_provider/ parsl/launchers parsl/providers/
 
 .PHONY: local_thread_test
-local_thread_test: $(DEPS) ## run all tests with local_thread config
+local_thread_test: ## run all tests with local_thread config
```
yadudoc, line 55: Removing DEPS here would result in tests possibly running without the test-requirements installed, no?

yadudoc: Resolved over chat. We are decoupling the tests from building to avoid having to rebuild everytime for every small test.

PR #1711, opened 2020-05-18
Rename Hub class because it clashes cognitively with MonitoringHub

benclifford:
There is now a MonitoringHub, containing a component called MonitoringRouter


ZhuozhaoLi:
There are some other unchanged variables, logs, exceptions, like `hub_address`, `hub_port`. Do you mind if I make a PR to your branch or change your branch directly?

benclifford:
I mostly-deliberately kept those: some of them are user-facing, and the user sees a `MonitoringHub` object (with a mostly hidden MonitoringRouter inside it). I'm fine with them being renamed, though.

PR #1712, opened 2020-05-18
Remove unused RESOURCE_INFO message tag

benclifford:
Resource messages are delivered via their own queue
and do not need a tag.

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -344,7 +344,7 @@ class DatabaseManager:
                 self._insert(table=NODE, messages=messages)
 
             """
-            RESOURCE_INFO messages
+            Resource info messages
 
             """
             messages = self._get_messages_in_batch(self.pending_resource_queue,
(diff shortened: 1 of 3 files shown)


ZhuozhaoLi:
Probably a more appropriate fix would be changing `TASK_INFO` in the line below to `RESOURCE_INFO` (that was a typo). 
https://github.com/Parsl/parsl/blob/aa545d7f38c6a7b6654b61c8f81ab59487bf2d5b/parsl/monitoring/monitoring.py#L561

I am fine with totally remove `RESOURCE_INFO`---this kind of messages are not specifically checked in `db_manager` because these messages come from a separate queue. But I was wondering if it would be useful to keep this `RESOURCE_INFO` tag in case we need to check it in the future.

benclifford:
That message type parameter never gets used for anything: message routing happens in a different way.

We should only keep unused code around if there is a clear reason for it.

ZhuozhaoLi:
> That message type parameter never gets used for anything: message routing happens in a different way.
> 
> We should only keep unused code around if there is a clear reason for it.

We may want to remove UDP and use only one ZMQ channel in the Hub (router) in the future. In that case, it would be useful to check the message type there, no? Or adding it back when we actually switch would be better.

benclifford:
> adding it back when we actually switch would be better.

this. but *if*, not *when*

PR #1714, opened 2020-05-18
Python packaging for Work Queue executor

trshaffer:
~~DEPENDS ON FIX IN #1708~~ 

This PR supersedes #1300. The `python_package_*` scripts in CCTools have been updated/improved, so this PR uses those versions. I also started the executor integration from scratch, as `master` has diverged from the previous PR. Some major differences from the previous iteration:
- Does not change the default behavior. This feature is for the specific case that a user has a working Conda environment for the app, and will deploy tasks to workers without a shared FS/configured software environment/Python interpreter. If users can avoid using this feature, they should as they'll get better performance.
- Only uses stdlib modules for analyzing source. Note that we use `conda-pack` to make packages relocatable, so this PR *does* assume that everything is running in a Conda environment and `conda-pack` is installled.
- `python_package_analyze` handles more edge cases, and captures software versions. Note that it doesn't make any attempt to detect/transfer imports not managed by Conda, so only stuff in the standard library or installed via pip/conda is allowed. In particular, users can't import other modules in the app. Functions to be shipped should have minimal dependencies.
- This PR caches as much as possible, since that's where most of the overhead comes from. The executor only analyzes each function body once per app run. Environment packages are cached in `$TMP`, so those can be reused across runs. The new `python_package_run` also lets multiple jobs reuse a single unpacked environment, reducing the job start overhead on workers.


parsl/executors/workqueue/executor.py
```diff
@@ -42,6 +44,13 @@
 
 logger = logging.getLogger(__name__)
 
+package_analyze_script = os.path.join(
```
yadudoc, line 47: Let's add these as entry_points in `parsl/setup.py`. It's not a good idea to mess with pythonpaths.

parsl/executors/workqueue/executor.py
```diff
@@ -739,8 +776,29 @@ def _serialize_function(self, fn_path, parsl_fn, parsl_fn_args, parsl_fn_kwargs)
...
+            subprocess.run([package_analyze_script, '-', spec.name], input=source_code, check=True)
+            with open(spec.name, mode='rb') as f:
+                pkg = os.path.join(pkg_dir, "pack-{}.tar.gz".format(hashlib.sha256(f.read()).hexdigest()))
+            if os.access(pkg, os.R_OK):
+                self.cached_envs[id(fn)] = pkg
+                return pkg
+            (fd, tarball) = tempfile.mkstemp(dir=pkg_dir, prefix='.tmp', suffix='.tar.gz')
+            os.close(fd)
+            with open(os.devnull, 'w') as devnull:
+                subprocess.run([package_create_script, spec.name, tarball], stdout=devnull, check=True)
```
yadudoc, line 798: You can use `subprocess.DEVNULL` directly here.

PR #1715, opened 2020-05-18
Adding a timeout for the call to get address

yadudoc:
Adds a timeout. Default set to 30s.

This avoids a hang on machines with firewalled network setups.

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -41,7 +41,7 @@ def address_by_query(timeout: float = 30) -> str:
        -----------
 
        timeout : float
-          Timeout for the request in seconds. Can be < 1. Default: 30s
+          Timeout for the request in seconds. Default: 30s
     """
     logger.debug("Finding address by querying remote service")
     response = requests.get('https://api.ipify.org', timeout=timeout)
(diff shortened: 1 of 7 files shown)


parsl/addresses.py
```diff
@@ -33,13 +33,18 @@ def address_by_route() -> str:
...
-def address_by_query() -> str:
+def address_by_query(timeout: float = 30) -> str:
     """Finds an address for the local host by querying ipify. This may
        return an unusable value when the host is behind NAT, or when the
        internet-facing address is not reachable from workers.
+       Parameters:
+       -----------
+
+       timeout : float
+          Timeout for the request in seconds. Can be < 1. Default: 30s
```
benclifford, line 44: > Can be < 1.

That's not notable for a float timeout

benclifford:
This PR failed CI just now with a probably unrelated error to do with installing the extreme scale executor MPI dependency, on three of the four builds.

```
Package libmpich-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
E: Unable to locate package mpich
E: Package 'libmpich-dev' has no installation candidate
```

PR #1716, opened 2020-05-18
renaming more hub to router

ZhuozhaoLi:
This renames more `hub` to `router` and is a PR merging to `benc-rename-hub` branch.

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -143,16 +143,16 @@ class MonitoringHub(RepresentationMixin):
         Parameters
         ----------
         hub_address : str
-             The ip address at which the workers will be able to reach the Hub.
+             The ip address at which the workers will be able to reach the Router.
         hub_port : int
-             The specific port at which workers will be able to reach the Hub via UDP. Default: None
+             The specific port at which workers will be able to reach the Router via UDP. Default: None
         hub_port_range : tuple(int, int)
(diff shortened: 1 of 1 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -143,16 +143,16 @@ def __init__(self,
         Parameters
         ----------
         hub_address : str
-             The ip address at which the workers will be able to reach the Hub.
+             The ip address at which the workers will be able to reach the Router.
```
yadudoc, line 146: This is confusing. hub_address is the address for the Router? While this is correct, it will be confusing for the user who doesn't know that the hub contains a router.

benclifford:
That was my original reason (aside from laziness) for keeping those parameters, now unambiguously referring to the `MonitoringHub` when they say `hub`

ZhuozhaoLi:
> This is confusing. hub_address is the address for the Router? While this is correct, it will be confusing for the user who doesn't know that the hub contains a router.

In @benclifford 's PR, Ben changed the `hub.log` to `monitoring_router.log` and there are a few appearance of `router` there---I was also worried about the naming being confused to the user. Maybe we do not expect the users to look at that `router` log to debug?

I can change `hub_address` to `router_address`, but I am afraid that is confused with `MonitoringHub` too.

ZhuozhaoLi:
This PR is a bit confusing, so closing it.

PR #1720, opened 2020-05-20
Support python 3.9

benclifford:
This PR makes changes so that CI can pass with Python 3.9.

No changes to core parsl code seem to be necessary to pass CI.

* This PR brings in upgrades for test/build dependencies that were not working with 3.9:  mypy, sphinx.

* One test used `platform`, but datatypes from `platform` cannot be passed around any more, so that test is changed to not use `platform`.


yadudoc:
Tagging this for next release.

benclifford:
This was failing mypy. I upgraded mypy to a more recent version to address this python 3.9 breakage: python/mypy#8627

benclifford:
The test suite hangs on parsl/tests/site_tests/test_provider.py.

Switching to LSST parsl which has better debugging info, I get this error:
```
ERROR    parsl.process_loggers:process_loggers.py:25 exception wrapper: caught an exception for thread HTEX-Queue-Management-Thread
  File "/home/benc/parsl/src/parsl/parsl/process_loggers.py", line 20, in wrapped
    r = fn(*args, **kwargs)
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/executor.py", line 404, in _queue_management_worker
    result = deserialize(msg['result'])
  File "/home/benc/parsl/src/parsl/parsl/serialize/facade.py", line 121, in deserialize
    result = self.methods_for_code[header].deserialize(payload)
  File "/home/benc/parsl/src/parsl/parsl/serialize/concretes.py", line 28, in deserialize
    data = pickle.loads(chomped)
TypeError: <lambda>() takes 6 positional arguments but 7 were given
```

This leads to https://bugs.python.org/issue42189 which suggest there is some cross version change to `platform`, which is used by that test case - in the linked bug this is to do with `deepcopy`, but in the above error it is to do with `pickle`. It's likely those are related.

I don't think there's any particular need to use `platform` in this test case, so a fix would probably be to do something more trivial there.

benclifford:
This now fails the doc build repeably:

```
Warning, treated as error:
error while formatting arguments for parsl.app.errors.MissingOutputs: type object 'list' has no attribute '_special'
Makefile:52: recipe for target 'html' failed
make: *** [html] Error 2
The command "make SPHINXOPTS=-W html" exited with 2.
```

benclifford:
For doc error see: https://bugzilla.redhat.com/show_bug.cgi?id=1814243 and https://github.com/sphinx-doc/sphinx/commit/46372726de435cbc6b70c99d95b7986f76563df4

yadudoc (approved):
These changes look good to go.

PR #1721, opened 2020-05-21
Can parsl build on OS X in CI?

benclifford:


yadudoc:
@benclifford, I'd tried something like this before.
The issue is that there is no standard way in travis to allow python version specification for OSX.
One mechanism is to specify xcode version like described [here](https://blog.travis-ci.com/2019-08-07-extensive-python-testing-on-travis-ci).

I tried a while back to do this using a brew-install of python virtualenv, which doesn't really cut it here because that still relies on system python.
 https://github.com/Parsl/parsl/commit/21e872a27a32c9c0cd474f7888170bf39cb34179#diff-354f30a63fb0907d4ad57269548329e3

danielskatz:
is this done now?

benclifford:
no - not by a long shot

danielskatz:
you want to not have to set the env var and do this the proper way?

benclifford:
I want the "all checks have failed" message to not happen on this PR - which is coming from travis CI not being able to run the parsl test suite in their CI os x environment. This PR is about automated per-PR testing, not about "has someone once managed to run a version of parsl on their os x box"

benclifford:
closing this - I don't care at the moment about OS X

Issue #1628, opened 2020-03-25
Make bash apps resilient to stalling when large stderr/stdout is produced using LocalChannel


annawoodard:
**Describe the bug**
Bash apps which produce too much output stall subprocess when using `LocalChannel`.

Note this was mentioned in #1444, but isn't fixed by #1471. The situation is better now because there is only one code path that breaks in this situation, but it is still possible to stall the process with a large enough output. (I believe the issue was just the large output, not so much that it was never read).

**To Reproduce**
I have tested this with the SlurmProvider+LocalChannel. Run SlurmProvider+LocalChannel with a bash app that produces a large amount of stdout. The stdout is never written, and the app never completes.

**Expected behavior**
The app should complete. What to do with the stdout/stderr is up for discussion. One possibility is to always pipe to /dev/null (as was done by @WardLT in #1620.) Another is to set bufsize (keyword arg to `subprocess.Popen`) [here](https://github.com/Parsl/parsl/blob/master/parsl/channels/local/local.py#L62-L69) and make it a configurable keyword arg in the `LocalChannel` constructor. I favor the latter. 

**Environment**
parsl master f183657

Originally reported by @WardLT.

benclifford:
It should also be possible to drain the pipes into variables while waiting for the process to complete, instead of at the end. This wouldn't need a new parameter.

yadudoc:
* We can use the select method (https://docs.python.org/3/library/select.html) or something similar to drain stdout/stderr while waiting for the process to complete. This is a bit complicated, and I'm not convinced we need it.
* We can set bufsize which protects against huge volumes of data, but might result in the loss of error messages at the end. Is there anything else we might lose by choosing this ?
* Use some sort of circular buffer that captures what happened at the end. I don't know if there's a stock solution for this available.

PR #1724, opened 2020-05-22
Local provider error handling

### LINKED ISSUE #1641: PR #1471 seems to go a few steps backwards
While trying to commit the rest of the split-up PR #1420, I ran into the fact that PR #1471 conflicts with it. 

In principle, the idea behind issue #1448 is the right one: make code more abstract. In practice, #1471 does it by also removing essential functionality. To use a bit of hyperbole, it's really nice when you can abstract all your code to "return True", but it is not very useful.

In particular, after #1471, tasks managed by the local provider appear to never be able to have the COMPLETED status, only RUNNING and FAILED. This was handled properly when using popen() locally. I would think that, should a single solution be adopted, we should adopt the correct one and fix the remote channels to properly report task status. Without the ability to distinguish between success and failure when running a worker, we cannot detect the type of problems that  #1420 does. Like ever.

Additionally, but this should be easily fixed, the channel interface does not support pull_file (although the SSH implementation does) method, which is needed for getting stdout/stderr files back. And, by the way, the current code munges stdout and stderr, whereas the special-cased one didn't.

### LINKED ISSUE #1706: Single node launcher has various problems
**Describe the bug**
The changes introduced in #1499 are utter nonsense. At least three major issues can be identified:
1. The syntax "$EXITCODE[$I]" as "the I-th element of the EXITCODE array" is incorrect. It, instead, tokenizes to $EXITCODE, "[", $I, "]". The correct bash syntax is "${EXITCODE[$I]}"
2. CMD() is launched using the background operator ("&"), which causes it to run in a subshell. That makes the assignment EXITCODES[$I]=$? inside CMD() invisible in the functions in the parent shell that need to use it (FAILONANY() and FAILONALL())
3. set -e is in effect, which is inherited by the execution of CMD(), which means that the line EXITCODES[$I]=$? is not even executed if the command fails.

**To Reproduce**
Some of this is apparent if one runs "/bin/false" with the local provider and notices that it succeeds.

**Expected behavior**
It is a bit difficult to say what *exactly* we want the behavior of launchers to be and whether errors in launcher code should be signaled separately from errors in the underlying command. This needs some thinking. This would suggest that an immediate fix of the above issues may not be useful. However, should that be done, the exit code propagation should work correctly in that executing "/bin/false" should cause the launcher to fail.

**Environment**
Doesn't matter.

**Distributed Environment**
Local.

hategan:
This PR adds back error reporting (via exit code and process streams) to the local provider (aka., fixes #1641). 

It also, hopefully, fixes #1706, adding a test that verifies that failing apps fail and succeeding ones succeed.

A test is also added that tests the correct functioning of error propagation for both the local and ssh channels. The test is fully automated and requires the ability of an SSHChannel to use a non default host keys file. Consequently, this PR adds a "host_keys_filename" parameter to the SSHChannel constructor.


parsl/launchers/launchers.py
```diff
@@ -6,9 +6,16 @@
 logger = logging.getLogger(__name__)
 
 
+def bool_to_num(b: bool):
```
benclifford, line 9: can you do this instead?
```
>>> print(int(True))
1
>>> print(int(False))
0
```

parsl/channels/ssh/ssh.py
```diff
@@ -26,7 +26,8 @@ class SSHChannel(Channel, RepresentationMixin):
 
     '''
 
-    def __init__(self, hostname, username=None, password=None, script_dir=None, envs=None, gssapi_auth=False, skip_auth=False, port=22, key_filename=None):
+    def __init__(self, hostname, username=None, password=None, script_dir=None, envs=None,
+                 gssapi_auth=False, skip_auth=False, port=22, key_filename=None, host_keys_file=None):
```
benclifford, line 30: host_keys_file is a new user-facing feature so maybe should be mentioned in the PR title.

parsl/providers/local/local.py
```diff
@@ -76,19 +76,70 @@ def status(self, job_ids):
...
-            retcode, stdout, stderr = self.channel.execute_wait('ps -p {} > /dev/null 2> /dev/null; echo "STATUS:$?" '.format(
-                self.resources[job_id]['remote_pid']), self.cmd_timeout)
-            for line in stdout.split('\n'):
-                if line.startswith("STATUS:"):
-                    status = line.split("STATUS:")[1].strip()
-                    if status == "0":
-                        self.resources[job_id]['status'] = JobStatus(JobState.RUNNING)
+            # This job dict should really be a class on its own
+            job_dict = self.resources[job_id]
+            if job_dict['status'] is not None and job_dict['status'].terminal:
```
benclifford, line 81: `if job_dict['status'] is not None ...` can be `if job_dict['status'] ...` ?

parsl/providers/error.py
```diff
@@ -68,3 +68,19 @@ def __init__(self, script_path, reason):
...
+
+    def __init__(self, task_name, message, stdout=None, stderr=None):
+        self.task_name = task_name
+        self.message = message
+        self.stdout = stdout
+        self.stderr = stderr
+
+    def __repr__(self):
+        # TODO: make this more user-friendly
+        return "Cannot launch task {0}: {1}; out={2}, err={3}".format(self.task_name, self.message,
```
benclifford, line 85: call these stdout and stderr for consistency with everywhere else

parsl/providers/local/local.py
```diff
@@ -76,19 +76,70 @@ def status(self, job_ids):
...
+                    status = JobStatus(JobState.RUNNING)
+                else:
+                    # not alive but didn't get to write an exit code
+                    if 'cancelled' in job_dict:
+                        # because we cancelled it
+                        status = JobStatus(JobState.CANCELLED)
+                    else:
+                        # we didn't cancel it, so it must have been killed by something outside
+                        # parsl; we don't have a state for this, but we'll use CANCELLED with
+                        # a specific message
```
benclifford, line 102: I think this case would also be hit if something failed to start really early on? I've encountered a case where an unexpected shell was being used, for example. I don't think that needs a change if so, other than perhaps in the comment, but checking my understanding of the logic.

parsl/channels/ssh/ssh.py
benclifford: use `filename`, not `file`, for consistency with `key_filename`.

parsl/providers/provider_base.py
```diff
@@ -26,9 +26,13 @@ def __new__(cls, value, terminal):
 class JobStatus(object):
     """Encapsulates a job state together with other details, presently a (error) message"""
 
-    def __init__(self, state: JobState, message: str = None):
+    def __init__(self, state: JobState, message: str = None, exit_code: int = -1,
```
benclifford, line 29: use None for exit_code. don't invent one when one doesn't exist.

parsl/tests/test_providers/test_local_provider.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+# It would probably be better, when more formalized site testing comes into existence, to
+# use a site-testing provided server/configuration instead of the current scheme
+def test_ssh_channel():
+    with tempfile.TemporaryDirectory() as config_dir:
+        sshd_thread, priv_key, server_port = _start_sshd(config_dir)
+        try:
+            with tempfile.TemporaryDirectory() as remote_script_dir:
+                # run this once so we add host keys and whatnot
```
benclifford, line 82: `SSHChannel` has this set:

> self.ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())

That should be adding host keys. Does something else break here?

parsl/tests/test_providers/test_local_provider.py
```diff
@@ -0,0 +1,202 @@
...
+        threading.Thread.__init__(self, daemon=True)
+        self.config_file = config_file
+        self.stop_flag = False
+        self.error = None
+
+    def run(self):
+        try:
+            # sshd needs to be run with an absolute path, hence the call to which()
+            p = subprocess.Popen([shutil.which('sshd'), '-D', '-f', self.config_file],
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
benclifford, line 116: I think this use of PIPE without being read until the end will lead to blocks on pipe writes. That problem exists elsewhere in parsl - https://github.com/Parsl/parsl/issues/1628

I'm unsure if that will make sshd hang, but it wouldn't surprise me.

parsl/providers/local/local.py
hategan: Good catch. I was under the impression that this will be routed through /bin/bash, but I cannot see why that would be so. I think this should probably be changed to /bin/bash -c ...

parsl/tests/test_providers/test_local_provider.py
hategan: This is a test and -D is "don't detach" not "debug".

parsl/launchers/launchers.py
hategan: If official python documentation says that this will always be the case, yes. On the other hand, the existing code does not require consulting anything else in order to ensure that (False, True) -> (0, 1).

Can we stop wasting time on such things? I am of the opinion that, as part of the review process, providing equivalent code that maybe to somebody looks better without justifying why it is so, is a pretty counterproductive way of going about it.

parsl/providers/local/local.py
```diff
@@ -76,19 +76,70 @@ def status(self, job_ids):
...
+                    return True
+                else:
+                    return False
+
+    def _read_job_file(self, script_path: str, suffix: str) -> str:
+        path = '{0}{1}'.format(script_path, suffix)
+        if self._should_move_files():
+            path = self.channel.pull_file(path, self.script_dir)
+
+        with open(path, 'r') as f:
```
yadudoc, line 140: It looks like this will only be run on status files which would be small. Can you confirm please?

parsl/channels/ssh/ssh.py
hategan: Good point. Renamed and added note in the PR description.

parsl/providers/local/local.py
```diff
@@ -158,26 +209,47 @@ def submit(self, command, tasks_per_node, job_name="parsl.localprovider"):
...
+        #   2. Launch the following sequence in the background:
+        #      a. the command to run
+        #      b. write the exit code of the command from (a) to the exit code file
+        #   3. Write the PID of the background sequence on stdout. The PID is needed if we want to
+        #      cancel the task later.
+        #
+        # We need to do the >/dev/null 2>&1 so that bash closes stdout, otherwise
+        # channel.execute_wait hangs reading the process stdout until all the
+        # background commands complete.
+        cmd = 'echo - >{0}.ec && {{ {{ bash {0} 1>{0}.out 2>{0}.err ; ' \
```
yadudoc, line 232: I worry that a non bash remote shell mess this up. Perhaps we should require bash in the channel.execute_wait calls.

parsl/providers/local/local.py
hategan: If you think that's more readable...

parsl/tests/test_providers/test_local_provider.py
hategan: Didn't try. I will shortly.

parsl/providers/local/local.py
hategan: No such guarantee. We can possibly change to a lazy read for stdout/stderr.

parsl/providers/local/local.py
hategan: See discussion above with Ben.

parsl/tests/test_providers/test_local_provider.py
hategan: Fails if the key file does not exist, but works if an empty one is created. I'll do that instead.

parsl/providers/local/local.py
hategan: I wrapped this in /bin/bash -c. That said, I also tried with /bin/dash -c, and it worked. I looked at it and it didn't seem to contain any bashisms. That said, it is hard to imagine a scenario, short of bash not being installed, in which being explicit about the shell is a bad thing.

parsl/providers/error.py
hategan: Done

parsl/providers/provider_base.py
hategan: Done

parsl/providers/local/local.py
hategan: Lazy reading is now being done

parsl/launchers/launchers.py
yadudoc: Docs do say that -> https://docs.python.org/release/3.0.1/reference/datamodel.html#the-standard-type-hierarchy. We just need to be careful not to do str(Bool) which will give you "True"/"False". bool_to_num is redundant.

hategan: Did both/either of you *knew* (as opposed to suspected) this before this conversation?

parsl/providers/local/local.py
hategan: After some more discussion, I added stdout_summary and stderr_summary properties in JobStatus. They read first and last 1k of the respective files.

By the way, the stdout/stderr properties are only used by test_local_provider.py, so the _summary methods are not really needed anywhere.

parsl/tests/test_providers/test_local_provider.py
yadudoc: @benclifford I'm guessing this one's resolved, can you confirm please.

parsl/launchers/launchers.py
yadudoc: Yes. I think the code will be cleaner without this function.

hategan: Can you *please* stop commenting on matters of style and making things slightly cleaner. I am asking because this has been happening consistently, it annoys the crap out of me, like really badly, while at the same time we miss actual serious bugs in the code.

parsl/tests/test_providers/test_local_provider.py
benclifford: I guess so. My main concern was if it made a lot of output, but as @hategan says -D is detach, not verbose debug output as I initially misread it.

ravihansa3000:
This breaks LocalProvider when using in combination with SSHChannel. Following is an exception stack trace;
@yadudoc 

2020-06-28 02:02:27 parsl.channels.ssh.ssh:215 [ERROR]  Remote file copy will overwrite a local file:/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Downloader-Output/remote/COAD_0/runinfo/001/submit_scripts/parsl.localprovider.1593280864.8464878.sh.ec
NoneType: None
2020-06-28 02:02:27 parsl.dataflow.flow_control:114 [ERROR]  Flow control callback threw an exception - logging and proceeding anyway
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/flow_control.py", line 112, in make_callback
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 63, in poll
    self._update_state()
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 69, in _update_state
    item.poll(now)
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 27, in poll
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/executors/status_handling.py", line 59, in status
    status = self._make_status_dict(job_ids, self._provider.status(job_ids))
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 87, in status
    str_ec = self._read_job_file(script_path, '.ec').strip()
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 142, in _read_job_file
    path = self._job_file_path(script_path, suffix)
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 138, in _job_file_path
    path = self.channel.pull_file(path, self.script_dir)
  File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/channels/ssh/ssh.py", line 216, in pull_file
    raise FileExists(None, self.hostname, filename=local_dest)
parsl.channels.errors.FileExists: Hostname:nus.nscc.sg, Reason:File name collision in channel transport phase:/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Downloader-Output/remote/COAD_0/runinfo/001/submit_scripts/parsl.localprovider.1593280864.8464878.sh.ec

hategan:
@yadudoc 

Do we really need this? (channels/ssh/ssh.py:196).

```python
        # Easier to check this than to waste time trying to pull file and
        # realize there's a problem.
        if os.path.exists(local_dest):
            logger.exception("Remote file copy will overwrite a local file:{0}".format(local_dest))
            raise FileExists(None, self.hostname, filename=local_dest)
```

yadudoc:
@ravihansa3000 Thanks for reporting this and tracing back to these changes.

@hategan I am guessing this is from the task_status_poller repeatedly fetching the exit code files. We should check whether sftp_client will overwrite, if the file exists locally rather than raise an error. Since it's just the provider using the channels, I think it should be safe to remove the exception raise and just log the overwrite.

hategan:
The first non-comment line in sftp_client.get() is:
```python
with open(localpath, "wb") as fl:
```

So it doesn't look like it throws an error. Perhaps, if we really expect that this will be needed, we could add an "overwrite_ok" parameter to channel.pull_file().

yadudoc:
Good idea :+1:, and let's set `overwrite_ok=True` as the default.

hategan:
> Good idea , and let's set `overwrite_ok=True` as the default.

I'm torn. In a predominantly concurrent system, I'd say it's probably safe to be safe unless you know what you're doing. This would translate to keeping overwrite_ok=False by default and manually setting it to True when the calling code can assert that it's ok to overwrite, as is the case with the exit code file.

But then ssh on the command line (and cp, and a bunch of other things) default to overwriting. From that perspective, the error above seems like unexpected behavior.

yadudoc:
Not allowing overwrites when we considered using channels to move app data, and silent overwrites are most likely wrong. Since the channels now really just serve provider + status_polling, this "should" be safe. We'd also incur a bit more overhead, copying files over itself when there's a shared-fs, (I don't know how safe it is to do this) and there isn't a config option for the user to indicate that to us now.

hategan:
> [...] when there's a shared-fs, (I don't know how safe it is to do this) [...]

What makes you think it would not be "safe"?

PR #1728, opened 2020-05-27
Adds type annotations to WorkQueueExecutor

btovar:
Solves #1694.


parsl/executors/workqueue/executor.py
```diff
@@ -167,25 +171,25 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
     """
 
     def __init__(self,
-                 label="WorkQueueExecutor",
```
benclifford, line 170: put `@typeguard.typechecked` on the front of this and the type annotations will be checked at runtime, as users pass in correct/incorrect values.

btovar: Added in 111b18246d8085446aea49291497db5bb0e45dfa

PR #1734, opened 2020-06-01
Adding a staging_required marker

yadudoc:
This differentiates tests that fail due to issue363 (stdout/stderr staging issue), from tests that could pass if a staging provider is available in cases where the system doesn't have a shared-FS. 

Adds markers into the site_testing branch.


ZhuozhaoLi:
I think there are some other tests that require staging provider but are marked as `issue363`. Do you mind if I make some changes here?

yadudoc:
@ZhuozhaoLi Sure please go ahead.

Issue #1659, opened 2020-04-28
parsl/htex vs multiprocessing spawn experience notes


benclifford:
This issue tracks some experiences I had trying to convert htex to use the `spawn` method of process launching (rather than the unix default of `fork`). I've encountered this twice so I'm adding notes here.

# Why do this?
parsl code is multithreaded. the python docs explicitly state that using the `fork` multiprocessing context is dangerous (see https://docs.python.org/3/library/multiprocessing.html). I've encountered problems in practice in two different situations (one with the htex interchange starting up, and one with the monitoring resource worker starting up) where the new processes deadlock. This is likely due to locks (for example, in `logging`) being held by another thread at the point of forking.

# What did I change?
I changed Queue and Process constructor calls in parsl/executors/high_throughput/executor.py to use a new `spawn` multiprocessing context

# What happened?
When the new process starts in a `spawn`ed, the new python process imports the original main python code. In the case of a parsl workflow, this is the original parsl workflow, and so the entire original workflow is run again in the new code. The recommended protection against this is that the original main have `if __name__ == '__main__':` around the main code. (see https://docs.python.org/3/library/multiprocessing.html)
This re-import doesn't happen using the `fork` context, because forking keeps everything from the parent process - the same sharing that causes lock problems also means that code imported in the parent process appears in the forked child process.

# Why am I encountering this now?
I think (unconfirmed) specifically I am hitting this now, even though this deadlock potential has existed for a long time, because I added a wrapper around every thread and process invocation to log start/end/exceptions. That means that the same piece of code (the wrapper) is being executed in every thread and process very close together in time, and so it is more likely that locks specific to the same code (maybe the logger for that particular module?) will conflict.

benclifford:
@danielskatz has encountered a situation that I think is related. python process dies with:
> objc[51699]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.

Setting `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` pushes the bug elsewhere so that you can startup ok.

Also see: https://bugs.python.org/issue33725

benclifford:
crossref PR #1738

benclifford:
[...]

PR #1738, opened 2020-06-05
Disable fork safety on mac

yadudoc:
This is required to get HTEX working on Mac. Related to #1659.

Warning note in FAQ/HTEX pending.

@danielskatz  Could you install this from this branch and test please ?


danielskatz:
as far as I can tell, this worked with my example program

PR #1751, opened 2020-06-12
Updating LocalProvider.max_blocks default to 1

yadudoc:
LocalProvider.max_blocks is set to 10 by default. This is a bad choice since this would easily result in Parsl overrunning a machine with too many manager/worker pools. This PR sets the default to 1.

### DIFF
--- a/parsl/providers/local/local.py
+++ b/parsl/providers/local/local.py
@@ -39,7 +39,7 @@ class LocalProvider(ExecutionProvider, RepresentationMixin):
                  launcher=SingleNodeLauncher(),
                  init_blocks=4,
                  min_blocks=0,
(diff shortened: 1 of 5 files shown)


ZhuozhaoLi:
wait, the init should be 1 also?

parsl/providers/local/local.py
```diff
@@ -39,7 +39,7 @@ def __init__(self,
                  launcher=SingleNodeLauncher(),
                  init_blocks=4,
```
ZhuozhaoLi, line 40: maybe change to 1 here?

yadudoc: Good catch.

PR #1752, opened 2020-06-12
fix to ignore compressed files in runinfo

danielskatz:


danielskatz:
I've run into a problem where I compressed/tarred a directory in the runinfo directory to send to @yadudoc to understand a problem, and where there is such a file (or any file with a name that can't be converted to an integer), it causes parsl to fail - this fixes it by only looking at filenames at are 3-digit integers.

parsl/dataflow/rundirs.py
```diff
@@ -23,7 +23,7 @@ def make_rundir(path):
         if not os.path.exists(path):
             os.makedirs(path)
 
-        prev_rundirs = glob(os.path.join(path, "[0-9]*"))
+        prev_rundirs = glob(os.path.join(path, "[0-9][0-9][0-9]"))
```
yadudoc, line 26: I wrote the previous code with mistaken assumption that `[0-9]*` would match with any number of digits. Buggy as it is, it would still work in cases where you have >1000 rundirs. This code here although better, would break if there are > 1000 runs. Although not perfect, what do you think about '[0-9]*[0-9]'?

danielskatz:
so ```current_rundir = os.path.join(path, '{0:03}'.format(x + 1))``` is going to work for values of more than 3 digits?

yadudoc:
Yes. Although, I had to test it to be sure: 
```
>>> print('{0:03}'.format(1))
001
>>> print('{0:03}'.format(10))
010
>>> print('{0:03}'.format(100))
100
>>> print('{0:03}'.format(1000))
1000
```

parsl/dataflow/rundirs.py
danielskatz, line 26:
```suggestion
        prev_rundirs = glob(os.path.join(path, "[0-9]*[0-9]"))
```

danielskatz:
ok, changed - I agree it's not great but it is better, and will solve at least my problem

danielskatz:
I think this is now ready to merge for someone who is authorized to do so (this does not include me)

danielskatz:
thanks!

parsl/dataflow/rundirs.py
benclifford: Use `+` to mean one or more

yadudoc: Ben `+` is not a wildcard supported by glob.

PR #1756, opened 2020-06-15
Remove unused current_capacity from provider interface

benclifford:


yadudoc (approved):
This looks reasonable.

PR #1762, opened 2020-06-15
Make task list appear under .../task/ not under .../app/

benclifford:
This makes the URL space layout a bit cleaner:
.../task/  lists all tasks, and
.../task/1 lists task 1.


ZhuozhaoLi:
This PR is actually one single word change---did you intentionally switch the order of two routes?

benclifford:
> did you intentionally switch the order of two routes?

yes, so that the two `/task/` handlers are next to each other in the code.

Issue #1592, opened 2020-03-04
Apps are hashed for memo in two different ways


benclifford:
**Describe the bug**
Two different mechanisms are used for hashing app functions for memoization.

The `task_record['fn_hash']` value is computed as `self.func_hash = md5(self.fn_source.encode('utf-8')).hexdigest()` in `parsl/app/app.py`.

The function is also hashed using the id_for_memo method, because the app function gets passed to a wrapper inside parsl and so is hashed as an argument to that wrapper. Issue #1591 describes issues related to that.

**Expected behavior**
Different methods of generating a hash lead to different rules for identity of apps between runs. There should be a single rule in use, probably whatever the rule used for id_for_memo is, and the separate `fn_hash` computation should go away.

benclifford:
Fixed by #1945

Issue #1591, opened 2020-03-04
Memoization of functions is awkwardly sensisitive to irrelevant source code changes


benclifford:
**Describe the bug**
Sometimes functions are passed as parameters to app functions, and so are used as material for the memoization hash (`task_record['hashsum']`). This does not work well, resulting in apparently-irrelevant changes to source code causing memoization to miss previously computed results.

This can happen both in when a user passes in a function (for example a wrapper function) explicitly, but more importantly happens when the raw app function is passed into one of the several internal parsl wrapper functions - so even if a workflow user does not pass in a function explicitly, function memoization happens.

The current master behaviour is to checkpoint functions using the default serialization-based mechanism, registered here:
```
@id_for_memo.register(types.FunctionType)
def id_for_memo_serialize(obj, output_ref=False):
```
which in turn uses `CannedFunction` from `parsl/executors/serialize/canning.py` which in turn uses the code pickling in `parsl/executors/serialize/codeutil.py`. 

`codeutil.py` includes lots of fields into this result:
```
    args = [co.co_argcount, co.co_nlocals, co.co_stacksize,
            co.co_flags, co.co_code, co.co_consts, co.co_names,
            co.co_varnames, co.co_filename, co.co_name, co.co_firstlineno,
            co.co_lnotab, co.co_freevars, co.co_cellvars]

```

so at the very least, the memoization for a function depends on the line number at which it is defined, and so adding a one line comment at the top of a source file invalidates the memoization hash of every function defined in that source file.

**Expected behavior**
This is up for debate.

My favoured option at the moment is that function memo hashing should be quite tolerant, maybe using just the function module/name, and not be dependent at all on how that function is defined. Dependencies on the code inside the function mean that changing a function definition directly causes memo invalidation, but changing any helper functions used by that function do not cause a memo invalidation. This is unintuitive.

I've modified the `lsst-dm-202002` branch to have a lighter form of function memo hashing that only looks at module and name, and have been using that with @tomglanzman

benclifford:
PR #1766 switches function memoization to using only function name and module (usually, the python source file, but I'm unsure how that works in notebooks)

yadudoc:
I agree that memoization should not be dependent on the line no of the function, or its docs, and various other notebook related variables. 

[...]

PR #1766, opened 2020-06-17
Memoize functions by only module:name, not content

benclifford:
I'm unhappy with how functions are memoized/checkpointed in parsl 1.0.0 and previous versions.

The traditional behaviour (before this PR) has been to serialize the function object, and use that serialised data as input to the memoization hash. That works badly, as described in #1591. This PR attempts to make this behaviour more consistent.

Discussion about the approach should happen in related issue #1591 

This PR should probably address the double function hashing described in #1592


benclifford:
closing this in favour of more recent work

PR #1767, opened 2020-06-17
Generate Gantt chart from status table rather than task table timestamps

benclifford:
The status table captures richer state information than the task
table timestamp columns.

The chart will now show: retries, the initial period when a task is in
pending state, and continue to the end of the workflow showing final
(success or failure) state.

Previous:

previous plot

New:

new plot


ZhuozhaoLi:
Do you have the code snippet on producing that gantt chart? I want to have a try there.

benclifford:
It is from the monitoring.db from: pytest parsl --config parsl/tests/configs/htex_local_alternate.py

PR #1770, opened 2020-06-18
Change python <3.6 deprecation warning to RuntimeError

benclifford:


parsl/dataflow/dflow.py
```diff
@@ -82,7 +82,8 @@ def __init__(self, config=Config()):
         logger.debug("Starting DataFlowKernel with config\n{}".format(config))
 
         if sys.version_info < (3, 6):
-            logger.warning("Support for python versions < 3.6 is deprecated and will be removed after parsl 0.10")
+            logger.error("Support for python versions < 3.6 stopped after parsl 1.0.0")
```
yadudoc, line 85: Is this necessary? With setup.py specifying `python_requires=">=3.6.0",` you won't be able to install parsl at all.

yadudoc:
I just tested this, and you do get an error when you attempt to install:
```
(parsl_py3.5) yadu@x:~/src/parsl$ pip install .
Processing /home/yadu/src/parsl
parsl requires Python '>=3.6.0' but the running Python is 3.5.6
```

parsl/dataflow/dflow.py
benclifford: i'm fine with removing it entirely

yadudoc: Let's do that.

PR #1774, opened 2020-06-19
Remove denormalised workflow duration from monitoring db

benclifford:
This value be computed from other fields in the database,
so should not be stored in the DB:

  duration = time_completed - time_began

Previously, the workflow duration was reported as None/NULL
when a workflow was not finished. This PR adds a helper
which provides NULL-aware subtraction for cases where
time_completed is null.

### DIFF
--- a/parsl/monitoring/visualization/templates/dag.html
+++ b/parsl/monitoring/visualization/templates/dag.html
@@ -10,7 +10,7 @@
 <ul>
     <li><b>Started:</b>  {{ workflow_details['time_began'] | timeformat }}</li>
     <li><b>Completed:</b>  {{ workflow_details['time_completed'] | timeformat }}</li>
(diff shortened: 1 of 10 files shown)


parsl/monitoring/visualization/templates/dag.html
```diff
@@ -10,7 +10,7 @@ <h2>{{ workflow_details['workflow_name'] }}</h2>
 <ul>
     <li><b>Started:</b>  {{ workflow_details['time_began'] | timeformat }}</li>
     <li><b>Completed:</b>  {{ workflow_details['time_completed'] | timeformat }}</li>
-    <li><b>Workflow duration:</b>  {{ workflow_details['workflow_duration'] | timeformat }}</li>
+    <li><b>Workflow duration:</b>  {{ (workflow_details['time_began'] - workflow_details['time_completed']) | durationformat }}</li>
```
ZhuozhaoLi, line 13: It should be `(workflow_details['time_began'], workflow_details['time_completed'])`

PR #1778, opened 2020-06-19
Remove unused fields from monitoring UDPRadio protocol

benclifford:


parsl/monitoring/monitoring.py
```diff
@@ -558,7 +554,7 @@ def monitor(pid,
             d['psutil_process_time_user'] += total_children_user_time
             d['psutil_process_time_system'] += total_children_system_time
             logging.debug("sending message")
-            radio.send(MessageType.TASK_INFO, task_id, d)
+            radio.send(d)
```
yadudoc, line 557: Don't we need a matching change on the receiver side?

benclifford: db_manager ignores all fields except the final field:

> self.pending_resource_queue.put(x[-1])

so anything or nothing can be sent before the final `d` without needing a receive-side change.

yadudoc: I see. Now that I look at this code, I see that we have a thread for each queue(priority, node, resource). I bet this is mostly redundant.

Issue #1780, opened 2020-06-21
Reconnect SSH channel when connection drops


ravihansa3000:
Currently, if the SSH connection drops (due to network-related issues, enforced connection close or inactivity) Parsl does not attempt to re-establish the connection. This feature request is to improve SSH channel by implementing a connection restore mechanism.

**Solution**
Check if the SSH connection transport is active before executing SSH commands. Attempt to re-establish the connection if not.

**Describe alternatives**
Set keepalive SSH config. This is not useful if the connection close is enforced by the SSH server-side.

**Additional context**
This feature is important for long-running Parsl jobs since SSH connection drops are not that uncommon.

PR #1781, opened 2020-06-21
SSH channel - reconnect if the transport is not active

ravihansa3000:
- Type: new feature (improvement for SSH channel)

- Implements #1780

- Description
Checks whether underlying SSH transport is active before executing SSH commands or performing SFTP operations. Attempts to re-establish the connection if it is not active.

### DIFF
--- a/parsl/channels/ssh/ssh.py
+++ b/parsl/channels/ssh/ssh.py
@@ -60,36 +60,52 @@ class SSHChannel(Channel, RepresentationMixin):
             self.ssh_client = paramiko.SSHClient()
         self.ssh_client.load_system_host_keys()
         self.ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        self.sftp_client = None
 
         self.envs = {}
         if envs is not None:
             self.envs = envs
 
-        try:
-            self.ssh_client.connect(
-                hostname,
-                username=username,
-                password=password,
-                port=port,
-                allow_agent=True,
-                gss_auth=gssapi_auth,
-                gss_kex=gssapi_auth,
-                key_filename=key_filename
-            )
-            t = self.ssh_client.get_transport()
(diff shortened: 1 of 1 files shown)


ravihansa3000:
@benclifford This will help us to run long-running Parsl jobs from a remote node over the SSH channel.

.gitignore
```diff
@@ -11,6 +11,7 @@ venv/*
 web/*
 api/*
 .idea/*
+.vscode/
```
benclifford, line 14: This should be a separate one-line PR

parsl/channels/ssh/ssh.py
```diff
@@ -119,7 +128,7 @@ def execute_wait(self, cmd, walltime=2, envs={}):
         '''
 
         # Execute the command
-        stdin, stdout, stderr = self.ssh_client.exec_command(
+        stdin, stdout, stderr = self._connection()[0].exec_command(
```
benclifford, line 131: I think this _connection()[0] style is really ugly, compared to the previously style. It is less clear than before what kind of connection is being used here. It is especially disappointing because there are two nicely named fields that _connection is maintaining (self.ssh_client, and self.sftp_client).

A nicer way might be to make two accessor functions, `self.valid_ssh_client()` and `self.valid_sftp_client()` which are something like this:

  def valid_sftp_client():
    self._connection()
    return self.sftp_client

That is, they check for the connection being valid, before returning that appropriate connection.

Then the above line would look like:
```
stdin, stdout, stderr = self.valid_ssh_client().exec_command(
```

ravihansa3000:
@benclifford I agree. Made changes as suggested.

ravihansa3000:
@benclifford do you need me to work on any other changes?

benclifford:
> @benclifford do you need me to work on any other changes?

No, i think this is fine. I was waiting to test it out on one of my systems but for various reasons that's been delayed, so I will merge this now.

PR #1783, opened 2020-06-24
Support for reporting asynchronous provider errors

### LINKED ISSUE #1035: htex/exex worker startup failure results in silent hangs

hategan:
# Description

This PR adds a `_fail_job_async` method to the `StatusHandlingExecutors` that fires when more jobs than a threshold have been submitted and they have all failed. When this failure condition is reached the executor will now shutdown and raise errors on all tasks.

This is the remainder of PR #1420. Fixes #1035

## Type of change


parsl/launchers/launchers.py
```diff
@@ -6,6 +6,10 @@
 logger = logging.getLogger(__name__)
 
 
+def bool_to_num(b: bool):
```
yadudoc, line 9: Maybe remove this.

parsl/executors/high_throughput/executor.py
```diff
@@ -573,19 +575,30 @@ def scale_out(self, blocks=1):
...
         return r
 
+    def _launch_block(self, external_block_id: str) -> Any:
+        if self.launch_cmd is None:
+            raise ScalingFailed(self.provider.label, "No launch command")
+        launch_cmd = self.launch_cmd.format(block_id=external_block_id)
+        internal_block = self.provider.submit(launch_cmd, 1)
+        logger.debug("Launched block {}->{}".format(external_block_id, internal_block))
+        if not internal_block:
+            raise(ScalingFailed(self.provider.label,
```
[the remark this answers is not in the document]
hategan: This PR has the infrastructure for error handling. It is only fully implemented at this moment in the local provider. Providers should, in their submit() method, report errors through exceptions rather than magical return values. However, updates to all providers to do so are not within the scope of this PR.

parsl/executors/status_handling.py
```diff
@@ -80,6 +95,20 @@ def bad_state_is_set(self):
     def executor_exception(self):
         return self._executor_exception
 
+    @property
+    def error_management_enabled(self):
+        return True
+
+    def handle_errors(self, error_handler: "parsl.dataflow.job_error_handler.JobErrorHandler",
+                      status: Dict[Any, JobStatus]) -> bool:
+        init_blocks = 3
```
yadudoc, line 104: It's a bit harder to read when there's some logic here in handle_errors and some over in the `simple_error_handler`, 
when this could just be consolidated in the JobErrorHandler.

PR #1786, opened 2020-06-30
Make channel overwrite check optional for pull_file and push_file

### LINKED ISSUE #1785: Fix `FileExists` error when running `LocalProvider` + `SSHChannel`
As reported by @ravihansa3000, a `FileExists` error is raised when running with `LocalProvider` + `SSHChannel`:

```
2020-06-28 02:02:27 parsl.channels.ssh.ssh:215 [ERROR] Remote file copy will overwrite a local file:/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Downloader-Output/remote/COAD_0/runinfo/001/submit_scripts/parsl.localprovider.1593280864.8464878.sh.ec
NoneType: None
2020-06-28 02:02:27 parsl.dataflow.flow_control:114 [ERROR] Flow control callback threw an exception - logging and proceeding anyway
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/flow_control.py", line 112, in make_callback
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 63, in poll
self._update_state()
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 69, in _update_state
item.poll(now)
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/dataflow/task_status_poller.py", line 27, in poll
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/executors/status_handling.py", line 59, in status
status = self._make_status_dict(job_ids, self._provider.status(job_ids))
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 87, in status
str_ec = self._read_job_file(script_path, '.ec').strip()
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 142, in _read_job_file
path = self._job_file_path(script_path, suffix)
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/providers/local/local.py", line 138, in _job_file_path
path = self.channel.pull_file(path, self.script_dir)
File "/home/akila/Documents/NUS_CSI/Parsl/parsl.git/parsl/channels/ssh/ssh.py", line 216, in pull_file
raise FileExists(None, self.hostname, filename=local_dest)
parsl.channels.errors.FileExists: Hostname:nus.nscc.sg, Reason:File name collision in channel transport phase:/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Downloader-Output/remote/COAD_0/runinfo/001/submit_scripts/parsl.localprovider.1593280864.8464878.sh.ec
```

This was introduced in #1724 and there is already discussion there with @hategan and @yadudoc, but I'm opening an issue so we don't lose track of this.

hategan:
# Description

The ssh channel checks if the destination file exists before transferring a file using pull_file. This can cause problems when one needs to repeatedly transfer the same file (such as when the file contains live data from a running task). Specifically, issue #1785 is a manifestation of this problem. This PR adds an overwrite_ok parameter to push_file and pull_file that can be used to guarantee that such a check will not be performed.

Setting overwrite_ok to False does not guarantee that no overwrite will happen. This is explained in the comments for the channel ABC. In principle, it should be possible to do open(fn, "xb") instead open(fn, "wb") to implement a check that is not subject to race conditions. However, the call to open the destination file is, in many cases, hidden inside libraries that implement transfers and, therefore, not straightforwardly modifiable.

The PR also modifies test_local_provider.py to run /bin/sleep 1 while polling for status every 0.1s, which should trigger issue #1785.

Fixes #1785 

## Type of change

Please check the options below:

- [x] Bug fix (non-breaking change that fixes an issue)

"Non-breaking" is meaningless. Most code change will change something in behavior, however subtle. What you might want to say here is "changed signature of documented method/class/etc." instead. And make it separate from bug fix/not bug fix. Because both are boolean and you can have all 4 combinations.

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code maintentance/cleanup


hategan:
Ben, Anna, if you have the time, can you comment on whether you think overwrite_ok should, by default, be False or True?

Setting it to "False" could potentially catch some concurrency issues. Setting it to "True" would more closely mimic what scp does and what the local channel does.

benclifford:
I don't really find any of the use-cases for using overwrite_ok=False anywhere very compelling, especially as real implementations won't do that mode very well (as @hategan has commented). Which makes me think overwrite_ok=True should be the default, or to not have that option at all.

annawoodard:
I also can't think of compelling cases for `overwrite_ok=False`, so I slightly favor removing the `overwrite_ok` option and always overwriting. Options are a burden to understand and maintain, so in my view we should only add them if we need them now, not if we might need them someday.

benclifford:
This PR changes more than channel overwrite checking.

hategan:
@benclifford Yeah, I need to double check, but I think I mistakenly branched this from the wrong branch.

@all, it seems like the consensus is for a simpler solution that removes the overwrite check from the ssh channel without any method signature changes; I'm all for that. However, I'm also in the process of moving and so am taking the week off. If anybody can do a PR with that, it would be great. If not, I'll try to see if I can find some time.

hategan:
Given #1800, I'm voting for deleting this PR and its branch.

benclifford:
closing in favour of merged #1800

PR #1789, opened 2020-07-04
fix #1788 by looping over "subdeps" and evaluating.

### LINKED ISSUE #1788: positional args aren't introspected and "sanitized"
running this in on our binder instance produces two different exceptions

```python
import parsl

parsl.load(config)


@python_app
def monte_carlo():
    from random import random
    from math import sin, cos
    u = random()
    return (sin(u)) / (u + cos(u) ** 2)


@python_app
def mean1(a, b, c):
    return (a + b + c) / len(fus)


@python_app
def mean2(inputs=[]):
    return sum(inputs) / len(inputs)


@python_app
def mean3(inputs=[]):
    return (inputs[0] + inputs[1] + inputs[2]) / len(inputs)


fus = [monte_carlo() for _ in range(3)]
integ = mean2(fus)
print("Average: {:.5f}".format(integ.result()))

fus = [monte_carlo() for _ in range(3)]
integ = mean3(fus)
print("Average: {:.5f}".format(integ.result()))
```

Call to `mean2` produces a type error from `int + AppFuture`. Call to `mean3` produces a type error from `AppFuture + AppFuture`.

makslevental:
# Description

added a loop to iterate over positional args that are themselves either lists or tuples

Fixes #1788 

## Type of change

Please check the options below:

- [x ] Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -606,7 +606,18 @@ class DataFlowKernel(object):
         # Replace item in args
         new_args = []
         for dep in args:
-            if isinstance(dep, Future):
+            if isinstance(dep, list) or isinstance(dep, tuple):
+                new_sub_args = []
+                for sub_dep in dep:
+                    if isinstance(sub_dep, Future):
+                        try:
+                            new_sub_args.extend([sub_dep.result()])
+                        except Exception as e:
+                            dep_failures.extend([e])
+                    else:
+                        new_sub_args.extend([sub_dep])
+                new_args.extend([new_sub_args])
+            elif isinstance(dep, Future):
                 try:
                     new_args.extend([dep.result()])
                 except Exception as e:


parsl/dataflow/dflow.py
```diff
@@ -606,7 +606,18 @@ def sanitize_and_wrap(self, task_id, args, kwargs):
         # Replace item in args
         new_args = []
         for dep in args:
-            if isinstance(dep, Future):
+            if isinstance(dep, list) or isinstance(dep, tuple):
+                new_sub_args = []
```
benclifford, line 610: does this mean that if I pass in a future of a tuple to an app, that app will see a `list` rather than a `tuple` as its parameter?

parsl/dataflow/dflow.py
```diff
@@ -606,7 +606,18 @@ def sanitize_and_wrap(self, task_id, args, kwargs):
...
+                for sub_dep in dep:
+                    if isinstance(sub_dep, Future):
+                        try:
+                            new_sub_args.extend([sub_dep.result()])
```
benclifford, line 614: Is the `sub_dep` future completed at this point for sure? If not, I think this introduces unexpected behaviour (perhaps blocking in the callback handling for other tasks, which can be from arbitrary executor threads, perhaps then deadlocking execution)

There is list of dependencies which the DataFlowKernel waits for, generated by `DataFlowKernel._gather_all_deps` and these changes to sanitize_and_wrap probably need to be reflected there.

yadudoc:
Let's make this a draft and mark this as a priority for the next release.

PR #1795, opened 2020-07-08
Rename and rearrange monitoring queues for clarity

benclifford:
# Description

This PR renames the variable names for most message structures in the monitoring database to make it clearer what goes into each queue. Monitoring behaviour should be unaffected.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -5,6 +5,8 @@ import os
 import time
 import datetime
 
+from typing import Any, Dict, Set
+
 from parsl.log_utils import set_file_logger
 from parsl.dataflow.states import States
 from parsl.providers.error import OptionalModuleMissing
@@ -223,7 +225,7 @@ class DatabaseManager:
         self.pending_node_queue = queue.Queue()
         self.pending_resource_queue = queue.Queue()
 
-    def start(self, priority_queue, node_queue, resource_queue):
+    def start(self, priority_queue, node_queue, resource_queue) -> None:
 
         self._kill_event = threading.Event()
         self._priority_queue_pull_thread = threading.Thread(target=self._migrate_logs_to_internal,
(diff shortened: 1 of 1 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -301,78 +311,83 @@ def start(self, priority_queue, node_queue, resource_queue):
...
-                            left_messages[msg['task_id']] = msg
-                if first_messages:
-                    self._insert(table=STATUS, messages=first_messages)
-                    self._update(table=TASK,
-                                 columns=['task_time_running',
-                                          'run_id', 'task_id',
-                                          'hostname'],
-                                 messages=first_messages)
+                            if msg['task_id'] in deferred_resource_messages:
+                                logger.error("Task {} already has a deferred resource message. Discarding previous message.".format(msg['task_id']))
```
ZhuozhaoLi, line 381: This is a good point. Do you see this quite often? I assume it would happen when the `resource_monitoring_interval` is very small. 

I wondered if it is appropriate to discard any of these messages. But anyway, this is not related to this PR.

parsl/monitoring/db_manager.py
```diff
@@ -251,40 +253,48 @@ def start(self, priority_queue, node_queue, resource_queue):
...
             """
             logger.debug("""Checking STOP conditions: {}, {}, {}, {}, {}""".format(
                               self._kill_event.is_set(),
                               self.pending_priority_queue.qsize() != 0, self.pending_resource_queue.qsize() != 0,
                               priority_queue.qsize() != 0, resource_queue.qsize() != 0))
 
-            # This is the list of first resource messages indicating that task starts running
-            first_messages = []
+            # This is the list of resource messages which can be reprocessed as if they
+            # had just arrived because the corresponding first task message has been
```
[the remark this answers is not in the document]
benclifford: no I mean *because*.

The resource messages did arrive *before* the corresponding first task messages, and so were placed into the `deferred_resource_messages` queue.

At this point in the code, we have processed the corresponding first task message, and so *because* of that, the code can process the messages as if they had just arrived.

parsl/monitoring/db_manager.py
benclifford: I have never seen it in real life but I wanted to log that this happened rather than silently discarding.

PR #1798, opened 2020-07-13
Add support for GPU kubernetes grids

MatthewBM:
# Description

Added option to specify number of GPUs to assign to pods in kubernetes provider

## Type of change

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -70,6 +70,8 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
     persistent_volumes: list[(str, str)]
         List of tuples describing persistent volumes to be mounted in the pod.
         The tuples consist of (PVC Name, Mount Directory).
+    n_gpus : int
+        number of CUDA-enabled GPUs to assign to pod
     """
     @typeguard.typechecked
     def __init__(self,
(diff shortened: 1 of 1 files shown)


benclifford:
My gut says that it would be to allow a user to specify arbitrary resource requests and limits in the configuration, rather than hard-coding specific choices for which request and limits are allow.

See how the slurm provider has a scheduler_options parameter to allow arbitrary slurm options to be specified.

That could be an `extra_request` dict and an `extra_limits` dict that are merged with the appropriate dict at the point that this PR adds in the n_gpus entries.

This would allow the same option to support other resource limits - I think `hugepages` and `ephemeral-storage` are available but unsupported by parsl at the moment, for example.

MatthewBM:
> My gut says that it would be better to allow a user to specify arbitrary resource requests and limits in the configuration, rather than hard-coding specific choices for which request and limits are allow.
> 
> See how the slurm provider has a scheduler_options parameter to allow arbitrary slurm options to be specified.
> 
> That could be an `extra_request` dict and an `extra_limits` dict that are merged with the appropriate dict at the point that this PR adds in the n_gpus entries.
> 
> This would allow the same option to support other resource limits - I think `hugepages` and `ephemeral-storage` are available but unsupported by parsl at the moment, for example.

benclifford:
@MatthewBM I think this functionality would be useful to have merged -- just suggesting a more generic possibility

MatthewBM:
Yes I agree I'll do some more work on it, I'm realizing that some kubernetes grids configurations may not have nvidia.com/gpu as a resource label at all so that code could cause exceptions there. I'll update this in the next few days.

PR #1801, opened 2020-07-14
Do not use "wait -n" in order to support bash versions older than 5.0.

### LINKED ISSUE #1797: Support older versions of Bash
Slurm submission scripts written by Parsl 0.7 use the Bash builtin command 'wait' with the -n option. I believe this option was introduced in a relatively recent version of Bash, perhaps 5.0 Many of the systems at my organization run Redhat Enterprise Linux 7, which uses Bash-4.2. This version of bash does not recognize the -n option to wait, and the script errors out.

Please consider implementing the slurm submission script without using wait -n to improve compatibility with older versions of bash, which I imagine are still in common use.

hategan:
# Description

The single node launcher introduced by PR #1724 uses "wait -n". The "-n" flag was introduced in bash v5. This PR rewrites that piece of code to not use said flag.

Fixes #1797 

## Type of change

### DIFF
--- a/parsl/launchers/launchers.py
+++ b/parsl/launchers/launchers.py
@@ -88,6 +88,7 @@ export CORES=$(getconf _NPROCESSORS_ONLN)
 [[ "{debug}" == "1" ]] && echo "Found cores : $CORES"
 WORKERCOUNT={task_blocks}
 FAILONANY={fail_on_any}
+PIDS=""
 
 CMD() {{
 {command}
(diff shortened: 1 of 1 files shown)


benclifford:
The single node launcher introduced by PR #1724 uses "wait -n". The "-n" flag was introduced in bash v5. This PR rewrites that piece of code to not use said flag.

Fixes #1797

Issue #949, opened 2019-05-21
Fold back in improvements from funcX executor


annawoodard:

yadudoc:
Related to #1013

PR #1806, opened 2020-07-21
New serialization system to replace IPP serialization

yadudoc:
# Description

This PR adds a modular serialization system that attempts a bunch of serialization methods. This is expected to replace `ipyparallel.serialize` which often uses it's own serialization methods which often fail esp (eg, some numpy objects).
 
Fixes # (issue)

Relates to #949 

## Type of change


parsl/serialize/facade.py
```diff
@@ -0,0 +1,163 @@
...
+
+    """
+
+    def __init__(self):
+        """ Instantiate the appropriate classes
+        """
+        headers = list(METHODS_MAP_CODE.keys()) + list(METHODS_MAP_DATA.keys())
+        self.header_size = len(headers[0])
+
+        self.methods_for_code = {}
```
benclifford, line 34: It's not clear to me what this code/data distinction is for.

benclifford:
There's a bunch of mess here that I won't comment on until you have a go at cleaning this up.

But I especially dislike the use of a magically allocated identifier number that needs to come from a fairly global registry to be unique. That's a barrier to modularity/extensibility.

Maybe it's a bit more asymmetric: if I have an instance of type x, I need to know which serializer to use to turn it into a byte stream. The instance of type x does not know that, though, but serializers can register which types they do know how to deserialize.

If I have a serialised byte stream, I need to know which deserializer to use to turn into into some object; but because parsl controls the byte stream format, it can add in any kind of annotations it want; such at the fully qualified name of the deserialiser class?

parsl/serialize/facade.py
yadudoc: If the object is not a callable (data), we only use pickle and dill. However if the object is code pickle/dill may not work straight-away, and we might need to code as text and do some extra handling. Secondly, I want to add caching for serialized code, and not for data.

benclifford: I don't understand either of those distinctions:

i) what about non-callable objects where pickle/dill don't work?
ii) what is special about a callable object that I am sending round that isn't special about a non-callable object that I'm sending around?

parsl/app/bash.py
```diff
@@ -123,6 +124,13 @@ def __init__(self, func, data_flow_kernel=None, cache=False, executors='all', ig
             if sig.parameters[s].default is not Parameter.empty:
                 self.kwargs[s] = sig.parameters[s].default
 
+        # update_wrapper allows remote_side_bash_executor to masquerade as self.func
+        # partial is used to attach the first arg the "func" to the remote_side_bash_executor
+        # this is done to avoid passing a function type in the args which parsl.serializer
+        # doesn't support
```
benclifford, line 130: I'm using functions as args (to bash apps) in the LSST DRP workflow. I think you would have to argue hard with me to convince me it is better to prohibit that and for me to move to a different style.

parsl/serialize/base.py
```diff
@@ -0,0 +1,105 @@
...
+        self.reason = reason
+
+    def __str__(self):
+        return self.reason
+
+    def __repr__(self):
+        return self.__str__()
+
+
+class RemoteExceptionWrapper:
```
benclifford, line 96: Maybe this should be de-duped against the `RemoteExceptionWrapper` in parsl.app.errors

parsl/serialize/concretes.py
```diff
@@ -0,0 +1,155 @@
...
+        return self.identifier + x
+
+    def deserialize(self, payload):
+        chomped = self.chomp(payload)
+        data = dill.loads(codecs.decode(chomped.encode(), 'base64'))
+        return data
+
+
+class json_base64(fxPicker_shared):
+    """ Json based serialization does not guarantee consistency on tuple/list types
```
benclifford, line 49: what do you mean by "consistency" here?

parsl/app/bash.py
yadudoc: I was worried about this. If the function can be serialized by pickle, this shouldn't break. Can you point me to your code ?

benclifford:
@yadudoc and I had a talk about this PR yesterday. I think the main thing I ended up concerned about is that it needs to be clear (for example, through CI tests) when the various non-`pickle`/`dill` serializers are used, as they seem to be for corner cases.

We also talked about using strings vs bytestrings as the serialisation form: this PR uses strings because funcx wants to put send serialised data around using non-8-bit-clean protocols; but bytestrings are probably a better serialisation form. I suggested a more traditional approach of serialising to/from 8-bit clean bytestrings in the serialiser code, and separately in the funcx code encoding those 8 bit bytestrings into something that can be transmitted through non-8-bit-clean code (eg base 64, or whatever). This keeps serialisation implementation separate from wire-protocol 8-bit-cleanliness.

parsl/serialize/concretes.py
yadudoc: Json serializer will convert tuples to lists.

yadudoc: Ben, I've updated the docstrings on the serializers including json.

parsl/serialize/base.py
yadudoc: Deduped.

parsl/app/bash.py
yadudoc: With the updates, the serializer will try pickle and if that fails - dill to try serialize your params. From testing pickle and dill should be able to handle most situations.

parsl/dataflow/memoization.py
```diff
@@ -1,7 +1,10 @@
 import hashlib
 from functools import singledispatch
 import logging
-from parsl.executors.serialize.serialize import serialize_object
+# from parsl.executors.serialize.serialize import serialize_object
+from parsl.serialize import ParslSerializer
```
benclifford, line 5: from a mostly style point of view, can these globals be defined in parsl.serialize and imported?

`from parsl.serialize import serialize_object`  reads much better to me than this block of code

parsl/serialize/concretes.py
```diff
@@ -0,0 +1,89 @@
...
+        x = dill.dumps(data)
+        return self.identifier + x
+
+    def deserialize(self, payload):
+        chomped = self.chomp(payload)
+        data = dill.loads(chomped)
+        return data
+
+
+class JsonSerializer(fxPicker_shared):
```
benclifford, line 65: If this class is specific to a particular user of parsl, (i.e. funcx), then perhaps this belongs in the funcx codebase: subclasses of fxPicker_shared can be defined anywhere, I think?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,190 @@
...
+        serialized = None
+        serialized_flag = False
+        kind = None
+        if callable(obj):
+            kind = 'callable'
+            for method in self.methods_for_code.values():
+                try:
+                    serialized = method.serialize(obj)
+                    # We attempt a deserialization to make sure both work.
+                    method.deserialize(serialized)
```
benclifford, line 99: why does this test only happen for `callable`s?

parsl/serialize/base.py
```diff
@@ -0,0 +1,68 @@
...
+import functools
+
+logger = logging.getLogger(__name__)
+
+# GLOBALS
+METHODS_MAP_CODE = {}
+METHODS_MAP_DATA = {}
+
+
+class fxPicker_enforcer(metaclass=ABCMeta):
```
benclifford, line 12: give this and fxPicker_shared a better a name?

benclifford: actually fxPicker_enforcer is not used, i think, so remove

benclifford: and move abstractmethods into fxPicker_shared

parsl/serialize/facade.py
```diff
@@ -0,0 +1,190 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ParslSerializer(object):
+    """
+    Information that we want to be able to ship around:
+
+
+    * Function run information ->
+    * Function id <---> Container id ? (is there a 1-1 mapping here?)
```
benclifford, line 14: is there?

parsl/serialize/base.py
yadudoc: Done, and it's now renamed to `SerializerBase`

parsl/serialize/facade.py
yadudoc: Removed.

parsl/dataflow/memoization.py
```diff
@@ -1,7 +1,8 @@
 import hashlib
 from functools import singledispatch
 import logging
-from parsl.executors.serialize.serialize import serialize_object
+# from parsl.executors.serialize.serialize import serialize_object
```
benclifford, line 4: dead comment

parsl/dataflow/memoization.py
```diff
@@ -167,12 +167,12 @@ def make_hash(self, task):
             t = t + [id_for_memo(outputs, output_ref=True)]   # TODO: use append?
 
         t = t + [id_for_memo(filtered_kw)]
-
         t = t + [id_for_memo(task['func_name']),
                  id_for_memo(task['fn_hash']),
                  id_for_memo(task['args'])]
 
         x = b''.join(t)
+        # hashedsum = hashlib.md5(x.encode('utf-8')).hexdigest()
```
benclifford, line 175: dead code

parsl/serialize/concretes.py
```diff
@@ -0,0 +1,61 @@
...
+    * classes defined in local context and not importable using a fully qualified name
+    * clojures, generators and coroutines
+    * [sometimes] issues with wrapped/decorated functions
+    """
+
+    _identifier = b'01\n'
+    _for_code = True
+    _for_data = True
+
+    def __init__(self):
```
benclifford, line 22: this init could go away because it does nothing other than the default behaviour, right?

parsl/serialize/concretes.py
```diff
@@ -0,0 +1,61 @@
...
+    * classes defined in local context and not importable using a fully qualified name
+    * functions that are wrapped/decorated by other functions/classes
+    * clojures
+    """
+
+    _identifier = b'02\n'
+    _for_code = True
+    _for_data = True
+
+    def __init__(self):
```
benclifford, line 51: remove this

parsl/serialize/facade.py
```diff
@@ -0,0 +1,183 @@
...
+        func: Function
+            A function to ship
+
+        args: Tuple/list of objects
+            positional parameters as a list
+
+        kwargs: Dict
+            Dict containing named parameters
+
+        buffer_threshold: Ignored
```
benclifford, line 57: there's no need to have this and item_threshold if they're ignored? retaining ipyparallel.serialize invocation compatibilty doesn't seem like a goal?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        return [self.deserialize(buf) for buf in self.unpack_buffers(packed_buffer)]
+
+    def serialize(self, obj, buffer_threshold=1e6):
+        """ Try available serialization methods one at a time
+
+        If all serialization methods fail we raise a TypeError. Ideally we should
+        reraise the exception from the methods we tried which might have more
+        useful info into why the object cannot be serialized
+        TODO ^
```
benclifford, line 81: TODO!

parsl/serialize/facade.py
```diff
@@ -0,0 +1,183 @@
...
+                    serialized = method.serialize(obj)
+                except Exception:
+                    logger.exception(f"Serialization method {method} did not work")
+                    continue
+                else:
+                    serialized_flag = True
+                    break
+
+        if serialized_flag is False:
+            # TODO : Replace with a SerializationError
```
benclifford, line 112: TODO!

parsl/serialize/facade.py
```diff
@@ -0,0 +1,183 @@
...
+           Payload object to be deserialized
+
+        """
+        header = payload[0:self.header_size]
+        if header in self.methods_for_code:
+            result = self.methods_for_code[header].deserialize(payload)
+        elif header in self.methods_for_data:
+            result = self.methods_for_data[header].deserialize(payload)
+        else:
+            raise Exception("Invalid header: {} in data payload".format(header))
```
benclifford, line 133: use some kind of serialization exception class here?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Parameters
+        ----------
+        buffers : list of \n terminated strings
+        """
+        packed = b''
+        for buf in buffers:
+            s_length = bytes(str(len(buf)) + '\n', 'utf-8')
+            print("buf type : ", type(buf))
+            print("s_lenght type : ", type(s_length))
```
benclifford, line 147: there shouldn't be prints in this code

test-requirements.txt
```diff
@@ -15,8 +15,8 @@ typeguard>=2.5
 paramiko
 dill
 tblib
-ipyparallel
 psutil>=5.5.1
 globus-sdk
 twine
 wheel
+pyzmq>=17.1.2
```
benclifford, line 22: what is this for?

yadudoc: pyzmq was a dependency in ipp, and we didn't have it in requirements till now. Now that IPP is completely removed, we have to specify this. I just noticed that ipykernel is still in there, and I'm removing it.

parsl/serialize/facade.py
yadudoc: Removed.

parsl/serialize/facade.py
yadudoc: Sorry, this is a bad comment. I've added functionality to limit buffer size, which is checked in the serialize methods.
I'm removing `item_threshold`

yadudoc: Sorry, this is a bad comment. I've added functionality to limit buffer size, which is checked in the serialize methods.
I'm removing `item_threshold`

parsl/serialize/facade.py
yadudoc: Replaced with raising the last exception raised by the serializer methods.

parsl/serialize/facade.py
yadudoc: Fixed.

parsl/executors/high_throughput/interchange.py
```diff
@@ -14,7 +14,8 @@
 import json
 
 from parsl.version import VERSION as PARSL_VERSION
-from ipyparallel.serialize import serialize_object
+from parsl.serialize import ParslSerializer
+serialize_object = ParslSerializer().serialize
```
benclifford, line 18: this can be a pure import of serialize_object, right?

parsl/app/bash.py
benclifford: so this comment should be more that function types are supported less well?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
+from parsl.serialize.concretes import *  # noqa: F403,F401
```
benclifford, line 1: I don't see a reason to import * here

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        if ParslSerializer.__singleton_instance is None:
+            ParslSerializer.__singleton_instance = object.__new__(cls, *args, **kwargs)
+        return ParslSerializer.__singleton_instance
+
+    def __init__(self):
+        """ Instantiate the appropriate classes
+        """
+        headers = list(METHODS_MAP_CODE.keys()) + list(METHODS_MAP_DATA.keys())
+        self.header_size = len(headers[0])
```
benclifford, line 25: this line makes me incredibly uncomfortable. it's assuming that all headers keys are the same length, or something like that. If there's a fixed value that all headers lengths need to be, hard code it.

or if you want to be more fancy, what happens when headers are different lengths?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+        serialized_flag = False
+        last_exception = None
+        if callable(obj):
+            for method in self.methods_for_code.values():
+                try:
+                    serialized = method.serialize(obj)
+                    # We attempt a deserialization to make sure both work.
+                    method.deserialize(serialized)
+                except Exception as e:
+                    logger.exception(f"Serialization method: {method} did not work")
```
benclifford, line 86: I think this is going to log an `ERROR` level log for every fallover-to-next-serializer serialisation that happens (multiple times per app invocation perhaps), even though no error condition exists. It should be debug or info level.

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+                    continue
+                else:
+                    serialized_flag = True
+                    break
+        else:
+            for method in self.methods_for_data.values():
+                try:
+                    serialized = method.serialize(obj)
+                except Exception as e:
+                    logger.exception(f"Serialization method {method} did not work")
```
benclifford, line 97: see above logger comment

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+                    last_exception = e
+                    continue
+                else:
+                    serialized_flag = True
+                    break
+
+        if serialized_flag is False:
+            # TODO : Replace with a SerializationError
```
benclifford, line 105: remove the TODO line

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            raise TypeError("Invalid header: {} in data payload. Buffer is either corrupt or not created by ParslSerializer".format(header))
+
+        return result
+
+    def pack_buffers(self, buffers):
+        """
+        Parameters
+        ----------
+        buffers : list of \n terminated strings
```
benclifford, line 134: this \n requirement is weird. It looks like the individual strings can contain \n anywhere in them because the \n isn't used to signify anything and packing/unpacking is done by transmitting the string length.
I don't think this really is a requirement?

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+        unpacked = []
+        while packed_buffer:
+            s_length, buf = packed_buffer.split(b'\n', 1)
+            i_length = int(s_length.decode('utf-8'))
+            current, packed_buffer = buf[:i_length], buf[i_length:]
+            unpacked.extend([current])
+
+        return unpacked
+
+    def unpack_and_deserialize(self, packed_buffer):
```
benclifford, line 158: this function is unused?

test-requirements.txt
benclifford: it's a duplicate with `requirements.txt`, though - it shouldn't need to be in test-requirements if it is also a core parsl requirement.

parsl/serialize/facade.py
```diff
@@ -0,0 +1,174 @@
...
+        buffer_threshold: Ignored
+            Limits buffer to specified size in bytes. Default is 1e6 bytes.
+        """
+        b_func = self.serialize(func, buffer_threshold=buffer_threshold)
+        b_args = self.serialize(args, buffer_threshold=buffer_threshold)
+        b_kwargs = self.serialize(kwargs, buffer_threshold=buffer_threshold)
+        packed_buffer = self.pack_buffers([b_func, b_args, b_kwargs])
+        return packed_buffer
+
+    def unpack_apply_message(self, packed_buffer, user_ns=None, copy=False):
```
benclifford, line 64: user_ns and copy are unused here - presumably leftovers from ipp compatibility attempt

PR #1809, opened 2020-07-22
Make wait_for_current_tasks wait for both AppFutures and DFK task status completion

benclifford:
# Description

AppFutures and the DFK state separately indicate that a task is complete, but
at not exactly the same time.

The AppFuture completing does not indicate that some end-of-task-completion processing
has happened, such as monitoring. Because of this, waiting for all AppFutures and then
exiting the process does not guarantee that all end-of-task-completion code will
happen.

This PR makes `dfk.wait_for_current_tasks` also waits for the DFK to indicate that the DFK task record also indicates that
the task is done, when waiting for all tasks to complete at shutdown using wait_for_current_tasks.

This is related to issue #1279, which more broadly discusses task completion issues.

## Type of change


parsl/dataflow/dflow.py
```diff
@@ -920,10 +921,14 @@ def wait_for_current_tasks(self):
             if task_id not in self.tasks:
                 logger.debug("Task {} no longer in task list".format(task_id))
             else:
-                fut = self.tasks[task_id]['app_fu']
+                task_record = self.tasks[task_id]  # still a race condition with the above self.tasks if-statement
```
yadudoc, line 924: Is there a race-condition you've observed here, can you clarify?

parsl/dataflow/dflow.py
```diff
@@ -920,10 +921,14 @@ def wait_for_current_tasks(self):
...
+                fut = task_record['app_fu']
                 if not fut.done():
-                    logger.debug("Waiting for task {} to complete".format(task_id))
                     fut.exception()
+                # now app future is done, poll until DFK state is final: a DFK state being final and the app future being done do not imply each other.
+                while task_record['status'] not in FINAL_STATES:
```
yadudoc, line 929: I believe it is unlikely that we'll be stuck in an infinite loop here, have you seen any hangs over here?

parsl/dataflow/dflow.py
benclifford: The race condition is when a task completes and is deleted from the task list in between checking that it is in the task list (the `if` statement) and assuming that it is in the task list (`task_record =`). I haven't encountered it in practice, so I just left a note in case someone does experience it later.

parsl/dataflow/dflow.py
benclifford: I haven't seen hangs. I think one would happen if the post completion code broke before setting the task record status.

yadudoc (approved):
Let's merge this.

PR #1812, opened 2020-07-23
Use real final state to color DAG visualization

benclifford:
Prior to this, a task state was inferred based on timing information
stored in monitoring.

This PR replaces that state inference with a lookup in the status
table which has more accurate information on task final states.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/visualization/plots/default/workflow_plots.py
+++ b/parsl/monitoring/visualization/plots/default/workflow_plots.py
@@ -203,7 +203,14 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
         groups_list = {app: (i, None) for i, app in enumerate(
             df_tasks['task_func_name'].unique())}
     else:
-        groups_list = {'Pending': (0, 'gray'), "Running": (1, 'blue'), 'Completed': (2, 'green'), 'Unknown': (3, 'red')}
+        groups_list = {"unsched": (0, 'rgb(240, 240, 240)'),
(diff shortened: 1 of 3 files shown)


parsl/monitoring/visualization/plots/default/workflow_plots.py
```diff
@@ -203,7 +203,13 @@ def workflow_dag_plot(df_tasks, group_by_apps=True):
...
             df_tasks['task_func_name'].unique())}
     else:
-        groups_list = {'Pending': (0, 'gray'), "Running": (1, 'blue'), 'Completed': (2, 'green'), 'Unknown': (3, 'red')}
+        groups_list = {"unsched": (0, 'gray'),
+                       "pending": (1, 'gray'),
+                       "running": (2, 'blue'),
+                       "dep_fail": (3, 'red'),
+                       "failed": (4, 'red'),
+                       "done": (5, 'green')
+                      }
```
ZhuozhaoLi, line 212: I see the gantt chart already has a color scheme for different states. It seems better to port that scheme here? Or we do not care to show different colors for different states?

benclifford: you are right - this should use the same colour scheme for consistency

benclifford: @ZhuozhaoLi I've changed to use a consistent colour scheme with the gantt chart, and added in a missing state.

PR #1817, opened 2020-07-28
Rearrange API reference documentation

benclifford:
This PR removes dev_docs.rst, which was mostly a collection of text generated from the python source code.

Some parts of this overlapped heavily with `reference.rst`, and classes mentioned in dev_docs.rst have been moved there.

reference.rst is also arranged into categories, similar to the way dev_docs.rst was organised before

Some of the human readable text from dev_docs.rst is moved into the docstrings of the relevant classes.

## Type of change

- Documentation update


docs/reference.rst
```diff
@@ -1,22 +1,54 @@
-Reference guide
-***************
+API Reference guide
+*******************
+
+Miscellaneous
```
kylechard, line 4: Perhaps core rather than misc?

benclifford: done

Issue #1754, opened 2020-06-13
Remove address hardcoding in config examples


yadudoc:
**Describe the bug**
v1.0.0 adds support for address auto-detection by trying all addresses. We should remove mentions of address hard-coding in the docs. Grepping for `address_by_query` and `address_by_hostname` in `parsl/parsl/configs` yields 25 hits combined. All of these can easily be removed. Some odd cases where an `address_by_interface` could be kept as an example for configuring for special cases.

PR #1833, opened 2020-08-06
address_query replaced in configs

meschw04:
# Description

Made to satisfy issue #1754 on hard-coding address queries in configs.

## Type of change

### DIFF
--- a/parsl/configs/Azure.py
+++ b/parsl/configs/Azure.py
@@ -26,7 +26,6 @@ config = Config(
     executors=[
         HighThroughputExecutor(
             label='azure_single_node',
-            address=address_by_query(),
             provider=AzureProvider(
                 vm_reference=vm_reference,
                 key_file='azure_key_file.json',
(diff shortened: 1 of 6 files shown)


yadudoc:
Hey @meschw04, there are a few more configs that explicitly use `address_by_query`/`address_by_hostname`. We could either merge this PR as is, or if you need more time to finish up the rest of the files  we could do that, or I could do the remainder. Let me know what you think.

parsl/configs/Azure.py
```diff
@@ -26,7 +26,6 @@
     executors=[
         HighThroughputExecutor(
             label='azure_single_node',
-            address=address_by_query(),
```
ZhuozhaoLi, line 29: Should remove the import `address_by_query` above.

yadudoc: ZZ, Azure uses address_by_query further down the config for the `storage_access` option for `RSyncStaging`. So we shouldn't remove the import.

ZhuozhaoLi: Oh, I missed that.

PR #1839, opened 2020-08-10
refactor the mechanism to guarantee at least one message sent for every task in monitoring

### LINKED ISSUE #1837: Monitoring Bug on ALCF Theta
**Describe the bug**

I get some errors when using monitoring on Theta, which I think are causing problems with executors. When using monitoring, my jobs sent to a `ThreadPoolExecutor` execute but I do not see a "Task Completed" notification and the jobs eventually stall:

```
2020-08-08 17:10:06,436 - root - ERROR - Exception reading IO counters for child memory_percent. Recorded IO usage may be incomplete
  File "/lus/theta-fs0/projects/CSC249ADCD08/colmena/env/lib/python3.7/site-packages/parsl/monitoring/monitoring.py", line 551, in monitor
    d['psutil_process_disk_write'] += child.io_counters().write_bytes
AttributeError: 'Process' object has no attribute 'io_counters'
```

**To Reproduce**

The following script computes on my desktop and stalls on the Theta login node:

```python
import logging

from concurrent.futures import as_completed

import parsl
from parsl import ThreadPoolExecutor, python_app
from parsl.monitoring import MonitoringHub

#logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO)

config = Config(
    executors=[
        ThreadPoolExecutor(label="local_threads", max_threads=1)
    ],
    monitoring=MonitoringHub(
        hub_port=55055,
        monitoring_debug=False,
    ),
    strategy=None,
)

parsl.load(config)


@python_app
def func(x):
    return x + 1


futures = [func(i) for i in range(32)]
for f in as_completed(futures):
    print(f.result())
```

**Expected behavior**

No scary error messages and my jobs all executing completely.

**Environment**
 - OS: Centos
 - Python version: 3.7
 - Parsl version: 1.0

**Distributed Environment**
- Where are you running the Parsl script from ? Launch node
- Where do you need the workers to run ? Some on compute nodes, some on launch node

ZhuozhaoLi:
# Description

Due to a process join wait (added in PR #1446 ) problem in monitor wrapper, Parsl hangs when monitoring is configured with `ThreadPoolExecutor`.

As a separate discussion on slack, we have decided to refactor the mechanism to guarantee at least one message sent for every task in monitoring. Instead of relying on the monitor process to send out the first message, `monitor_wrapper` sends out the first message with minimal info (no psutil info) to indicate task running before the actual task starts.  PR implements this mechanism.

Fixes #1837 . Relevant to PR #1446 . 

## Type of change


benclifford:
> Due to an indefinite process join wait in monitor wrapper

I don't understand the cause of this - can you describe more?

ZhuozhaoLi:
> > Due to an indefinite process join wait in monitor wrapper
> 
> I don't understand the cause of this - can you describe more?

It means that the original `p.join()` will wait indefinitely if the `monitor` process does not join properly. That was the root cause of issue #1837 : a task hung on there waiting for monitor process to join, but the monitor logs showed that they all exited (the last line of the logs). It was most likely because there was data inside the `command_q` that was not `get`.

ZhuozhaoLi:
Editing the description and title for this PR as we decided to use a different approach to guarantee that at least one message for one task is sent to monitoring

ZhuozhaoLi:
@benclifford I resolved the conflicts with `master`

WardLT:
Is this one ready to go? I'm currently holding on to a custom branch of Parsl with these fixes, and would like to get back to using the main branch.

ZhuozhaoLi:
> Is this one ready to go? I'm currently holding on to a custom branch of Parsl with these fixes, and would like to get back to using the main branch.

Yes, I think this is good to go. I will pin someone to review this. Just to confirm: is this patch working properly on your custom branch? Thanks @WardLT

parsl/monitoring/monitoring.py
```diff
@@ -300,24 +300,32 @@ def monitor_wrapper(f,
...
+            try:
+                monitor(os.getpid(),
+                        task_id,
+                        monitoring_hub_url,
+                        run_id,
+                        logging_level,
+                        sleep_dur,
+                        first_message=True)
+            except Exception:
+                pass
```
benclifford, line 313: Why is it OK to silently discard exceptions here?

benclifford (approved):

parsl/monitoring/monitoring.py
```diff
@@ -559,17 +576,10 @@ def monitor(pid,
...
-            if msg == "Finished":
-                logging.info("Received task finished message. Ending the monitoring loop now.")
-                break
-        except queue.Empty:
-            logging.debug("Have not received any message.")
-
         logging.debug("sleeping")
         time.sleep(sleep_dur)
+
+    logger.info("Monitor exiting")
```
benclifford, line 585: This statement is unreachable (and so upsets  `mypy` in the benc-mypy branch). This process doesn't exit by reaching the end of its main function; instead it is going to exit by being terminated from outside by the monitor wrapper process. Putting this statement in a `finally` block might cause it to run in that case, though.

PR #1841, opened 2020-08-10
Up the buffer_threshold and raise soft warnings

yadudoc:
Raising buffer limit form 1MB to 128MB and switching to a warning ins…tead of exception on exceeding threshold

# Description

Current serialization methods, used a 1MB buffer size threshold, which is too small as reported by @WardLT. 
This PR updates the threshold to 128 MB, and instead of raising an exception, instead logs a warning to the logs
so that we can catch a situation where parsl's send buffers or the interchange buffers might run out of memory.
The warning is added so that it can help debug a situation where the OS might terminate the processes.

## Type of change


benclifford:
This is failing parsl/tests/test_regression/test_1480.py in CI

yadudoc:
@benclifford Tests are fixed, and it's passing CI.

benclifford (approved):

PR #1845, opened 2020-08-12
Allow user to specify the resource for each task with workqueue executor

ZhuozhaoLi:
# Description

This PR allows users to specify the resource usage for each task in workqueue executor.

Users can define an app as:

```
@python_app
def double(x, parsl_resource_specification={}):
    return x * 2
```

Then when the app is submitted, the resource specification is like:

```
spec = {'cores': 2, 'memory': 1000, 'disk': 1000}
fut = double(n, parsl_resource_specification=spec)
```

## Type of change

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -343,9 +343,13 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
                                           "three resources to be specified simultaneously: cores, memory, and disk, "
                                           "and only takes these three resource types.")
 
-            cores = resource_specification['cores']
-            memory = resource_specification['memory']
-            disk = resource_specification['disk']
+            for k in keys:
+                if k.lower() == 'cores':
+                    cores = resource_specification[k]
+                elif k.lower() == 'memory':
+                    memory = resource_specification[k]
+                elif k.lower() == 'disk':
+                    disk = resource_specification[k]
 
         self.task_counter += 1
         task_id = self.task_counter
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -60,7 +60,7 @@ logger = logging.getLogger(__name__)
 
 
 # Support structure to communicate parsl tasks to the work queue submit thread.
-ParslTaskToWq = namedtuple('ParslTaskToWq', 'id category env_pkg map_file function_file result_file input_files output_files')
+ParslTaskToWq = namedtuple('ParslTaskToWq', 'id category cores memory disk env_pkg map_file function_file result_file input_files output_files')
 
 # Support structure to communicate final status of work queue tasks to parsl
 # result is only valid if result_received is True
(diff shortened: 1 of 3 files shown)


parsl/executors/workqueue/executor.py
```diff
@@ -321,6 +321,32 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+        memory = None
+        disk = None
+        if resource_specification and isinstance(resource_specification, dict):
+            logger.debug("Got resource specification: {}".format(resource_specification))
+
+            acceptable_resource_types = ['cores', 'memory', 'disk']
+            keys = list(resource_specification.keys())
+            if len(keys) != 3:
+                logger.error("Task resource specification requires "
+                             "three resources to be specified simultaneously: cores, memory, and disk")
```
annawoodard, line 334: Why do we require all three to be specified? @btovar should correct me if I'm wrong, but I think we can allow users to provide a guess for one but not the others?

parsl/tests/test_error_handling/test_resource_spec.py
```diff
@@ -11,19 +13,42 @@ def double(x, parsl_resource_specification={}):
...
+    fut = double(n, parsl_resource_specification=spec)
+    try:
+        fut.result()
+    except UnsupportedFeatureError:
+        assert not isinstance(executor, WorkQueueExecutor)
+    except Exception as e:
+        assert isinstance(e, ExecutorError)
+
+    # Specify resources with wrong types
+    # cpus is incorrect
```
annawoodard: for clarity consider changing to:

Resource specification has incorrect keys: `cpus` should be `cores`

parsl/tests/test_error_handling/test_resource_spec.py
```diff
@@ -11,19 +13,42 @@ def double(x, parsl_resource_specification={}):
...
-    spec = {'cores': 2, 'memory': '1GiB'}
+    executors = parsl.dfk().executors
+    executor = None
+    for label in executors:
+        if label != 'data_manager':
+            executor = executors[label]
+            break
+
+    # Specify incorrect number of resources
+    spec = {'cores': 2, 'memory': 1000}
```
annawoodard, line 24: As noted above, not sure why we need to require all three to be specified

parsl/executors/workqueue/executor.py
annawoodard: Testing this PR, I see that if it's not specified, WQ will ask for everything available, which will usually result in one task/worker. Users expecting the constraint to only apply to the specified resource will be surprised. So in light of that, I think it makes sense to require all three to be specified, as implemented here.

parsl/tests/test_error_handling/test_resource_spec.py
annawoodard: resolved above

parsl/executors/workqueue/executor.py
```diff
@@ -321,6 +321,32 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+                raise ExecutorError(self, "Task resource specification requires "
+                                          "three resources to be specified simultaneously: cores, memory, and disk, "
+                                          "and only takes these three resource types.")
+
+            if not all(k.lower() in acceptable_resource_types for k in keys):
```
benclifford, line 339: is it right to have `.lower()` here? what happens one of the keys is "CORES"?

ZhuozhaoLi: The `lower()` here is meant to allow users to not worry about the case? so if a user does something like `parsl_resource_specification={'Cores', 'MEMORY', 'Disk'}`, that should also work.

ZhuozhaoLi: I missed that check in the later code. I will fix that

ZhuozhaoLi:
> The implementation looks fine to me, but I think we need more documentation of this new feature. I see an example in the configuration section, but maybe we should add a few more details. For example, can any units be used, or only GiB? Can you omit the units?

@annawoodard I updated the docs in the configuration section. Feel free to tweak my poor English.

PR #1849, opened 2020-08-19
Add JOBNAME to local provider 

WardLT:
# Description

Unlike the other providers, LocalProvider does not set a JOBNAME environment variable. This causes havoc with some launchers that use the JOBNAME variable.

## Type of change

### DIFF
--- a/parsl/providers/local/local.py
+++ b/parsl/providers/local/local.py
@@ -206,7 +206,7 @@ class LocalProvider(ExecutionProvider, RepresentationMixin):
         script_path = "{0}/{1}.sh".format(self.script_dir, job_name)
         script_path = os.path.abspath(script_path)
 
-        wrap_command = self.worker_init + '\n' + self.launcher(command, tasks_per_node, self.nodes_per_block)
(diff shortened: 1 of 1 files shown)


parsl/providers/local/local.py
```diff
@@ -206,7 +206,7 @@ def submit(self, command, tasks_per_node, job_name="parsl.localprovider"):
         script_path = "{0}/{1}.sh".format(self.script_dir, job_name)
         script_path = os.path.abspath(script_path)
 
-        wrap_command = self.worker_init + '\n' + self.launcher(command, tasks_per_node, self.nodes_per_block)
+        wrap_command = self.worker_init + f'\nJOBNAME=${job_name}\n' + self.launcher(command, tasks_per_node, self.nodes_per_block)
```
benclifford, line 209: the other providers `export` this - that means it will be visible to other processes. Maybe that doesn't affect things in this particular case but it is more consistent.

WardLT: Sure, I'll make the change.

PR #1853, opened 2020-08-25
Allow HTEX to set CPU affinity

### LINKED ISSUE #1344: HTEX forcing thread affinity
I have issues around multi-threaded Parsl apps interfering with each other when placed on the same node. Instead of creating my own logic to prevent apps from attempting to use the same CPU cores, it could be nice for Parsl to enforce affinity itself (e.g., with the [`os` module](https://docs.python.org/3/library/os.html#interface-to-the-scheduler)).

It is related to my issues resulting from Tensorflow's CPU/GPU affinity (#751)

For Parsl to force explicity HTEX workers to use certain CPUs.

Having users make affinity logic themselves.

**Additional context**
I'm going to work up a solution in my own parsl app and then submit it for review.

WardLT:
# Description

This PR introduces an option for HTEX workers to set their CPU affinity on launch. Setting CPU affinity can prevent multithreaded tasks from interfering with each other (e.g., Tensorflow tasks). 

Fixes #1344 

## Type of change

- Documentation update

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -129,6 +129,13 @@ class HighThroughputExecutor(StatusHandlingExecutor, RepresentationMixin):
     max_workers : int
         Caps the number of workers launched by the manager. Default: infinity
 
+    cpu_affinity: string
+        Whether or how each worker process sets thread affinity. Options are "none" to forgo
+        any CPU affinity configuration, "block" to assign adjacent cores to workers
+        (ex: assign 0-1 to worker 0, 2-3 to worker 1), and
+        "alternating" to assign cores to workers in round-robin
+        (ex: assign 0,2 to worker 0, 1,3 to worker 1).
+
     prefetch_capacity : int
         Number of tasks that could be prefetched over available worker capacity.
         When there are a few tasks (<100) or when tasks are long running, this option should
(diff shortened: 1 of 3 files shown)


WardLT:
Any advice for how to test this? I was thinking of setting the `htex_local_alternate` test configuration to use one of the affinity settings.

benclifford:
@wardlt if it is very likely that every system will support affinity, then it makes sense to add it into htex_local_alternate. That won't test if affinity is actually happening, though.

Or, make a local mode test that actually checks (somehow) that affinity is happening.

WardLT:
I placed the test in the `sites` folder, which is where I put the "worker_info" test I made a while back (for reasons I cannot remember). Is that OK?

benclifford:
> I placed the test in the sites folder, which is where I put the "worker_info" test I made a while back (for reasons I cannot remember). Is that OK?

@wardlt yeah that's fine - the structure of test directories is fairly vague and what really matters is that pytest can find the test.

benclifford:
@WardLT you need to merge parsl master into this (ideally using the github user interface) so that it can be tested in CI before being merged into parsl master. Usually there's a button for me to do that myself, but not for this PR (perhaps because it is coming from a different organisation's repo)

WardLT:
Done

PR #1854, opened 2020-09-02
Invoke pytest in tests/ so SKIPPED summaries are displayed at end of run

benclifford:
If pytest is invoked for the higher level parsl/ directory, the skipped summary
is not output.

I think this difference is because conftest.py is located in tests/

This PR increases visibility of what is being skipped.

# Description

Fixes # (issue)

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
Looks reasonable.

PR #1855, opened 2020-09-02
Test case for stage-in/stage-out bug

benclifford:
This PR introduces a test for a bug I've noticed with certain file staging patterns, where a stage-out causes an unwanted stage-in to happen. I will write up an issue. The fix I have for this will appear in a different PR #1856 and should make this test pass and mergeable.


parsl/tests/test_staging/test_elaborate_noop_file.py
```diff
@@ -0,0 +1,82 @@
...
+                storage_access=[NoOpTestingFileStaging(allow_stage_in=False)]
+            )
+        ],
+        run_dir=get_rundir()
+    )
+
+    parsl.load(no_stageout_config)
+
+    # test with no staging
+    touch("test.1", outputs=[File("test.1")]).result()
```
yadudoc, line 50: It is not clear what test.1 vs test.2 does. Both invocations look the same?

parsl/tests/test_staging/test_elaborate_noop_file.py
```diff
@@ -0,0 +1,82 @@
...
+
+    parsl.load(no_stageout_config)
+
+    # TODO create a file not using a task
+
+    f = open("test.3", "a")
+    f.write("test")
+    f.close()
+
+    app_test_in(File("test.3")).result()
```
yadudoc, line 79: Do we need an invocation that needs stage_out to raise an exception to test the alternate path?

parsl/tests/test_staging/test_elaborate_noop_file.py
benclifford: This should be addressed in 8566c4402  - I changed the invocation to test what the comment says.

parsl/tests/test_staging/test_elaborate_noop_file.py
benclifford: These test cases were intended to test that exceptions *don't* happen but I think it would make sense to test that exceptions from providers *do* propagate correctly too.

benclifford: I've added two test cases for that in 2ef6ca05d

yadudoc (dismissed):
Looks good now. Let's merge this.

benclifford:
@yadudoc this had a merge conflict with #1915 which is now resolved - please reapprove

yadudoc (approved):

PR #1856, opened 2020-09-02
Generalise staging inhibiting to work in more cases

benclifford:
When a task has input or output files, then file staging elaboration
may add in new tasks to perform that input or output staging.

Those new tasks should not themselves be further elaborated by
the staging mechanism, otherwise inappropriate staging behaviour
might be added in around the staging tasks themselves.

Prior to this inhibition, this inhibition of staging happens in
two ways, differently for stage-in and stage-out.

If the task is assigned to the 'data_manager' executor,
then no staging-in elaboration is performed on that task.

If a task has a 'staging_inhibit_output=True' keyword argument,
then no staging-out elaboration is performed on that task.

It turns out that check for the 'data_manager' executor is
not a very good test, for a couple of reasons:

  Stage-out tasks can run on arbitrary executors, just like stage-in
  tasks. It just so happens that no staging providers do that:
  HTTP/FTP have arbitrary executor stage-in tasks, but not arbitrary
  stage-out tasks; and globus staging tasks always run in the
  'data_manager' executor. If a stage-out task from some other
  staging provider runs on an arbitrary executor, then stage-in
  elaboration will not be inhibited, and a new stage-in task may
  be generated to stage in the file that is being staged out. This
  is wrong. PR#1855 adds a test to demonstrate this.

  Upcoming work for @join_apps wants to use that executor for
  executing user supplied tasks for which staging might be expected
  to happen.

The core of this PR is a change around like 524 of dflow.py
which changes stage-in elaboration to be controlled by the presence
of a boolean inhibit keyword arg, as for staging-out elaboration.

The rest of the PR renames that keywrd arg from staging_inhibit_output
to a more general and internal-looking _parsl_inhibit_staging, and
adds that annotation onto the globus stage-out code where the previous
mechanism would have inhibited it.

## Type of change

### DIFF
--- a/parsl/data_provider/ftp.py
+++ b/parsl/data_provider/ftp.py
@@ -25,7 +25,7 @@ class FTPSeparateTaskStaging(Staging, RepresentationMixin):
         else:
             file.local_path = file.filename
         stage_in_app = _ftp_stage_in_app(dm, executor=executor)
-        app_fut = stage_in_app(working_dir, outputs=[file], staging_inhibit_output=True, parent_fut=parent_fut)
+        app_fut = stage_in_app(working_dir, outputs=[file], _parsl_staging_inhibit=True, parent_fut=parent_fut)
         return app_fut._outputs[0]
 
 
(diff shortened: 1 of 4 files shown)


yadudoc (approved):
Code changes look good. Ben and I had a discussion about these changes over a call.

PR #1858, opened 2020-09-04
Rename data_manager executor for more general use

benclifford:
An upcoming PR to implement monad-like joins needs to execute tasks locally inside the submitting
process, just like the data_manager local executor is used for when performing some staging tasks
(globus ones).

This PR renames that local executor from data_manager to _parsl_internal to reflect that intended
broader use.

## Type of change

- Breaking change:  any external file staging providers which expect to use `data_manager` as an executor name need to change to use `_parsl_internal`
- Code maintentance/cleanup


yadudoc (approved):
:+1:  Clarifying that this threadpool isn't just for data management opens it up to be used for other things like status polling.

PR #1859, opened 2020-09-05
Fixes #1857 by setting remote func.__name__ manually

### LINKED ISSUE #1857: Remote function name is not resolved correctly | task_func_name: wrapper
**Describe the bug**
The stdout and stderr log filenames of Parsl apps should contain the function name of the corresponding app (Parsl 1.0.0 behavior). In master branch version (ec09fdc), the function name is appended as 'wrapper'. This unexpected behavior is only noticed in bash_apps. It works as expected for python_apps.

**To Reproduce**
1. Setup Parsl latest (ec09fdc) with Python 3.7 on cluster
2. Run a test script with a bash_app named "my_bash_app"
3. Wait until tasks are finished
4. See generated Parsl stdout, stderr log files.
5. Check parsl.log for 'task_func_name': 'my_bash_app'. However, 'task_func_name': 'wrapper' is seen

**Expected behavior**
1. 'task_func_name': 'my_bash_app' in Parsl log
2. "my_bash_app" string contained in stdout and stderr log filenames.

**Environment**
 - OS: RHEL6.1
 - Python version: 3.7 (Anaconda 4.8.2)
 - Parsl version: ec09fdc2099e1795f41210bd1a159633b601cee3

**Distributed Environment**
- PBS Pro in NSCC's ASPIRE1 (HTEX)

ravihansa3000:
# Description

Manually set `func.__name__` to fix Parsl stdout log filename issue when `parsl.AUTO_LOGNAME` is set in bash_apps
Added a test case to validate the stdout format when `parsl.AUTO_LOGNAME` is set

Fixes #1857 

## Type of change

### DIFF
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -129,6 +129,7 @@ class BashApp(AppBase):
         # this is done to avoid passing a function type in the args which parsl.serializer
         # doesn't support
         remote_fn = partial(update_wrapper(remote_side_bash_executor, self.func), self.func)
+        remote_fn.__name__ = self.func.__name__
         self.wrapped_remote_function = wrap_error(remote_fn)
 
     def __call__(self, *args, **kwargs):
--- a/parsl/tests/configs/local_threads.py
+++ b/parsl/tests/configs/local_threads.py
@@ -1,9 +1,13 @@
 from parsl.config import Config
 from parsl.executors.threads import ThreadPoolExecutor
+from parsl.tests.utils import get_rundir
 
 
 def fresh_config():
-    return Config(executors=[ThreadPoolExecutor()])
+    return Config(
+        executors=[ThreadPoolExecutor()],
+        run_dir=get_rundir(),
+    )
 
 
 config = fresh_config()
--- a/parsl/tests/test_bash_apps/test_basic.py
+++ b/parsl/tests/test_bash_apps/test_basic.py
@@ -3,6 +3,8 @@ import os
 import pytest
 import shutil
 import time
+import random
+import re
 
 import parsl
 from parsl import File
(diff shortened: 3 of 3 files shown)


ravihansa3000:
@benclifford This seems to have fixed memoization issues as well. When using `partial`, I don't see a better way to preserve the func name other than manually setting it.

benclifford:
this is something @yadudoc should look at

benclifford:
Maybe this would be better implemented by the app name being stored explicitly by the app decorator independently from the __name__ of any particular callable.

The code at the moment assumes that the many wrapped callables we have are always named after the original app even as the code we want to execute changes quite a lot from that actual raw app code. We could stop making that assumption, and treat the app name as something completely different than the __name__ of intermediate callables.

That might involve:

1. dfk.submit getting a new app_name parameter which is stores in the task record instead of asking for the __name__ of the callable
2. the bash and python app decorators being modified to pass the app name when they call dfk.submit

I think this is not a lot of work to do and would be cleaner, I think - @yadudoc ?

yadudoc:
@ravihansa3000, Thanks for posting these changes. I think it is a reasonable expectation that `func.__name__` will pass through unmangled through our app wrapping. So, I think this is the nicer approach. 

@benclifford, I worry that we'll be adding a duplicate mechanism for `func.__name__` which is only visible to parsl. While this would make logging app names more reliable, I think we might be better off not adding new code for something that should just work. If any wrapper mangles the name, it's a bug. [functools.update_wrapper](https://docs.python.org/2/library/functools.html#functools.update_wrapper) exists for this reason.

yadudoc (approved):
I think these changes look good.

PR #1860, opened 2020-09-08
Implement @join_app

### LINKED ISSUE #1796: @join_app decorator for launching apps from inside other apps
parsl 1.0.0 supports launching apps from within other (ThreadPool) apps, but lacks a good way to
depend on the results of those apps. This issue talks about that and proposes a solution.

For the avoidance of doubt, this issue does not address launching apps inside a remote worker
in some kind of remote mini-DFK to avoid the main DFK being aware of those apps.

## Motivating example

There are "sensors" which must be processed and assembled into "patches", and all patches assembled into a "mosaic".
The list of which sensors must be assembled into "patches" is not know ahead of time and is expensive to compute.
There are many patches which must all be processed in this way.

For each patch:
1. Generate list of sensors for this patch, one parsl app.
2. Process sensors for this patch - one parsl app per sensor
3. When all sensors are processed, assemble all the processed data into a patch.
4. Some post processing happens on the patch
Once all patches are done:
5. Make mosaic of all patches

Here are some implementation attempts and the weaknesses I see in them.

## Attempt 1:

```
post_process_futures = []
for patch in patch_list:
  sensor_list_future = generate_sensor_list_app(patch)
  sensor_futures = []
  for sensor in sensor_list_future.result():
    sensor_futures.append(process_sensor(sensor))
  assembled_future = assemble_sensors(*sensor_futures)
  post_process_futures.append(post_process(assembled_future))
mosaic_future = make_mosaic(*post_process_futures)
mosaic_future.result()
```

Weaknesses:

*  Only one patch is processed at a time. The outer for loop blocks on job completion repeatedly
before all of the apps are submitted, forcing unnecessarily serialised execution ordering: each
loop will block waiting for generate_sensor_list_app to complete.

## Attempt 2

Put the whole loop body into an app.

```
patch_futures = []
for patch in patch_list:
  patch_futures = process_patch(patch)
mosaic_future = make_mosaic(*patch_futures)
mosaic_future.result()

@python_app(executors=['local-thread-pool']
def process_patch():
  sensor_list_future = generate_sensor_list_app(patch)
  sensor_futures = []
  for sensor in sensor_list_future.result():
    sensor_futures.append(process_sensor(sensor))
  assembled_future = assemble_sensors(*sensor_futures)
  post_process_future = post_process(assembled_future)
  post_process_future.result()
```

In this attempt, all of the for-loop bodies are launched without blocking (as process_patch invocations),
and potentially can run concurrently.

Weaknesses:

Each process_patch app must stay executing (occupying a thread pool worker) for the entire duration of
the tasks that it has launched, because it blocks waiting for completion of the post_process step, by
calling post_process_future.result().
It does this so that the process_patch app completes after the post_process step. If this call to
.result() was not there, the process_patch app invocation would complete too early, and make_mosaic
could run before post_process is complete.

Because of this, there can be serialization and deadlock issues: there needs to be one local-thread-pool
worker available for every patch to be processed simultanously, that will for the most part be sitting
idle waiting for final results. If there are fewer, then patch processing will be serialised due to
lack of workers, although to a lesser extent than in attempt 1.

But worse, if any of the launched apps also used local-thread-pool, then the workflow can deadlock if there
insufficient workers:
launched process_patch apps will be waiting for other apps to complete, but those apps cannot complete
because process_patch apps are occupying all of the local-thread-pool workers.

This leads to a principle: apps should not block on other apps

## Attempt 3:
No blocking inside apps

```
sensor_list_futures = []
for patch in patch_list:
  future = generate_sensor_list_app(patch)
  future.patch = patch
  sensor_list_futures.append(future)

post_process_futures=[]
for future in concurrent.futures.as_completed(sensor_list_futures):
  patch = future.patch
  sensor_futures = []
  for sensor in future.result()
    sensor_futures.append(process_sensor(sensor)
  assembled_future = assemble_sensors(*sensor_futures)
  post_process_futures.append(post_process(assembled_future))

mosaic_future = make_mosaic(*post_process_futures)
mosaic_future.result()
```

weaknesses:

There is some loss of the sense of data flow expressed in syntax, obscured by the implementation of a rudimentary
task scheduler that only knows about sensor list futures.

This is *still* blocking on the second for loop needing to complete before later apps can be launched,
and that for loop only completes when all sensor lists have been generated - this is earlier than
in previous examples, but there is a still blocking there.

This impedes compositionality: this code could not be placed into a function and (for example) run
inside another for loop that loops over multiple datasets: the function would block each time waiting
for sensor list generation, rather than processing each data set's sensor list generations
concurrently.

Anything that blocks the execution thread on future completion (for example, `.result()`
or `.as_completed()`) is the enemy.

## Proposed and prototyped join syntax:

This is an attempt to move some of the cases where blocking and ad-hoc task scheduling happens in
the workflow into parsl dependency handling. The only blocking should happen at the very end of the
workflow, so that the main process does not end until work is completed. Other than that, nothing else
should block on app completion.

```
post_process_futures=[]
for patch in patch_list:
  sensor_list_future = generate_sensor_list_app(patch)
  sensors_future = process_sensors(sensor_list_future)
  assembled_future = assemble_sensors(sensors_future)
  post_process_futures.append(post_process(assembled_future))

mosaic_future = make_mosaic(*post_process_futures)
mosaic_future.result()

@join_app
def process_sensors(sensor_list):
  sensor_futures = []
  for sensor in sensor_list:
    sensor_futures.append(process_sensor(sensor))
  return combine_app(*sensor_futures)

@python_app
def combine_app(*args):
  pass # do nothing, but only after all args are complete
```

The new prototype language feature here is a new kind of app called a @join_app, which runs python code locally inside the workflow process and then performs special processing on the result, described below.

There is also a helper app (which already works in parsl 1.0.0) which, given a list of input futures,
completes when all of those futures complete, without any further processing. This constructs a
barrier future, depending on an arbitrary list of other futures.

Without @join_app, in parsl 1.0.0, the process_sensors app could be implemented as  a @python_app running on a thread local executor. It would invoke the
individual process sensors apps, and then immediately return. The value returned would be a Future,
the AppFuture from invoking combine_app.
process_sensors returns a Future containing a Future.

assemble_sensors then depends on receiving a value from process_sensors. This would happen almost
immediately, with the value being the Future returned by combine_app. But that's no good - we want
to block assemble_sensors from executing until combine_app has finished execution. Instead, we only
block until we get the AppFuture for combine_app.

We could put in a `.result()` somewhere to extract that value - but remember that calling `.result()`
is bad, for reasons outlined above.

What could help is a way to say that if an app returns a Future, the AppFuture doesn't complete
until that returned Future has completed. In the case of process_sensors above, the process_sensors
AppFuture should not complete until the returned Future completes, in this case the AppFuture
returned by combine_app. That is what the special post-processing behaviour of @join_app does.

This allows more naunced dependencies to be expressed which can help with:
* increased concurrency - helping with strong scaling
* more focused error propagation - allowing more of an ultimately failing workflow to complete
* more useful monitoring information

This has been prototyped in LSST work in two forms, with the above syntax prototyped in the
`lsst-dm-202005` branch.

## Terminology

The term `join` comes from use of monads in functional programming, especially Haskell.

## Merge plan

- PR #1856 - generalise file staging inhibit, allowing reuse of data_manager executor for other parsl internal tasks
- rename data_manager executor to parsl_internal
- main join_app PR

benclifford:
# Description

    Add @join_apps
    
    @join_apps allow a python app to launch more apps and then complete
    only after the launched apps are completed, without occupying a
    worker waiting.
    
    These are described more in docs/userguide/joins.rst
    which is added by this patch.
    
    This PR adds a new state after the body of a python app completes,
    but before the join future is complete, which is only entered
    if an app is a join app.
    
    This PR modifies the recursive fibonacci test case (which previously
    blocked workers and so would deadlock easily) to use join apps.
    
    This PR adds a field to the monitoring database to represent the
    app to be joined on, which is similar to but not the same as a regular
    dependency: a dependency indicates what must be completed for this
    app to start, but joins field indicates what must be completed for
    this app to complete.

Fixes #1796 

## Type of change


docs/userguide/joins.rst
```diff
@@ -0,0 +1,200 @@
...
+Motivating example
+------------------
+
+Here is a motivating example which shows ways in which launching apps from inside a
+`@python_app` is insufficient.
+
+Consider a workflow where there are "sensors" which must be processed and assembled
+into "patches", and then all patches assembled into a "mosaic".
+
+The list of which sensors must be assembled into which "patches" is not know ahead
```
danielskatz, line 20:
```suggestion
The list of which sensors must be assembled into which "patches" is not known ahead
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,200 @@
...
+them into the task graph. They can be specified using the `@join_app` decorator.
+
+Although apps can be launched from within a `@python_app` in any `ThreadPoolExecutor`,
+the `Future` objects for those launched apps cannot easily be used outside of the
+launching `@python_app`.
+
+Motivating example
+------------------
+
+Here is a motivating example which shows ways in which launching apps from inside a
```
danielskatz, line 14:
```suggestion
Here is a motivating example that shows ways in which launching apps from inside a
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,200 @@
...
+`@python_app` is insufficient.
+
+Consider a workflow where there are "sensors" which must be processed and assembled
+into "patches", and then all patches assembled into a "mosaic".
+
+The list of which sensors must be assembled into which "patches" is not know ahead
+of time and is expensive to compute.
+
+There are many patches which must all be processed in this way.
```
danielskatz, line 23:
```suggestion
There are many patches that must all be processed in this way.
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,200 @@
...
+There is some loss of the sense of data flow expressed in syntax, obscured by
+the implementation of a rudimentary in-workflow task scheduler that only knows about sensor list futures.
+
+This is *still* blocking on the second for loop needing to complete before later apps can be launched,
+and that for loop only completes when all sensor lists have been generated - this is earlier than
+in previous examples, but there is still blocking there.
+
+This impedes compositionality: this code could not be placed into a function and (for example) run
+inside another for loop that loops over multiple datasets: the function would block each time waiting
+for sensor list generation, rather than processing each data set's sensor list generations
```
danielskatz, line 144:
```suggestion
for sensor list generation, rather than processing each dataset's sensor list generations
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,200 @@
...
+A join_app looks quite like a python app, but should return a future, rather than a value.
+After the python code has run, the app invocation will not complete until that future has
+completed, and the return value of the `@join_app` will be the return value (or exception)
+from the returned future.
+
+This example uses a helper app called `combine` which, given a list of input futures,
+completes when all of those futures complete, without any further processing. This constructs a
+barrier future, depending on an arbitrary list of other futures.
+
+This allows more naunced dependencies to be expressed which can help with:
```
danielskatz, line 191:
```suggestion
This allows more naunced dependencies to be expressed that can help with:
```

benclifford:
I have seen a bunch of python interpreter crashes (not exceptions) in testing both in CI and on my laptop around joinapp. An example is:

> parsl/tests/test_python_apps/test_fibonacci_recursive.py
> Makefile:64: recipe for target 'htex_local_alternate_test' failed
> make: *** [htex_local_alternate_test] Aborted (core dumped)

This needs proper investigation before I feel comfortable with this PR being merged

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+=========
+
+Join apps allows an app to define a sub-workflow: the app can launch other apps
+and incorporate them into the main task graph. They can be specified using the
+`join_app` decorator.
+
+Usage
+-----
+
+A `join_app` looks quite like a python app, but should return a ``Future``,
```
danielskatz, line 11: do you mean `python_app` instead of python_app?  ie a parsl decorated python app or just a regular python app?

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+rather than a value. After the python code has run, the app invocation will not
+complete until that future has completed, and the return value of the `join_app`
+will be the return value (or exception) from the returned future.
```
danielskatz, line 14: say how this is different than a parsl-decorated python_app?

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+    if x > 0:
+      return type_one(x)
+    else:
+      return type_two(x)
+
+  # process(10).result() == 20
+  # process(-3).result() = 6
+
+The join app ``process`` looks very much like a normal Python function
+which launches apps would. When invoked with literal numbers as above,
```
danielskatz, line 43:
```suggestion
that launches apps. When invoked with literal numbers as above,
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+    time.sleep(120)
+    return 10
+
+  # process(later()) = 20
+
+If ``process`` was a regular Python function, it would be able to inspect
+the result of ``later`` - that result wouldn't exist for another 120 seconds.
+
+A join app would defer its execution until ``later`` was complete, just like
+any other Parsl app that is passed a ``Future``.
```
danielskatz, line 65: again, at this point it's not clear why something different than a python_app is needed

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+1. Generate list of sensors for this patch, in a single parsl app.
+2. Process sensors for this patch - one parsl app per sensor
+3. When all sensors are processed, assemble all the processed data into a patch.
+4. When assembly is complete, do some post processing on the patch
+
+Once all patches are done:
+
+5. Make mosaic of all patches
+
+Here are some implementation attempts and the weaknesses I see in them.
```
danielskatz, line 96:
```suggestion
Here are some possible implementations and their shortcomings.
```

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+idle waiting for final results. If there are fewer workers, then patch processing will be
+serialised due to lack of workers, although to a lesser extent than in attempt 1.
+
+But worse, if any of the launched apps also used the same `ThreadPoolExecutor`, then
+the workflow can deadlock:
+launched process_patch apps will be waiting for other apps to complete, but those apps cannot start
+because process_patch apps are occupying all of the `ThreadPoolExecutor` workers.
+
+This leads to a principle: apps should not block on other apps; instead any blocking of execution
+should happen inside parsl's dependency mechanism.
```
danielskatz, line 169: Is this a more general principle than just in these examples? Something that we should use to rethink some of other tutorials examples?

docs/userguide/joins.rst
```diff
@@ -0,0 +1,260 @@
...
+
+* increased concurrency - helping with strong scaling
+* more focused error propagation - allowing more of an ultimately failing workflow to complete
+* more useful monitoring information
+
+Terminology
+-----------
+
+The term "join" comes from use of monads in functional programming, especially Haskell.
+
```
danielskatz, line 260: what would happen if we replaced all python_apps by join_apps?  (Or simply changed the implementation of python_app to the implementation of join_app?

docs/userguide/joins.rst
benclifford: yes. (it was in the parslfest talk on structuring workflows too, i think)

danielskatz:
Just to repeat this here, rather than as an embedded comment:

What would happen if we replaced all python_apps by join_apps? (Or simply changed the implementation of python_app to the implementation of join_app?)

benclifford:
> What would happen if we replaced all python_apps by join_apps? (Or simply changed the implementation of python_app to the implementation of join_app?)

The implementation is the same base, but join apps have an additional waiting stage on the end after the python bit of the app completes: the python part of a python app can say `return 3` to return the value 3. the python part of a join app returns a Future, which can complete later on, and that might contain the value 3.

So to return a `3` directly from a join app, you need some adapter that turns `3` into a Future that (eventually/right away) has the value 3 stored inside it.

@python_app
def immediate(x):
  # on the calling side, this app will give a Future which will contain the supplied value straight away without needing to compute anything
  return x

@join_app
def foo():
  return immediate(3)  # this wraps the value 3 in a Future, so that the join app unwrapping of the Future gets the 3 back.

is the same as

@python_app
def foo():
  return 3

docs/userguide/joins.rst
benclifford: and there's even a userguide chapter on structuring workflows that this could form part of.

danielskatz:
I still find this unclear
> Python apps encapsulate pure Python code, while Bash apps wrap calls to external applications and scripts, and Join apps allow composition of other apps to form sub-workflows.

since the reason for wrapping apps (with python_app) to allow composition.

What are "other apps"? python code? calls to external applications?

Why do we need something different than the @python-app and @bash-app?

Maybe "sub-workflows" needs some brief explanation?

danielskatz:
maybe something about dynamic apps, or dynamic workflow could be mentioned?

> join apps are useful when a workflow needs to launch some apps, but it doesn't know what those apps are until some earlier apps are completed.

seems important here

danielskatz:
I like the new simpler examples.

danielskatz:
What would happen in the example if all the apps were join apps instead of some being python apps?

benclifford:
> I understand the technical answer, I'm more asking in theory - is there a way we can have fewer types of apps?

there can be fewer types of apps if we smash down that lump and bring it up somewhere else... (that's the immediate() function in my previous answer) - the unwrapping of two futures into one future is a lump that isn't going to go away though. (mumble citation of some papers about monads from the functional programming world).

separately we *could* get rid of bash apps, by turning the bash specific functionality into a helper function that you run inside a python app - because bash app (internally) is a python app with some helper code added on - but that helper function behaviour has to be there.

I think in both cases, it's a usability question: is presenting this stuff as different decorators the right way?

A few years ago, there was a single app decorator, parameterised with a type-of-app parameter, and then we moved to bash_app and python_app because it felt much cleaner.

docs/userguide/joins.rst
benclifford: whats a "regular python app"  that isn't parsl decorated?

docs/userguide/joins.rst
benclifford: this should be explained now later on in the examples

docs/userguide/joins.rst
danielskatz: I meant that maybe this should be "`python_app`" instead of "python app"

yadudoc (approved):
@benclifford I've read through the code a couple of times, and it looks good. I've not gone over the docs as closely and will do that separately.

PR #1868, opened 2020-09-23
Set WQ category to func name, not qualname

### LINKED ISSUE #1867: WQ category is always the same
**Describe the bug**

The WQ category is always being set according to the function's `__qualname__` [here](https://github.com/Parsl/parsl/blob/master/parsl/executors/workqueue/executor.py#L409), and the function is being wrapped [here](https://github.com/Parsl/parsl/blob/master/parsl/app/errors.py#L154). For some reason, the qualname is not being updated correctly, so the category is always `wrap_error.<locals>.wrapper`:

```
1600806505778540 2352716 CATEGORY wrap_error.<locals>.wrapper MAX {"cores":[18,"cores"],"memory":[2464,"MB"],"disk":[2,"MB"]}
1600806505778556 2352716 CATEGORY wrap_error.<locals>.wrapper MIN {"cores":[3,"cores"],"memory":[100,"MB"],"disk":[2,"MB"]}
1600806505778565 2352716 CATEGORY wrap_error.<locals>.wrapper FIRST MAX_THROUGHPUT {"cores":[3,"cores"],"memory":[100,"MB"],"disk":[2,"MB"]}
```
According to the [docs](https://docs.python.org/3/library/functools.html#functools.wraps), `functools.wraps` should update the qualname as well as the name. But this patch fixes the issue, does anyone object to just setting according to name?

```diff
(gf) annawoodard@daenerys:/cephfs/users/annawoodard/parsl/parsl$ git diff executors/workqueue/
diff --git a/parsl/executors/workqueue/executor.py b/parsl/executors/workqueue/executor.py
index 52aa5c2..accae4e 100644
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -407,6 +407,7 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
         # Create message to put into the message queue
         logger.debug("Placing task {} on message queue".format(task_id))
-        category = func.__qualname__ if self.autocategory else 'parsl-default'
+        category = func.__name__ if self.autocategory else 'parsl-default'
         self.task_queue.put_nowait(ParslTaskToWq(task_id,
                                                  category,
                                                  cores,
```

annawoodard:
The function qualname is getting lost when the function is
wrapped for error propagation; `__name__` is updated explicitly and safe
to use here.

Fixes #1867. 

## Type of change

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -407,6 +407,7 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
         # Create message to put into the message queue
         logger.debug("Placing task {} on message queue".format(task_id))
         category = func.__qualname__ if self.autocategory else 'parsl-default'
+        category = func.__name__ if self.autocategory else 'parsl-default'
         self.task_queue.put_nowait(ParslTaskToWq(task_id,
                                                  category,
                                                  cores,
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -406,7 +406,6 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
 
         # Create message to put into the message queue
         logger.debug("Placing task {} on message queue".format(task_id))
-        category = func.__qualname__ if self.autocategory else 'parsl-default'
         category = func.__name__ if self.autocategory else 'parsl-default'
         self.task_queue.put_nowait(ParslTaskToWq(task_id,
                                                  category,


parsl/executors/workqueue/executor.py
```diff
@@ -407,6 +407,7 @@ def submit(self, func, resource_specification, *args, **kwargs):
         # Create message to put into the message queue
         logger.debug("Placing task {} on message queue".format(task_id))
         category = func.__qualname__ if self.autocategory else 'parsl-default'
+        category = func.__name__ if self.autocategory else 'parsl-default'
```
ZhuozhaoLi, line 410: maybe remove the line before this? they are duplicate

annawoodard: @ZhuozhaoLi good point, fixed

ZhuozhaoLi (approved):

PR #1876, opened 2020-09-29
Logging more manager status (e.g., when a manager becomes inactive) to monitoring

ZhuozhaoLi:
# Description

Prior to this PR, master monitoring only logs when a manager is registered. This PR logs more manager status to the monitoring table. Specifically, it now logs the uid, block_id, active and inactive timestamp, and last heartbeat time for each manager.

## Type of change

- Code maintentance/cleanup


parsl/executors/high_throughput/interchange.py
```diff
@@ -282,10 +285,27 @@ def migrate_tasks_to_internal(self, kill_event):
...
+                                         datetime.datetime.now(),
+                                         self._ready_manager_queue[manager]))
+
     def _command_server(self, kill_event):
         """ Command server to run async command to the interchange
         """
         logger.debug("[COMMAND] Command Server Starting")
+
+        # Need to create a new ZMQ socket for command server thread
+        monitoring_enabled = False
```
benclifford, line 301: This block means that each manager will have a two monitoring TCP connections to the monitoring hub; and the second connection will only be used in the case of HOLD_WORKER commands.

I feel uncomfortable with that extra load on the hub side of things: this would be thousands of extra connections in the case of LSST DRP, for example.

Maybe there's a way to send that shutdown message from the main thread, or a similar message - for example, combined with the code that reports `bad_managers` to monitoring?

benclifford (changes requested):

parsl/executors/high_throughput/interchange.py
benclifford: After a meeting with @ZhuozhaoLi, I realise this is only one connection per interchange, not per note, so I withdraw this object.

benclifford (approved):

PR #1881, opened 2020-10-06
Allow users to specify only two types of resource for WQEX when autolabel=True

ZhuozhaoLi:
# Description

Prior to this PR, we require users to specify all three types of resources: cores, memory, and disk. But actually if `autolabel=True`, users can specify only two of them and WQ can automatically fill the remaining resources. This PR allows users to specify only two types of resource for WQEX when `autolabel=True`.

## Type of change


parsl/executors/workqueue/executor.py
```diff
@@ -329,19 +329,21 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
                              "three types of resources: cores, memory, and disk")
-                raise ExecutorError(self, "Task resource specification requires "
-                                          "three resources to be specified simultaneously: cores, memory, and disk, "
-                                          "and only takes these three resource types.")
+                raise ExecutorError(self, "Task resource specification only accepts "
+                                          "three types of resources: cores, memory, and disk")
+
+            if len(keys) != 3:
+                if not self.autolabel:
+                    logger.error("Rnning with `autolabel=False`. In this mode, "
```
yadudoc, line 340: s/Rnning/Running

parsl/executors/workqueue/executor.py
```diff
@@ -329,19 +329,21 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
             if not all(k.lower() in acceptable_resource_types for k in keys):
                 logger.error("Task resource specification only accepts "
                              "three types of resources: cores, memory, and disk")
-                raise ExecutorError(self, "Task resource specification requires "
-                                          "three resources to be specified simultaneously: cores, memory, and disk, "
-                                          "and only takes these three resource types.")
+                raise ExecutorError(self, "Task resource specification only accepts "
+                                          "three types of resources: cores, memory, and disk")
+
+            if len(keys) != 3:
```
yadudoc, line 338: Might make sense to convert this to set comparisons to make this a bit more clearer.

yadudoc (approved):
Looks much cleaner.

PR #1883, opened 2020-10-30
WQEX updates

trshaffer:
# Description

Minor updates to the WQ executor to address some usability issues.
- The situations in which `autocategory=False` would be beneficial are rare (very large number of apps, each called rarely, very different resource utilization). It should be turned on by default in almost all cases, so this makes that the default.
- Use new `python_package_analyze` to scan `exec_parsl_function.py`. The remote runner has dependencies of its own, so it's necessary to scan both the app body and the runner.
- Add option to specify extra packages to pack up. Useful if e.g. the Python app `exec()`s a Perl program provided by Conda, which can't be picked up by analyzing imports.

## Type of change

cc @ZhuozhaoLi


trshaffer:
Hmmm.... looks like this isn't safe to merge yet.

At some point the CCTools Ubuntu build broke (Ben's looking into it), so the tarball used for testing will 404. The version bump in 64d811b will break the WQ tests only *after* this is merged. I'll update when the CCTools build is sorted out

trshaffer:
Ugh....looks like the new CCTools build doesn't match the Travis environment, this is turning into ugly dependency management. Could CCTools be obtained via Conda or by building from source?

ZhuozhaoLi:
> Ugh....looks like the new CCTools build doesn't match the Travis environment, this is turning into ugly dependency management. Could CCTools be obtained via Conda or by building from source?

Looks more like a path issue to me?

benclifford:
@trshaffer  > Could CCTools be obtained via Conda or by building from source?

There aren't any particular requirements on how pre-reqs like cctools should be installed in CI. Some soft goals are: not complicated; fast; and close to how an end-user is told to install this stuff.

If you can do it in a better way than it is being done now, do that.

trshaffer:
We discussed it at the meeting over at ND, and decided that continuing with a tarball build is the best way to go. We'll just need to make sure that we have a Travis-compatible build going forward.

So this is OK to review/merge

ZhuozhaoLi (approved):

PR #1887, opened 2020-11-09
Add basic test of monitoring db behaviour

benclifford:
# Description
This tests that the monitoring database as seen through SQL looks OK after running one app.

## Type of change

- Code maintentance/cleanup


ZhuozhaoLi (approved):
A very useful test to ensure that monitoring works properly

PR #1895, opened 2020-11-19
Add monitoring db test that asserts memo-related behaviour

benclifford:
## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/tests/test_monitoring/test_memoization_representation.py
+++ b/parsl/tests/test_monitoring/test_memoization_representation.py
@@ -1,0 +1,80 @@
+
+import logging
+import os
+import parsl
+import pytest
+import sqlalchemy
+
+logger = logging.getLogger(__name__)
+
+from parsl.tests.configs.htex_local_alternate import fresh_config
+
+
+@parsl.python_app(cache=True)
+def this_app(x):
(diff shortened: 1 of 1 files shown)


parsl/tests/test_monitoring/test_memoization_representation.py
```diff
@@ -0,0 +1,81 @@
...
+    f2 = this_app(17)
+    assert f2.result() == 18
+
+    logger.info("invoking and waiting for result (3/4)")
+    f3 = this_app(4)
+    assert f3.result() == 5
+
+    logger.info("invoking and waiting for result (4/4)")
+    f4 = this_app(4)
+    assert f4.result() == 5
```
ZhuozhaoLi, line 42: I don't mind adding extra invocation. But I wonder about the purpose to invoke one more time here. If it is necessary, maybe another invocation with `17` would be better?

benclifford: It is to give some more distinction in the database between the this_app(4) tasks and the this_app(17) tasks, which made debugging things a bit easier for me.

yadudoc (approved):
The test looks good.

PR #1899, opened 2020-11-23
Remove task future from executor once completed

WardLT:
# Description

Changes HTEX to remove references to a task after it completes. Could otherwise lead to memory issue if task outputs are large.

Fixes # (issue) [Would you like me to open one?]

## Type of change


WardLT:
For reference, I do observe this to be a problem when I have task sizes on the order of a few 100MB.

yadudoc (approved):
@WardLT Thanks for debugging and pushing this fix. This dict is only used to track the future <-> task_id mapping and should be safe to pop from.

PR #1902, opened 2020-11-25
Take _get_messages_in_batch parameters from self implicitly

benclifford:
self._get_messages_in_batch was always called with batching parameters
coming from self, so invocations look cleaner if that method looks up
the batching parameters directly.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Looks good to me.

PR #1903, opened 2020-12-07
Log choice of executor for a task as soon as it is chosen

benclifford:
Previously this was not logged until after the task had been
successfully submitted.

## Type of change


yadudoc (commented):
This is soon followed by a log message of the form "Task <TID> launched on <executor>", so I'm not sure what this adds?

benclifford:
it's not followed by that message when the task isn't successfully launched - you end up with a stack trace for some executor without which executor configuration tried to launch it

yadudoc:
Okay, that makes sense. Perhaps this should be a debug log line.

benclifford:
@yadudoc i've change the log level for the new message

yadudoc (approved):

PR #1909, opened 2021-01-05
Make task record garbage collection optional.

benclifford:
# Description

LSST has a usecase where we want to examine all tasks that have been attempted,
and where we don't encounter the scaling problems that garbage collection was
intended to solve.

The default behaviour remains as before, but set garbage_collect = False when invoking
Config() to disable.

## Type of change


yadudoc:
I am not sure I follow. If you want to track tasks and not lose them, why don't you track the futures in the Parsl script?

benclifford:
>  If you want to track tasks and not lose them, why don't you track the futures in the Parsl script?

That involves making a global list that needs to be updated at every point that an app is invoked, which is a bunch of boilerplate when there are a lot of different apps being invoked.

Parsl is capable of making that global list for the user, as a parsl feature, so that the user doesn't have to maintain that list manually.

This PR adds that feature.

yadudoc:
Alright, I don't think adds too much as long as the default behavior is to garbage collect.

PR #1910, opened 2021-01-05
Do not discard OSErrors at MonitoringRouter construction

benclifford:
# Description

If monitoring fails to initialize, it should cause
an exception in the top level workflow rather than the
workflow running without monitoring.


ZhuozhaoLi:
After removing the try-catch, this raises exception in the `router_starter` process, but it neither logs anywhere about the exception nor stops the monitoring?

yadudoc (commented):
I suspect Zhuozhao's comment about the exception not propagating back is true here. Have you tested this? Since the `MonitoringRouter` is started on a separate process, there's an `exception_q` param that could be used to communicate the exception to the parent process. We probably need to use that here.

benclifford:
Yeah, I agree that this won't propagate to the main workflow.

It will propagate up to router_starter, though, which is the right place to deal with construction exceptions: that code already deals with `exception_q` handling for exceptions from `MonitoringRouter.start()`.

I've done some other work to deal with MonitoringRouter's constructor raising exceptions, so I will put that into this PR too, and then at least the exceptions will go into the `exception_q`.

ZhuozhaoLi:
Yea, the exception should at least be propagated back to `exception_q`.

benclifford:
I've modified this to send MonitoringRouter construction/initialization errors along comm_q. There is already a synchronous wait on that channel during DFK.__init__ to get port numbers, so it can also be used to synchronously forward the failure to create port numbers.

I've tested this by adding an explicit `raise` at line 486 of monitoring.py, and I see that causes the monitoring tests to fail explicitly at DFK initialization.

yadudoc (approved):
These last changes should bubble up errors during the MonitoringRouter start to the main context. This looks right, let's merge this.

ZhuozhaoLi (approved):
Looks good to me

PR #1915, opened 2021-01-05
Use parsl inbuilt runinfo uniquification, not  runinfo_XXXXXX for pytest

benclifford:
## Description

There is no reason at the moment that the test suite should be doing uniquification in
a way that parsl cannot do automatically:

1. There are comments about needing this when running multiple tests in parallel. The
test suite at the moment isn't expected to run in parallel - at the very least:

1.1 there is no separation of working files and global variables between tests

1.2 running pytest with -nX fails immediately because the --config parameter does not
work

2. There were also problems a year or two ago with a style of writing workflows which
initialised a DFK as soon as a module was loaded. That meant a DFK would be created on
workers when deserialisation happened, and that caused a race condition with multiple
workers trying to create the same rundir/NNN/ when they shouldn't have created a worker
at all. The solution to this is "don't do that", as in, don't load a DFK by default in
a module that will be loaded on a worker. The test suite as it is now does not have
that bad behaviour.

I've had this patch in my LSST dev stack since April 2020 and haven't observed any
problems.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This is great :) We shouldn't have to run the tests in parallel, so that use case is moot.

PR #1916, opened 2021-01-05
CI-test that exec_done implies tries have a running time in monitoring

benclifford:
Running time is set on a try row through a different mechanism than the other columns
in a try row, because it comes from a wrapper around the executing task rather than
from the submit site.

The code that sets this has been very fragile in the past, so it warrants an explicit
test.

## Type of change

- Testing


ZhuozhaoLi (dismissed):

benclifford:
@ZhuozhaoLi this had a merge conflict with #1877 which I have resolved. Please re-approve.

ZhuozhaoLi (approved):

PR #1917, opened 2021-01-05
Fix monitoring "db locked" errors occuring at scale

benclifford:
This PR fixes and tests against a serious database locking problem encountered in LSST DESC.

If a parsl monitoring database update is attempted when someone else has a read lock on one of the tables involved, then the parsl code would previously fail, and all monitoring database recording would stop at that point.

This happens when making expensive queries against a monitoring database that is also receiving fresh monitoring data from a running workflow.

This PR makes the code retry updates repeatedly (at one second intervals). There is no maximum number of tries/timeout: database write attempts will stop when the workflow is explicitly killed.

This PR assumes that OperationalErrors will eventually go away if retried enough times. This is reasonable when those operational errors are transient lock conflicts from queries. There may be other OperationalErrors where this is not true, although I have no encountered any.

There's a further problem that this PR does not address - that similar errors related to locking the entire database can happen at the point that parsl monitoring asserts that the monitoring tables should exist. This will manifest if the database is locked at the point that monitoring starts up at the beginning of a workflow run.

## Type of change


yadudoc (approved):
I think this solution looks good. Maybe @ZhuozhaoLi could take a look as well.

parsl/monitoring/db_manager.py
```diff
@@ -539,7 +544,20 @@ def _migrate_logs_to_internal(self, logs_queue: queue.Queue, queue_tag: str, kil
...
     def _update(self, table: str, columns: List[str], messages: List[Dict[str, Any]]) -> None:
         try:
-            self.db.update(table=table, columns=columns, messages=messages)
+            done = False
+            while not done:
+                try:
+                    self.db.update(table=table, columns=columns, messages=messages)
+                    done = True
+                except sa.exc.OperationalError as e:
+                    # This code assumes that an OperationalError is something that will go away eventually
```
ZhuozhaoLi, line 553: This assumption may be really bad? Probably set a maximum retry like 5 times?

benclifford: Maximum retry counts don't solve the problem this PR is intended to address: if I have a big query running for 10 minutes, I want parsl to buffer the data rather than discard it.

Perhaps it is fine to keep trying through the duration of the workflow run, perhaps with some backoff to once a minute or so, with better handling of shutdown - the code at the moment won't shut down if it gets stuck in an operationalerror loop.

PR #1918, opened 2021-01-05
Make checkpoint_files = get_all_checkpoints() by default

benclifford:
This is the sane default for this parameter in almost all
checkpointing situations that I've encountered.

The previous behaviour of loading no checkpoint files, but
checkpointing new activity can be obtained by writing:

checkpoint_files = []

This will be a breaking change for anyone expecting the previous default behaviour.

## Type of change


yadudoc:
The changes look right, but I want to give others a chance to weigh in on whether it makes sense to have this be the default behavior. I worry that use-cases that have apps that depend on libraries in development will get silently skipped.

yadudoc (dismissed):
The changes look right, and I'm approving this PR. After a brief discussion over on slack, it doesn't look like there are any complaints about the change to the default behavior. 

Since @annawoodard and @ravihansa3000 have more experience with checkpointing in practice, I'm hoping they'd chime in soon, if they have any reservations against this change.

@benclifford it looks like there's a merge conflict that needs fixing.

parsl/dataflow/dflow.py
```diff
@@ -152,7 +152,13 @@ def __init__(self, config=Config()):
             self.monitoring.send(MessageType.WORKFLOW_INFO,
                                  workflow_info)
 
-        checkpoints = self.load_checkpoints(config.checkpoint_files)
+        if config.checkpoint_files is not None:
+            checkpoints = self.load_checkpoints(config.checkpoint_files)
+        elif config.checkpoint_files is None and config.checkpoint_mode is not None:
+            checkpoints = self.load_checkpoints(get_all_checkpoints(self.run_dir))
+        else:
+            checkpoints = {}
```
ravihansa3000, line 160: shouldn't this be `checkpoints = []` ?

benclifford: `load_checkpoints` returns a dict (of arguments->result mappings) - what this is defaulting to empty is that dict, not defaulting the list of what to load.

yadudoc: This is correct, `load_checkpoints` returns the memoized task table.

yadudoc (approved):

PR #1919, opened 2021-01-05
Generally refresh checkpoint docstrings

benclifford:
## Type of change

- Documentation update


danielskatz (commented):
It's a little strange to me that this changes docs but not code.
Has the code already been changed and the docs were not?

benclifford:
> Has the code already been changed and the docs were not?

Yes, pretty much: one docstring is just plain wrong and probably a cut and paste error from the initial implementation. Another deals with the change of parsl configuration from JSON to a python `Config` object which happened years ago.

danielskatz (approved):

PR #1921, opened 2021-01-05
Weaken too-strong assertion in thread pool test and re-enable

benclifford:
There is no guarantee that a task will run on every thread in
the threadpool; but no more than the expected number of threads
should be observed.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Looks good.

PR #1922, opened 2021-01-05
Fix, tidy, re-enable outputs= related test

benclifford:
The test was previously not waiting for tasks to finish,
which was a race condition.

## Type of change

- Code maintentance/cleanup
- Testing


yadudoc (approved):
This looks reasonable. Adding the tag for requiring a shared-fs makes sense.

Issue #1885, opened 2020-11-08
Possible race condition in htex scale down


benclifford:
**Describe the bug**
I've seen the below test failure as a one-off in `--config local` tests in the CI on several occasions.

My gut feeling is there's a race condition here - either in the test, or in parsl itself. I'm noting this for later investigation.

The scale down behaviour here is relatively new. Tagging @yadudoc who implemented it.

```
parsl/tests/test_scaling/test_scale_down.py F
=================================== FAILURES ===================================
________________________________ test_scale_out ________________________________
    @pytest.mark.local
    def test_scale_out():
        dfk = parsl.dfk()
    
        # Since we have init_blocks = 0, at this point we should have 0 managers
        print("Executor : ", dfk.executors['htex_local'])
        print("Before")
        print("Managers   : ", dfk.executors['htex_local'].connected_managers)
        print("Outstanding: \n", dfk.executors['htex_local'].outstanding)
        assert len(dfk.executors['htex_local'].connected_managers) == 0, "Expected 0 managers at start"
    
        fus = [sleeper(i) for i in [3, 3, 25, 25, 50]]
    
        for i in range(2):
            fus[i].result()
    
        # At this point, since we have 1 task already processed we should have atleast 1 manager
        print("Between")
        print("Managers   : ", dfk.executors['htex_local'].connected_managers)
        print("Outstanding: \n", dfk.executors['htex_local'].outstanding)
>       assert len(dfk.executors['htex_local'].connected_managers) == 5, "Expected 5 managers once tasks are running"
E       AssertionError: Expected 5 managers once tasks are running
E       assert 3 == 5
E        +  where 3 = len([{'active': True, 'block_id': '1', 'idle_duration': 0, 'manager': '884d27969893', ...}, {'active': True, 'block_id': '...486f', ...}, {'active': True, 'block_id': '2', 'idle_duration': 0.0022573471069335938, 'manager': 'a19000850efd', ...}])
[...]

PR #1926, opened 2021-01-05
Correct create_monitoring_info args in ParslExecutor

ZhuozhaoLi:
# Description

`create_monitoring_info` has a block_id_type parameter, but that isn't the case for ParslExecutor.

This PR corrects the issue.

## Type of change

### DIFF
--- a/parsl/executors/base.py
+++ b/parsl/executors/base.py
@@ -103,7 +103,7 @@ class ParslExecutor(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def create_monitoring_info(self, status: Dict[object, JobStatus]) -> List[object]:
+    def create_monitoring_info(self, status: Dict[object, JobStatus], block_id_type: str) -> List[object]:
         """Create a monitoring message for each block based on the poll status.
 
         :return: a list of dictionaries mapping to the info of each block
--- a/parsl/executors/base.py
+++ b/parsl/executors/base.py
@@ -102,13 +102,12 @@ class ParslExecutor(metaclass=ABCMeta):
         """
         pass
 
-    @abstractmethod
     def create_monitoring_info(self, status: Dict[object, JobStatus], block_id_type: str) -> List[object]:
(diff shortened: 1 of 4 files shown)


parsl/executors/base.py
```diff
@@ -103,7 +103,7 @@ def scaling_enabled(self) -> bool:
         pass
 
     @abstractmethod
-    def create_monitoring_info(self, status: Dict[object, JobStatus]) -> List[object]:
+    def create_monitoring_info(self, status: Dict[object, JobStatus], block_id_type: str) -> List[object]:
```
benclifford, line 106: this signature makes that parameter compulsory, which is probably the right thing to do.

But not every call to create_monitoring_info supplies a block_id_type, I think

benclifford (changes requested):

parsl/executors/status_handling.py
```diff
@@ -155,9 +152,6 @@ def set_bad_state_and_fail_all(self, exception: Exception):
     def status(self):
         return {}
 
-    def create_monitoring_info(self, status: Dict[object, JobStatus]) -> List[object]:
-        return []
-
```
benclifford, line 160: These implementations in this file might be necessary (or move them into the provider base class rather than making the base class method abstract) because not all concrete executors implement this method.

benclifford (changes requested):

parsl/executors/status_handling.py
ZhuozhaoLi: Yea, you are right. I realized that via CI again. I added these back.

parsl/executors/base.py
ZhuozhaoLi: Added `block_id_type` to every call

benclifford:
looks like this hit issue #1885 in CI

ZhuozhaoLi:
> looks like this hit issue #1885 in CI

It passed this time.

benclifford (approved):

Issue #1137, opened 2019-07-15
Better ec2 installation of correct parsl version


benclifford:
At present, the ec2/aws template.py installs the latest release version of parsl using pip.

When running from development source, this often installs an incompatible version of parsl.

This is a problem for any users who attempt our very common "you have a bug; please use the latest `master` version of parsl" approach; and makes development and testing much harder.

A better way to install the same version of source as is being submitted from would be nice. For example, @annawoodard has commented on slack about how some condor-based installations install a worker-side parsl that is a copy of the submit-side python install tree.

benclifford:
After some experimentation, I would be more comfortable if it was left to either the base AMI or the worker init to install the correct parsl version, rather than the template having a hard-coded often-incorrect install. The choice of base AMI and worker init is more easily configured by a user.

yadudoc:
I agree, both AMI and worker_init are far more easier for the user. We should remove the whole bunch of installs around here: https://github.com/Parsl/parsl/blob/94de10275a0190fd7e59c5d5c2261fb81feeab72/parsl/providers/aws/template.py#L6

Assuming debian/ubuntu is another poor choice that we've had at least 1 user complain about.

PR #1928, opened 2021-01-07
Install correct Parsl version on EC2 via SCP from local installation

knagaitsev:
# Description

Fixes # https://github.com/Parsl/parsl/issues/1137

This is still somewhat a draft, but I am quite new to Parsl and am hoping to get some feedback. Maybe there are better solutions to this issue, but this is just what I was experimenting with. I've also only tested the AWS remote provider so far in familiarizing myself with Parsl, but possibly this solution could be applicable to other providers.

I first added a new AWS option `upload_parsl` which defaults to False, but follows the below steps if True.

This approach installs the user's locally installed Parsl on the EC2, uploading it with SCP. It follows these steps:

1. Wait for EC2 to start running, then have an additional delay to allow for SSH connection
2. Connect to the EC2 via SSH for SCP (using options I added `ssh_username` and `ssh_key_filename` if needed)
3. Find path to the local Parsl install, then SCP it to the EC2 in the `/tmp` directory
4. Do a normal `pip3 install parsl` on the EC2 in case something fails
5. When the local Parsl upload is done, find and replace this pip3 installed version on the EC2 with the uploaded version located in `/tmp`
6. Run `process_worker_pool.py` as usual

Edit: One other detail to consider is that the upload function takes a long time, maybe instead of stalling the main thread while the EC2 boots up and the SCP is being done, it should happen as a separate async task

## Type of change


parsl/providers/aws/aws.py
```diff
@@ -2,7 +2,12 @@
 import logging
 import os
 import time
+import io
 from string import Template
+from pathlib import Path
+
+from paramiko import SSHClient, AutoAddPolicy
+from scp import SCPClient
```
Loonride, line 10: I will move these dependencies such that they are not needed unless AWS is being used, like with `boto3` dep

parsl/providers/aws/install_uploaded_parsl.py
```diff
@@ -0,0 +1,22 @@
...
+        pass
+    copy_tree(orig, dest)
+
+try:
+    orig_parsl_path = '/tmp/parsl'
+    dest_parsl_path = parsl.__path__[0]
+    move(orig_parsl_path, dest_parsl_path)
+
+    orig_dist_path = '/tmp/parsl-1.0.0.dist-info'
+    dest_dist_path = str((Path(dest_parsl_path) / '../parsl-1.0.0.dist-info').resolve())
```
Loonride, line 19: still need to change this hard-coded version

parsl/providers/aws/aws.py
```diff
@@ -564,6 +579,52 @@ def status(self, job_ids):
 
         return all_states
 
+    def upload_local_parsl(self, instance):
+        parsl_root = Path(__file__).parent / '../..'
+        parsl_pip_dist = parsl_root / '../parsl-1.0.0.dist-info'
```
Loonride, line 584: Need to change hard-coded version

benclifford:
Hi.

* One approach I've thought about is to ask the local install which version of parsl it has installed, and then allow the installer template to use that version string in the `pip` install command (or however you want to install parsl).

The main reason I didn't pursue that is because most of my work is done with development versions of parsl which don't have a clearly defined version number, but instead might be identified by a git commit hash - although that ID is also identifiable and usable in a pip install command.

That moves away from making assumptions about how parsl is installed locally and remotely.

- There is also some work with the workqueue executor to do automatic installs of the remote environment, beyond just parsl. - tagging @dthain here in case he has any good pointers here, as I haven't done anything with it.

My feeling at the moment is that parsl should be using capabilities of other things like workqueue to do this, where there is a more serious goal to make this stuff work, rather than attempting to do this within parsl code.

dthain:
Hi @benclifford -- The problem of moving parsl from here to there is not unlike that of moving the function environment from here to there.  They both boil down to observing what is installed locally, then either moving the module files, creating a virtual environment and moving that, or doing a fresh install on the other side.  @trshaffer has been working on all of these and can perhaps pipe up about the best idea here.

Loonride:
I think I'll resume working on this once I have tackled a few smaller Parsl issues, just to familiarize myself more with the codebase.

yadudoc:
@Loonride, it might be best to move this to a draft while we sort out how to make progress on this.

yadudoc:
I'm closing this since we've not had any progress on this for a while.

PR #1931, opened 2021-01-11
Tidy strategy parameters

benclifford:
# Description

Unused fields are removed, and _general_strategy now requires
an explicit strategy type (which was already the case in all
invocations)

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -130,14 +130,11 @@ class Strategy(object):
         for executor in executors:
             self.executors[executor.label] = {'idle_since': None, 'config': executor.label}
 
-    def _strategy_noop(self, status: List[ExecutorStatus], tasks, *args, kind=None, **kwargs):
+    def _strategy_noop(self, status: List[ExecutorStatus], tasks):
         """Do nothing.
 
         Args:
             - tasks (task_ids): Not used here.
-
-        KWargs:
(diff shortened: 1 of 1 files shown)


parsl/dataflow/strategy.py
```diff
@@ -154,13 +151,13 @@ def unset_logging(self):
...
+    def _strategy_simple(self, status_list, tasks):
+        self._general_strategy(status_list, tasks, strategy_type='simple')
 
-    def _strategy_htex_auto_scale(self, status_list, tasks, *args, kind=None, **kwargs):
-        self._general_strategy(status_list, tasks, *args, strategy_type='htex', kind=None, **kwargs)
+    def _strategy_htex_auto_scale(self, status_list, tasks):
+        self._general_strategy(status_list, tasks, strategy_type='htex')
 
-    def _general_strategy(self, status_list, tasks, *args, strategy_type='simple', kind=None, **kwargs):
+    def _general_strategy(self, status_list, tasks, *, strategy_type):
```
ZhuozhaoLi, line 160: `*` also needs to be removed, I assume.

benclifford: The `*` is to keep strategy_type be a kwarg, like it was before this PR.

ZhuozhaoLi (approved):
Looks good

PR #1932, opened 2021-01-11
Test that monitoring resource rows are recorded for a long task

benclifford:
## Type of change

- Testing


ZhuozhaoLi (approved):
I saw 3 messages on my local test. `>1` should be reasonable to set here for a 3-second task with 1s monitoring interval.

PR #1933, opened 2021-01-11
htex task prioritisation prototype

benclifford:
This patch has been in LSST parsl since around July 2020.

It allows explicit prioritisation of queued htex tasks, overriding the implicit queue order that was sometimes confusing/undesired in the complicated submission patterns of LSST work.

This has been used in a couple of ways: to prioritise/group all tasks of a particular type, and to prioritise/group all apps associated with a particular subset of data.

I'm putting it up for review as a draft PR because I haven't reviewed this for general usability.


benclifford:
Closing this in favour of using Work Queue for interesting task management, see #2067 which wires in the Work Queue priority field. The work queue priority type is simpler than the type restriction used in this PR: this prototype let you use any comparable objects; work queue uses a double.

PR #1936, opened 2021-01-13
Add much more aggressive type checking for monitoring

benclifford:
# Description

One of the main points of this is to work towards more explicit descriptions
of what goes over each queue.

Another is that this code has been poorly tested in CI in the past, and
typechecking provides more resilience there.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -309,7 +309,7 @@ class MonitoringHub(RepresentationMixin):
                         monitoring_hub_url: str,
                         run_id: str,
                         logging_level: int,
-                        sleep_dur: int) -> Callable:
+                        sleep_dur: float) -> Callable:
         """ Internal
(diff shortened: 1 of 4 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -304,17 +303,17 @@ def close(self):
...
-                        run_id,
-                        logging_level,
-                        sleep_dur):
+    def monitor_wrapper(f: Any,
+                        try_id: int,
+                        task_id: int,
+                        monitoring_hub_url: str,
+                        run_id: str,
+                        logging_level: int,
+                        sleep_dur: int) -> Callable:
```
ZhuozhaoLi, line 312: This should be `float`? the type for `MonitorHub` is currently `float`

benclifford: yeah, probably. I've changed that

ZhuozhaoLi (approved):
Looks good

PR #1944, opened 2021-01-17
Rework hub_channel usage to be more obviously thread safe

benclifford:
Instances of hub_channel aren't threadsafe. This PR moves away from having
one magic self.* channel, and another as a local in a thread, to creating
all channels as locals in threads.

monitoring_enabled is replaced by hub_channel becoming an optional, with
None signalling the same as !monitoring_enabled.

_send_monitoring_info gets a new parameter, hub_channel, which means it can
now be used in more places, reducing some code duplication.

This is in preparation for some upcoming PRs which allow the htex messaging
connections to be used for sending monitoring messages from the worker
side, instead of the monitoring system running its own monitoring
network connections.

This PR should not affect monitoring behaviour.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -181,14 +181,6 @@ class Interchange(object):
         self.hub_address = hub_address
         self.hub_port = hub_port
 
-        self.monitoring_enabled = False
-        if hub_address and hub_port:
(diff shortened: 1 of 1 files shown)


ZhuozhaoLi (approved):
LGTM

PR #1945, opened 2021-01-17
Memoize functions like any other value

benclifford:
Prior to this PR, inputs for memo hashes are generated in two places:

* in app construction, producing a func_hash by attempting to get the
  function source code

* id_for_memo hashes various things, including arguments and the app name

When the function is wrapped by some pieces of parsl, such as
remote_side_bash_executor or some file staging code, then the function is
replaced by a new function, and the original function is passed as an
additional argument.

This has the effect of id_for_memo hashing also hashing the function, in
addition to the app construction hash, and this time hashing it in a
different way - by using serialisation of the function object.

This hash-by-serialisation method was extremely fragile and would change for
probably undesirable reasons, such as source code changes elsewhere in
the source file.

This PR removes the first explicit func_hash construction, and makes all
hash generation happen with id_for_memo.

It replaces the serialisation-based hashing of functions with the source based
method previously used in the app's generation of func_hash. This now applies
to both the base function, and to functions passed as arguments.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/app/app.py
+++ b/parsl/app/app.py
@@ -4,8 +4,6 @@ The App class encapsulates a generic leaf task that can be executed asynchronous
 """
 import logging
 from abc import ABCMeta, abstractmethod
-from inspect import getsource
-from hashlib import md5
 from inspect import signature
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


yadudoc (changes requested):
@benclifford, I think these changes make sense. The memoization module is a better place for this logic than having some of these bits scattered in the app modules. One of the reasons for having the fn_hash attached to the app was to ensure that the function body isn't hashed for every app invocation. I think the same benefits can be had by adding a `@functool.lru_cache()` decorator. What do you think?

benclifford:
I think that kind of caching should work ok for functions.

PR #1946, opened 2021-01-20
Add decorator to catch exceptions coming out of threads and processes

benclifford:
# Description

When a thread or process raises an exception out of its top level function,
that exception does not get logged automatically. This sometimes results in
the threads appearing to silently stop without any reported error.

This patch adds a @wrap_with_logs decorator which can be used on
thread/process top level functions (or any other function) and which logs
such exceptions as they propagate upwards.

This won't catch cases where logging is sent to a logger name where
logging has not been configured for that name yet - for example, exceptions
happening before interchange or monitoring database manager have
configured their logging.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/process_loggers.py
+++ b/parsl/process_loggers.py
@@ -1,0 +1,39 @@
+import logging
+import threading
+import functools
+
+from typing import Callable, Optional
+
+
+def wrap_with_logs(fn: Optional[Callable] = None, target: str = __name__) -> Callable:
+    """Calls the supplied function, and logs whether that
+    function raised an exception or terminated normally.
+
+    This is intended to be used around the top level functions of
+    processes and threads, where exceptions would normally not
+    go to a log.
+    """
+
+    def decorator(func):
+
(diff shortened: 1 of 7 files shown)


parsl/process_loggers.py
```diff
@@ -0,0 +1,39 @@
...
+            thread = threading.current_thread()
+            name = f"{func.__name__} on thread {thread.name}"
+            logger = logging.getLogger(target)
+
+            try:
+                r = func(*args, **kwargs)
+                logger.debug("Normal ending for {}".format(name))
+                return r
+            except Exception:
+                logger.error("Exceptional ending for {}".format(name), exc_info=True)
```
yadudoc, line 31: This sounds oddly like a positive outcome? Maybe this could instead be `"{} failed with an Exception"`

yadudoc (approved):
Changes look good. I have a minor comment on the wording of a logline.

PR #1947, opened 2021-01-25
Make monitoring more resilient to spurious network traffic

benclifford:
In practice on cori.nersc.gov on large runs, some unexpected network
traffic arrived at monitoring ports. This would sometimes crash
the monitoring system. This PR makes monitoring more likely to ignore
such traffic rather than crashing.

This PR adds some tests to generate spurious traffic.

Notes to reviewers:
I'm a little unclear about the correct way to report that
exceptions happened in the database loop: at the moment, this
ignores such exceptions and then reports a final exception
when monitoring finishes.

Another of my concerns is that there are other causes of exceptions,
not only spurious messages, and this may not be the right
handling for such messages. (for example, does a particular
socket closed exception result in the loop going into a 100%
CPU loop?)

This change most makes indentation changes. `git diff --ignore-space-change -r origin/master` can give an easier to read diff.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -451,6 +451,21 @@ class DatabaseManager:
                     self._insert(table=NODE, messages=node_info_messages)
 
                 """
+                BLOCK_INFO messages
+
+                """
+                block_info_messages = self._get_messages_in_batch(self.pending_block_queue)
+                if block_info_messages:
+                    logger.debug(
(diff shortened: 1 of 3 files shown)


ZhuozhaoLi (approved):
LGTM. Adding two `try-catch` blocks to catch the exception in monitoring loop. The file changed page is very ugly, but there is an option to hide white space changes.

PR #1949, opened 2021-01-26
Add per-executor resource monitoring toggle

benclifford:
This allows an executor implementor (not a user) to indicate
that parsl resource monitoring is not compatible with that
executor - this incompatibility arises with the
ThreadPoolExecutor as process spawning and thread spawning
do not play nicely together in python.

With this PR, ThreadPoolExecutor can be used in a configuration
with monitoring enabled.

ThreadPoolExecutor is needed for a couple of features
even when most work is done through some other executor such
as HTEX:

* the existing globus data provider

* still unmerged @join_app functionality

# Description

Fixes # (issue)

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (dismissed):
These changes look good.

ZhuozhaoLi (approved):

yadudoc (approved):
I think these changes look good.

Issue #1948, opened 2021-01-25
Monitoring failed to capture FAILED blocks info in database


ZhuozhaoLi:
**Describe the bug**

In PR #1877, we log block polling status in monitoring. However, as reported on slack, this incurs a new error when Parsl fails to request a new block due to some provider errors (e.g., job limit). 

```
2021-01-20 17:54:11.478 parsl.dataflow.flow_control:114 [ERROR]  Flow control callback threw an exception - logging and proceeding anyway
Traceback (most recent call last):
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/dataflow/flow_control.py", line 112, in make_callback
    self.callback(tasks=self._event_buffer, kind=kind)
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/dataflow/task_status_poller.py", line 99, in poll
    self._update_state()
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/dataflow/task_status_poller.py", line 106, in _update_state
    item.poll(now)
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/dataflow/task_status_poller.py", line 46, in poll
    self.send_monitoring_info(self._status, block_id_type='internal')
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/dataflow/task_status_poller.py", line 52, in send_monitoring_info
    block_id_type=block_id_type)
  File "/home1/05340/mcflynn/miniconda3/envs/sra_workflow/lib/python3.7/site-packages/parsl/executors/high_throughput/executor.py", line 600, in create_monitoring_info
    d['block_id'] = self.removed_block_mapping.pop(id)
KeyError: '30'
```

After some investigations, this is not quite relevant with monitoring itself. Instead, this is because the block id type in HTEX is quite messy.

[...]

PR #1952, opened 2021-01-26
Make block id type consistent in the status returned from executor status polling

ZhuozhaoLi:
# Description

This PR refactors the key types in `status` returned from executor status polling method `executor.status()`.

Before this PR, executor status returns a `status` dict, where the key of a normal running block is a provider job id, but the key of a failed block is the block id counter in executor internally. This inconsistency in block id type causes errors in monitoring (issue #1948 ). This PR addresses this inconsistency in block id type---for all blocks, their keys are the block id counter in executor internally.  

This PR also renames `external_block_id` to `block_id`, and `internal_block_id` to `job_id` for ease to understand.

Note: it does not fix the issue in #1948 completely yet---upcoming PRs along with this PR will fix that.

## Type of change


parsl/executors/high_throughput/executor.py
```diff
@@ -714,8 +714,15 @@ def scale_in(self, blocks=None, block_ids=[], force=True, max_idletime=None):
...
-        return list(self.blocks.values())
+    def _get_block_and_job_ids(self) -> Tuple[List[object], List[object]]:
+        # Not using self.blocks.keys() and self.blocks.values()
+        # The dictionary may be changed during invoking this function
+        # As scale_in and scale_out are invoked in multiple threads
+        block_ids = list(self.blocks.keys())
+        job_ids = []
+        for bid in block_ids:
+            job_ids.append(self.blocks[bid])
+        return block_ids, job_ids
```
benclifford, line 725: There is a mypy error here. Here are two questions:

1) what is the type of block IDs? (I think probably str, not object)

2) what is the type of job IDs? (maybe also always a str?)


This fixes mypy for me, but it might not be the right fix:
index 54f3d19ba..dc8285c29 100644
@@ -714,12 +714,12 @@ class HighThroughputExecutor(StatusHandlingExecutor, RepresentationMixin):
 
         return self._filter_scale_in_ids(to_kill, r)
 
-    def _get_block_and_job_ids(self) -> Tuple[List[object], List[object]]:
+    def _get_block_and_job_ids(self) -> Tuple[List[str], List[object]]:
         # Not using self.blocks.keys() and self.blocks.values()
         # The dictionary may be changed during invoking this function
         # As scale_in and scale_out are invoked in multiple threads
         block_ids = list(self.blocks.keys())
-        job_ids = []
+        job_ids = []  # type: List[object]
         for bid in block_ids:
             job_ids.append(self.blocks[bid])
         return block_ids, job_ids
diff --git a/parsl/executors/status_handling.py b/parsl/executors/status_handling.py
index 994d30c7d..f27c891e7 100644
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -52,7 +52,7 @@ class StatusHandlingExecutor(ParslExecutor):
             return self._provider.status_polling_interval
 
     @abstractmethod
-    def _get_block_and_job_ids(self) -> Tuple[List[object], List[object]]:
+    def _get_block_and_job_ids(self) -> Tuple[List[str], List[object]]:
         raise NotImplementedError("Classes inheriting from StatusHandlingExecutor must implement "
                                   "_get_block_and_job_ids()")

parsl/executors/high_throughput/executor.py
```diff
@@ -615,27 +615,27 @@ def scale_out(self, blocks=1):
...
+                r.append(block_id)
             except Exception as ex:
-                self._fail_job_async(external_block_id,
-                                     "Failed to start block {}: {}".format(external_block_id, ex))
+                self._fail_job_async(block_id,
+                                     "Failed to start block {}: {}".format(block_id, ex))
         return r
 
-    def _launch_block(self, external_block_id: str) -> Any:
+    def _launch_block(self, block_id: str) -> Any:
```
benclifford, line 629: can the `Any` on this line turn into the correct type for a job id?

parsl/executors/status_handling.py
```diff
@@ -52,25 +52,25 @@ def status_polling_interval(self):
...
-    def _fail_job_async(self, job_id: Any, message: str):
+    def _fail_job_async(self, block_id: Any, message: str):
         """Marks a job that has failed to start but would not otherwise be included in status()
         as failed and report it in status()
         """
-        if job_id is None:
-            job_id = "failed-block-{}".format(self._generated_job_id_counter)
-            self._generated_job_id_counter += 1
-        self._simulated_status[job_id] = JobStatus(JobState.FAILED, message)
+        if block_id is None:
```
benclifford, line 63: Can block_id ever be None here?

benclifford (changes requested):

parsl/executors/status_handling.py
ZhuozhaoLi: Well, before this PR, although it was `job_id`, but when this method is invoked, the input is `block_id` actually---it is just a terminology fix for this line.

In terms of whether it can be `None`, for HTEX case, it would never be. But this is a based method in `StatusHandlingExecutor`, so this is just to ensure the method is invoked in the right way here? If you think it is okay to remove this, I can remove that `if` check

ZhuozhaoLi: You can check how `_fail_job_async` is invoked in master at https://github.com/Parsl/parsl/blob/4f47fb4cf6ec8c7135056c4ec3a335ee6a798635/parsl/executors/high_throughput/executor.py#L625-L626

parsl/executors/high_throughput/executor.py
ZhuozhaoLi: I think the job id can be either `None` or `str`?

benclifford: hm I looked in provider_base.py and the ID coming back from a provider is defined to be `Any` there. So using `str` for a job ID anywhere is wrong.

benclifford: (or this PR could change the API for providers to insist that job IDs are str)

ZhuozhaoLi:
@benclifford Are there any other changes you want in this PR?

benclifford:
> @benclifford Are there any other changes you want in this PR?

This PR gives job IDs type `str`, but providers can return any type as a job ID, not only string.

ZhuozhaoLi:
> > @benclifford Are there any other changes you want in this PR?
> 
> This PR gives job IDs type `str`, but providers can return any type as a job ID, not only string.

I changed the HTEX one to `Any`

parsl/executors/status_handling.py
```diff
@@ -22,22 +22,22 @@ def __init__(self, provider):
         self._simulated_status = {}
         self._executor_bad_state = threading.Event()
         self._executor_exception = None
-        self._generated_job_id_counter = 1
+        self._generated_block_id_counter = 1
         self._tasks = {}  # type: Dict[object, Future]
 
-    def _make_status_dict(self, job_ids: List[Any], status_list: List[JobStatus]) -> Dict[Any, JobStatus]:
-        """Given a list of job ids and a list of corresponding status strings,
+    def _make_status_dict(self, block_ids: List[Any], status_list: List[JobStatus]) -> Dict[Any, JobStatus]:
```
benclifford, line 28: block IDs are str, not Any

parsl/executors/status_handling.py
```diff
@@ -22,22 +22,22 @@ def __init__(self, provider):
...
+        """Given a list of block ids and a list of corresponding status strings,
         returns a dictionary mapping each job id to the corresponding status
```
benclifford, line 31: Does this return a dictionary with job id or block id keys?

benclifford (changes requested):

parsl/executors/status_handling.py
ZhuozhaoLi: with block_id as keys. This is what this PR is supposed to fix, make all the keys consistent since `block_id` should never be None, while `job_id` could be

benclifford: ok - the docstring is wrong - it still talks about job IDs

benclifford: both of the `Any` types should turn into str too

ZhuozhaoLi: Yea, I noticed them and fixed.

benclifford: the docstring is still wrong

parsl/executors/status_handling.py
```diff
@@ -52,25 +52,25 @@ def status_polling_interval(self):
...
-        if job_id is None:
-            job_id = "failed-block-{}".format(self._generated_job_id_counter)
-            self._generated_job_id_counter += 1
-        self._simulated_status[job_id] = JobStatus(JobState.FAILED, message)
+        if block_id is None:
+            block_id = "failed-block-{}".format(self._generated_block_id_counter)
+            self._generated_block_id_counter += 1
+        self._simulated_status[block_id] = JobStatus(JobState.FAILED, message)
 
     def status(self) -> Dict[object, JobStatus]:
```
benclifford, line 68: status returns a dict of block IDs -> job statuses here, now, I think?

benclifford (changes requested):

parsl/executors/status_handling.py
ZhuozhaoLi: Okay, tried to fix all the `status` instances in commit https://github.com/Parsl/parsl/pull/1952/commits/8e3e23c366b8872845abbcf57329537e3aa4f43e

parsl/executors/high_throughput/executor.py
```diff
@@ -714,8 +714,15 @@ def scale_in(self, blocks=None, block_ids=[], force=True, max_idletime=None):
 
         return self._filter_scale_in_ids(to_kill, r)
 
-    def _get_job_ids(self) -> List[object]:
-        return list(self.blocks.values())
+    def _get_block_and_job_ids(self) -> Tuple[List[str], List[Any]]:
+        # Not using self.blocks.keys() and self.blocks.values()
```
benclifford, line 718: what does this comment mean? the code *does* use self.block.keys()

benclifford (changes requested):

parsl/dataflow/task_status_poller.py
```diff
@@ -43,7 +43,7 @@ def poll(self, now: float):
         if self._should_poll(now):
             self._status = self._executor.status()
             self._last_poll_time = now
-            self.send_monitoring_info(self._status, block_id_type='internal')
+            self.send_monitoring_info(self._status, block_id_type='external')
```
benclifford, line 46: I think external and internal have been replaced by "block" and "job" so this parameter should be tidied up too: call it 'block' and 'job' perhaps

benclifford:
I'm a little unclear about what is going on with scale_out and scale_in in PollItem in task_status_poller.py vs the ID types that ParslExecutor is documented as returning in the docstrings.

parsl/dataflow/task_status_poller.py
ZhuozhaoLi: Fixed

ZhuozhaoLi:
> I'm a little unclear about what is going on with scale_out and scale_in in PollItem in task_status_poller.py vs the ID types that ParslExecutor is documented as returning in the docstrings.

Fixed the docstrings in the latest commits

parsl/executors/high_throughput/executor.py
ZhuozhaoLi: Wanted to say it does not use `self.blocks.keys()` and `self.blocks.values()` simultaneously.
E.g. `self.block` may be changing when invoking `self.blocks.keys()` and `self.blocks.values()` in between, and thus they returns different lengths.

benclifford:
I've been merging this repeatedly against my `benc-mypy` branch to look for relevant conflicts, and it looks like they are happy together now, which is good.

benclifford (approved):

PR #1955, opened 2021-02-01
Rewrite function comparison paragraph

benclifford:
It isn't about modules.

## Type of change

- Documentation update

### DIFF
--- a/docs/userguide/checkpoints.rst
+++ b/docs/userguide/checkpoints.rst
@@ -55,7 +55,7 @@ its signature, its body, or even the docstring within the body)
 will invalidate cached values. 
 
 However, Parsl does not traverse the call graph of the app function,
-so changes to functions called by an app will not invalidate
+so changes inside functions called by an app will not invalidate
 cached values.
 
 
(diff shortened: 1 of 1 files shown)


docs/userguide/checkpoints.rst
```diff
@@ -53,9 +53,10 @@ Parsl determines app equivalence by storing the a hash
 of the app function. Thus, any changes to the app code (e.g., 
 its signature, its body, or even the docstring within the body)
 will invalidate cached values. 
-Further, Parsl does not traverse imported modules, and thus
-changes to modules used by an app will not invalidate cached
-values.
+
+However, Parsl does not traverse the call graph of the app function,
+so changes to functions called by an app will not invalidate
```
danielskatz, line 58:
```suggestion
so changes inside functions called by an app will not invalidate
```

danielskatz: This is just a suggestion - I think, but am not sure, that it's a little clearer

danielskatz (approved):

PR #1958, opened 2021-02-04
Fix bad args for ScalingFailed constructor

benclifford:
In one place, the constructor was called with the wrong number of args,
which would result in a different exception when trying to raise this.

This PR makes provider label optional to support that use, where there
is no provider available.

This PR adds type annotations on the ScalingFailed __init__()

# Type of change


yadudoc (approved):
This looks right.

PR #1959, opened 2021-02-04
Remove unused executor errors

benclifford:
Looks like ControllerError is a leftover from IPP,
and InsufficientMPIRanks is from some now discarded
MPI work.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Nice!

PR #1960, opened 2021-02-04
Fix executor error repr vs str wrong-way-round-ness

benclifford:
str should be more informal/human readable, and repr should be more
python syntax

see:
  https://docs.python.org/3/reference/datamodel.html#object.__str__
  https://docs.python.org/3/reference/datamodel.html#object.__repr__

Before this PR (eg for a ScalingFailed error, which inherits its
str/repr behaviour from executorerror):

>>> str(ex)
('local', 'Attempts to provision nodes via provider has failed')
>>> repr(ex)
Executor local failed due to Attempts to provision nodes via provider has failed

After this, the object default repr is used instead of overriding,
and where repr() was previously overridden to provide human readable text,
that is changed to be the str() representation.

So, the example ScalingFailed error looks like:

>>> str(ex)
'Executor local failed due to: Attempts to provision nodes via provider has failed'
>>> repr(ex)
"ScalingFailed('local', 'Attempts to provision nodes via provider has failed')"

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/errors.py
+++ b/parsl/executors/errors.py
@@ -12,8 +12,8 @@ class ExecutorError(ParslError):
         self.executor = executor
         self.reason = reason
 
-    def __repr__(self):
-        return "Executor {0} failed due to {1}".format(self.executor, self.reason)
+    def __str__(self):
+        return "Executor {0} failed due to: {1}".format(self.executor, self.reason)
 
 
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
This looks reasonable.

PR #1963, opened 2021-02-09
Add typeguard to user-facing app decorators

benclifford:
This PR changes imports in file staging providers to prevent an import loop introduced by the import of DataFlowKernel into the decorator module.

`import M` is more robust against import loops than `from M import x`

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Looks reasonable.

PR #1973, opened 2021-02-17
Unify two impls of OptionalModuleMissing exception class

benclifford:
There was one at the parsl top level, and one in providers, but
these were not used uniformly:

* the monitoring system used the provider version even though
it is nothing to do with the provider system.

* the top level error was not listed in the API documentation

This will break any code that was expecting specific classes for these errors, but I've never seen any in real life.
 
## Type of change

- Code maintentance/cleanup

### DIFF
--- a/docs/stubs/parsl.providers.error.OptionalModuleMissing.rst
+++ b/docs/stubs/parsl.providers.error.OptionalModuleMissing.rst
@@ -1,6 +1,0 @@
-parsl.providers.error.OptionalModuleMissing
-===========================================
-
-.. currentmodule:: parsl.providers.error
-
-.. autoexception:: OptionalModuleMissing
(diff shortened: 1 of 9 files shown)


docs/stubs/parsl.providers.error.OptionalModuleMissing.rst
```diff
@@ -1,6 +0,0 @@
-parsl.providers.error.OptionalModuleMissing
```
ZhuozhaoLi, line 1: Looks like `OptionalModuleMissing` in provider is removed. So this file should be removed as well?

benclifford: which file? `parsl.providers.error.OptionalModuleMissing.rst` *is* removed.

ZhuozhaoLi: Okay. I missed that.

ZhuozhaoLi (approved):

PR #1976, opened 2021-02-19
Allow passing GPU requirements to WQ executor

trshaffer:
# Description

Update WQ executor to allow specifying GPUs.

CCTools just released version 7.2.0, which improves handling for GPUs. This PR allows Parsl apps to specify GPUs in their resource requirements, which will be passed down to WQ.

Fixes # (issue)

## Type of change

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -337,16 +337,17 @@ class WorkQueueExecutor(NoStatusHandlingExecutor):
         if resource_specification and isinstance(resource_specification, dict):
             logger.debug("Got resource specification: {}".format(resource_specification))
 
+            required_resource_types = set(['cores', 'memory', 'disk'])
             acceptable_resource_types = set(['cores', 'memory', 'disk', 'gpus'])
             keys = set(resource_specification.keys())
-            if self.autolabel:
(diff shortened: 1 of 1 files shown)


parsl/executors/workqueue/executor.py
```diff
@@ -333,25 +333,25 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+                    raise ExecutorError(self, message)
 
             elif keys != acceptable_resource_types:
                 logger.error("Running with `autolabel=False`. In this mode, "
                              "task resource specification requires "
                              "three resources to be specified simultaneously: cores, memory, and disk")
                 raise ExecutorError(self, "Task resource specification requires "
-                                          "three resources to be specified simultaneously: cores, memory, and disk, "
-                                          "and only takes these three resource types, when running with `autolabel=False`. "
+                                          "three resources to be specified simultaneously: cores, memory, and disk. "
```
benclifford, line 354: this branch of the `if` fires when one of the (now) four resource types is missing, I think?

trshaffer: You're right, I redid those checks.

Also, WQ will be changing its requirements in the near future, so we'll be able to drop the "all 3 required" rule then

benclifford (approved):

PR #1978, opened 2021-02-25
Workaround for missing /sbin/ip command

radiantone:
# Description

Some linux distros do not come with /sbin/ip command and thus a workaround is needed for the function address_by_route.

Fixes # (issue)

## Type of change
Code

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -7,7 +7,6 @@ so some experimentation will probably be needed to choose the correct one.
 """
 
 import logging
-import os
 import platform
 import requests
 import socket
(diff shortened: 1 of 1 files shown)


benclifford:
Hi. This looks useful.

Perhaps it would be good to remove `ip` entirely - I think the socket code you added will work in strictly more circumstances than calling out to `ip`.

benclifford:
CI failing with these two linter errors that need fixing. Otherwise, this is ready to merge, I think.

```
flake8 parsl/
parsl/addresses.py:10:1: F401 'os' imported but unused
parsl/addresses.py:39:1: W293 blank line contains whitespace
Makefile:41: recipe for target 'flake8' failed
make: *** [flake8] Error 1
The command "make test" exited with 2.
```

benclifford (approved):

PR #1992, opened 2021-03-15
Add units to walltime human readable message

benclifford:
This makes it clearer what time period is actually being represented.

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
:+1:

PR #1995, opened 2021-03-18
Mark some tests that do not work for sites without a shared file system

ZhuozhaoLi:
# Description

Some tests currently failed on sites without a shared file system (e.g., k8s). This PR marks those tests to exclude them when doing site tests.

## Type of change

### DIFF
--- a/parsl/tests/site_tests/site_config_selector.py
+++ b/parsl/tests/site_tests/site_config_selector.py
@@ -73,3 +73,5 @@ def fresh_config():
         # We should skip this.
 
     return config
+
+config = fresh_config()
--- a/parsl/tests/site_tests/site_config_selector.py
+++ b/parsl/tests/site_tests/site_config_selector.py
@@ -74,4 +74,5 @@ def fresh_config():
 
     return config
 
+
 config = fresh_config()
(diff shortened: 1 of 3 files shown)


benclifford:
This fails flake8

parsl/tests/site_tests/site_config_selector.py
```diff
@@ -73,3 +73,6 @@ def fresh_config():
         # We should skip this.
 
     return config
+
+
+config = fresh_config()
```
benclifford, line 78: What is this for?

benclifford (changes requested):
new test labels are good.

I don't see what the `fresh_config` change is for.

parsl/tests/site_tests/site_config_selector.py
ZhuozhaoLi: To do `make site_test`, we need this `site_config_selector` to return a `config`, otherwise it returns a `config` not found error

benclifford: looks like this was removed accidentally in #1720

benclifford (approved):

PR #2001, opened 2021-03-30
Switched the order of serializers to DillSerializer and PickleSerializer

yongyanrao:
Switched the order of serializers to DillSerializer and PickleSerializer

@benclifford We wanted to prioritize dill serializer, because dill is more reliable in some cases. For example, with the following snippet
```
    dfk = parsl.load(config)
    exe = dfk.executors['Midway_HTEX_multinode']
    f = exe.submit(square, None, 2)
    print(f.result())
```
The original order of serializers would give an error message of
`AttributeError: Can't get attribute 'square' on <module '__main__' from '/home/yrao2/.conda/envs/parsl_probe/bin/process_worker_pool.py'>`

### DIFF
--- a/parsl/serialize/concretes.py
+++ b/parsl/serialize/concretes.py
@@ -6,29 +6,6 @@ logger = logging.getLogger(__name__)
 from parsl.serialize.base import SerializerBase
 
 
-class PickleSerializer(SerializerBase):
-    """ Pickle serialization covers most python objects, with some notable exceptions:
-
-    * functions defined in a interpretor/notebook
-    * classes defined in local context and not importable using a fully qualified name
-    * clojures, generators and coroutines
-    * [sometimes] issues with wrapped/decorated functions
-    """
-
-    _identifier = b'01\n'
-    _for_code = True
-    _for_data = True
-
-    def serialize(self, data):
-        x = pickle.dumps(data)
-        return self.identifier + x
-
-    def deserialize(self, payload):
-        chomped = self.chomp(payload)
-        data = pickle.loads(chomped)
-        return data
-
-
 class DillSerializer(SerializerBase):
     """ Dill serialization works on a superset of object including the ones covered by pickle.
     However for most cases pickle is faster. For most callable objects the additional overhead
@@ -41,7 +18,7 @@ class DillSerializer(SerializerBase):
     * clojures
     """
 
-    _identifier = b'02\n'
+    _identifier = b'01\n'
     _for_code = True
     _for_data = True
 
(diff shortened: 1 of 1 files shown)


yadudoc (approved):

benclifford:
What is going on here that is different from regular app submission, that makes this break but does not break app submission?

yongyanrao:
I was doing some config testing. With the following submission, the original order would also have the same issue.
```
            block_id = executor.start()
            fut = executor.submit(square, None, 3)
            print(fut.result())
```

benclifford:
but does parsl app execution work with the same config file?

yongyanrao:
Yes, I used the `site_test` configs on their corresponding sites.

yongyanrao:
@benclifford Let us look at the following two examples to compare `dill` and `pickle`.
- Example 1: with one file `main_only.py`
```
# main_only.py
import dill
import pickle

def square(x):
    return x*x

if __name__ == "__main__":

    dill_serialized = dill.dumps(square)
    pickle_serialized = pickle.dumps(square)

    dill_deserialized = dill.loads(dill_serialized)
    pickle_deserialized = pickle.loads(pickle_serialized)

    print(dill_deserialized(5))
    print(pickle_deserialized(5))
```
`python main_only.py` would run without any issue.
- Example 2: with two files in the same directory, `main_lib.py` and `lib.py`
```
# main_lib.py
import dill
import pickle
from lib import dill_serialize, pickle_serialize

if __name__ == "__main__":

    dill_deserialized = dill.loads(dill_serialize())
    pickle_deserialized = pickle.loads(pickle_serialize())

    print(dill_deserialized(5))
    print(pickle_deserialized(5))
```
```
# lib.py
import dill
import pickle

def square(x):
    return x*x

def dill_serialize():
    return dill.dumps(square)

def pickle_serialize():
    return pickle.dumps(sqaure)
```
`python main_lib.py` would show that `dill` part worked and `pickle` part did not work with error message of `NameError: name 'sqaure' is not defined`. This is because `pickle.dumps()` merely returns a reference to the original function, but `dill.dumps()` encodes the function body. The difference shows up when the function is defined in a different module. So we think `dill` is more reliable for serialization, and decide to prioritize it in our serializer.

The reason we did not see this kind of behavior before is that with the `@python_app` decorator, whether the function is defined in the same module or not will not matter. Both Example 3 and Example 4 as below work.
- Example 3: with one file `main_only_decorator.py`
```
# main_only_decorator.py
import dill
import pickle
from parsl.serialize import ParslSerializer

@python_app
def square(x):
    return x*x

if __name__ == "__main__":

    parsl_serializer = ParslSerializer()
    parsl_packed = parsl_serializer.pack_apply_message(square, 5, None)
    func, args, kwargs = parsl_serializer.unpack_apply_message(parsl_packed, locals())
    print(func)
    print(args)
    print(kwargs)
```
`python main_only_decorator.py` would work.
- Example 4: with two files in the same directory `main_lib_decorator.py` and `lib_decorator.py`
```
# main_lib_decorator.py
from parsl.serialize import ParslSerializer
from lib_decorator import parsl_serialize


if __name__ == "__main__":

    parsl_serializer = ParslSerializer()
    func, args, kwargs = parsl_serializer.unpack_apply_message(parsl_serialize(), locals())
    print(func)
    print(args)
    print(kwargs)
```
```
# lib_decorator.py
from parsl import python_app
from parsl.serialize import ParslSerializer

@python_app
def square(x):
    return x*x

def parsl_serialize():
    parsl_serializer = ParslSerializer()
    return parsl_serializer.pack_apply_message(square, 5, None)
```
`python main_lib_decorator.py` would also work. The decorator makes this happen.

yadudoc:
This PR is based on a bunch of faulty assumptions. Since this works hasn't moved in the past 6 months, I'm closing this.

PR #2002, opened 2021-03-31
Add exception handling for submit() of providers

yongyanrao:
Add exception handling for submit() of providers: make submit() raise consistent exception across providers


benclifford (commented):
I think the parts of this PR which put in testing for result codes and raising an exception are good.

I don't like the parts which take an exception replace it with a different exception, like this:
```
        except Exception as e:
            raise SubmitException(job_name,
                                  f'Submission of command failed due to {e}')
```

This style means that someone trying to debug an error later on loses access to stack trace information about what happened further inside the code.

Maybe it would be good to make SubmitException able to store the entire exception which it is hiding in that case.

Or maybe, SubmitException shouldn't happen here: we already know that something is wrong because an exception is already being raised, so there is no need to raise a different exception.

yadudoc (commented):
We cannot merge this before we have exception handling added to the executors/scaling logic. Right now this PR will break any code that uses these providers.

parsl/providers/aws/aws.py
```diff
@@ -648,11 +649,15 @@ def submit(self, command='sleep 1', tasks_per_node=1, job_name="parsl.aws"):
...
-                                    self.nodes_per_block)
-        [instance, *rest] = self.spin_up_instance(command=wrapped_cmd, job_name=job_name)
+        try:
+            job_name = self.generate_aws_id()
+            wrapped_cmd = self.launcher(command,
+                                        tasks_per_node,
+                                        self.nodes_per_block)
+            [instance, *rest] = self.spin_up_instance(command=wrapped_cmd, job_name=job_name)
+        except Exception as e:
+            raise SubmitException(job_name,
```
yadudoc, line 659: I think the better option here would be to add an `logger.exception(...DESCRIPTION..)` followed by a `raise` so that the exception is logged and left to the layer above to deal with.

parsl/providers/cobalt/cobalt.py
```diff
@@ -196,7 +197,10 @@ def submit(self, command, tasks_per_node, job_name="parsl.cobalt"):
             self.resources[job_id] = {'job_id': job_id, 'status': JobStatus(JobState.PENDING)}
         else:
             logger.error("Submission of command to scale_out failed: {0}".format(stderr))
-            raise (ScaleOutFailed(self.__class__, "Request to submit job to local scheduler failed"))
+            raise SubmitException(job_name,
```
yadudoc, line 200: You should still report the class name so that you can identify which provider is failing.

parsl/providers/slurm/slurm.py
```diff
@@ -232,8 +233,11 @@ def submit(self, command, tasks_per_node, job_name="parsl.slurm"):
                     job_id = line.split("Submitted batch job")[1].strip()
                     self.resources[job_id] = {'job_id': job_id, 'status': JobStatus(JobState.PENDING)}
         else:
-            print("Submission of command to scale_out failed")
             logger.error("Retcode:%s STDOUT:%s STDERR:%s", retcode, stdout.strip(), stderr.strip())
```
yadudoc, line 236: Please fix this to use `logger.exception`

yadudoc (commented):
Minor comments.

PR #2009, opened 2021-04-10
Use zsh compatible install syntax

benclifford:
Without '-marks, zsh tries to resolve our given example installer
commandlines as wildcards, resulting in errors like this:

$ pip install parsl[monitoring]
zsh: no matches found: parsl[monitoring]

## Type of change
- Documentation update


yadudoc (approved):
This looks reasonable.

Issue #2014, opened 2021-04-22
dfk.tasks[id] elimination


benclifford:
Since the introduction of `dfk.tasks` garbage collection (#1512), knowing a task ID is no longer enough to be able to get a task's record from the DFK. This has resulted in occasional bugs to do with assuming otherwise, and awkward logic vs checkpointing (itself buggy, #1996).

Probably a better way is to pass around python level references to task structures, and only use task ID values when leaving the DFK (for example, monitoring or log messages); any use of dfk/self.tasks[id] should be considered suspicious.

This then allows self.tasks records to be eliminated earlier, gives more confidence that a task record will be accessible when expected, and allows the Python garbage collector to deal with garbage collection of task records.

Potentially this could eliminate the need for self.tasks entirely, although that isn't my immediate goal.

One thing to be aware of is that previously code which only stored a task ID (which could disappear) might now be holding a python-level reference to a task record, and keeping that task record around longer than previously. I don't think this is an issue, though.

At time of writing, I've had a few patches being tested in the `desc` release for a while.

This issue should document progress on this.

benclifford:
I think #2033 is a similar kind of problem but in HTEX task management, not  DFK task management.

benclifford:
I've also seen this happen in CI, which I believe is related. This test isn't very good - it checks that checkpoint files are written out, but does not check that the relevant tasks are written. What happens I think is that the checkpoint call will race with the post-task-completion code: sometimes tasks will be wiped from the DFK by the post task completion, and no tasks will be checkpointed; sometimes the checkpoint code will get there first, meaning that there will be two attempts to wipe the task from the DFK (once in checkpointing, and once in task completion code).

The use of checkpointing when it is supposed to be disabled is invalid for this test, though, so really this test should be rewritten to test something better.

```
parsl/tests/test_checkpointing/test_python_checkpoint_1.py F
=================================== FAILURES ===================================
________________________ test_initial_checkpoint_write _________________________
n = 2
    @pytest.mark.local
    def test_initial_checkpoint_write(n=2):
        """1. Launch a few apps and write the checkpoint once a few have completed
        """
        parsl.load(config)
        results = launch_n_random(n)

>       cpt_dir = parsl.dfk().checkpoint()
parsl/tests/test_checkpointing/test_python_checkpoint_1.py:34:
[...]

PR #2015, opened 2021-04-22
Eliminate self.tasks[id] calls from joinapp callback

benclifford:
# Description

see issue #2014 

## Type of change


yadudoc (approved):
:+1:

PR #2016, opened 2021-04-22
Eliminate self.tasks[id] from exec done callback

benclifford:
# Description

See issue #2014 

## Type of change


yadudoc (approved):
This looks right.

PR #2018, opened 2021-04-22
eliminate self.tasks[id] from launch_if_ready

benclifford:
# Description

see #2014

## Type of change


yadudoc (approved):
This looks good to go!

PR #2019, opened 2021-04-22
Put viz test in a script

benclifford:
This means that the same test used in CI can also be used on the commandline,
and with bisection, 'git bisect run parsl/tests/test-viz'

## Type of change

- Code maintentance/cleanup


benclifford:
This probably has some interaction with #2078 as this PR changes travis behaviour, not github actions behaviour

yadudoc (changes requested):
As you've already noted, this PR does not have changes for github actions. Ideally we add that here before we merge.

yadudoc (approved):
Looks good now.

PR #2027, opened 2021-04-22
Fix broken (by #????) task time invoked in visualisation

benclifford:
this was failing silently (wrt CI) prior to this

TODO: look up which broke this (by removing schema model line?)

## Type of change


benclifford:
closing this because its not high enough priority for me - will re-open based on desc worker later potentially.

PR #2029, opened 2021-04-22
Describe monitoring protocols better

benclifford:
This PR adds type annotations, asserts and comments to better
describe the existing protocols used between monitoring components.

This is the first PR in a series which attempts to rationalise those protocols.

This PR should not change any protocol or behaviour except error handling when
a format is violated.

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
I think these look good.

PR #2030, opened 2021-04-22
Tidy tasks_per_node in strategy

benclifford:
The use of tasks_per_node case in general strategy is a mess:
the ExtremeScale case will never fire in the current extreme scale implementaiton,
because an extreme scale executor is also a high throughput executor, and so the
earlier htex case will fire instead.

It is possible that extreme scale scaling was broken because of this case. This
patch should not make it either better or worse, because it only eliminates dead
code.

When an executor is not an htex instance, no cases match, but no error is raised
here, and so tasks_per_node is never assigned. Later on (line 206) use of
tasks_per_node is an error.

This entire case is removed, and executor.workers_per_node is always used.

This opens the way to other executors that are not HighThroughputExecutor being scaled by this strategy - specifically, the workqueue executor in the `desc` branch.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
:+1:

Issue #2050, opened 2021-04-30
parsl.executors.FluxExecutor for the Flux resource manager


jameshcorbett:
**Is your feature request related to a problem? Please describe.**
No, it isn't.

**Describe the solution you'd like**
I would like to add an executor subclass that uses the [Flux](https://github.com/flux-framework) resource manager to launch Parsl tasks.

I have a branch that implements most of the core functionality, but it is missing tests and a lot of filling-in of features and functionality.

I think Parsl could also benefit from a FluxLauncher and FluxProvider, which I could also implement if it was helpful.

**Describe alternatives you've considered**
N/A

**Additional context**
Part of the [ExaWorks](https://github.com/ExaWorks/) collaboration with @kylechard, @hategan, and @yadudoc

jameshcorbett:
Closed due to PR merging.

PR #2051, opened 2021-04-30
parsl.executors.FluxExecutor

jameshcorbett:
# Description

#2050 

Adds an executor which leverages the Flux resource manager to launch apps and return futures.

Fixes # (2050)

## Type of change


parsl/app/bash.py
```diff
@@ -65,7 +65,7 @@ def open_std_fd(fdname):
 
     returncode = None
     try:
-        proc = subprocess.Popen(executable, stdout=std_out, stderr=std_err, shell=True, executable='/bin/bash')
+        proc = subprocess.Popen(executable, stdout=std_out, stderr=std_err, shell=True, executable='/bin/bash', close_fds=False)
```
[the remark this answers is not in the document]
jameshcorbett: Yeah, it will affect all executors in theory. In practice, though, do they check all their open fds, or do something else that would be affected by this change? 

The problem is that Flux provides MPI to jobs by letting them inherit a FD and setting an environment variable defining what the FD is. So in `MPI_Init` the MPI implementation will read the environment variable and go to open the FD, which fails if `close_fds=True`. 

There may be other ways to provide MPI, though, so the problem is specific to that specific method of providing it, and to executors that launch MPI jobs.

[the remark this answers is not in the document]
benclifford: I think its fine - was interested in having the reasoning recorded here in case something interesting shows up later.

benclifford:
There are a few problems with mypy type checking. From easiest to hardest:

Because flux isn't installed in the test environment, tell mypy to stop trying to typecheck flux modules - eg in mypy.ini, add:

```
[mypy-flux.*]
ignore_missing_imports = True
```

(or if you are intending to make CI test the flux executor, then this would go away when you install that)

HighThroughputExecutor's process worker pool doesn't pass mypy type checking - prior to this PR, it was ignored by mypy so didn't cause failures, but this PR imports code from inside process_worker_pool to re-use. That makes me a little uncomfortable.

In another branch, I have a bunch of typechecking work which fixes the process worker pool to typecheck properly - your PR is motivation for me to get that piece put on master, so I will make a separate PR for that now.

benclifford:
> In another branch, I have a bunch of typechecking work which fixes the process worker pool to typecheck properly - your PR is motivation for me to get that piece put on master, so I will make a separate PR for that now.

That PR, #2052, is merged into `master` now

jameshcorbett:
Thanks---rebased and pushed.

jameshcorbett:
I told mypy to stop checking Flux modules like you suggested. I do think it would be good to make CI test the executor. I think all we would need to do is install Flux and then get the tests running inside of a Flux instance. I could write some tests that assume an active Flux instance, or at least that Flux is in the PATH. Would it be feasible then to get Flux installed on the machine used for CI?

From a quick look just now, the CI seems to have failed just now for other reasons that don't have anything to do with my changes.

jameshcorbett:
I don't know how much everyone is up on Flux, so I will try to explain the executor. Every app submitted to the executor is run as an individual Flux job. Although technically, each app is wrapped up in that `execute_parsl_task` script that I added in my first commit, and that script is what is actually submitted as the Flux job.

(For bash apps I think it would be more efficient to just forget the wrapper script and submit the bash app directly to Flux. Any thoughts on whether that is a good idea and feasible? It would break a lot of the abstraction.)

Flux jobs are fairly heavyweight. I think with the latest Flux version, I've seen no more than ~50 jobs/sec on our clusters, regardless of resources, and I think that's pretty normal. So at the moment, I don't think the FluxExecutor is a great fit for users who want to run tons and tons of fast-running jobs.

There is a big caveat to the performance numbers, however. It's possible to improve Flux's performance almost infinitely by duplicating it. This comes at two big costs: resource fragmentation and code complexity, because you (usually) need a way of communicating and load-balancing between all the different Flux instances. I think Parsl may already have most of this logic, though? In any case, the current executor I have here doesn't do any duplication---there's just a single Flux instance.

For a use-case that consists primarily of small, fast python functions, I don't think the FluxExecutor is a good fit because I suspect you already have better executors for that kind of workflow. I'm not sure what Flux would add to that use case. But if you want to maximize Flux's performance I will need to do a lot more work to implement the load-balancing and duplicating and so on.

Right now, however, where I think the FluxExecutor would be most useful would be on apps that are a little slower and need a dedicated and nontrivial resource set (like MPI jobs), and especially when the resource requirements vary across apps. Flux (and the FluxExecutor here) would handle that kind of app no problem.

The resource options for each app (accepted keys in the `parsl_resource_specification`) are:
- num_tasks: the number of tasks to launch (MPI ranks for an MPI job), default 1
- cores_per_task: cores per task, default 1
- gpus_per_task: gpus per task, default 1
- num_nodes: if > 0, evenly distribute the allocated cores/gpus across N nodes. Does _not_ give the job exclusive access to those nodes. This option only affects distribution.

As for the availability of Flux, it is fairly portable and easy to build, and it is currently used on LLNL and ORNL machines running Slurm and LSF. So for Parsl, this executor wouldn't necessarily just be an LLNL-Slurm-only executor or something like that. However, Flux has known problems on Cray machines (like ALCF's Theta or LANL's Trinitite/Trinity) because of vendor-specific extensions to Cray's MPI libraries. The consequence is that if you want to run an MPI job under Flux on those machines, as of right now the only way to do it is to compile the MPI application with OpenMPI rather than the default Cray MPI. I am not sure when that problem is likely to be resolved, if at all.

-------------
Sorry for the information dump, I just want to make sure that everyone on the Parsl team knows what they're getting and what they're _not_ getting. If people are happy with what I've implemented now and just want some tests and that's it, I could do that no problem.

Tagging @kylechard @yadudoc and @hategan in case they can help figure out the path forward.

benclifford:
> (For bash apps I think it would be more efficient to just forget the wrapper script and submit the bash app directly to Flux. Any thoughts on whether that is a good idea and feasible? It would break a lot of the abstraction.)

There are various parsl features that rely on running python code on the worker: some file staging, monitoring, time limiting. 

There's also a theoretical possibility (but I've never seen it in real usage) that the python bit of a bash app looks at its environment to decide what bash command to run. That would also break - but I've never seen this done in real life.

The whole execution environment is built around executing python functions (some of which happen to fork out as bash apps) - that's pretty fundamental to everything parsl does as a python library for remotely executing python code. I think putting in deep changes to turn parsl core into something which can also run command lines, not python code, is a direction that we shouldn't be pursuing.

> Flux jobs are fairly heavyweight. I think with the latest Flux version, I've seen no more than ~50 jobs/sec on our clusters, regardless of resources, and I think that's pretty normal. So at the moment, I don't think the FluxExecutor is a great fit for users who want to run tons and tons of fast-running jobs.

I think thats fine -- things like the work queue executor are relatively heavy in their per-task startup, and make up for it by being better than parsl's earlier executors in other ways - primarily through resource allocation. I think its important to be clear about the costs and benefits and "why should you choose executor A over executor B" for all of our executors, and you've written text for that in this PR.

> There is a big caveat to the performance numbers, however. It's possible to improve Flux's performance almost infinitely by duplicating it. This comes at two big costs: resource fragmentation and code complexity, because you (usually) need a way of communicating and load-balancing between all the different Flux instances. I think Parsl may already have most of this logic, though? In any case, the current executor I have here doesn't do any duplication---there's just a single Flux instance.

There's nothing that does any load balancing between multiple executors in parsl data flow kernel: it sends tasks to an executor, and lets the executor do its own magic to get tasks executed.

jameshcorbett:
I forgot to mention that each Future is associated with several files---another reason the executor shouldn't be used for large numbers of futures.

> The whole execution environment is built around executing python functions (some of which happen to fork out as bash apps) - that's pretty fundamental to everything parsl does as a python library for remotely executing python code. I think putting in deep changes to turn parsl core into something which can also run command lines, not python code, is a direction that we shouldn't be pursuing.

OK, good to know so that possibility can be ruled out.

> There's nothing that does any load balancing between multiple executors in parsl data flow kernel: it sends tasks to an executor, and lets the executor do its own magic to get tasks executed.

In that case, it sounds like as a future optimization, the executor could be turned into a front-end for multiple sub-executors, and the top-level executor (the only executor exposed to the Parsl data flow kernel) would just be responsible for load balancing between them. For certain workflows that would improve throughput enormously, but at the moment, given current Flux APIs, it would require that the user specify:

1. The number of partitions or sub-executors (technically, the full hierarchy of Flux instances, like '5x6x12', but for simplicity let's just say the number of partitions/sub-executors).
2. The number of nodes available to Flux
3. The number of cores and gpus per node

parsl/executors/flux/execute_parsl_task.py
```diff
@@ -0,0 +1,43 @@
...
+    result = None
+    # open and deserialize the task's pickled input package
+    with open(args.input, "rb") as f:
+        fn_buf = f.read()
+        logging.info("Read input pickle file")
+        try:
+            result = execute_task(fn_buf)
+        except Exception as e:
+            logging.exception("Parsl task execution failed:")
+            result = e
```
jameshcorbett, line 31: As it stands, there is no way to distinguish between an exception thrown by a callable and one returned. So if the callable raises an exception, you would fetch it with ``future.result()`` and not ``future.exception()`` as one might expect. Is this behavior consistent with the other Parsl executors? I'm guessing not, just want to make sure.

parsl/executors/flux/executor.py
```diff
@@ -0,0 +1,203 @@
...
+            num_nodes=resource_spec.get("num_nodes"),
+            cores_per_task=resource_spec.get("cores_per_task", 1),
+            gpus_per_task=resource_spec.get("gpus_per_task"),
+        )
+        jobspec.cwd = os.getcwd()  # should be self.working_dir?
+        jobspec.environment = dict(os.environ)
+        jobspec.stdout = os.path.abspath(
+            os.path.join(self.working_dir, tid + "_stdout_stderr.txt")
+        )
+        jobspec.stderr = jobspec.stdout
```
jameshcorbett, line 184: Should stdout and stderr be separated?

parsl/executors/flux/executor.py
```diff
@@ -0,0 +1,203 @@
...
+        with open(infile, "wb") as infile_handle:
+            infile_handle.write(fn_buf)
+        jobspec = flux.job.JobspecV1.from_command(
+            command=[sys.executable, _worker_path, "-i", infile, "-o", outfile],
+            num_tasks=resource_spec.get("num_tasks", 1),
+            num_nodes=resource_spec.get("num_nodes"),
+            cores_per_task=resource_spec.get("cores_per_task", 1),
+            gpus_per_task=resource_spec.get("gpus_per_task"),
+        )
+        jobspec.cwd = os.getcwd()  # should be self.working_dir?
```
jameshcorbett, line 179: What should be the CWD for the apps?

parsl/executors/flux/executor.py
```diff
@@ -0,0 +1,247 @@
...
+        flux_handle_kwargs: Mapping = {},
+    ):
+        super().__init__()
+        self.label = label
+        self.working_dir = os.path.abspath(working_dir) if working_dir else os.getcwd()
+        self.managed = managed
+        self._flux_executor = None
+        self.flux_handle_args = flux_handle_args
+        self.flux_handle_kwargs = flux_handle_kwargs
+        self._provider = LocalProvider()
```
jameshcorbett, line 154: Flux isn't the default resource manager on any systems yet, so currently the primary way to use Flux is to request an allocation from the local resource manager (like Slurm or LSF) and then launch Flux within that allocation (e.g. srun --ntasks-per-node=1 --mpi=none flux start COMMAND``). How would that work out with the various provides offered by Parsl? I forced ``LocalProvider`` to require that users already have acquired their resource allocation and launched Flux, but that cuts down dramatically on the potential use-cases for this executor.

parsl/executors/flux/executor.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+        except Exception as unknown_err:
+            future_wrapper.set_exception(unknown_err)
+    else:  # the job exited abnormally
+        future_wrapper.set_exception(
+            AppException(f"Parsl task exited abnormally: returned {returncode}")
+        )
+
+
+class FluxExecutor(NoStatusHandlingExecutor, RepresentationMixin):
```
jameshcorbett, line 101: Currently the class inherits from NoStatusHandlingExecutor for convenience only. Flux provides status updates on all submitted jobs, so it would be pretty trivial to provide that on this executor. However, I see that ``Executor.status()`` is supposed to return a mapping from strings to ``JobStatus`` objects. What are the strings supposed to be? I already have a uuid for each Future returned from the executor, so it would make sense to use that string. However, how is the user supposed to figure out which uuid goes with which Future?

parsl/executors/flux/executor.py
jameshcorbett: One option would be to accept the CWD and environment through the resource spec.

parsl/executors/flux/executor.py
```diff
@@ -0,0 +1,247 @@
...
+            * gpus_per_task: gpus per task, default 1
+            * num_nodes: if > 0, evenly distribute the allocated cores/gpus
+              across the given number of nodes. Does *not* give the job exclusive
+              access to those nodes; this option only affects distribution.
+        *args:
+            positional arguments for the callable
+        **kwargs:
+            keyword arguments for the callable
+        """
+        tid = str(uuid.uuid4())
```
jameshcorbett, line 205: I wonder if it might be easier to use and debug if the task IDs were 0, 1, 2, 3, ..., and every executor instance was given a unique workdir (so, unique workdirs instead of unique file names). As it stands, with all the files produced, it can be very hard to remember which app/future goes with which UUID, because the UUIDs are so long. There would just be less to memorize.

jameshcorbett:
Dropping the draft tag. I left a bunch of inline comments where I need someone more familiar with Parsl to help out. Otherwise, and aside from the Flux hierarchy performance thing I mentioned above (which should be left for another time), as far as I know this is ready to go.

parsl/executors/flux/execute_parsl_task.py
yadudoc: If the call raised, `future.result()` will raise the same exception. This is consistent with the `concurrent.futures` behavior: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result. `future.exception()` on the other hand returns the exception and does not raise it.

parsl/executors/flux/executor.py
yadudoc: Scaling and status handling in Parsl is fairly messy. The executor doesn't need to be a StatusHandlingExecutor if we don't expect to plug into Parsl's mechanism to gauge available resource (nodes/compute) and the current task load and then try to request through the executor to get additional nodes. Ultimately the question is whether we want any dynamic resource provisioning and if so whether Flux or Parsl should be making the provisioning decisions.

yadudoc: If we just want a fixed-sized set of resources (the equivalent of 1 batch job) and just schedule tasks within it, `NoStatusHandlingExecutor` is probably the right choice.

parsl/executors/flux/execute_parsl_task.py
jameshcorbett: Right, that's what I thought. I think I inherited this code from somewhere else and just wanted to make sure it worked correctly. Thanks for clarifying!

parsl/executors/flux/executor.py
yadudoc: We should probably chat about this. We could essentially have the executor define a flux-specific `launch_cmd` that is then handed off to a parsl provider that will do the instantiation of Flux within a batch job.

parsl/executors/flux/executor.py
yadudoc: You could use an atomic counter, and it's probably nicer than UUIDs.

parsl/executors/flux/executor.py
yadudoc: I feel like I care about this because I've always had it these separate, but I do not know whether a user would care. Does flux default to merging the streams?

parsl/executors/flux/executor.py
jameshcorbett: Right, I was thinking something like that would be a good idea. The actual command will depend on the resource manager, and possibly the platform, though.

For instance, on LLNL Slurm machines we have to add the ``--mpibind=off`` option when we launch Flux in order to get decent performance.

The goal is generally to launch a single instance of Flux on each node in a batch job.

The executor can also connect to a remote Flux instance, but I think LocalProvider would cover that use case.

parsl/executors/flux/execute_parsl_task.py
hategan: I think this behavior is not consistent with the general contract of `Future`. The idea being that `Future.result()` should behave as if the computation wrapped by the future was invoked synchronously. That is, if said computation  throws an exception instead of returning a value, so should `Future.result()`.

jameshcorbett: Right, I think we're all agreed on what's supposed to happen.

parsl/executors/flux/executor.py
hategan: If you're grepping through a log file, there's no way a UUID is anything else but what it's supposed to be. If, on the other hand, you're grepping for `1`, you might get a few more extra hits.

parsl/executors/flux/executor.py
jameshcorbett: No, Flux defaults to storing it in its KVS. I'm happy to make them separate, although I'm used to have them united on the terminal, which is why I united them here. I think we should do whatever Parsl does elsewhere, if it does anything similar.

parsl/executors/flux/executor.py
jameshcorbett: Flux doesn't currently have any way to request additional nodes or resources, I believe, although I know that is an area of research for the Flux team. I think you would have to request more nodes from the local resource manager and then launch Flux in the new allocation.

I mentioned in another comment that the executor can connect to remote Flux instances, so it might not be too hard to have a executor which communicates with multiple allocations (and hence multiple Flux instances). But it would take some work to get the load-balancing right across allocations, unless Parsl already has code to handle that.

I think that functionality, though, like the functionality of having a hierarchy of Flux instances controlled by the executor, could be added later.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -464,17 +464,8 @@ def execute_task(bufs):
...
-        exec(code, user_ns, user_ns)
-
-    except Exception as e:
-        logger.warning("Caught exception; will raise it: {}".format(e), exc_info=True)
-        raise e
-
-    else:
-        # logger.debug("[RUNNER] Result: {0}".format(user_ns.get(resultname)))
-        return user_ns.get(resultname)
+    exec(code, user_ns, user_ns)
```
jameshcorbett, line 467: The exception was actually clobbered because ``logger`` wasn't defined. I thought the log message was not super useful anyway so I just removed it and the whole try/except block.

parsl/executors/flux/execute_parsl_task.py
jameshcorbett: Added a commit that should resolve this problem.

parsl/executors/flux/executor.py
benclifford: I've done some work to pull out HighThroughputExecutors block provisioning code into a superclass so it can also be used by the WorkQueueExecutor. The abstraction there is roughly that that kind of executor wants to execute task by having blocks of workers allocated by a batch system, but the nature of those workers is executor specific (so in the HightThroughputExecutor case, it is our process_worker_pool.py script; in the case of workqueue, it is workqueue's own `workqueue_worker` command. That abstraction also deals with deciding (in a simple way) when more blocks should be requested.

This abstraction doesn't prevent a user from saying "don't allocate any blocks at all - I will do this myself", which makes sense if a workqueue user wants to use some other provisioning tools provided by workqueue.

It sounds like potentially that abstraction fits your flux implementation too - if so, I can push on getting this into `master` a bit harder.

jameshcorbett: That sounds great, I think that would fit well with this executor. However, what about load balancing? When a new app is submitted to the executor, how would the executor decide what blocks to forward the app onto? Is there any kind of abstraction for that or would that be up to me to implement? That could be pretty hard if each app has a variable amount of resources (cores and gpus).

benclifford: > When a new app is submitted to the executor, how would the executor decide what blocks to forward the app onto?

In the HighThroughputExecutor/WorkQueue model, that is one of the primary purposes of the executor implementation - they have some component (different in each) that the workers all connect with.

jameshcorbett: And the workers report to the executor how much they are being utilized, so the executor knows which worker to pass new apps to?

benclifford: there's usually some kind of communication like that but it's executor-specific exactly what is going on there.

jameshcorbett: Right, that makes sense. I'm just not sure there is great Flux functionality for figuring out whether a Flux instance is being underutilized or not. Off the top of my head the only thing I can think of is going by the number/percentage of PENDING jobs. I will talk to the Flux team about this.

parsl/executors/flux/executor.py
jameshcorbett: We agreed in a conversation to separate them.

parsl/executors/flux/executor.py
yadudoc: Let's punt on this until a user complains that they want to set cwd per app.

parsl/executors/flux/executor.py
jameshcorbett: For reference: https://github.com/flux-framework/flux-core/discussions/3656

parsl/executors/flux/executor.py
jameshcorbett: Based on our talk on Tuesday, I updated the implementation so that it properly supports Parsl's execution model and doesn't assume that it is running inside a Flux instance on the target resources. Depending on the setup, users may want to provider overrides to a Launcher instance, or to mess with the executor's ``launch_cmd``. I don't know if there will be any more user-friendly way of doing it, because it is so dependent on the specific platform (LLNL Slurm vs LANL Slurm, for instance).

yadudoc:
@jameshcorbett Could you check the flake8 errors from CI here: https://travis-ci.com/github/Parsl/parsl/jobs/510817580#L544

jameshcorbett:
I believe I have fixed all of the issues but I cannot quite get the CI to pass. Each time one of the builds seems to error out.

benclifford:
> I cannot quite get the CI to pass. 

the last run seems to have passed!

The CI has a few tests that fail often in circumstances I've been unable to diagnose.

yadudoc:
@jameshcorbett I just tested this out on quartz, and it looks like the tests are working! :tada: At this point I'd recommend merging this PR, and following it up with a separate documentation PR that covers when, where and how to use the `FluxExecutor`.

yadudoc (approved):
:+1:

parsl/tests/test_flux.py
```diff
@@ -0,0 +1,170 @@
...
+)
+
+try:
+    import flux.job.executor  # noqa: F401
+except ImportError:
+    FLUX_AVAIL = False
+else:
+    FLUX_AVAIL = True
+
+require_flux = pytest.mark.skipif(
```
benclifford, line 20: I'm not super fond of this style of magic autodetection of if a test should happen. Elsewhere we use explicit pytest `-k` arguments to disable tests we expect to not pass (see eg. `-k not cleannet and not issue363` in the Makefile). Every time I've encountered this style of testing, I've encountered tests incorrectly passing because this code will incorrectly not test flux when I've installed it incorrectly.

jameshcorbett: That makes sense to me. There are three separate sets of tests in the module:

1. Tests that can run any time
2. Tests that require an installation of flux to be in `PATH`
3. Tests that must be run inside of a Flux instance, e.g. `flux start python -m pytest test_module.py`

3 implies 2. Should I split 1 and (2, 3) into separate modules, and then add a target "flux_test" for explicitly checking tests of type 2 and 3?

Also I'm a little confused about `-k`/`-m` because it looks like (according to pytest documentation) `-k` is for matching tests by name and `-m` is for matching tests by marker, and yet `-k` seems to be used in Parsl for matching tests by marker... am I missing something?

benclifford: huh, interesting. I hadn't noticed that -k vs -m thing.

I think you should get this PR merged (by clicking "update from master" and then waiting for a parsl committer to merge it, without putting in any further changes), and then fiddle with tests separately.

I don't think the layout of the test modules matters too much - just ensure the ones that require flux behaviour have a proper tag attached to them to make them disabled in the CI where necessary. Adding a flux test makefile target would be ok too but not mandatory.

PR #2053, opened 2021-05-03
Remove unused AppBase status field

benclifford:
This was initialized to 'created' and then never changed.

It was creating minor confusion wrt DFK task record status fields,
which it looks like this was once intended to be similar to.

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
:+1: Looks reasonable

PR #2054, opened 2021-05-03
Test combine() pattern in joinapps

benclifford:
Most of the test code for this was already defined, but it was not
actually used in a test.

The combine() pattern is the way joinapps can wait for several
tasks to complete, rather than one.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This test looks good.

PR #2056, opened 2021-05-05
clarify that max workers is per node

benclifford:
## Type of change

- Documentation update


ZhuozhaoLi:
Is it better to even make it a breaking change? i.e., change it to `max_workers_per_node`

benclifford:
maybe. this PR is not intended to achieve that.

yadudoc (approved):
@benclifford I agree with the documentation change. Unfortunately, most users don't read the docs, and it helps to have the config option be more descriptive. Would you be open to a separate PR adding a new `max_workers_per_node` option in addition and deprecating max_workers in the release after?

benclifford:
i'm fine with someone else doing that

PR #2057, opened 2021-05-05
Remove unused walltime from LocalProvider

benclifford:
## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Looks like this option is indeed obsolete.

PR #2059, opened 2021-05-05
Disable WQ-level retries with an option to re-enable

benclifford:
# Description

Previously by default, Work Queue will retry tasks that fail at the WQ level (for example, because of worker failure) an infinite number of times, inside the same parsl-level execution try.

That hides the repeated tries from parsl (so monitoring does not report start/end times as might naively be expected for a try, and parsl retry counting does not count).

This PR changes the default so that WQ will only execute a task once (in line with other executors), leaving parsl to handle any retries.

A new config parameter, max_retries, allows WQ level retries to be specified (back to infinity, or to any other number) so as to get previous behaviour if that is desired.

## Type of change


yadudoc (approved):
I'm going to assume that WQ support this functionality and that this was already tested. The changes look reasonable.

PR #2060, opened 2021-05-05
String formatting

### LINKED ISSUE #2045: MAINT: improve string formatting
String formatting is not consistent and can be improved.

Parsl doesn't support Python earlier than 3.6 meaning all strings that use `.format()` or `%` if any can be safely replaced with f-strings without breaking any backward compatibility. That would make the code shorter and more readable.
I can fix this with a series of PRs, say, one per file.

I don't see any alternatives other than ignoring the problem.

vkhodygo:
# Description
This is a draft of a cumulative PR to switch string formatting to `f-strings` where acceptable.
In addition, it fixes the only bug that I introduced in #2055 that is: `logging` module doesn't follow [PEP-0498](https://www.python.org/dev/peps/pep-0498/#rationale) due to many reasons and using `f-strings` as an argument works in general but is not officially supported (TIL).

I also found a few spots where similar issues had been introduced at some point before, they've been fixed.
Finally, there was a place where a string template and its argument were passed to `print()` separately as `print('%s', x)` rather than a single entity `print('%s' % x)`, resulting in both of them being printed independently. 

Note that logging doesn't follow the same standard throughout the code base but I didn't touch it as this is a matter of personal preference in the first place (and rarely can be a performance issue).

Fixes #2045

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -58,7 +58,9 @@ def address_by_query(timeout: float = 30) -> str:
         logger.debug("Address found: {}".format(addr))
         return addr
     else:
-        raise RuntimeError("Remote service returned unexpected HTTP status code {}".format(response.status_code))
+        msg = (f"Remote service returned unexpected HTTP status code "
+               f"{response.status_code}")
+        raise RuntimeError(msg)
 
 
 def address_by_hostname() -> str:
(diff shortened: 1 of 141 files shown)


vkhodygo:
Hi @benclifford,

Could you please take a look at this? There are some conflicts here but that should be fixable. I know there is a lot of files to review, sorry, I tend to get knee-deep into certain issues.

benclifford:
@vkhodygo I can't resolve the merge conflicts in your branch because it is in your repo, not the main parsl repo.

kylechard:
Closing due to the outstanding merge conflicts.

PR #2061, opened 2021-05-08
Eliminate self.tasks[id] calls from launch_task

benclifford:
See issue #2014 

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This looks reasonable

PR #2063, opened 2021-05-08
Fix and test wrong type handling for joinapp returns

benclifford:
prior to this, returning the wrong type was resulting in a
hang.

after this, returning the wrong type gives a TypeError - for
example:

TypeError: join_app body must return a Future, got <class 'list'>

## Type of change


yadudoc (approved):
This looks reasonable.

PR #2068, opened 2021-05-10
Workflow-pluggable retry scoring

benclifford:
When a task fails, instead of causing a retry "cost" of 1 (the previous behaviour), this PR allows that cost to be determined by a user specified function which is given some context about the failure.

After this PR there is a distinction between the accumulated retry cost, and the number of tries (which were previously the same number). This change is reflected in the monitoring database, with both values available.

Other considerations:
 - the retry scorer, if given the task record, is not prohibited from fiddling with
   the task record. What sort of things are plausible there?
    - changing the executor label for the task, so that the job retries somewhere
      else - for example some executor that is more expensive (slower to schedule
      but longer tasks?)
    - retry for a time period rather than for a number of counts - eg so as to not
         burn through retry units very fast in a fast failing transient case -
         so "quick fails" are discounted

## Type of change


docs/userguide/exceptions.rst
```diff
@@ -110,3 +112,57 @@ as they are unaffected by task C's failure.
...
+``example_retry_handler`` should be a function defined by the user that will
+compute the retry cost for a particular failure, given some information about
+the failure.
+
+For example, the following handler will give a cost of 1 to all exceptions,
+except when a bash app exits with unix exitcode 9, in which case the cost will
+be 100. This will have the effect that retries will happen as normal for most
+errors, but the bash app can indicate that there is little point in retrying
+by exiting with exitcode 9.
+
```
yadudoc, line 157: We should leave a note here that if the `retry_handler` itself fails, the cost is 1. Just to clarify that failure in the handler is treated similar to a failure in the app but with a pre-determined cost, just so that user's would avoid setting bad `retries` counts.

yadudoc (dismissed):
@benclifford I think this looks good to merge, but I wanted to check with you if it makes sense to update the docs to clarify the cost of retry_handler failure.

docs/userguide/exceptions.rst
benclifford: If the retry_handler itself fails, the whole task is aborted so it isn't really correct to say the cost is 1. Really, the cost is infinity - where infinity means "something big enough to stop any more tries happening".
```
  task_record['fail_cost'] = self._config.retries + 1
```

I'll add in a note that says something like "if the retry handler fails, no further tries are attempted".

benclifford: @yadu I've added a documentation note for this in 313ef3b9c2bd2f49aebb3328803b454019ebb33b - can you check, and reapprove if ok?

yadudoc: Alright this makes more sense.

yadudoc (approved):
Looks good now.

PR #2080, opened 2021-06-03
Remove task_id param from memo functions, as whole task record is available

benclifford:
When these functions need the task id, they can extract it from
the task record.
see #2014 

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Okay, this makes sense.

PR #2081, opened 2021-06-03
Remove unneeded task_id param from sanitize_and_wrap

benclifford:
See #2014

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This looks good. I also see that the sanitize_and_wrap docs include a `func` option in args which is now obsolete.

PR #2082, opened 2021-06-03
Remove documentation that interchange is walltime aware

benclifford:
## Type of change

- Documentation update


yadudoc (approved):
Okay, this is good to remove to avoid any confusion.

PR #2084, opened 2021-06-03
Do not unwrap joinapp future exceptions unnecessarily

benclifford:
An AppFuture will always present its exception as a future.exception(),
not as a RemoteWrapper. RemoteWrappers are used at the executor future
layer.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -371,10 +371,8 @@ class DataFlowKernel(object):
 
         outer_task_id = task_record['id']
 
-        try:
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
This makes sense. Wouldn't it have been easier to do `res = inner_future.result()` and leave the rest as is?

benclifford:
> Wouldn't it have been easier to do res = inner_future.result() and leave the rest as is?

@yadudoc i think just personal preference for representing control flow choices using `if` rather than try/except in an expected case.

PR #2085, opened 2021-06-03
Index task_hashsum to give cross-run query speedup

benclifford:
Practical experience with wstat has shown this index
to give great speedup when making queries which match
up tasks between runs based on their checkpointing
hashsum.

## Type of change


yadudoc (approved):
Sounds reasonable.

PR #2086, opened 2021-06-07
Refresh the sanitize_and_wrap docstring

benclifford:
## Type of change

- Documentation update


parsl/dataflow/dflow.py
```diff
@@ -692,19 +692,24 @@ def check_dep(d):
         return depends
 
     def sanitize_and_wrap(self, args, kwargs):
-        """This function should be called only when all the futures we track have been resolved.
+        """This function should be called when all dependency futures for a task
```
danielskatz, line 695: "dependency futures" sounds a little odd to me - have we defined this elsewhere?  Would "futures that are dependent on {something}" be better?  Or maybe "futures on which this the result of this function is dependent"?  (Sorry, I'm not sure which meaning is intended, which is my concern.)

benclifford: A Task [A] depend on other tasks [B, C] completing. The way this is represented internally is that task [A] is passed two futures f_B and f_C that represent the run status of tasks B, C.

This function should be called (by another part of the DFK) when f_B and f_C have completed.

It's probably fine to say "when all dependencies have completed"

yadudoc: I think saying "when all dependencies have completed" is clear. Let's make that change and get this merged!

danielskatz (approved):

PR #2087, opened 2021-06-07
autogenerate sphinx stubs rather than requiring manual update each PR

benclifford:
I'm slightly wary about this PR, because I can't tell which version of sphinx is used by our readthedocs.io setup, and I think we've had some version conflicts there before.

But this removes a chunk of needless ritual when making PRs, so I think it is worth trying.

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
Why didn't we do this sooner!

yadudoc (approved):
Reapproving

PR #2089, opened 2021-06-09
Document WorkqueueExecutor project_name remote reporting better

dthain:
# Description

This PR causes the WQ executor to use a project name of "parsl-wq-{UUID}" by default, unless the user picks another one.  This allow the standard WQ reporting infrastructure to uniquely identify each worklfow separately.

## Type of change

### DIFF
--- a/parsl/configs/wqex_local.py
+++ b/parsl/configs/wqex_local.py
@@ -1,7 +1,7 @@
 from parsl.config import Config
 from parsl.executors import WorkQueueExecutor
 
-import uuid;
+import uuid
 
 config = Config(
     executors=[
@@ -13,7 +13,7 @@ config = Config(
             # which can be viewed here:  http://ccl.cse.nd.edu/software/workqueue/status
 
             # To disable status reporting, comment out the project_name.
-            project_name="parsl-wq-"+str(uuid.uuid4()),
+            project_name="parsl-wq-" + str(uuid.uuid4()),
 
             # The port number that Work Queue will listen on for connecting workers.
             port=9123,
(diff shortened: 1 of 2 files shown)


benclifford:
does this cause parsl to start reporting usage stats to a remote server that is not owned by the user?

dthain:
Yep, this posts telemetry to the global catalog server:
http://catalog.cse.nd.edu:9097

benclifford:
insert tired old #777 dd3eefc2edaebc779d1eb8ad3ca7ceb4b2f94789 debate

dthain:
Aha, that is helpful context.  Well, I think it makes sense for WQ to just follow what Parsl is doing.  How about we activate the default project name only if the user did the top level opt-in?  (And are users actually doing this?)

benclifford:
yeah i think it makes sense to inherit usage tracking behaviour from the parsl config. That might need a tweak to the executor `def start()` API to get the configuration information through.

someone else - @kylechard ?  - could comment on how much info we actually get from parsl-level usage tracking.

dthain:
Hmm, the more I read about it, I think the Parsl reporting and the WQ reporting have somewhat different purposes.  As I understand it, the Parsl reporting is fully anonymous and only for usage tracking by the project. The WQ reporting is not anonymous, but it is made publically visible for the purposes of rendezvous and status monitoring by the user, as well as usage tracking by the project.

So perhaps the better way to address this is via the documentation for the WQ executor: simply state how the reporting works, how they can see/use the information, and request that they turn it on.  How does that sound?

benclifford:
yeah that would be fine too I think

yadudoc (approved):
I think this looks good. As long as it's opt-in and we are clear about what is being reported, we should be safe :)

PR #2092, opened 2021-06-23
Assert that there should be no doc stubs in version control

benclifford:
# Description

PR #2087 autogenerates stubs, but does not stop accidentally committed stubs appearing in version control.
PR #2051 accidentally adds such a stub.

This PR removes that incorrect stub and replaces the old CI stub hygiene test (which checked that stubs were up to date) with a new one (which checks that there are no stubs).

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
This is nice!

PR #2094, opened 2021-06-23
Configure sphinx to put in full documentation for each method

benclifford:
Before this, only a small table of summary information was given for each
class, and per-method docstrings were not presented.

This PR also fixes several errors in docstring text that were previously
not a problem, because those pieces of docstring were not used.

Attached are before/after screenshots showing the extra information:

pre
post

## Type of change

- Documentation update


yadudoc (approved):
Docs definitely look nicer! Rest of the doc changes look reasonable.

PR #2098, opened 2021-06-29
docs: add Flux/LLNL TOSS3 configuration

jameshcorbett:
# Description

Updates the documentation to add a config for LLNL TOSS3 machines. TOSS3 machines are commodity machines with no GPUs (with one or two exceptions). They all run Slurm.

Fixes # (issue)

## Type of change

- Documentation update

### DIFF
--- a/docs/userguide/configuring.rst
+++ b/docs/userguide/configuring.rst
@@ -514,6 +514,19 @@ using the `CobaltProvider`. This configuration assumes that the script is being 
 .. literalinclude:: ../../parsl/configs/theta.py
 
 
+TOSS3 (LLNL)
+------------
+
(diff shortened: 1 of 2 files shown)


benclifford (approved):

jameshcorbett:
I tried to add a config file for LLNL's BlueOS machines (like Sierra, the counterpart to Summit) but I ran into considerable trouble---first building/installing a version of Cryptography that would satisfy `paramiko` and then in statussing with `LSFProvider` which I didn't investigate very closely. I would be happy to investigate more later, but only once there is a real use-case.

yadudoc (approved):
This looks good to merge.

PR #2099, opened 2021-06-30
Pull os x multiprocessing code into a single module

benclifford:
There are a few pieces of os x compatibility code in the parsl codebase which deal with multiprocessing.{Process, Queue}.

This PR attempts to consolidate them in one place.

Most parsl processes need to be launching using multiprocessing fork method, no matter which platform.

This PR introduces ForkProcess, which will always fork, no matter which platform (or fail, when forking is not available).
Because this always forks, there is no need to switch behaviour based on operating system type.

This PR also introduces SizedQueue which is a Queue which can have its size taken. The default linux queue implements this directly, but a wrapper is needed for os x. Using SizedQueue will work on both platforms.

This should also make parsl more resistant to the global multiprocessing process type being set to anything other than fork (which happens by default on os x but is also possible on other platforms).

## Type of change


benclifford:
this needs a conflict resolution with some monitoring type re-arrangements, which i will do now

parsl/monitoring/monitoring.py
```diff
@@ -323,19 +324,24 @@ def wrapped(*args: List[Any], **kwargs: Dict[str, Any]) -> Any:
...
-                            args=(os.getpid(),
-                                  try_id,
-                                  task_id,
-                                  monitoring_hub_url,
-                                  run_id,
-                                  logging_level,
-                                  sleep_dur),
-                            name="Monitor-Wrapper-{}".format(task_id))
-                p.start()
+                pp = ForkProcess(target=monitor,
```
yadudoc, line 330: Isn't the var `pp` redundant here?

yadudoc (approved):
This looks reasonable.

parsl/monitoring/monitoring.py
benclifford: It's to do with an awkwardness in type-checking vs Optional:

p is Optional[Process], which means p might be None, which means p.start() is not valid.

Previously, `mypy` could figure out that p would never be None because it is directly assigned a Process() value, and so p.start() is safe.

With the abstraction of `ForkProcess`, `mypy` can't narrow the type of `p` down any more.

It *can* figure out that `pp` is safe to call .start() on, apparently, or perhaps just that pp is `Any`.

It's a frustrating clash of this python abstraction style vs static checking.

(an alternative here would be to put a type checking override in... I don't really have any preference)

yadudoc: Alright, just wanted to understand better what's going on. This is okay.

PR #2101, opened 2021-07-02
Updating citation to HPDC paper

kylechard:
# Description

Updating the suggested citation.

## Type of change

- Documentation update


danielskatz (commented):
I don't actually see any changes of content here, other than formatting. Do we need this change/PR?

PR #2103, opened 2021-07-13
Balsam LDRD

radiantone:
# Description

Balsam2 executor, singularity/container app types, balsam examples.

Fixes # (issue)

## Type of change

- Documentation update - TBD


benclifford (commented):
@radiantone do you want a review of this with the intention of merging it into master branch parsl?

parsl/app/bash.py
```diff
@@ -108,13 +108,17 @@ def __init__(self, func, data_flow_kernel=None, cache=False, executors='all', ig
...
                 self.kwargs[s] = sig.parameters[s].default
 
-        # update_wrapper allows remote_side_bash_executor to masquerade as self.func
-        # partial is used to attach the first arg the "func" to the remote_side_bash_executor
-        # this is done to avoid passing a function type in the args which parsl.serializer
-        # doesn't support
-        remote_fn = partial(update_wrapper(remote_side_bash_executor, self.func), self.func)
-        remote_fn.__name__ = self.func.__name__
-        self.wrapped_remote_function = wrap_error(remote_fn)
+        if 'BalsamExecutor' not in executors:
```
yadudoc, line 111: I'm concerned that this check would break support for multiple executors within a single run.

radiantone: Elaborate? Maybe give me a test case to implement?

radiantone:
@benclifford Yeah, I think that is the idea. But there are still some pending tweaks I think. So a few more changes will come

benclifford:
@radiantone Here are some broad comments.  I haven't gone through the code line-by-line.

I'll be much harsher on code going into core parsl, rather than going into the balsam executor modules.

## Decorators

1. There are a couple of container decorators introduced. Are these balsam specific or are they intended to be a general container decorator for every parsl executor?

2. There's a BalsamApp class that looks like half of an implemented decorator - is this something that should be deleted? I can't see how it is used otherwise.

## Examples

There are lots of nice examples. They don't look easily runnable by an arbitrary user though.

3. I really would like people to be able to run the `pytest` suite with all of our executors as a basic test of functionality - for example in CI we run four different configurations (involving threads, HighThroughputExecutor, Work Queue). There's a configuration file:  `parsl/config/balsam.py`. Should I expect `pytest parsl/tests/ --config parsl/config/balsam.py` to succeed (if I'm in the right environment)? It looks like the example files have lots of parameters to BalsamExecutor that parsl/config/balsam.py doesn't have.

4. I think a top level examples/ directory is the wrong place for these examples - they aren't overall parsl examples. A few of candidate locations: parsl/tests/integration/balsam/  (the integration directory is ignored by the test runner), `parsl/tests/manual_tests/balsam`,  `docs/userguide/examples/balsam`.

## Unused code?

5. There is a reference to parsl.executors.chrome - what is this?

6. There's a SingularityProvider - I can't see this doing anything singularity specific. I can't see it used in the examples. Is this dead code?

7. There's a BalsamProvider - what does this do?

## Core Special-casing

There's a special casing in the bash app code that specialises behaviour in the case that an executor is labelled `BalsamExeuctor`. I'm massively against putting per-executor special casing into the parsl core, and would like to understand what this is trying to do and explore if there is a non-invasive way to achieve this.

kylechard:
Closing for now as not passing tests and still needs some discussion around the right model for integration.

PR #2104, opened 2021-07-16
Send heartbeats via results connection

### LINKED ISSUE #2040: Parsl stuck after 30 minutes on HTCondor
**Describe the bug**
Using Parsl with HighThroughputExecutor and CondorProvider, Parsl becomes stuck after 30 minutes of no results from a manager.

**To Reproduce**
1. Run the following script with HTCondor available
  <details>
    <summary>Python script</summary>
    

```python
#!/usr/bin/env python3

import os
import parsl
from tqdm import tqdm


def get_parsl_config(num_jobs):
    condor_init = "source {}/environment.sh\n".format(os.environ["HOME"])
    provider = parsl.providers.CondorProvider(
        init_blocks=num_jobs,
        max_blocks=num_jobs,
        worker_init=condor_init
    )
    parsl_executor = parsl.executors.HighThroughputExecutor(
        label="HTCondor",
        address=parsl.addresses.address_by_hostname(),
        max_workers=1,
        provider=provider,
    )
    parsl_config = parsl.config.Config(
        executors=[parsl_executor],
    )
    return parsl_config


@parsl.python_app
def work(sleep):
    import time
    time.sleep(sleep)
    return sleep

parsl_config = get_parsl_config(4)
parsl.load(parsl_config)
futures = [work(i + 28 * 60) for i in range(0, 4 * 60, 60)]
results = []
for future in tqdm(futures):
    results.append(future.result())

print(results)
```
  </details>

The `~/environment.sh` script that is executed here just sets `PATH`, `PYTHONPATH` and related variables to allow executing `process_worker_pool.py` with the needed libraries available on my particular setup.

 2. Wait for the HTCondor jobs to start and wait 30 minutes.
 3. The script will be stuck at 50%. Only the first two `future.result()` calls will return (the ones with 28 and 29 min of sleep), the others won't.
 
I also tested it with with various other sleep times. If the sleep time is below 30 min, it works. If it is above or equal it gets stuck. The HTCondor jobs continue running fine and the workers report having finished all the tasks.

**Expected behavior**
All four calls should return after a waiting time of 31 min and the script should print `[1680, 1740, 1800, 1860]`.

**Environment**
 - OS: CentOS Linux release 7.9.2009
 - Python version 3.6 and 3.8
 - Parsl version 1.0.0 and 1.1.0

**Distributed Environment**
- Where are you running the Parsl script from? Login node
- Where do you need the workers to run? HTCondor cluster nodes that run the same OS and software as the login node

**Additional Information**
From some debugging I did it looks to me like the connection `result_outgoing` of the Manager stops working, but I'm not sure if this is the entire problem.

jrueb:
This makes the results connection not being dropped even if there are no results for a longer time span.

# Description

The PR will make the process_worker_pool of the high_throughput executor send heartbeats via the results connection. Previously heartbeats were only sent via the tasks connection. The timing for these new heartbeats also determined by `heartbeat_period`, just as the tasks heartbeats. This way the results connection will stay active and is not dropped. The results connection being dropped is a large issue, as the manager will stay active via the results heartbeats but will be unable to send results, possibly causing Parsl to hang.

Fixes #2040 

## Type of change


benclifford:
not sure why this hasn't run in our CI

benclifford:
This looks good. It needs to run in CI, though, which for some reason didn't happen. Can you push some trivial commit to your branch to see if that wakes it up?

benclifford:
@jrueb yep that commit seems to have made it start CI. thanks.

benclifford (dismissed):

benclifford (approved):

benclifford:
cool thanks

PR #2078, opened 2021-06-01
Adding github actions for CI

yadudoc:
# Description

This PR adds a github actions workflow template that replicates our current CI setup in Travis.

Fixes #2077 

## Type of change

Choose which options apply, and delete the ones which do not apply.

- New feature (non-breaking change that adds functionality)
- Code maintentance/cleanup

PR #2105, opened 2021-07-28
Fix CI broken by dependency package changes

benclifford:
Recently (since the PR #2078) something has changed in dependency packages which made `pip` unable to find a valid set of package versions to install.

The version pin on `coverage` that is removed by this PR was conflicting with
pytest-cov > 2.10. Previously pip was able to easily find a solution
to that by using an earlier pytest-cov, but as of some recent
ecosystem change, it was unable to find a solution.

When manually trying to install pytest-cov 2.11.0 on my laptop
with a working parsl dependency stack already installed:

```
The conflict is caused by:
    The user requested coverage==4.5.4
    pytest-cov 2.11.0 depends on coverage>=5.2.1
```

When building in CI, this resulted in pip spending "forever" (i.e. longer
than the test timeout) trying to find a solution.

This PR removes that pin of `coverage` -- coverage is only used by pytest-cov so we should trust it to require the correct version.

The consequent version change in the `coverage` package broke coverage checking with `--config local`, and so this PR introduces a `.coveragerc` that appears to fix that problem.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+concurrency=multiprocessing,thread
+parallel=True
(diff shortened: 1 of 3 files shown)


yadudoc (approved):
:+1:

Issue #1878, opened 2020-10-06
Is Parsl on Windows supported?


bilke:
Is Parsl supported on Windows?

**To Reproduce**
Run a simple Parsl workflow (e.g. the one the [Quickstart-guide](https://parsl.readthedocs.io/en/stable/quickstart.html#getting-started)) on Windows:

```
....\.venv\lib\site-packages\parsl\addresses.py", line 14, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

The module `fcntl` [exists on Linux only](https://docs.python.org/3.7/library/fcntl.html).

**Environment**
 - OS: windows
 - Python version 3.7
 - Parsl version 1.0.0

benclifford:
I don't think we can put much effort into supporting Windows. If you're encountering bugs as simple as this, it sounds like no one has cared enough in the past to report this.

I think there will be a bunch of small things like this bug which should be straightforward to fix if you wanted to: nothing in parsl.addresses is strictly required to use parsl, as all that is in there is address autodetection code.

However, there are more serious problems that I can imagine, for example, the high throughput executor being able to run on Windows, because of the way it uses `multiprocessing`. (see https://github.com/Parsl/parsl/issues/1659)

bilke:
Thank you Ben for the response. I perfectly understand that Windows is not an OS you target. But maybe this should be stated somewhere that Windows is simply not supported.

jonoharms:
When importing on windows, I get the error:

>>> import parsl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Applications\Anaconda3\envs\py310\lib\site-packages\parsl\__init__.py", line 22, in <module>
    from parsl.app.app import bash_app, join_app, python_app
  File "C:\Applications\Anaconda3\envs\py310\lib\site-packages\parsl\app\app.py", line 12, in <module>
    from parsl.dataflow.dflow import DataFlowKernel
  File "C:\Applications\Anaconda3\envs\py310\lib\site-packages\parsl\dataflow\dflow.py", line 23, in <module>
    from parsl.config import Config
  File "C:\Applications\Anaconda3\envs\py310\lib\site-packages\parsl\config.py", line 7, in <module>
    from parsl.executors.base import ParslExecutor
  File "C:\Applications\Anaconda3\envs\py310\lib\site-packages\parsl\executors\__init__.py", line 4, in <module>
[...]

PR #2107, opened 2021-07-30
Improve support for Windows

WardLT:
# Description

Changes made in my efforts to get Parsl working minimally on Windows.

- Do not crash when fcntl is not found.

Addresses #1878, though I would not claim I will fix it completely.

## Type of change

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -10,7 +10,10 @@ import logging
 import platform
 import requests
 import socket
-import fcntl
+try:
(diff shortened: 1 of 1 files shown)


benclifford (approved):
I didn't try this on windows myself so I'm hoping you did.

I guess there are windows-specific ways of getting host IP addresses in Python, which could go in `parsl/addresses.py` if/when it turns out the other methods aren't sufficient.

PR #2108, opened 2021-08-02
Fix incorrect order of manager and interchange versions in error text

benclifford:
## Type of change


danielskatz (approved):
not sure how this error got in to begin with

PR #2110, opened 2021-08-18
Reduce the size of recursive fibonacci joinapp testing

benclifford:
This test was using about 1/3rd of the testing time of the
WQ executor - I haven't profiled for other executors -
which is a lot for a fairly obscure feature.

Rough timings on my laptop: before: 100s, after 10s

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Yep, this makes sense.

PR #2111, opened 2021-08-20
Upgrade CI to use recent ubuntu, as old version was deprecated

benclifford:
The Ubuntu version in use in our Github Actions workflows (16.04)
is deprecated by github now. In another project (Data Station), this
has already started causing problems.

Ubuntu 20.04 is the most recent Long Term Support (LTS) ubuntu version
supported by github actions, so this PR switches our CI to use that.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Changes here look good. It's also nice that we get upgraded to py3.8 as the default.

PR #2113, opened 2021-08-28
Allow use of WorkQueue running_time_min resource constraint

benclifford:
# Description

WQ recently introduced an additional resource constraint: workers can be aware of their remaining wall time, and tasks can be constrained to only go to workers with sufficient remaining time.

This has been used with success on the `desc` parsl branch for a few months.

cc. @dthain 

## Type of change

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -59,7 +59,8 @@ logger = logging.getLogger(__name__)
 
 
 # Support structure to communicate parsl tasks to the work queue submit thread.
-ParslTaskToWq = namedtuple('ParslTaskToWq', 'id category cores memory disk gpus priority env_pkg map_file function_file result_file input_files output_files')
+ParslTaskToWq = namedtuple('ParslTaskToWq',
+                           'id category cores memory disk gpus priority running_time_min env_pkg map_file function_file result_file input_files output_files')
 
 # Support structure to communicate final status of work queue tasks to parsl
 # result is only valid if result_received is True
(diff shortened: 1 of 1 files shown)


yadudoc (commented):
@benclifford I'm not clear on how this feature works, so it might be useful to have a chat. The changes here to allow specification makes sense.

It would be useful to document the behavior this option enables along with this PR. For eg. what happens if all tasks need more time than available on workers? Do the workers get their walltimes from the scheduler or from the config? If for whatever reason tasks cannot be accommodated, do we get exceptions?

benclifford:
One way that would be nice for documenting all of the work queue resource specifications would be to link them to the actual work queue documentation, rather than trying to explain them in the parsl documentation.

I think that's probably this? https://cctools.readthedocs.io/en/latest/work_queue/#task-resources  @dthain might have a better option

benclifford:
In the meantime, the in-parsl documentation for resource specifications is outdated and spotty. I think fixing parsl_resource_specification documentation should not delay this PR.

dthain:
Yes, that is a good link for describing WQ resources and what they mean, please go ahead.

However, just thinking about this from the end-user perspective, the code pattern in our documentation (`t.specify_cores(4)`) isn't how resources are expressed in a parsl function definition, so be careful not to send them in the wrong direction...

benclifford:
Notes for documenting:

> For eg. what happens if all tasks need more time than available on workers? 

Work Queue waits for more workers to appear that have the time available.

> Do the workers get their walltimes from the scheduler or from the config?

You have to tell the work queue workers the walltime they should be using for resource allocation on the work queue worker commandline. That walltime might be different than the walltime that you give the scheduler, to accomodate startup costs.

>  If for whatever reason tasks cannot be accommodated, do we get exceptions?

Work Queue doesn't really have a concept of making a final decision about not accomodating tasks: they're delayed until the future when hopefully they can be. If you submit a task that need more min_time than you will ever run workers with, that is going to result in a hang forever.

dthain:
We have been mulling the latter problem lately, and are planning to add a feature that emits a text warning if a given task is unable to match to any worker after a reasonable amount of time.

(This could be due to any sort of resource requirement.  For example, the task might be asking for 8 cores, but all workers have a max of 4 cores.)

As you point out, this is a warning rather than a hard error, because WQ doesn't know the overall intent of the user, and doesn't know if more (larger) workers will connect in the future.  And, it's not suitable as an API.

Another thing we might consider: add a "max queue wait time" to a task, that causes it to be returned to the submitted if it sits in the queue without a match for too long.  (Or possibly a "max overall everything" time.)

FYI @btovar @cpreciad

yadudoc (approved):
Assuming that we'll update the `parsl_resource_specification` docs separately to cover the new options, this is good to go.

PR #2115, opened 2021-08-28
Fix a race condition in the local channel

### LINKED ISSUE #2102: grid engine provider hangs when qstat output is large
**Describe the bug**

From Quentin Le Boulc'h;

```
I still have the issue with the qstat command from Parsl which is hitting the 60s timeout. When I have 1000 jobs running every qstat fails, I can see many processes accumulating and Parsl get completely stucked.
I checked I can reproduce the issue by myself running the subprocess command you are using. With several hundreds of jobs running the proc.wait() is waiting forever. However a manual qstat returns the output in less than a second. And after killing the proc.wait()  I can check that the stdout is actually here so the command has succeeded.
I can also reproduce the issue with a different command than `qstat`, like a simple script returning several hundreds of lines.
Could you check if there is some issue with your subprocess command?
See https://docs.python.org/3/library/subprocess.html#popen-objects for instance: “This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use Popen.communicate() when using pipes to avoid that.”
```

**Expected behavior**
This should not hang

**Environment**
in2p3, parsl desc branch

benclifford:
This was showing up as a hang when the qstat command launched by
the grid_engine provider returned a very large output: the process
would not exit until stdout had been read enough to empty buffers,
but stdout would not be read until the process had been exited.

Python provides `communicate` to deal with exactly this sort of
complexity, so this PR switches to using that.

Found by Quentin Le Boulc'h

Fixes #2102 

## Type of change

### DIFF
--- a/parsl/channels/local/local.py
+++ b/parsl/channels/local/local.py
@@ -68,9 +68,7 @@ class LocalChannel(Channel, RepresentationMixin):
                 shell=True,
                 preexec_fn=os.setpgrp
             )
-            proc.wait(timeout=walltime)
-            stdout = proc.stdout.read()
-            stderr = proc.stderr.read()
+            (stdout, stderr) = proc.communicate(timeout=walltime)
             retcode = proc.returncode
 
         except Exception as e:


yadudoc (approved):
This makes sense!

PR #2118, opened 2021-09-01
Make changes for CI reliability

benclifford:
This CI makes a couple of changes for CI reliability:

Recently CI tests have been taking longer than 40 minutes of walltime, and this PR extends the 40 minute time limit to 60 minutes.

This PR also copies runinfo/ from test runs. This is for debugging ongoing hangs in `--config local` tests that I have been unable to reproduce on my laptop, but which occurred in travis CI before and are still occuring in github actions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,7 +74,7 @@ jobs:
         make coverage
 
     - name: Archive runinfo logs
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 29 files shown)


benclifford:
turns out this PR doesn't archive runinfo/ on test failure, which is specifically what I want it to do - I'm not so interested in runinfo/ on test success. I will fiddle some more.

PR #2119, opened 2021-09-06
Fix htex scale down breakage due to overly aggressive result heartbeat

benclifford:
PR #2104 introduced a result channel heartbeat. However, this heartbeat
resulted in the idle time of managers being updated on each heartbeat.

Because of this, no manager ever appears to be idling for very long.

And because of this, the scale down code will not select any manager
for scaling down.

This PR makes result channel messages not update the idle time if
a manager already has an idle time: a result channel message (either
a result or a heartbeat) now means:
"I am (still) idle", rather than "I am idle starting from now".

There is no CI test for this (before or after). See issue #1885 

## Type of change

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -534,7 +534,7 @@ class Interchange(object):
 
                     self.results_outgoing.send_multipart(b_messages)
                     logger.debug("[MAIN] Current tasks: {}".format(self._ready_manager_queue[manager]['tasks']))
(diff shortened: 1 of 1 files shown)


benclifford:
Quentin Le Boulc'h has tested this fix applied to the `desc` branch and it fixed scaling down for his 1200 block run.

benclifford:
cc @jrueb  - does this change make sense to you?

jrueb:
Looks good to me.

yadudoc (approved):
This looks correct.

Issue #2125, opened 2021-09-23
'partition' argument needed in SlurmProvider although it is documented as optional


QLeB:
Using SlurmProvider without specifying 'partition' returns
`TypeError: missing a required argument: 'partition'`
Although it is supposed to be optional:
`Slurm partition to request blocks from. If none, no partition slurm directive will be specified.`

benclifford:
Which version of parsl are you using? it looks like this was supposed to be fixed in 4a726d1d5a613d0f4641ad07714b1371009c1fe6, issue #518.

benclifford:
oh, I see what's happening.

There are two different meanings of optionality floating around.

If you specify a parameter `partition=None`, then this should work for you.

Probably that documentation should be clarified to indicate that `partition=None` is a required parameter.

QLeB:
That's the latest 'desc' branch. Yes you are right it's working with `partition=None`. Thanks!

PR #2126, opened 2021-09-23
Improve docs and defaults for slurm partition and account parameters.

benclifford:
This adds a default to the partition parameter of None, to make it
behave like the account parameter.

The documentation for partition and account parameters is made a
bit more verbose to distinguish between omitting a parameter and
specifying a python None value.

See issue #2125 

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
Doc lines are clearer now. :+1:

danielskatz (approved):

PR #2129, opened 2021-09-30
Fix type error when job status output is large

benclifford:
When job status output is large, parsl tries to create an abbreviated
summary.

Before this PR, when that code path runs, this error happens:

    head = f.read(JobStatus.SUMMARY_TRUNCATION_THRESHOLD / 2)
TypeError: argument should be integer or None, not 'float'

There is a cross-python behaviour change here: in python 2,
/ 2 returns an int, but in python3 it returns a float.

python 2: 1/2 gives 0
>>> type(1/2)
<type 'int'>

python 3:  1/2 gives 0.5
>>> type(1/2)
<class 'float'>

This PR makes the code round-down to an int.

## Type of change


yadudoc (approved):
Looks safe, since the change is a typecast to int.

PR #2131, opened 2021-09-30
Make monitoring hub exit condition more explicit

benclifford:
Prior to this PR, the monitoring hub decided it was time to exit
by relying on the absence of a python version string in WORKFLOW_INFO
messages that it was sent.

This PR replaces that with an explicit exit field, which makes
the shutdown logic easier to read. It should not change any other
behaviour.

This is part of an ongoing sequence of PRs attempting to simplify monitoring message flow.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Looks reasonable.

PR #2144, opened 2021-11-04
Update CONTRIBUTING.rst

danielskatz:
# Description

trying to make CONTRIBUTING friendlier

## Type of change

- Documentation update

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,11 +1,13 @@
 Where to start
 --------------
 
-We welcome contributions of any type (e.g., bug fixes, new features, reporting issues, documentation, etc).  If you're looking for a good place to get started you might like to peruse our current Git issues (those marked with `help wanted <https://github.com/Parsl/parsl/labels/help%20wanted>`_ are a good place to start).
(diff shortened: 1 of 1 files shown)


danielskatz:
It would also be good to add

* Expected time frame for response
* That we express gratitude to contributors in the language we use
* Who decides whether to accept my contribution? How are decisions about the project made?
* What is the contributor’s responsibility after submitting a PR?
* Is the contribution within the roadmap/vision of the project?

But I'm not sure exactly how to do this

yadudoc (approved):
Adding the slack link in there makes sense. :+1:

yadudoc:
@danielskatz Would be good to cover the items you raised in the parsl meeting.

benclifford:
> Expected time frame for response
> Who decides whether to accept my contribution? How are decisions about the project made?
> What is the contributor’s responsibility after submitting a PR?
> Is the contribution within the roadmap/vision of the project?
> But I'm not sure exactly how to do this

I think attempting to write anything about the above in a contributing document when there isn't consensus on the answers would be a frustrating experience that will result in text that is confusing and wrong. Though they seem valid points that could benefit from consensus building.

PR #2149, opened 2021-11-24
Update LSF provider to make it more friendly for different LSF-based computers

ZhuozhaoLi:
# Description

The current LSF provider is used only on Summit, but can't be used on some other computers that use a slightly different LSF.

For example, the SUSTech LSF-based HPC cluster differs in the following aspects.

(1) It requests resources based on `the number of cores`, rather than the number of nodes `#BSUB -nnodes ${nodes}` on Summit. 
```
#BSUB -n 80 ##number of total cores
#BSUB -R "span[ptile=40]" ##40 cores per node
```

(2) The job submission requires to specify the queue `#BSUB -q short ##queue name`.

(3) It does not have a project specification `#BSUB -P project`.

(4) `Bsub` command requires a redirection symbol `<`, i.e., `bsub < job_script`.

This PR updates the LSF provider to make it more friendly to general LSF providers.
The PR also adds an override option to mpirun launcher to allow binding a manager to each node.

### DIFF
--- a/parsl/providers/lsf/lsf.py
+++ b/parsl/providers/lsf/lsf.py
@@ -75,7 +75,7 @@ class LSFProvider(ClusterProvider, RepresentationMixin):
     bsub_redirection: Bool
         Should a redirection symbol "<" be included when submitting jobs, i.e., Bsub < job_script.
     request_by_nodes: Bool
-        Request by nodes or request by cores per block. 
+        Request by nodes or request by cores per block.
         When this is set to false, nodes_per_block is computed by cores_per_block / cores_per_node.
         Default is True.
     """
--- a/parsl/providers/lsf/lsf.py
+++ b/parsl/providers/lsf/lsf.py
@@ -56,6 +56,8 @@ class LSFProvider(ClusterProvider, RepresentationMixin):
         Walltime requested per block in HH:MM:SS.
     project : str
         Project to which the resources must be charged
+    queue : str
+        Queue to which to submit the job request
     scheduler_options : str
         String to prepend to the #SBATCH blocks in the submit script to the scheduler.
     worker_init : str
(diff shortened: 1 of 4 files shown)


parsl/providers/lsf/lsf.py
```diff
@@ -93,8 +112,26 @@ def __init__(self,
...
+
+        # Update scheduler options
+        self.scheduler_options = scheduler_options + "\n"
+        if project:
+            self.scheduler_options += "#BSUB -P {}\n".format(project)
+        if queue:
+            self.scheduler_options += "#BSUB -q {}\n".format(queue)
+        if request_by_nodes:
+            self.scheduler_options += "#BSUB -nnodes {}\n".format(nodes_per_block)
+        if not request_by_nodes and cores_per_block and cores_per_node:
```
benclifford, line 130: this `if` won't properly lead the user into setting cores_per_block/cores_per_node if they omit it - it will instead mean that neither an -nnodes or -n gets added.

A better structure might be:
if request by noes: 
     set -nnodes
else:
     assert cores_per_block not none, "useful error/hint"
     assert cores_per_node not none, "useful error/hint"
     add -n and -R

parsl/providers/lsf/lsf.py
```diff
@@ -64,11 +72,19 @@ class LSFProvider(ClusterProvider, RepresentationMixin):
         :class:`~parsl.launchers.SrunLauncher`, or
         :class:`~parsl.launchers.AprunLauncher`
     move_files : Optional[Bool]: should files be moved? by default, Parsl will try to move files.
+    bsub_redirection: Bool
+        Should a redirection symbol "<" be included when submitting jobs, i.e., Bsub < job_script.
```
benclifford, line 76: for consideration later - not affecting this review - do most LSF systems what `<` or not?
If LSF generally wants `<`, then the default should be to include `<`.

parsl/providers/lsf/lsf.py
ZhuozhaoLi: Done. I have updated the `if` logic for `request_by_nodes`

benclifford (approved):

PR #2150, opened 2021-11-24
Use proper subclass of Exception when executor is in a bad state.

benclifford:
This follows the general Parsl style, which defines different
exception classes for different errors.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This is a nice improvement.

PR #2151, opened 2021-11-25
Make sshd channel test errors more obvious

benclifford:
I have repeatedly hit this error:

```
    def _start_sshd(config_dir: str):
        server_config, priv_key, port = _init_sshd(config_dir)
        sshd_thread = SSHDThread(server_config)
        sshd_thread.start()
        time.sleep(1.0)
        if not sshd_thread.is_alive():
>           raise Exception('Failed to start sshd: {}'.format(sshd_thread.error))
E           Exception: Failed to start sshd: expected str, bytes or os.PathLike object, not NoneType
```

which does not point clearly to a resolution.

This commit adds in both logging so that pytest --log-cli-level=ERROR will
report stack traces for exceptions such as the above "not NoneType",
and adds an assertion for the common case that is triggering this for me,
which is that sshd cannot be found on the $PATH.

In the case that sshd is not on the $PATH, that error is now reported as:

```
ERROR    parsl.tests.test_providers.test_local_provider:test_local_provider.py:138 SSHDThread exception from run loop
  File "/home/benc/parsl/src/parsl/parsl/tests/test_providers/test_local_provider.py", line 119, in run
    assert sshpath is not None, "can find sshd executable"
AssertionError: can find sshd executable
assert None is not None
```

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/tests/test_providers/test_local_provider.py
+++ b/parsl/tests/test_providers/test_local_provider.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import pathlib
 import pytest
@@ -13,6 +14,8 @@ from parsl.channels import LocalChannel, SSHChannel
 from parsl.launchers import SingleNodeLauncher
 from parsl.providers import LocalProvider
 from parsl.providers.provider_base import JobState
+
+logger = logging.getLogger(__name__)
 
 
 def _run_tests(p: LocalProvider):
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
This makes sense. Hopefully, this can help differentiate cases where the test system is at fault. Would've been nicer if this was a test fixture that skipped the tests.

benclifford:
>  Would've been nicer if this was a test fixture that skipped the tests.

surely by now you know i hate automatic "I detected this test would fail so I didn't test it so it wouldn't fail" logic in test suites.

yadudoc:
You're right, it's better for the test to just fail if the testing env is messed up rather than magically skip it.

PR #2153, opened 2021-11-26
Make missing worker test cleanup DFK at end

benclifford:
Without this, that DFK continues to exist and emit logs while other tests are happening.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
:+1:

PR #2158, opened 2021-12-02
Tidying of Block Status Updates

kommav:
# Description

This change only reports changes to block status (tidiness)

Fixes # (2121)

## Type of change

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -57,7 +57,7 @@ class Database:
     Base = declarative_base()
 
     def __init__(self,
-                 url: str = 'sqlite:///monitoring.db',
+                 url: str = 'sqlite:///runinfomonitoring.db',
                  ):
 
         self.eng = sa.create_engine(url)
@@ -251,7 +251,7 @@ class Database:
 
 class DatabaseManager:
     def __init__(self,
-                 db_url: str = 'sqlite:///monitoring.db',
+                 db_url: str = 'sqlite:///runinfo/monitoring.db',
                  logdir: str = '.',
                  logging_level: int = logging.INFO,
                  batching_interval: float = 1,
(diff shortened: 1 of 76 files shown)


benclifford (approved):

parsl/monitoring/db_manager.py
```diff
@@ -463,17 +463,28 @@ def start(self,
 
                 """
                 BLOCK_INFO messages
-
                 """
                 block_info_messages = self._get_messages_in_batch(self.pending_block_queue)
+                # Creation of dictionary to store previous block messages to avoid duplicate messages
+                UBM = {}  # type: Dict[Any, Any]
```
benclifford, line 469: having this initialisation of `UBM` here means that it is reset to empty on every iteration of the processing loop which begins around line 340. That means that this code will forget everything on each iteration of that loop - so it will forget that it has already added, for example, block 0 in an earlier iteration and add it again.

Probably the right thing to do is move this `UBM = {}` line to before the while loop starting around line 343

parsl/monitoring/db_manager.py
```diff
@@ -463,17 +463,28 @@ def start(self,
...
-                    # block_info_messages is possibly a nested list of dict (at different polling times)
-                    # Each dict refers to the info of a job/block at one polling time
+                    # block_info_messages is a nested list of dict (at different polling times)
+                    # block_msg is a list of messages (each message is a dict)
+                    # Each dict refers to the info of a job / block at one polling time
                     block_messages_to_insert = []  # type: List[Any]
                     for block_msg in block_info_messages:
-                        block_messages_to_insert.extend(block_msg)
+                        for this_msg in block_msg:
+                            this_id = this_msg['block_id']
```
benclifford, line 479: a block ID isn't enough to identify a block in a particular run - block_id is only relative to a particular parsl executor which you can probably find identified in this_msg['executor_label'].

Using just the block ID means that block 0 of one executor and block 0 of another executor will get confused in this code.

You could maybe make this_id a tuple  `(this_msg['executor_label'], this_msg['block_id'])` and index UBM that way.

benclifford (changes requested):

benclifford:
this was implemented in a different way in #2401 so closing

PR #2166, opened 2021-12-07
Upgrade mypy

benclifford:
This is to get generally improved typechecking compared to the previous version.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
:+1:

PR #2167, opened 2021-12-07
Allow join_apps to return multiple Futures

benclifford:
# Description

To make a join_app wait for multiple component apps, return a list of Futures from the join app, instead of a single Future.

Previously, to wait for multiple launched Futures inside an app, it was necessary to make a separate `combine` app which waits for them all the complete.
    
Use of this style of additional trivial app caused awkwardness in monitoring and checkpointing: it often did not have a clear separate identity based on parameters, which made it hard to distinguish "different" `combine` calls from each other in the monitoring and checkpointing databases. This was also boilerplate that seems nicer to not force on the user.

This PR also moves sending monitoring updates more strictly before adding callbacks: in the case of fast executing tasks, a race condition would mean that tasks would complete (and callbacks fire) before handling of the join app completion code finished. This resulted in two exec_done messages, rather than one joining and one exec_done message.

## Type of change


yadudoc (commented):
@benclifford I've gone through the code, but this is fairly tricky.  think it might be best to discuss over a call to make sure I understand this properly.

parsl/dataflow/dflow.py
```diff
@@ -352,21 +355,36 @@ def handle_exec_update(self, task_record, future):
...
+                        task_record['joins'] = joinable
+                        task_record['join_lock'] = threading.Lock()
+                        for inner_future in joinable:
+                            # TODO: typechecking and error setting here - perhaps
+                            # should put this and the one-future case inside a
+                            # try and perform the error handling there in an
+                            # except block? (it would be ok to go joining->failed
+                            # which doesn't happen in the type error case but
+                            # does happen in the joined-tasks fail case)
+                            # For now, this assert will cause a DFK hang
```
yadudoc, line 378: Could we trigger a shutdown if the assert fails? Better to fail loudly than hang.

parsl/dataflow/dflow.py
```diff
@@ -352,21 +355,36 @@ def handle_exec_update(self, task_record, future):
...
+                        task_record['status'] = States.joining
+                        task_record['joins'] = joinable
+                        task_record['join_lock'] = threading.Lock()
+                        joinable.add_done_callback(partial(self.handle_join_update, task_record))
+                    elif isinstance(joinable, list):  # TODO: should this be list or arbitrary iterable?
                         task_record['status'] = States.joining
-                        task_record['joins'] = inner_future
-                        inner_future.add_done_callback(partial(self.handle_join_update, task_record))
+                        task_record['joins'] = joinable
+                        task_record['join_lock'] = threading.Lock()
```
yadudoc, line 370: To avoid a race-condition between adding the callbacks and the processing in handle_join_update, shouldn't we have this set the callback in a locked context?

parsl/dataflow/dflow.py
```diff
@@ -379,38 +397,83 @@ def handle_exec_update(self, task_record, future):
...
+            # now we know each joinable Future is done
+            # so now look for any exceptions
+            e = None
+            if isinstance(joinable, Future):
+                if joinable.exception():
+                    e = joinable.exception()
+            elif isinstance(joinable, list):
+                for future in joinable:
+                    if future.exception():
+                        e = future.exception()
```
yadudoc, line 442: Not sure about this. The futures in the list are ordered, so I wonder if we'll lose info on all exceptions after the first one in the list?

parsl/dataflow/dflow.py
```diff
@@ -379,38 +397,83 @@ def handle_exec_update(self, task_record, future):
...
-            e = inner_app_future.exception()
-            logger.debug("Task {} failed due to failure of inner join future".format(outer_task_id))
-            # We keep the history separately, since the future itself could be
-            # tossed.
-            task_record['fail_history'].append(repr(e))
-            task_record['fail_count'] += 1
-            # no need to update the fail cost because join apps are never
-            # retried
+            if e:
+                logger.debug("Task {} failed due to failure of an inner join future".format(outer_task_id))
```
yadudoc, line 447: I'm a bit worried that this debug log line will keep printing the exception from the first task to fail in the list, instead of presenting new info on each subsequent failure.

parsl/dataflow/dflow.py
```diff
@@ -379,38 +397,83 @@ def handle_exec_update(self, task_record, future):
...
+                task_record['fail_count'] += 1
+                # no need to update the fail cost because join apps are never
+                # retried
 
-            task_record['status'] = States.failed
-            self.tasks_failed_count += 1
-            task_record['time_returned'] = datetime.datetime.now()
-            with task_record['app_fu']._update_lock:
-                task_record['app_fu'].set_exception(e)
+                task_record['status'] = States.failed
```
yadudoc, line 455: I'm not sure about the logic here, it looks like if there are multiple failed joinables the parent task is set multiple times?

yadudoc (commented):
@benclifford  I'm not sure I followed the logic in `handle_join_update`, so I have a few questions. It might also help to add a few tests for cases where there are multiple failures.

parsl/dataflow/dflow.py
benclifford: probably i should figure out proper handling here

parsl/dataflow/dflow.py
benclifford: This was modelled on the behaviour of regular dependency failure handling which also has what I thought was a benign race condition.

However, on trying to reproduce that in regular dependency handling, I don't think that the regular dependency handling is benign - I can get invalid state exceptions. I'll open a separate issue for this, as it is present in master, not just this PR.

I'll also move this PR back to draft state as I'd like that issue resolved first.

benclifford:
I marked this PR as draft because yadu's review of join dependency failure handling has revealed a race condition in the general parsl dependency failure handling, which I have opened an issue on: https://github.com/Parsl/parsl/issues/2227

I would like that issue fixed in a way that also works for this PR, before continuing work on this PR.

benclifford:
This PR will also need updating against #2192 status handling, if/when that is merged

parsl/dataflow/dflow.py
benclifford: I think it's fine to add callbacks outside of a locked context: the callbacks are safe to fire instantly (when one or more joining applications have already completed) because nothing else happens after this except adding more callbacks.

There is no need to wait for all callbacks to be added before letting the first callback fire and handle things.

parsl/dataflow/dflow.py
benclifford: I added a new error, `JoinError`, which looks similar to DependencyError - it captures all of the exceptions and tids for failed joining apps.

parsl/dataflow/dflow.py
benclifford: This line will only fire once - the first time the callback runs with all joining apps completed. After that, callbacks will never get this far because they will exit earlier due to the app no longer being in `joining` state.

parsl/dataflow/dflow.py
benclifford: the bad dependency handling code problem is detailed and fixed here: issue #2227 and PR #2288
I still need to investigate the behaviour for joins in this PR after that fix.

benclifford: due to the join_lock around all of `handle_join_update`, and the early test in `handle_join_update` to end the callback without doing anything if the join_app task is not in `States.joining`, I think this updating behaviour will now only happen once: the first time the callback is called when all joining apps are completed.

On earlier invocations, the callback will exit earlier because of this test:
```
            if isinstance(joinable, list):
                for future in joinable:
                    if not future.done():
                        return
```

and on later invocations because of this test:

```
            if task_record['status'] != States.joining:
                return
```

benclifford:
I've noticed that in some situations, some tasks get two `exec_done` statuses recorded in the monitoring database. I didn't dig into what was going on there but this should be fixed before merge (and maybe make some stronger tests of monitoring database behaviour)

PR #2168, opened 2021-12-07
Dispatch monitoring messages on tag type, instead of queue name

benclifford:
This means monitoring messages are handled based on their
declared tag type, rather than how they are received by the
database manager.

Specifically, RESOURCE_INFO messages are now handled more
like other messages.

This is groundwork for future PRs which will, for example,
allow remote tasks to deliver their resource messages over
htex's existing channels, rather than via UDP.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
I believe this is correct.

PR #2169, opened 2021-12-07
Test large channel data

benclifford:
# Description

Regression test for #2210.
The local channel was hanging if the specified command gave too
much output, due to a race condition between process exiting and
pipes filling up.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
It's not great that we aren't asserting anything like the volume of output. But this looks reasonable since we mostly care about the case where it hangs and timeout is triggered.

PR #2170, opened 2021-12-07
Remove ipp logging hack in PR #204

benclifford:
# Description

This removes an IPP specific logging hack which lives inside the strategy code. IPP is no longer used in the parsl codebase, so the hack should not be necessary. I've been using this since April and haven't seen a problem.

This is part of a series of PRs tidying up logging behaviour.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This was added to manage an unconventional logging setup in IPP, this should be obsolete now that IPP is no longer used in Parsl.

PR #2171, opened 2021-12-07
ENH kill condor jobs at shutdown

beckermr:
# Description

Fixes # (issue)

## Type of change

- Documentation update
- Code maintentance/cleanup

### DIFF
--- a/parsl/providers/condor/condor.py
+++ b/parsl/providers/condor/condor.py
@@ -3,6 +3,7 @@ import os
 import re
 import time
 import typeguard
+import atexit
 
 from parsl.channels import LocalChannel
 from parsl.providers.provider_base import JobState, JobStatus
@@ -132,6 +133,8 @@ class CondorProvider(RepresentationMixin, ClusterProvider):
         self.worker_init = worker_init + '\n'
         self.requirements = requirements
         self.transfer_input_files = transfer_input_files
+
+        atexit.register(_kill_all_jobs, self)
 
     def _status(self):
         """Update the resource dictionary with job statuses."""
@@ -332,6 +335,11 @@ def _chunker(seq, size):
             res = []
     if res:
         yield res
+
+
+def _kill_all_jobs(condor_provider):
+    job_ids = list(condor_provider.resources.keys())
+    condor_provider.cancel(job_ids)
 
 
 if __name__ == "__main__":


benclifford:
this should already be handled by the scaling in code automatically scaling in all of the blocks at dfk cleanup (which fires atexit) - without this PR, do you see attempts at that happening in your parsl.log?

beckermr:
We see the attempts but the scaling code is missing jobs and leaving them running.

beckermr:
I think managers on jobs that don't connect back are left running.

benclifford:
if the scaling in code isn't scaling in some blocks, that sounds like it would be a problem when running other providers too - can you attach a parsl.log to an issue with this happening?

benclifford:
hm - that isn't how it's meant to work - or rather, managers do get told to shut down, but then blocks should also be killed by the provider mechanism, separately, independently of what any processes inside those blocks are doing.

the shutdown code in parsl is fairly error-prone, though, so I wouldn't be surprised if something is going awry there.

esheldon:
Note there were 100 jobs/workers running but it only tried to kill 74, and there were 74 orphaned jobs.

[parsl.log](https://github.com/Parsl/parsl/files/7669697/parsl.log)

esheldon:
sorry, 100 - 74 = 26 orphaned jobs

benclifford:
ok send a log and lets trace whats happening to those.

benclifford:
(a whole parsl.log to start with)

esheldon:
I attached it above

benclifford:
oops, yep, sorry

benclifford:
ok, this is suspiciously similar numbers:

```
Executor HighThroughputExecutor has 8 active tasks, 73/27 running/pending blocks, and 74 connected workers
```

The scaling code thinks there are 27 pending blocks... maybe those are the ones that it isn't trying to kill.

I'm not sure:

i) why those are pending rather than running -- that information should be nothing to do with what the manager process is doing.

Is condor setting failed jobs back to pending in the condor job queue? (if so, that's probably confusing to parsl)

ii) why those blocks aren't being passed to condor_rm - it makes sense that every block should be killed, no matter what state they are in.

I'll have a look through the code to see if anything is obvious there.

benclifford:
ok, I'm wrong -- the scale in code in the high throughput executor only scales in blocks that have had a successful manager registration, from a quick look at 'def scale_in' in parsl/executors/high_throughput/executor.py

If so, that's going to be a problem with any provider, not just condor, and it should be straightforward to reproduce.

cc @yadudoc @ZhuozhaoLi who are closer to this piece of the code than me

beckermr:
Thank you so much @benclifford for looking into this! 

I wonder if this PR should be merged anyways to ensure cleanup if there is a bug or issue somewhere else though? We could add some signal handling as well. The goal is to never orphan a job and having more than once method by which we attempt to kill jobs is useful I think.

PR #2182, opened 2022-01-14
Allocate block ids uniquely

### LINKED ISSUE #2181: Block number re-use
**Describe the bug**
Parsl will reuse block the latest block number, if that block ends and is recreated.
This is confusing when examining logs and monitoring, trying to understand which actual block is being referred to.

This is specifically caused some confusion when looking at logs for a long-standing CI hang in `--config local` tests.

**To Reproduce**
Watch an execution which scales down a block and scales up again.

**Expected behavior**
Blocks should be uniquely numbered - for example by a monotonically increasing counter.

benclifford:
Previously the code around block ID allocation was a bit fuzzy and
it looks like it could sometimes allocate the same block ID several
times in the presence of failure.

This is at least causing confusion in debugging frequent hangs in
CI.

A second ID allocator was used when a block ID had not already been
allocated, for simulating failures. This ID allocation code is changed
in this PR to use the same sequence counter.

Fixes #2181 

## Type of change

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -9,6 +9,7 @@ import parsl  # noqa F401
 from parsl.executors.base import ParslExecutor
 from parsl.executors.errors import ScalingFailed
 from parsl.providers.provider_base import JobStatus, ExecutionProvider, JobState
+from parsl.utils import AtomicIDCounter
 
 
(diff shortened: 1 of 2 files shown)


yadudoc (approved):
This looks correct, but I haven't tested this on my side to confirm.

PR #2183, opened 2022-01-14
Make manager phrasing consistent in log messages.

benclifford:
The before message referred to manager; the after
message referred to worker.

This patch makes them the same - using manager

## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
This is clearer, although we still name the command "HOLD_WORKER". Probably that can be fixed separately.

yadudoc (approved):

PR #2184, opened 2022-01-14
Elaborate JobStatus docstring to include stdout/err info

benclifford:
## Type of change

- Documentation update


yadudoc (commented):
The positional arg: `state` is missing from the docs. The rest of it looks right.

yadudoc (changes requested):
Missing doc line for `state`.

yadudoc (approved):
:+1:  Looks good now!

PR #2190, opened 2022-01-15
Start bash as if it is a login shell

WardLT:
# Description

Needed for the module environment on ThetaGPU to be functional

## Type of change


benclifford:
I'm a bit suspicious about what this overrides from the job environment in favour of the login environment, but maybe it's ok.

benclifford (approved):

Issue #2178, opened 2022-01-12
error in cleanup do to iterating over dictionary that changes size


beckermr:
**Describe the bug**
When parsl is cleaning up, it is iterating through a diction ary that can change size and this is bad. 

```
parsl.executors.high_throughput.interchange.ManagerLost: Task failure due to loss of manager abe0713dddb1 on host spool0757.sdcc.bnl.gov
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/gpfs/mnt/gpfs02/astro/workarea/beckermr/parsl/parsl/dataflow/dflow.py", line 1023, in atexit_cleanup
    self.cleanup()
  File "/gpfs/mnt/gpfs02/astro/workarea/beckermr/parsl/parsl/dataflow/dflow.py", line 1066, in cleanup
    self.log_task_states()
  File "/gpfs/mnt/gpfs02/astro/workarea/beckermr/parsl/parsl/dataflow/dflow.py", line 951, in log_task_states
    for tid in self.tasks:
RuntimeError: dictionary changed size during iteration
```

**To Reproduce**

Make a run fail very badly and it will happen eventually I hope.

**Expected behavior**
I expect not an error.

**Environment**
 - OS: linux
 - Python version: 3.9
 - Parsl version: 1.1.0 but master

**Distributed Environment**
- Where are you running the Parsl script from ? Login node
- Where do you need the workers to run ? condor system

beckermr:
Switching to 

```
for tid in list(self.tasks):
```

will fix this error.

benclifford:
I think there's more awfulness here than just this.

(One of) the reasons the dictionary changes is because tasks can be removed from self.tasks by some code that assumes it is safe to remove code from this dict.

So even if the key list is atomically copied (not sure if list() does things atomically) I think there is still scope for the subsequent lookup of 'self.tasks[tid]' to fail.

```
        for tid in self.tasks:
            keytasks[self.tasks[tid]['status']] += 1
```

Most of this code was originally written when self.tasks[] would only grow, and tasks would never be removed.

#1512 introduced code to remove tasks when they're no longer needed, and there is a long history of bugs like this appearing because of that.

This whole way of counting tasks at the end for some states, and counting them during execution for other states, is pretty ugly:

```
        for tid in self.tasks:
            keytasks[self.tasks[tid]['status']] += 1
        # Fetch from counters since tasks get wiped
        keytasks[States.exec_done] = self.tasks_completed_count
[...]

PR #2192, opened 2022-01-17
Eliminate tasks[id] when reporting final status counts

benclifford:
This replaces unsafe use of self.tasks[] in shutdown along with a handful of ad-hoc state counting variables with a more regular counting mechanism for all task states that never references self.tasks.

see issues #2178 for this particular report, 
and #2014 for overview of tasks[id] elimination.

## Type of change


benclifford:
conflicts with PR #2167 so keeping draft until that's merged

yadudoc (approved):
Okay, this looks correct.

PR #2193, opened 2022-01-17
Make retry tests wait for all launched tasks before finishing

benclifford:
Without this, the tasks continue to run while subsequent local
tests happen. This is at the least confusing in logs - I haven't
checked if it introduces any actual testing problems.

## Type of change

### DIFF
--- a/parsl/tests/test_error_handling/test_retries.py
+++ b/parsl/tests/test_error_handling/test_retries.py
@@ -55,14 +55,15 @@ def test_fail_nowait(numtasks=10):
         fu = sleep_then_fail(sleep_dur=0.1)
         fus.extend([fu])
 
+    # wait for all tasks to complete before ending this test
+    [x.exception() for x in fus]
+
     try:
         [x.result() for x in fus]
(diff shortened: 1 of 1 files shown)


yadudoc (commented):
This test will not wait for all tasks if the assertion fails. It's not clear what exactly the second wait accomplishes.

benclifford:
The remaining tasks continue executing after the test function finishes, and their resulting DFK log traffic is interleaved with subsequent test log traffic in the next DFK in the same python process. This PR adds a barrier to make sure they are all finished before moving onto the next test - `--config local` tests should not leave tasks running after they claim to be finished by returning.

benclifford:
most of this was merged as part of #2205 but yadu's comment on ordering of results still needs fixing here

yadudoc:
@benclifford I hope you don't mind me converting this to draft. At this point, this PR is a placeholder for changes that aren't ready.

yadudoc (approved):
Alright, this looks correct now. Let's merge.

PR #2196, opened 2022-01-20
Make scale in of htex_auto_scale more effective

### LINKED ISSUE #2195: Strategy htex_auto_scale removing blocks far too slow
**Describe the bug**
The "htex_auto_scale" strategy is very slow to remove blocks in case there are more slots than tasks. It removes one block per 5 seconds at most.

**To Reproduce**
1. Run the test script
2. Check parsl.log and wait for the first handful of tasks are completed
3. The logs shows that it will need around a minute after the idle time out to remove the blocks, even though all tasks but one have been finished. Most of the blocks are idle.

**Expected behavior**
All idle blocks are removed as soon as possible. This becomes really important if you are dealing with, say 500 blocks and most of them are idle.

**Environment**
Parsl 1.2.0, Python 3.8

**Test script**
```python
import parsl


local_htex = Config(
    executors=[
        HighThroughputExecutor(
            label="htex_Local",
            worker_debug=True,
            max_workers=1,
            provider=LocalProvider(
                channel=LocalChannel(),
                init_blocks=2,
                max_blocks=20,
            ),
        )
    ],
    strategy="htex_auto_scale",
    max_idletime=3,
)
parsl.load(local_htex)


@python_app
def long():
    import time
    time.sleep(10000)


@python_app
def short():
    import time
    time.sleep(1)


a = long()
b = [short() for i in range(20)]
a.result()
```

jrueb:
# Description

For htex_auto_scale instead of removing only 1 block at most in case there are more slots than tasks, remove try to remove many blocks so that the number of slots matches the number of tasks. This is of course still limited by how many blocks have reached the configured max_idletime.

Fixes #2195

## Type of change

### DIFF
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -262,7 +262,10 @@ class Strategy(object):
                     logger.debug("More slots than tasks")
                     if isinstance(executor, HighThroughputExecutor):
                         if active_blocks > min_blocks:
-                            exec_status.scale_in(1, force=False, max_idletime=self.max_idletime)
+                            exec_status.scale_in(
+                                active_blocks - active_tasks // tasks_per_node // nodes_per_block,
+                                force=False, max_idletime=self.max_idletime
+                            )
 
                 elif strategy_type == 'simple':
                     # skip for simple strategy
--- a/parsl/dataflow/strategy.py
+++ b/parsl/dataflow/strategy.py
@@ -242,8 +242,8 @@ class Strategy(object):
                 # Case 2b
                 else:
                     # logger.debug("Strategy: Case.2b")
-                    excess = math.ceil((active_tasks * parallelism) - active_slots)
-                    excess_blocks = math.ceil(float(excess) / (tasks_per_node * nodes_per_block))
+                    excess_slots = math.ceil((active_tasks * parallelism) - active_slots)
+                    excess_blocks = math.ceil(float(excess_slots) / (tasks_per_node * nodes_per_block))
                     excess_blocks = min(excess_blocks, max_blocks - active_blocks)
                     logger.debug("Requesting {} more blocks".format(excess_blocks))
                     exec_status.scale_out(excess_blocks)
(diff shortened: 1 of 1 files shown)


parsl/dataflow/strategy.py
```diff
@@ -262,7 +262,10 @@ def _general_strategy(self, status_list, tasks, *, strategy_type):
                     logger.debug("More slots than tasks")
                     if isinstance(executor, HighThroughputExecutor):
                         if active_blocks > min_blocks:
-                            exec_status.scale_in(1, force=False, max_idletime=self.max_idletime)
+                            exec_status.scale_in(
+                                active_blocks - active_tasks // tasks_per_node // nodes_per_block,
```
benclifford, line 266: I think this calculation should look similar to the calculation around line 246: both of them arg calculating some kind of "target number of blocks" and it makes me uncomfortable that they don't look exactly the same (eg `ceil`, `min` and `parallelism`)

other than that, this looks like the right thing to be doing.

jrueb:
Adjusted the computation. Indeed I didn't take parallelism and min_blocks into the calculation.

yadudoc (approved):
@jrueb I think the logic here is correct. I have a minor recommendation about variable naming but otherwise, this PR is good to go. Thanks for writing this up! :)

parsl/dataflow/strategy.py
```diff
@@ -262,7 +262,10 @@ def _general_strategy(self, status_list, tasks, *, strategy_type):
                     logger.debug("More slots than tasks")
                     if isinstance(executor, HighThroughputExecutor):
                         if active_blocks > min_blocks:
-                            exec_status.scale_in(1, force=False, max_idletime=self.max_idletime)
+                            excess = math.ceil(active_slots - (active_tasks * parallelism))
```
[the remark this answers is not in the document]
yadudoc: Sure, that isn't a bad idea.

parsl/dataflow/strategy.py
benclifford: this difference in calculation manifested as issue #3696

PR #2198, opened 2022-01-20
More aggressively shut down DFKs in  tests

benclifford:
this PR is to exercise CI with more aggressive shutdown of DFKs during `--config local` testing.

prior to this PR, there was a lot of tangled overlap in logging, making it unclear exactly which part of which test is hanging and when


benclifford:
merged as part of #2205 - closing without explicit merge

PR #2203, opened 2022-01-25
Explicitly shut down htex after bad state test

benclifford:
Without this, the interchange process continues running for the duration
of the pytest lifetime, overlapping with other tests.

This makes CI hangs harder to debug.

# Type of change

- Code maintentance/cleanup


benclifford:
merged as part of #2205 - closing without explicit merge

PR #2204, opened 2022-01-25
Two concurrent fixes to packaging and CI

benclifford:
* Restrict pandas version for parsl-visualization to one that has previously been used

A recently released version of pandas, 1.4, doesn't work with the version of sqlite-alchemy that is installed by `pip install parsl[monitoring]`. This PR restricts the version of pandas to earlier versions, which would have been installed before this week.

Without this, parsl-visualize was failing in CI:

```
ImportError: Pandas requires version '1.4.0' or newer of 'sqlalchemy' (version '1.3.24' currently installed).
```

* Remove coverage checking from CI - this is little used and seems to be the source of a lot of mysterious hangs that have been happening in CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9"]
     runs-on: ubuntu-20.04
-    timeout-minutes: 60
+    timeout-minutes: 30
 
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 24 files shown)


yadudoc (approved):
The pandas version pinning seems reasonable. coverage causing hangs seems very unusual, but we've seen worse. I see no harm in trying.

yadudoc:
Looks like one element in the CI matrix managed to hang.

benclifford:
the content of this PR was merged as part of #2205 - closing this without further merge necessary

PR #2205, opened 2022-01-25
CI fixing megablast

benclifford:
CI was down to less than 50% of builds passing, and then to 100% failure with a new release of pandas.

This PR contains a number of distinct changes which I have tried for fixing and debugging CI, none on which were in themselves completely reliable.

This PR has passed 6 times in a row with no code changes, so the reliability looks like it is massively increased. I still think there is some broken behaviour, as the reason that 6 is only 6 is because the 7th to last test hung, and all that's changed since then is addition of some logging around the suspicious area.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9"]
     runs-on: ubuntu-20.04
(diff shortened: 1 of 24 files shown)


yadudoc (approved):
Timeout change to 30 mins is nice. Removing coverage shouldn't be a problem. Changes to shutdown signature and additional log lines look reasonable.

PR #2208, opened 2022-01-28
Add python 3.10, drop python 3.6

benclifford:
# Description
Python 3.10 was released 2021-10-04  (see https://www.python.org/dev/peps/pep-0619/)
Python 3.6 was end-of-lifed 2021-12-23 (see https://www.python.org/dev/peps/pep-0494/#lifespan)

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
   main-test-suite:
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     runs-on: ubuntu-20.04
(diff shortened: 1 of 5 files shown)


benclifford:
as is usual with python upgrades for parsl, this is failing with pytest internals. I'll make a separate PR to upgrade pytest, which has been bothering me for other uses too.

benclifford:
see #2209 for a pytest update that should merge before further work on this PR

benclifford:
This PR has passed all 4 tests required by the branch: 3.7 .. 3.10, but github is configured to note that it is missing a 3.6 test - which it deliberately is.

yadudoc (approved):
This is reasonable. We should make a note of this in the announcement for next release.

PR #272, opened 2018-05-11
Tests #128

annawoodard:
This PR implements a change in testing philosophy: instead of writing multiple tests to run the same code on different configurations, the configurations and code are now factorized. By default pytest will run all of the tests on all of the configurations in `parsl/parsl/tests/configs`.  To run a specific config, use `pytest tests --config config.py`. Configurations for remote sites which require user-specific options are collected in `user_opts.py`. Any sites that haven't been configured will be skipped. In this way, as you configure sites, you can gradually add to the sites that you test. I removed a lot of redundancy, but running all of the tests on all of the configurations currently takes a long time, and sometimes hangs. For that reason, for now I recommend running `pytest tests --basic`, which only runs local_ipp and local_threads. I will open separate issues about speeding up tests, pruning remaining redundancies, and fixing broken tests and those that do not assert anything (I have skipped these for now; a summary of skipped tests will be printed when you run the tests). 

Fixes #128, fixes #192, fixes #157, fixes #159.

Example output:
```
> ~/ci/parsl/parsl/tests> pytests --basic
==================================================== test session starts =====================================================
platform darwin -- Python 3.6.2, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /Users/awoodard/ci/parsl/parsl/tests, inifile: pytest.ini
plugins: xdist-1.22.2, timeout-1.2.1, ordering-0.5, lazy-fixture-0.4.0, forked-0.2
collected 290 items / 8 skipped

test_aalst_patterns.py ssssssssssssssssssssssssss                                                                      [  8%]
test_cleanup.py .                                                                                                      [  9%]
test_docker_multisite.py s                                                                                             [  9%]
test_factory.py ..                                                                                                     [ 10%]
test_module.py ss                                                                                                      [ 11%]
test_swift.py sss                                                                                                      [ 12%]
test_thread_parallelism.py ss                                                                                          [ 12%]
test_bash_apps/test_apptimeout.py .                                                                                    [ 13%]
test_bash_apps/test_basic.py ..                                                                                        [ 13%]
test_bash_apps/test_error_codes.py ..s.s.                                                                              [ 15%]
test_bash_apps/test_file_bug_1.py .                                                                                    [ 16%]
test_bash_apps/test_multiline.py .                                                                                     [ 16%]
test_bash_apps/test_pipeline.py ..                                                                                     [ 17%]
test_bash_apps/test_stdout.py ..                                                                                       [ 17%]
[...]

PR #2209, opened 2022-01-28
Eliminate pytest nose and upgrade pytest version

benclifford:
# Description

Older pytest was causing problems in a couple of places: python 3.10 support, exaworks development

Newer pytest changes/removes how it deals with `nose` test compatibility; but parsl stopped doing nose tests in #272 in 2018, so this PR removes those vestiges.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,6 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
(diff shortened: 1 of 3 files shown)


yadudoc (approved):
I think this looks right.

PR #2211, opened 2022-01-30
Add type annotations for provider_base.py and fix some bugs there

benclifford:
# Description

Most of these annotations have been brewing in the `benc-mypy` branch for a while. My recent work on exaworks psij, which has much stronger type-checking, has pushed me to want these annotations in mainstream parsl while implementing a parsl provider which interfaces with psij.

Implementing this interface with psij and using these type annotations has revealed a bug when a job status is created with a `None` stdout or stderr: a default None stdout or stderr path causes an exception when accessing the `stdout` or `stderr` property. This is fixed to return `None` in that case.

~~This PR uses some python >=3.7 syntax, so probably can't be merged unless Python 3.6 support is dropped - see PR #2208 - keeping this PR as draft until then~~

## Type of change

- Bugfix

### DIFF
--- a/parsl/providers/provider_base.py
+++ b/parsl/providers/provider_base.py
@@ -29,8 +29,8 @@ class JobStatus(object):
     """Encapsulates a job state together with other details, presently a (error) message"""
     SUMMARY_TRUNCATION_THRESHOLD = 2048
 
-    def __init__(self, state: JobState, message: str = None, exit_code: Optional[int] = None,
-                 stdout_path: str = None, stderr_path: str = None):
+    def __init__(self, state: JobState, message: Optional[str] = None, exit_code: Optional[int] = None,
+                 stdout_path: Optional[str] = None, stderr_path: Optional[str] = None):
         self.state = state
         self.message = message
         self.exit_code = exit_code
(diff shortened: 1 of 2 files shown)


parsl/providers/provider_base.py
```diff
@@ -29,52 +32,54 @@ class JobStatus(object):
...
         return self._read_file(self.stdout_path)
 
     @property
-    def stderr(self):
+    def stderr(self) -> Optional[str]:
         return self._read_file(self.stderr_path)
 
-    def _read_file(self, path):
+    def _read_file(self, path: Optional[str]) -> Optional[str]:
+        if path is None:
```
yadudoc, line 66: This is awkward. Can path be None?

benclifford: it can because JobStatus can be constructed with `None` paths for stderr and stdout - and is so by default.

This test is already present in the similar _read_summary call lower down.

Pretty much all of the `JobStatus` class behaves as if you pass a None path in the constructor, then all of the file related calls will return a None too - eg `stderr()` or `stderr_summary()`.

I'm not trying to move away from that in this PR, but trying to document it in the types and make behaviour more consistent.

PR #2213, opened 2022-01-30
Change docstring example to htex or WQ, rather than IPP/Swift/T

benclifford:
## Type of change

- Documentation update
- Code maintentance/cleanup


yadudoc (approved):
Good to scrub every mention of deprecated components :)

PR #2221, opened 2022-02-02
Allow configurable `-l select` options in PBSPro provider

### LINKED ISSUE #2220: PBSPro scheduler options ignored
**Describe the bug**
The PBSPro provider template adds `${scheduler_options}` to the bottom of the list of #PBS commands. However, PBSPro seems to only consider the first #PBS option and ignore any later competing ones. This means specifying a new select option with gpus is superseded by the default one and ignored. We can resolve this by moving the user defined scheduler options to the top of the #PBS list in the template.

**To Reproduce**
Use the PBSPro provider (e.g., at ALCF's edge testbed for Polaris) and try to specify a new select option:`-l select=1:ncpus=32:ngpus=1`. These options are ignored in favor of the default specified above in the template.

**Expected behavior**
I would like scheduler_options to be put first so they can be used to enter specific options.

**Environment**
 - Parsl 1.2.0

**Distributed Environment**
- Where are you running the Parsl script from ? ALCF
- Where do you need the workers to run ? Compute nodes

ryanchard:
# Description

Moves the PBSPro scheduler option to the top of the template. 

Fixes #2220 

## Type of change

### DIFF
--- a/parsl/providers/pbspro/pbspro.py
+++ b/parsl/providers/pbspro/pbspro.py
@@ -43,6 +43,9 @@ class PBSProProvider(TorqueProvider):
         Walltime requested per block in HH:MM:SS.
     scheduler_options : str
         String to prepend to the #PBS blocks in the submit script to the scheduler.
+    select_options : str
+        String to append to the #PBS -l select block in the submit script to the scheduler. This can be used to
+        specify ngpus.
     worker_init : str
         Command to be run before starting a worker, such as 'module load Anaconda; source activate env'.
     launcher : Launcher
@@ -54,6 +57,7 @@ class PBSProProvider(TorqueProvider):
                  account=None,
                  queue=None,
                  scheduler_options='',
+                 select_options='',
                  worker_init='',
                  nodes_per_block=1,
                  cpus_per_node=1,
@@ -81,6 +85,7 @@ class PBSProProvider(TorqueProvider):
         self.template_string = template_string
         self._label = 'pbspro'
         self.cpus_per_node = cpus_per_node
+        self.select_options = select_options
 
     def submit(self, command, tasks_per_node, job_name="parsl"):
         """Submits the command job.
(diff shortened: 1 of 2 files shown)


benclifford:
Does this override this line entirely:

> #PBS -l select=${nodes_per_block}:ncpus=${ncpus}

?

and if so, maybe there should be a parameter to allow additional selects to be added on this line...

yadudoc (commented):
This bothers me endlessly. Some implementations/installations use different ordering, and we might be better served by placing the options both at the top and bottom assuming one would be ignored. Alternatively we could do something like `scheduler_opt_placement = {'top', 'bottom'}`. 

@benclifford what do you think?

yadudoc:
I'm okay with going with this now, since the only supported/tested target site is polaris.

benclifford:
@yadudoc is it `-l select` overrides that you are doing or other PBS overrides?

If pre and postfix behaviour genuinely is different per site, then a scheduler pre options and scheduler post options which need to be configured according to how your site wants things is one way to do it - and, the `scheduler_options` field as a single field does not make sense, as something that magically will be pasted in the "right" place.

benclifford:
of the fields that the pbs provider sets, `-l select` is the only one i'd expect someone be expecting to override with their own behaviour, right?

ryanchard:
Good question on whether that is the only field people would need to override. I originally assumed there would be others, but perhaps not. Anything else you'd want to customize probably won't need to overwrite an existing, non-configurable field in the template. In my case I just I want to specify GPUs with an addition of `:ngpus=2` to the end of the select -l statement. An option to extend the existing select field would also work.

benclifford:
@ryanchard what i think should happen here is that the pbs provider gets a new option for setting/overriding the `-l select` field, rather than hoping (unreliably) that placing a second `-l select` elsewhere in the job definition will work.

parsl/providers/pbspro/pbspro.py
```diff
@@ -119,6 +124,12 @@ def submit(self, command, tasks_per_node, job_name="parsl"):
         job_config["worker_init"] = self.worker_init
         job_config["user_script"] = command
 
+        # Add a colon to select_options if one isn't included
+        if self.select_options and not self.select_options.startswith(":"):
```
benclifford, line 128: if I put a `:` on the front, this silently ignores the options rather than erroring?

benclifford:
btw 

> I'm okay with going with this now, since the only supported/tested target site is polaris.

@yadudoc i have other users using pbspro that are not polaris

parsl/providers/pbspro/pbspro.py
benclifford: oh ignore this i misunderstood

PR #2222, opened 2022-02-03
Update hook advice in CONTRIBUTING.md

benclifford:
This takes into account that additional cheap linting is available
with mypy and the __init__ directory checker script, and that tests
are no longer invoked with nose.

## Type of change

- Documentation update


yadudoc (approved):
Looks reasonable.

PR #2224, opened 2022-02-07
Correct misleading docstring on confusing hub port vs hub port range

benclifford:
Although the naming and previous docstring suggest that the two parameters
work together to define a single port which either is specified explicitly
or comes from the supplied range, this is not true.

Two listening ports are used:

* one for the UDP receiver for message from workers

and

* one for ZMQ queue from from the htex interchange (or any other executor
  that wants it)

The UDP port is either `hub_port` or an arbitrary system-allocated port,
ignoring hub_port_range.

The ZMQ queue port is always chosen from `hub_port_range`, which has a
default value so a user is not required to specify it. `hub_port` is never used
for this port.

## Type of change

- Documentation update

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -143,10 +143,17 @@ class MonitoringHub(RepresentationMixin):
         hub_address : str
              The ip address at which the workers will be able to reach the Hub.
         hub_port : int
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
:+1:

PR #2225, opened 2022-02-08
adds wraps to monitoring so functions have their names back

tphung3:
# Description

Callables lose their name attributes when monitoring is turned on as the wrapper function for monitoring doesn't wrap callables correctly. More details can be found in this [issue](https://github.com/cooperative-computing-lab/cctools/issues/2813). Issue is now fixed using wrap decorator from functools module.

## Type of change

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -6,6 +6,7 @@ import time
 import typeguard
 import datetime
 import zmq
+from functools import wraps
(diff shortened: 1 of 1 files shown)


danielskatz:
Thanks for the contribution @tphung3 - it's great to see new people joining the community!

tphung3:
No problem. Thanks for merging my PR!

PR #2226, opened 2022-02-09
Adds support for specifying callables' categories in WorkQueue executor

tphung3:
# Description

Before callables when submitted to WorkQueue (WQ) executor are categorized using their name attribute. Now callables can explicitly set their categories before submitted to WQ.

Fixes [this issue](https://github.com/cooperative-computing-lab/cctools/issues/2813).

## Type of change


yadudoc (approved):
This looks reasonable.

PR #2235, opened 2022-03-28
Upgrade Sphinx to latest

benclifford:
This is motivated by a failure in CI with the previous version,
vs Jinja2.

See: https://github.com/sphinx-doc/sphinx/issues/10291

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -13,6 +13,6 @@ types-python-dateutil
 types-requests
 types-six
 sqlalchemy-stubs
-Sphinx==3.4.1
+Sphinx==4.5.0
 twine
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
The only odd bit is setting the label to "undefined". I think that's okay. Everything else here looks good to go.

benclifford:
> The only odd bit is setting the label to "undefined". I think that's okay. Everything else here looks good to go.

sphinx wants label to be defined, because elsewhere the documentation references label being defined.

so it's a bit of "sphinx seems stricter than the python runtime"

PR #2238, opened 2022-03-31
Fix type ambiguity between executor object and executor label in scaling failed, along with TODOs

benclifford:
I've seen the ambiguity that lead to this in the wild with LSST - naming a provider when an executor should have been named, in a ScalingFailed error.

## Type of change


yadudoc (approved):
This looks correct. Although we don't enforce setting of executor.label, at least there's a default from the base.

PR #2241, opened 2022-03-31
Tidy provider.submit() failure errors

benclifford:
Use logger, not print

Do not use !

Refer to submit, not scale out

## Type of change


yadudoc (changes requested):
This wording seems redundant `Submission of command to submit failed`. Maybe something like `Job submission to {self=Provider} failed` is clearer?

Replacing prints with log lines is all good.

PR #2243, opened 2022-03-31
Make str of BashExitFailure be clearer about fields

benclifford:
Before:

('calibrate', 1)

After:

'bash_app calibrate failed with unix exit code 1'

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This is definitely cleaner!

PR #2246, opened 2022-04-03
syntax level tidying in cluster provider write_script

benclifford:
## Type of change

- Code maintentance/cleanup


yadudoc (approved):
I'm confused as to why it was written this way. The fix looks correct.

PR #2247, opened 2022-04-03
Remove monitoring CPU percentages, which are broken

### LINKED ISSUE #2206: CPU percent monitoring values are broken
**Describe the bug**
The task resource monitoring code attempts to report CPU percentage. This doesn't work properly and usually reports 0.

This code at least does not pay attention to the CPU percent values of child processes, unlike other CPU-time related values.

This code uses `psutil` cpu percent code which reports on accumulated system and user time since it was last invoked. 
Resource monitoring in parsl master is not expected to report all values because it transmits that data over UDP. This lossy 
method of transmitting values is an argument for reporting cumulative values, not derived differences between potentially missing sample points.

Accumulated system and user time is already reported in monitoring, and those more raw metric are more resilient to lost UDP packets: their values still correctly accumulate. 

My opinion is that the cpu percentage fields be removed entirely, and any "fraction of resource usage" values be calculated as a derived value on the query side.

**Expected behavior**
More clearly defined cumulative metrics that survive loss of UDP and include child processes.

benclifford:
Fixes #2206

## Type of change

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -224,8 +224,6 @@ class Database:
             'resource_monitoring_interval', Float, nullable=True)
         psutil_process_pid = Column(
             'psutil_process_pid', Integer, nullable=True)
-        psutil_process_cpu_percent = Column(
-            'psutil_process_cpu_percent', Float, nullable=True)
         psutil_process_memory_percent = Column(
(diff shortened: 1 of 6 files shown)


yadudoc (approved):
I've only looked closely at the changes to monitoring.py and the changes there look correct. The remainder looks safe

PR #2248, opened 2022-04-03
Rename tasks/app to tries/app to reflect retries

benclifford:
This plot counts tries, not tasks.

## Type of change


yadudoc (approved):
Looks reasonable.

Issue #2252, opened 2022-04-22
Exception in checkpoint restore test in CI


benclifford:
**Describe the bug**

I have seen this once in CI, when merging an apparently unrelated PR. 

This is in the code that tests restoring from checkpoints, in python 3.10, merging into `master` 9d7ff034001118cc26e1b1940972ac51a7863484

parsl/tests/test_checkpointing/test_python_checkpoint_2.py 

It looks like a code path is taken that believes that a value is checkpointed, but then later on that checkpoint is missing from a different structure.

I'm noting here in case it comes up again, so we can track it.

```
parsl/tests/test_checkpointing/test_python_checkpoint_2.py:21: in launch_n_random
    return [i.result() for i in d]
parsl/tests/test_checkpointing/test_python_checkpoint_2.py:21: in <listcomp>
    return [i.result() for i in d]
/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/concurrent/futures/_base.py:439: in result
    return self.__get_result()
/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/concurrent/futures/_base.py:391: in __get_result
    raise self._exception
parsl/dataflow/dflow.py:284: in handle_exec_update
    res = self._unwrap_remote_exception_wrapper(future)
parsl/dataflow/dflow.py:469: in _unwrap_remote_exception_wrapper
    result = future.result()
/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/concurrent/futures/_base.py:439: in result
    return self.__get_result()
/opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/concurrent/futures/_base.py:391: in __get_result
    raise self._exception
parsl/dataflow/dflow.py:1165: in checkpoint
    r = self.memoizer.hash_lookup(hashsum).result()
```

**To Reproduce**

I haven't been able to reproduce this.

**Expected behavior**
The test should pass

**Environment**
CI

benclifford:
I can reliably recreate this race condition by adding in the following two sleep statements:

```
diff --git a/parsl/dataflow/dflow.py b/parsl/dataflow/dflow.py
index 873e3be674..210823b082 100644
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -483,6 +483,8 @@ class DataFlowKernel(object):
 
         """
 
+        time.sleep(15)
+
         task_id = task_record['id']
 
[...]

PR #2253, opened 2022-04-22
Rework checkpointing to avoid race conditions and incorrectly checkpointed exceptions

benclifford:
Although mostly this doesn't happen, in the case of issue #2252,
a KeyError was being raised by hash_lookup, and this was being
placed into the checkpoint result rather than propagated.

This PR reworks that code to not use hash_lookup at all, and
remove that entire exception code path.

This does not address #2252 but makes the behaviour a bit more
consistent.

## Type of change

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -1158,15 +1158,11 @@ class DataFlowKernel(object):
                         t = {'hash': hashsum,
                              'exception': None,
                              'result': None}
-                        try:
-                            # Asking for the result will raise an exception if
-                            # the app had failed. Should we even checkpoint these?
-                            # TODO : Resolve this question ?
-                            r = self.memoizer.hash_lookup(hashsum).result()
-                        except Exception as e:
-                            t['exception'] = e
-                        else:
-                            t['result'] = r
+
+                        fut = self.memoizer.hash_lookup(hashsum)
+                        assert fut.done()
+                        assert fut.exception() is None
+                        t['result'] = fut.result()
 
                         # We are using pickle here since pickle dumps to a file in 'ab'
                         # mode behave like a incremental log.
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -1158,15 +1158,11 @@ def checkpoint(self, tasks=None):
...
-                            # TODO : Resolve this question ?
-                            r = self.memoizer.hash_lookup(hashsum).result()
-                        except Exception as e:
-                            t['exception'] = e
-                        else:
-                            t['result'] = r
+
+                        fut = self.memoizer.hash_lookup(hashsum)
+                        assert fut.done()
+                        assert fut.exception() is None
```
yadudoc, line 1164: Aren't we already checking this in the condition above on line 1152?

yadudoc (commented):
I agree that we shouldn't checkpoint exceptions. It looks like there's checks on that already in the code, so I'm not sure what the assertion adds.

parsl/dataflow/dflow.py
benclifford: The test on line 1152 checks  self.tasks[task_id]['app_fu'], not self.memoizer.hash_lookup(hashsum)

Is the result of the hash_lookup meant to be the same as self.tasks[task_id]['app_fu'] ? I think probably so?
 
In which case that hash_lookup call is probably unnecessary?

benclifford: (i.e. the task record should be the source of truth here, not the hash lookup table?)

benclifford:
I rearranged this code to use the task record as source of truth, instead of expecting the value to have been placed in the memo lookup table. This should eliminate the race condition in #2252, which my first commits in this PR did not do.

benclifford:
I still need to test that against my race-condition-provoking patch in #2252, though

benclifford:
I have validated this PR against the reproducible failure case in #2252

PR #2259, opened 2022-05-12
only query active slurm jobs

### LINKED ISSUE #1588: slurm provider can lose track of blocks if squeue polling isn't fast enough, leading to parsl hang
**Describe the bug**
The slurm provider can lose track of slurm jobs/blocks in certain circumstances. It will not report any status change for those blocks, and so if a block has finished but the current recorded status is that the block is running, then the block will be reported as running forever.
This can result in parsl hanging forever, reporting one block available, as it believes it has the ability to make progress when it does not, like this:

```
2020-03-03 06:57:54.444 parsl.dataflow.strategy:199 [DEBUG]  Executor worker-nodes has 109 active tasks, 1/0 running/pending blocks, and 0 connected workers
...
2020-03-03 08:48:29 parsl.dataflow.strategy:199 [DEBUG]  Executor worker-nodes has 109 active tasks, 1/0 running/pending blocks, and 0 connected workers
```

There is a race condition between the provider polling for jobs with `squeue` and slurm moving jobs from `C` (completed) state to not being known/listed.

If that move from `C` to non-existent happens before `slurm._status()` successfully polls that job to discover completion, then further polls with `squeue` will exit with non-zero unix exit code, and this will be ignored silently by `slurm.py` as a slurm failure, forever, rather than as a job completion.

It looks like this may have been introduced around 37250848b8bbf6bdbed3b3ec2f54d5482b79ba5f

Here is an example of such a poll:
```
bxc@cori02:~> squeue --job 28509629
slurm_load_jobs error: Invalid job id specified
bxc@cori02:~> echo $?
1
```

Jobs which are executing will die with `ManagerLost` and cause progress, but jobs which are queued for that executor but waiting to be executed will sit, waiting, forever.

**To Reproduce**
This needs a delay in polling long enough for the completed job to disappear from the slurm queue.

**Expected behavior**
parsl should never hang without making progress.

at the very least, failures in executing slurm commands should be logged as WARNINGs rather than silently ignored.

**Environment**
cori. `lsst-dm-202002` branch of parsl.

aclowes:
# Description

When squeue is called with at least one job ID which is active, it returns successfully. If it is called with only jobs that have already completed, it returns an error. This situation happens if no blocks are currently running.

The same logic (checking if the existing status is terminal) is used in the [local](https://github.com/Parsl/parsl/blob/master/parsl/providers/local/local.py#L77) and [kubernetes](https://github.com/Parsl/parsl/blob/master/parsl/providers/kubernetes/kube.py#L208) providers.

Fixes #1588

## Type of change

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -138,15 +138,21 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         Returns:
               [status...] : Status list of all jobs
         '''
-        job_id_list = ','.join(self.resources.keys())
+        job_id_list = ','.join(
+            [jid for jid, job in self.resources.keys() if not job['status'].terminal]
+        )
+        if not job_id_list:
+            logger.debug('No active jobs, skipping status update')
+            return
+
         cmd = "squeue --job {0}".format(job_id_list)
-        logger.debug("Executing sqeueue")
+        logger.debug("Executing %s", cmd)
         retcode, stdout, stderr = self.execute_wait(cmd)
-        logger.debug("sqeueue returned")
+        logger.debug("sqeueue returned %s %s", stdout, stderr)
 
         # Execute_wait failed. Do no update
         if retcode != 0:
-            logger.warning("squeue failed with non-zero exit code {} - see https://github.com/Parsl/parsl/issues/1588".format(retcode))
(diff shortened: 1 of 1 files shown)


benclifford (approved):

benclifford:
This is definitely a better thing to do.
I have not convinced myself that it does fix the race condition in #1588 though.

PR #2261, opened 2022-05-18
Performance fix for slow manager capacity adverts

yadudoc:
# Description

With long-running tasks and low activity on the HTEX manager, the poller period updates upwards on the manager.
While the task_puller thread does increasing longer polls for tasks, it also takes longer to send out capacity advertisements.
This leads to delays in getting new tasks out. This PR resets the poll period on result activity as well by having the 
result_pusher thread signal the task_puller thread to wake it from long-polls.

Adds a new zmq messaging based `SignalSender` and `SignalReceiver` class to HTEX/zmq_pipes that allows the result pusher thread to wake up  the task puller thread from a long poll. 

Fixes # (issue)

Perf issues reported by @WardLT 

## Type of change


WardLT:
What do we need to merge this?

benclifford:
After poking at this with @yadudoc we are mostly of the opinion that we can add the extra messaging in this PR, and make a further simplificiation in the interchange code, so that the interchange keeps the right number of tasks delivered to each manager, without needing the manager to request them each time. (instead, effectively, a manager sending a result message will trigger the next task to be sent).

The current code I'm looking at is here:

```
                    real_capacity = min(m['free_capacity'],
                                        m['max_capacity'] - tasks_inflight)
```

which sends to a manager the minimum of the number of tasks it requested, and the number of tasks the interchange knows the manager can cope with.

I am proposing changing this and always sending `m['max_capacity'] - tasks_inflight` without any communication from the manager about its current free_capacity.

benclifford:
Closing in favour of #2588 which attempts to solve the same root problem, but by removing code rather than adding code.

PR #2263, opened 2022-05-24
Fix slurm provider type error broken by PR #2259

benclifford:
PR #2259 introduced a type error which was not found by CI as we have no slurm testing in CI.

cc @aclowes 

## Type of change


aclowes:
@benclifford sorry about that, thanks for fixing it!

yadudoc (approved):
Looks good to me.

PR #2265, opened 2022-05-25
Transfer chained exception from remote

jrueb:
# Description

Python does not pickle an exception chain and so this needs to be handled by hand. Do this by saving the cause exception as another `RemoteExceptionWrapper` object.

Chained exceptions allow raising a more specific exception with additional information while still keeping the entire information of the initial exception (like `raise SpecificException("additional information") from e`). However if this happens remotely with Parsl, the original exception is lost. With this change, the entire exception with all its causes are transferred.

Example script:
```python
import parsl

local_htex = Config(
    executors=[
        HighThroughputExecutor(
            label="htex_Local",
            worker_debug=True,
            cores_per_worker=1,
            provider=LocalProvider(
                channel=LocalChannel(),
                init_blocks=1,
                max_blocks=1,
            ),
        )
    ],
    strategy=None,
)

parsl.load(local_htex)


@python_app
def raise_errors():
    print("Hello")
    try:
        raise ValueError("error1")
    except ValueError as e:
        raise RuntimeError("error2") from e


raise_errors().result()
```
Output before:
```
  File "htex_parsl.py", line 36, in <module>
    raise_errors().result()
[...]
  File "htex_parsl.py", line 33, in raise_errors
    raise RuntimeError("error2") from e
RuntimeError: error2
```
Output now:
```
  File "htex_parsl.py", line 31, in raise_errors
    raise ValueError("error1")
ValueError: error1

  File "htex_parsl.py", line 36, in <module>
[...]
  File "htex_parsl.py", line 33, in raise_errors
    raise RuntimeError("error2") from e
RuntimeError: error2
```

## Type of change

### DIFF
--- a/parsl/app/errors.py
+++ b/parsl/app/errors.py
@@ -1,6 +1,6 @@
 """Exceptions raised by Apps."""
 from functools import wraps
-from typing import Callable, List, Union, Any, TypeVar
+from typing import Callable, List, Union, Any, TypeVar, Optional
 from types import TracebackType
 
 import dill
@@ -112,11 +112,11 @@ class BadStdStreamFile(ParslError):
 
 
 class RemoteExceptionWrapper:
-    def __init__(self, e_type: type, e_value: Exception, traceback: TracebackType) -> None:
+    def __init__(self, e_type: type, e_value: BaseException, traceback: Optional[TracebackType]) -> None:
 
         self.e_type = dill.dumps(e_type)
         self.e_value = dill.dumps(e_value)
-        self.e_traceback = Traceback(traceback)
+        self.e_traceback = None if traceback is None else Traceback(traceback)
         if e_value.__cause__ is None:
             self.cause = None
         else:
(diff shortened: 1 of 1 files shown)


jrueb:
Last commit was mainly to satisfy mypy. It was complaining about the `traceback` parameter in the `__init__` of `RemoteExceptionWrapper` not being of optional type, as an exception in general is not required to have a traceback.

benclifford (approved):

parsl/app/errors.py
```diff
@@ -128,10 +133,19 @@ def reraise(self) -> None:
...
         logger.debug("Reraising exception of type {}".format(t))
 
-        v = dill.loads(self.e_value)
-        tb = self.e_traceback.as_traceback()
+        v = self.get_exception()
+
+        reraise(t, v, v.__traceback__)
 
-        reraise(t, v, tb)
+    def get_exception(self) -> Exception:
```
benclifford, line 140: the return value of this can be a BaseException, I think? (because __init__ takes a BaseException, so e_value would contain a serialised BaseException).

This doesn't cause a type error here because dill.dumps/loads doesn't preserve types, but in some other work I've done that causes a problem.

jrueb: Yes, you are right, BaseException would have been correct.

PR #2287, opened 2022-06-10
Log full exception at rundir creation failure

benclifford:
## Type of change


yadudoc (approved):
This makes sense :+1:

PR #2289, opened 2022-06-16
Introduce a running_ended state

benclifford:
## Type of change


benclifford:
This is failing in CI at `parsl-visualize` in a way that I can't recreate on my laptop. I will continue debugging.

benclifford:
Broken parsl-visualize was a separate environment related problem fixed in #2290

PR #2293, opened 2022-06-17
slurm: have squeue be explicit about the format

PaulPrice:
# Description

If the user has SQUEUE_FORMAT set, the results from squeue
might not be in the expected format. Changed the code to be
explicit about this format from squeue. This also allows a
briefer format with simpler parsing because we can dump the
header and columns we don't care about.

## Type of change


yadudoc (approved):
Hi @PaulPrice, Thanks for these code changes. Requesting only the job id and state code explicitly is a good improvement.

benclifford:
i've just tested this out on cori and it works for me

PR #2294, opened 2022-06-17
Add option to specify stream for stream logger

ehusby:
# Description

Adds a `stream` keyword argument to `set_stream_logger` function so that the user can specify whether the `StreamHandler` directs to `sys.stderr` or `sys.stdout`.
For my work on a TACC Frontera project, I want this stream to be directed to `sys.stdout` instead of the default `sys.stderr` for the instantiated `StreamHandler`.

## Type of change

### DIFF
--- a/parsl/executors/extreme_scale/mpi_worker_pool.py
+++ b/parsl/executors/extreme_scale/mpi_worker_pool.py
@@ -439,13 +439,15 @@ def start_file_logger(filename, rank, name='parsl', level=logging.DEBUG, format_
     logger.addHandler(handler)
 
 
-def set_stream_logger(name='parsl', level=logging.DEBUG, format_string=None):
+def set_stream_logger(name='parsl', level=logging.DEBUG, format_string=None, stream=None):
     """Add a stream log handler.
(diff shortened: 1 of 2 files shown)


benclifford:
This looks good.

From a quick look at the code, it looks like only the parsl logutils version of set_stream_logger is used - and the mpi_worker_pool.py version could be removed entirely. Does that match your understanding?

ehusby:
@benclifford Yes, that does match. I see where I fumbled a bit of the PR description calling out only `mpi_worker_pool.py` so that was confusing. Just adjusted the PR description to fix that!

I don't see any references to the mpi_worker_pool version in the master branch, and only a change to the log_utils version is necessary to me. If there's no development needing the mpi_worker_pool version, then perhaps that can be removed as you say.

benclifford (approved):

benclifford:
I tested this out on the commandline to check this really does change where things go.

benclifford:
@ehusby I have removed set_stream_logger from mpi_worker_pool.py in PR #2299

PR #2297, opened 2022-06-22
Relax sqlalchemy version constraint

### LINKED ISSUE #2296: parsl[monitoring] pins sqlalchemy to <1.4
**Describe the bug**
Parsl[monitoring] currently requires sqlalchemy <1.4, making it difficult to integrate with other packages that depend on newer versions.

This issue is related to #1994, solved in fd8072f7, where it was found that sqlalchemy-utils 0.36.8 was incompatible with squalchemy 1.4.0.

However the newest sqlalchemy-utils 0.38.2 does not seem to break with sqlalchemy 1.4.37. Indeed,

`pytest parsl/tests/test_monitoring/ --config local`

succeeds with 4/4 passed.

**To Reproduce**
n/a

**Expected behavior**
Parsl[monitoring] should depend on the newest version that does not explicitly break functionality

**Environment**
- OS: `Linux 5.17.9-arch1-1 #1 SMP PREEMPT Wed, 18 May 2022 17:30:11 +0000 x86_64 GNU/Linux`
- Python 3.10.4
 - Parsl 1.2.0

**Distributed Environment**
- Where are you running the Parsl script from?
 Currently testing on a local machine. Will deploy on a cluster as part of a larger service whose web interface could benefit from newer sqlalchemy

- Where do you need the workers to run ?
 Workers will run mostly on a slurm cluster, although the issue is relative to the master only, the only one to have the monitoring

jacopo-exact:
Relax sqlalchemy version constraints.

# Description

As newest versions of sqlalchemy appear to be compatible with sqlalchemy-utils and with the monitoring subsystem, this PR relaxes the constraint on sqlalchemy.

Fixes #2296

## Type of change

- Code maintentance/cleanup: dependency versioning


jacopo-exact:
I removed the problem at its root. With the last commit there is no more dependency on sqlalchemy-utils

benclifford (approved):

benclifford:
> This PR should also allow newer versions of Pandas, right?

I think probably the pandas version can be relaxed, but I'd like to try out parsl-visualize by hand - it's pretty fragile.

PR #2301, opened 2022-06-22
Remove remaining traces of IPyParallel

### LINKED ISSUE #1548: Remove IPyParallel executor

benclifford:
Fixes #1548

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Makes sense to wipe everything IPP related at this point.

PR #2305, opened 2022-06-23
Remove unused fields from RESOURCE_INFO protocol messages

benclifford:
This means instead of a tuple of one useful field and two unused
fields, the tuple can be dropped and only the one useful field,
the message, sent.

## Type of change

- Code maintentance/cleanup


danielskatz:
seems reasonable

PR #2322, opened 2022-06-29
Remove CI explicit installs

benclifford:
If these are needed they should be in test-requirements.txt

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
This looks reasonable. I'm not sure why we had installs of cython though.

PR #2323, opened 2022-06-29
Increase CI concurrency

benclifford:
# Description

This is to decrease walltime for a single CI run.

This ends up needing a bunch of target dependency work, as the existing makefile+workflow assumes a particular linear ordering not expressed in makefile targets.

## Type of change

- Code maintentance/cleanup


benclifford:
this is long abandoned - closing

PR #2368, opened 2022-07-26
Make pytest auto-selector fail on unknown site, rather than trivially passing

benclifford:
The current behaviour is when site cannot be identified, to instead pick a configuration
that runs on the localhost, without being particularly clear that a site-specific
configuration is not being used.

This can lead to false test passes.

This change forces users to see that a site-specific configuration has not been
selected, rather than giving them a false test pass.

## Type of change


benclifford:
this exposes behaviour where site_tests/test_site.py runs in `--config local` mode, which is probably not the right place to be running this test: either it should run as a specified config (which might happen to be the site selector, or might not) or it should run with a config that can run locally.

PR #2369, opened 2022-07-26
Refresh config.strategy and .max_idletime docstrings

benclifford:
## Type of change

- Documentation update


benclifford:
dupe of #2362 from last week

PR #2388, opened 2022-08-04
Performance improvements to Theta config

WardLT:
# Description

Quick change the Theta example. The current one has issues with CPU masking

## Type of change

- Documentation update


benclifford (approved):

benclifford:
Quick change the Theta example. The current one has issues with CPU masking

PR #2389, opened 2022-08-04
Use heartbeat period for HTEX worker watchdog

benclifford:
Previously, this used 1000 x the tight loop poll period.
    
This was awkward to understand and describe.
    
The heartbeat period is more relevant as a configuration
option for setting the time scale on which broken components
will be discovered.

## Type of change


benclifford:
I am uncomfortable with this being so frequent. Perhaps the right time period to use for worker watchdog should be related to the worker heartbeat period, rather than the tight loop poll period - @yadudoc ?

benclifford:
i have modified this PR to use the heartbeat period instead of the tight loop poll period

yadudoc (approved):
Not sure what we get from slowing this down from 10s -> 30s. But, I agree that the heartbeat period is logically the more appropriate one to use here.

PR #2392, opened 2022-08-08
Implement TaskRecord typed dict for task records

benclifford:
# Description

This does not change the representation of tasks records as dictionaries, but allows the structure of those dictionaries to be better expressed.

This change allows the type checker to check that fields added to a task record have been declared (in taskrecord.py) and allows docstrings to be associated with each declared field.

This PR adds most usages of `TaskRecord` that are present in the `benc-mypy` branch.

## Type of change

- Code maintentance/cleanup


benclifford:
stuck on a circular import which is removed by #2394

PR #2401, opened 2022-08-10
Only send block monitoring status on change

benclifford:
Sending every block on every iteration is extremely expensive when runs
are made with very many blocks (for example, some HTC configurations).

This PR reduces the amount of network traffic and the number of rows
recorded in the monitoring database block table.

## Type of change


benclifford:
obsoletes #2158

PR #2402, opened 2022-08-15
Bump version to 1.3.0a0

yadudoc:
# Description

Only change here is a version bump to `1.3.0a0` for putting an alpha out on Pypi to unblock `funcx`


benclifford:
this PR has been overtaken by events

PR #2422, opened 2022-09-05
Remove unused interchange SHUTDOWN codepath

benclifford:
This is never used - or at least, there is no other
SHUTDOWN literal in the parsl codebase.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Seems reasonable.

PR #2423, opened 2022-09-05
Remove STOP/kill event handling from interchange task_puller

benclifford:
The interchange is terminated by process termination, not by
STOP messages, so these code paths are unused.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
We don't send `STOP` messages from the executor, removing this code should be fine.
Looks like it is remnant code from when we wanted to launch the interchange remotely and needed a mechanism to signal it to terminate.

PR #2424, opened 2022-09-05
Remove apparently unused notify from flowcontrol

benclifford:
## Type of change

- Code maintentance/cleanup


yadudoc (approved):
We don't use this anymore, should be safe to remove.

Issue #2264, opened 2022-05-24
ClusterProvider+ThreadPoolExecutor to avoid OOM errors when running Parsl with MPI tasks


WardLT:
**Is your feature request related to a problem? Please describe.**
Workflows that include MPI tasks require creating an executor that sits on the launch node of an HPC, which is the root of my problem. What that executor does is sit on the launch node and deploy BashApps which involve a subprocess call to invoke a MPI app.. This works great, right up until I try running large number tasks concurrently. HTEx creates a new process for each worker, which can use a lot of memory on a launch node (~100 MB per process * 500 processes = 50 GB memory!) 

**Describe the solution you'd like**
Use a ThreadPoolExecutor instead of HTEx, which has a much lower memory cost. However, I cannot use TPEx with a ClusterProvider (it only runs with a LocalProvider).

**Describe alternatives you've considered**
- Use Parsl with the experimental RCT backend. I'm not yet prepared to commit to this solution given the complex installation procedure for RCT
- Use ThreadPoolExecutor with LocalProvider. I'd much rather have Parsl manage (re-)submitting jobs with the scheduler

**Additional context**

WardLT:
#2433 fixes this issue by creating a "thread" option for HTEx

PR #2433, opened 2022-09-27
Support using spawn instead of fork for HTEx

WardLT:
# Description

Polaris, in particular, has an interaction between MPI and fork which makes HTEx worker pools fail to function when used with an MPILauncher. This PR adds the ability for HTEx to launch workers by spawning a new process rather than forking.

## Type of change

- Documentation update

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -32,6 +32,8 @@ from parsl.providers import LocalProvider
 
 logger = logging.getLogger(__name__)
 
+_start_methods = ['fork', 'spawn', 'thread']
+
 
 class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
     """Executor designed for cluster-scale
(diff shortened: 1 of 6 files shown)


WardLT:
While I'm thinking about it, adding a "Thread" option for launching workers could help with #2264 . Seems like we'd have all kinds of issues with logging for each thread though 🤔

benclifford:
What's the logging issue?

With `fork`, the parent logger already appears in each worker; the worker code conveniently avoids logging to that logging name, but in general a worker can log into the manager log. (I think - it's caused me trouble before)

If it's to do with globals, maybe there are other global problems to be cautious about?

WardLT:
There could be no logging issue. I haven't reviewed the code yet to figure out if starting a worker alters "basicConfig" and messes with everyone-elses logger

benclifford:
I think some resource monitoring stuff wont' work (for the same reason that it doesn't work in the thread pool executor) - the resource monitoring is done by process. (see parsl.monitoring.monitor_wrapper)

I think it's fine for various things to not work. better if some test for invalid config options raises an exception at initialization time for known bad combos (i.e. write the machine readable version of the warning in the documentation)

WardLT:
Sounds great. I'll add that in while it's on the brain

WardLT:
Just confirmed that this runs on at least 256 nodes with 1024 works on Polaris!

PR #2441, opened 2022-09-28
Github action to publish to testpypi

yadudoc:
GIthub action that publishes to testpypi. Plan is to test this over there and then update to push to pypi.
This PR is missing changes to `tag_and_release.sh` that will come separately.


yadudoc:
With #2442 merged, this is ready to go in. We'll only be pushing to TestPypi with this. I'll add a follow up that switches to Pypi once the automation works as expected.

yadudoc:
Following a discussion with @benclifford, this PR needs changes to the trigger, URL target, and token in the next round. For now, this should be sufficient to test whether this works as intended.

PR #2448, opened 2022-10-03
Follow up to get release automation point at Pypi

yadudoc:
# Description

* Point at pypi instead of test.pypi
* Do scheduled push to pypi at 09:42 on Wednesdays from main.

Fixes # (issue)

## Type of change


yadudoc:
*HOLD* until user announcement goes out. @kylechard @benclifford

Issue #2453, opened 2022-10-15
Multiple Work Queue executors collide in function_dir


benclifford:
**Describe the bug**
The Work Queue executor creates working files inside rundir/NNN/function_data, and rundir/NNN/package_data

If two (or more) work queue executors are used in the same run, then both executors try to create their function data in this directory.

This causes conflicts because the same filename is used by both executors.

**To Reproduce**
Run attached script.

Get this error:

```
Traceback (most recent call last):
  File "/home/benc/parsl/src/parsl/wq2.py", line 22, in <module>
    parsl.load(config)
  File "/home/benc/parsl/virtualenv-3.9/lib/python3.9/site-packages/typeguard/__init__.py", line 1033, in wrapper
    retval = func(*args, **kwargs)
  File "/home/benc/parsl/src/parsl/parsl/dataflow/dflow.py", line 1332, in load
    cls._dfk = DataFlowKernel(config)
  File "/home/benc/parsl/src/parsl/parsl/dataflow/dflow.py", line 179, in __init__
    self.add_executors(config.executors + [parsl_internal_executor])
  File "/home/benc/parsl/src/parsl/parsl/dataflow/dflow.py", line 1025, in add_executors
    block_ids = executor.start()
  File "/home/benc/parsl/src/parsl/parsl/executors/workqueue/executor.py", line 312, in start
    os.mkdir(self.function_data_dir)
FileExistsError: [Errno 17] File exists: '/home/benc/parsl/src/parsl/runinfo/075/function_data'
```

**Expected behavior**
It should be possible to use multiple work queue executors in one run.

**Environment**
my laptop, parsl master ca7507d63a6b5b6ad9562c3432d96fc4e5be063d

benclifford:
Reproducing case:

```
$ cat wq2.py 
import parsl
from parsl import python_app
from parsl.config import Config
from parsl.executors import WorkQueueExecutor

from parsl.data_provider.http import HTTPInTaskStaging
from parsl.data_provider.ftp import FTPInTaskStaging
from parsl.data_provider.file_noop import NoOpFileStaging

from parsl.providers import LocalProvider

ax = WorkQueueExecutor(label='ax', port=9000,
                                             storage_access=[FTPInTaskStaging(), HTTPInTaskStaging(), NoOpFileStaging()],
                                             provider=LocalProvider(init_blocks=1, min_blocks=1, max_blocks=1))

bx = WorkQueueExecutor(label='bx', port=9001,
[...]

PR #2455, opened 2022-10-15
Fix conflicting paths in multiple WQ executors.

benclifford:
Previously Work Queue Executor working files were placed under the rundir top level. This resulted in conflicts when more than one Work Queue Executor is used in a run.

This moves those files down one level into a per-executor directory.

Fixes issue #2453

## Type of change


benclifford:
cc @btovar @dthain

PR #2459, opened 2022-10-25
Wrapper for using Parsl as a concurrent.futures Executor

WardLT:
# Description

Adds a wrapper for Parsl that lets you use it in the same way as a `concurrent.futures.Executor`. 

For example,

```python
    with ParslPoolExecutor(my_config) as exc:
        # Test a single submit
        future = exc.submit(f, 1)
        assert future.result() == 2

        # Make sure the map works
        results = list(exc.map(f, [1, 2, 3]))
        assert results == [2, 3, 4]

        # Make sure only one function was registered
        assert exc.app_count == 1    
```

I'll add documentation for this a revised intro page and "quickstart" in another PR, if that's OK. 

## Type of change


parsl/dataflow/dflow.py
```diff
@@ -1047,6 +1047,7 @@ def wait_for_current_tasks(self) -> None:
 
         logger.info("Waiting for all remaining tasks to complete")
 
+        # TODO (wardlt): Can this be replaced with conccurent.futures.wait?
```
benclifford, line 1050: yes I think so

benclifford: There's two pieces of logic here - one is the waiting for app_fu to complete, and one is waiting for that status in the task record to be a final state. Parsl does not cleanly define which of those really means a task is all completed. See issue https://github.com/Parsl/parsl/issues/1279 for related discussion about task completion.

It's important in this `wait_for_current_tasks` call that both of those completion states have been reached for each task.

parsl/concurrent/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+            return self._app_cache[fn]
+        app = PythonApp(fn, data_flow_kernel=self.dfk)
+        self._app_cache[fn] = app
+        return app
+
+    def submit(self, fn, *args, **kwargs):
+        app = self._get_app(fn)
+        return app(*args, **kwargs)
+
+    # TODO (wardlt): This override can go away when Parsl supports cancel
```
benclifford, line 56: what happens when the timeout is reached? this method raises an exception but parsl continues to execute all the stuff?

parsl/dataflow/dflow.py
WardLT: Oh dang, that's interesting about the two termination conditions. I'll retract my TODO comment

parsl/concurrent/__init__.py
WardLT: Correct. See any way around that? Or, should I warn people in the docstring?

benclifford: docstring

WardLT: Done

parsl/tests/sites/test_concurrent.py
```diff
@@ -0,0 +1,48 @@
...
+from parsl import Config, HighThroughputExecutor
+from parsl.concurrent import ParslPoolExecutor
+
+
+def f(x):
+    return x + 1
+
+
+# TODO (wardlt): For reasons I don't understand, re-using a configuration is a problem. This creates a fresh one
+def make_config(run_dir):
```
benclifford, line 14: in the rest of the test suite, this is called `fresh_config`

benclifford: > # TODO (wardlt): For reasons I don't understand, re-using a configuration is a problem. This creates a fresh one

A Config() object in parsl is not a description of the components (such as executors and launchers) that will be used in a run. It *is* the components that will be used.

So, for example, a config might contain a high throughput executor, and that high throughput executor will be configured with script directories etc, started up, used and then shut down. None of the parsl code is built on the basis that after shutting various components down, you can reconfigure them and start them up again - which is what happens when you re-use a config.

The *description* of the configuration, which *can* be re-used, is the code that generates a Config() object - and the way to re-use that is to put it in a function, so that you can generate a fresh config object when you need to start a new DFK.

In that sense, the "configuration file" equivalent is not the Config object itself, but the source code that generates a config object.

WardLT:
Are those errors in the CI transient?

PR #2461, opened 2022-10-28
Support python 3.11 in CI

benclifford:
## Type of change


benclifford:
waiting for 3.11 to appear in github actions

benclifford:
another error that has appeared here is that the rendering of task summaries has become less helpful due to a change in enumeration stringification in python 3.11:

This assertion is failing with python 3.11:

> AssertionError: assert 'Tasks in state States.exec_done: 1' in 'INFO 

because the output has switched to using numeric values, like this:

> Tasks in state 3: 1

This is a worse way of outputting things for users to see, so it would be good to fix.

Issue #2469, opened 2022-11-16
Requesting a way to monitor long running Parsl jobs


astro-friedel:
One of the projects I am using Parsl in submits long running jobs (days to weeks long). The jobs consists of many time-steps and outputs files after a user specified number of steps. The purpose of these files is to checkpoint the process in case it needs to be restarted, but can also be used to give the status of the computations (e.g. working as expected, unknown issue causing unexpected, but non-fatal results, etc.)

In the current Parsl system there is no way to see the status of the job other than waiting/running/done. But by looking for these output files and processing them we could get a very detailed look at the status. The workflow of this monitoring is as follows:

- The user gives the monitor system a file type to look for and a directory to look in, and a callback to call if/when files are found
- The monitor periodically scans the directory for files of the given type
    - If files are found, then call the callback function
    - If not, then sleep until the next scan

**Potential Solutions**
- Write our own code to do this. We have done this, but feel that other projects may like to do this type of monitoring, and that an implementation inside of Parsl would be cleaner.
- Change the Parsl workflow to do shorter jobs, process the output files, and restart using the checkpoint files. This is infeasible for any workflow where there is significant overhead in (re)starting a job.
- Incorporate our code into Parsl and let it handle the background work and the calling of the callback functions.

The preferred solution is the third one as it would provide a cleaner implementation and also provide the capability to a wider audience.

PR #2470, opened 2022-11-16
File monitoring subsystem #2469

astro-friedel:
# Description

This PR introduces a file monitoring subsystem to Parsl. The idea behind File Monitoring is to have a mechanism to provide the status of a job, beyond the typical waiting/running/done
status. Specifically, this is best suited for long running jobs (days, weeks, etc.) that output files periodically, such
as at the end of a timestep. The user specifies what files to look for and a function(s) to call when these files are found.
The `parsl.monitoring.FileMonitor` class provides the interface for specifying the patterns and callbacks.

The file monitoring system will periodically scan the file system (within the given root directory) for any file(s)
of the given file type(s) or regex pattern(s). If any are found they are sent to the user specified callback function(s).
The infrastructure tracks files that have been found previously and only sends newly found files to the callback. The
system also tries to verify that the file(s) are no longer being written to. On shared file systems, there is no direct
mechanism to do this, so it works under the assumption that any file with a modification timestamp that is at least
a specified (by user) seconds old will be considered. If a process is expected to periodically write to a file then
the ``sleep_dur`` parameter of the FileMonitor class should be set to a larger value.

The file monitoring system is given a list of regex style patterns and/or a list of file types to use for searching for
files. For regex style patters they should be strings preceeded by **r** as they are compiled into regex objects by the
file monitoring system (see the `re <https://docs.python.org/3/library/re.html>`_ Python module documentation for
regex specifics). For file type patterns a list of file suffixes should be given. Suffixes with and without an asterisk
are acceptable (e.g. ``pdf``, ``.pdf``, and ``*.pdf`` are all equivalent). File types will have the given ``path`` prepended
to them.

The callback functions have only a few restrictions on them

    1. The function should take a single argument that is a list of the detected files (string, including full path)
    2. The function should return either ``None``, a string like value, or something that can be cast to a string with ``str()``
    3. Any files produced by the function need to be handled by the user (transfer, etc.)

If an email address is provided to the `parsl.monitoring.FileMonitor` instance then the return value of the each callback
function is sent as the body of an email to that address. Currently, attachments are not supported.

The `parsl.monitoring.FileMonitor` has been added as an optional argument to both app decorators, allowing for the use of file monitoring to be flexible at the individual app level.

Fixes # 2469

## Type of change

- Documentation update


benclifford:
My main concern here is that I don't think the core of parsl should have anything to do with sending email. The file monitoring code provides a callback for users to do arbitrary things, and that is probably the right place for them to send an email if they want.

(A motivating question: why should a user need to patch the source code of parsl to send their notifications via Slack instead of email?)

In that model, the email sending code should exist as a helper that can be easily invoked by users inside their callback: the callback should do all the work, rather than doing half the work and then returning some instructions to do more work.

I'm also concerned that the email sending code is overly optimistic about its ability to autoconfigure itself to deliver outbound email; and that this will generate support load for you and a poor reputation for parsl. Most email sending porgrams have multiple configuration fields for outbound email, and I'm not convinced that the attempts at autoconfiguration in this PR are general enough to work in most places. Moving this code into helper modules would allow this code to be labelled as more experimental / less supported distinct from the file monitoring callback code.

parsl/monitoring/file_monitoring.py
```diff
@@ -0,0 +1,488 @@
...
+        The working directory for this process to run in. Default is ``None``, meaning the working directory is
+        inherited from the calling process.
+    email: str, optional
+        Email address to send callback results to
+    max_proc: int, optional
+        The maximum size of the multiprocessing Pool for the file processing. Default is 4, but the
+        value is adjusted to be ``min(max_proc, len(patterns))`` so as to not make the pool larger than
+        needed.
+    """
+    try:
```
benclifford, line 304: Does this code run on the submit side or on workers?
(aka does the user callback code run on the submit side or the worker side?)

I think it's running in a worker?

parsl/monitoring/file_monitoring.py
```diff
@@ -0,0 +1,488 @@
...
+                            pat = '.' + pat
+                        pat = '*' + pat
+                    temppat[i] = pat
+            for ft in temppat:
+                self.patterns.append((ft, False))
+        if not self.patterns:
+            raise Exception("Either pattern or filetype must be given")
+        if isinstance(callback, list):
+            if len(self.patterns) != len(callback):
+                raise Exception("Pattern list is not the same size as function list")
```
benclifford, line 439: there's a different style you could maybe use that makes this true by construction:

instead of three lists of callback, pattern, filetype, you instead have one list, with the contents being a tuple: (callback, pattern, filetype)

that's maybe also easier for users when they edit their code - instead of three separate lists they need to keep aligned manually, there's a single list and inserting an element in that list does not break the other two.

parsl/monitoring/file_monitoring.py
astro-friedel: Yes, on the worker, so perhaps the max_proc default should be reduced to 1 or 2 so as to not overload the worker. I should also add the caveat in the docs that any callbacks should be lightweight or risk slowing the worker.

astro-friedel:
I debated about not having the email stuff in there. After putting it in and working with it for a bit now. I agree that it should be left up to the user to implement.

parsl/monitoring/file_monitoring.py
benclifford: Another caveat to document, perhaps, is where side-effects of callbacks have their effect: the side-effects can only happen on the worker and cannot (for example) update a progress bar on the submitting process.

astro-friedel: I updated the documentation to make this clearer

astro-friedel:
I was wondering if this PR will get a review again? (I will work to bring it up to date with main)

PR #2480, opened 2022-11-21
Remove histogram of mean CPU time

benclifford:
This removed plot measures the mean CPU time across all resource readings for each task, and then makes a histogram of how many tasks fall into each bucket.

However, the mean cumulative CPU time across all resource readings (for a given task) doesn't really have a lot of meaning. The maximum value plot, which makes more sense, is retained by this PR.

# Description

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/visualization/templates/resource_usage.html
+++ b/parsl/monitoring/visualization/templates/resource_usage.html
@@ -37,9 +37,6 @@
 <div class="container">
   <div class="row">
     <div class="col-6" style="height: 300px;">
-{{ user_time_distribution_avg_plot|safe }}
-    </div>
-    <div class="col-6" style="height: 300px;">
 {{ user_time_distribution_max_plot|safe }}
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


WardLT:
Why doesn't it have much meaning? Is it that your are measuring the average over all resources even though only a subset of those resources is being used for the task?

benclifford:
@WardLT 

Take one single parsl app invocation (aka parsl task)

Imagine the (monotonically increasing) measured cumulative cpu usage values, as recorded every so often by the resource monitoring system as a  0.1 cpu seconds, 0.2 cpu seconds, 0.25 cpu seconds, 0.35 cpu seconds, 0.45 cpu seconds then exits.

I can interpret max = final reading = 0.45s  as "this task used 0.45 CPU-seconds to do its work".

 I can use that to understand how many cpu-hours my full workflow of a million of these tasks might take.

And I can look on a histogram of such values and see that there is some distribution of those task costs: that all tasks (for example) tend to use around the same amount of CPU time (or perhaps not -- the histogram will tell me)

But, I don't know what meaning I can gather from  mean(0.1,0.2,0.25,0.35,0.45) = 0.27s for a particular task.

WardLT:
Oh! Yea, that average isn't what I thought it would be (average CPU utlization). 

I'm on board with dropping this.

PR #2487, opened 2022-11-23
Don't deserialize during serialization

benclifford:
It is unlikely that this will detect any problems serialization problems, and it is expensive under heavy loads - roughly doubling the cost of serializing complex objects, if they are not already in the cache.

## Type of change

- Code maintentance/cleanup


benclifford:
@yadudoc wrote this code originally - I think our online discussions resulted in a consensus that this line could go away.

PR #2502, opened 2022-12-09
Fix bug in securityContext user_id and group_id

manning-ncsa:
# Description

If the `user_id` and `group_id` are specified in the Kubernetes provider configuration, the following error is thrown:

```
  File "/src/main.py", line 101, in <module>
    print(f'''[{cur_time()}] python replica {idx} result: {replicas[idx].result()}''')
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.8/dist-packages/parsl/dataflow/dflow.py", line 288, in handle_exec_update
  File "/usr/local/lib/python3.8/dist-packages/parsl/dataflow/dflow.py", line 481, in _unwrap_remote_exception_wrapper
    result = future.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
parsl.executors.errors.BadStateException: Executor kube-htex failed due to: 1. Failed to start block 0: (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Audit-Id': 'db01ee74-a678-41b2-bf0a-136df2fcb804', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'e10947a4-ddac-4479-94a9-fb5acbb34225', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'f4449e23-8f8f-4a4a-9115-be7233c0ea63', 'Date': 'Fri, 09 Dec 2022 19:40:30 GMT', 'Content-Length': '288'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod in version \"v1\" cannot be handled as a Pod: json: cannot unmarshal string into Go struct field SecurityContext.spec.containers.securityContext.runAsGroup of type int64","reason":"BadRequest","code":400}
```

The `client.V1SecurityContext()` function expects integers while Parsl expects strings.

Fixes # (issue)

## Type of change

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -244,8 +244,8 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
 
         security_context = None
         if self.user_id and self.group_id:
-            security_context = client.V1SecurityContext(run_as_group=self.group_id,
-                                                        run_as_user=self.user_id,
+            security_context = client.V1SecurityContext(run_as_group=int(self.group_id),
+                                                        run_as_user=int(self.user_id),
                                                         run_as_non_root=self.run_as_non_root)
 
         # Create the enviornment variables and command to initiate IPP


benclifford:
This looks ok to merge.

But maybe it would be better to change the types of `KubeProvider.__init__`:

```                 user_id: Optional[str] = None,```

to be an int - it only ever gets passed through to `V1SecurityContext` so I think it makes sense to expose that expectation of an int - and have it type-checked by the existing typeguard decorator around `KubeProvider.__init__`

manning-ncsa:
That makes more sense to me. It seemed so odd that the type was string that I assumed there was some reason for it!

Issue #2174, opened 2021-12-28
Ctrl-C handling is generally unreliable


benclifford:
**Describe the bug**
When pressing ctrl-C in various situations, parsl tends to hang. This is especially obvious to me in `pytest`, one of the main environments in which I run parsl code.

For example, pressing ctrl-C in a recent pytest run has left pytest waiting for parsl to shut down before exiting, while the parsl strategy code continues to attempt to launch (failing) blocks to run its one outstanding job on htex, apparently forever.

Other misbehaviour happens with monitoring - for example, see issue https://github.com/Parsl/parsl/issues/2123

In general, parsl's handling of shutdown like this is pretty poor and needs a serious review and fixup to ensure that the various component processes can cope with being killed by means other than successful execution shutdown.

PR #2503, opened 2022-12-09
Fall back to load_incluster_config() when load_kube_config() fails

manning-ncsa:
# Description

See https://github.com/kubernetes-client/python/issues/1005 for a good description of the issue.

Fixes # (issue)

## Type of change

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -104,7 +104,10 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         if not _kubernetes_enabled:
             raise OptionalModuleMissing(['kubernetes'],
                                         "Kubernetes provider requires kubernetes module and config.")
-        config.load_kube_config()
+        try:
+            config.load_kube_config()
+        except:
+            config.load_incluster_config()
 
         self.namespace = namespace
         self.image = image


benclifford:
This is failing flake8 with:

```
flake8 parsl/
parsl/providers/kubernetes/kube.py:109:9: E722 do not use bare 'except'
```

parsl/providers/kubernetes/kube.py
```diff
@@ -104,7 +104,10 @@ def __init__(self,
         if not _kubernetes_enabled:
             raise OptionalModuleMissing(['kubernetes'],
                                         "Kubernetes provider requires kubernetes module and config.")
-        config.load_kube_config()
+        try:
+            config.load_kube_config()
+        except:
```
benclifford, line 109: This line is caught by flake8: it catches all exceptions.

In practice in parsl, this sort of thing has caused problems such as ctrl-C handling: https://github.com/Parsl/parsl/issues/2174

Maybe it should be more narrowly `kubernetes.config.config_exception.ConfigException`

manning-ncsa: Sounds reasonable.

dannielshalev: yes it is, try this try block, and the issue persists
```
    try:
        config.load_incluster_config()
    except ConfigException as e:
        logging.exception(e)
        config.load_config()

```

tcbonds:
@manning-ncsa You still working on this? Running into the same problem right now.

manning-ncsa:
@tcbonds No, I am not working on this anymore. Good luck!

mfisher87:
@benclifford myself and @trey-stafford are dealing with this today, and this looks simple enough for us to re-implement in a superceding PR so we're going to go for that :)

benclifford:
closing because #3357 is a replacement

PR #2505, opened 2022-12-12
Changing environment variables for SYCL devices

cms21:
# Description

This PR changes the SYCL environment variable for gpu discovery from SYCL_DEVICE_FILTER to ZE_AFFINITY_MASK.  I've been doing tests with SYCL devices and I found that setting both caused conflicts in some programs.  I was told in private communications that SYCL_DEVICE_FILTER has never been used or tested.  I removed it and added ZE_AFFINITY_MASK.

Fixes # (issue)

## Type of change


WardLT (approved):
LGTM! Thanks, @cms21 ! 

As far as I know, the SYCL filter is not used (the private communication 'twas I) and Christine also pointed out it is deprecated anyway.

PR #2507, opened 2022-12-13
RADICAL-Pilot executor for Parsl (RPEX)

AymenFJA:
# Description

Adding RADICAL-Pilot (RP) as an executor for Parsl. The PR brings RPs capabilities to Parsl, such as heterogeneous resource management and heterogeneous MPI tasks execution, bulk submission, etc.

Fixes # (issue)
None

## Type of change

### DIFF
--- a/parsl/tests/test_bash_apps/test_error_codes.py
+++ b/parsl/tests/test_bash_apps/test_error_codes.py
@@ -138,25 +138,3 @@ def run_app(test_fn, err_code):
     os.remove('std.err')
     os.remove('std.out')
     return True
-
-
-if __name__ == '__main__':
-
-    parser = argparse.ArgumentParser()
-    parser.add_argument("-d", "--debug", action='store_true',
-                        help="Count of apps to launch")
-    args = parser.parse_args()
-
-    exit(0)
-
-    if args.debug:
-        parsl.set_stream_logger()
-
-    for test in test_matrix:
-        try:
-            run_app(test, test_matrix[test]['exit_code'])
-        except AssertionError as e:
-            print("Test {0:15} [FAILED]".format(test.__name__))
-            print("Caught error : {0}".format(e))
-        else:
-            print("Test {0:15} [SUCCESS]".format(test.__name__))


benclifford:
This is failing near the start of the test suite, in `mypy`:

```
parsl/executors/radical/executor.py:13: error: Cannot find implementation or library stub for module named "radical.pilot"
parsl/executors/radical/executor.py:13: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
parsl/executors/radical/executor.py:13: error: Cannot find implementation or library stub for module named "radical"
parsl/executors/radical/executor.py:14: error: Cannot find implementation or library stub for module named "radical.utils"
```

The first set of errors you need to either install type stubs for radical, or more likely, tell mypy that radical doesn't have any - see the end of `/mypy.ini` where other packages are labelled this way:

```

[mypy-setproctitle.*]
ignore_missing_imports = True
```

then for this error:


```
parsl/executors/radical/executor.py:76: error: Need type annotation for "future_tasks" (hint: "future_tasks: Dict[<type>, <type>] = ...")
```

add a type annotation like this one in htex to indicate what is going to go inside the dict.

```
parsl/executors/high_throughput/interchange.py:        self._ready_managers: Dict[bytes, ManagerRecord] = {}
```

Alternatively you can disable mypy type checking for parsl/executors/radical/ in mypy.ini, but I discourage that.

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,388 @@
...
+                 label: str = 'RPEX',
+                 bulk_mode: bool = False,
+                 resource: str = None,
+                 login_method: str = None,
+                 walltime: int = None,
+                 managed: bool = True,
+                 cores: Union[int, float] = float('inf'),
+                 gpus: Optional[int] = 0,
+                 worker_logdir_root: Optional[str] = ".",
+                 partition: Optional[str] = " ",
```
benclifford, line 65: The `Optional` annotation here means that I can pass in `None` as a value, distinct from the default value of " ".

Is that what you intend? If the value always has to be a string, never None, and " " is the right default value, remove the `Optional` annotation here.

Same for other lines nearby which use `Optional`.

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,388 @@
...
+
+                    return func, args, task_type
+
+            # @python_app from parsl
+            else:
+                task_type = inspect.getsource(func).split('\n')[0]
+                if PYTHON in task_type:
+                    task_type = PYTHON
+                else:
+                    task_type = ''
```
benclifford, line 250: is this an error case?

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,388 @@
...
+            # push task to rp submit thread
+            self._bulk_queue.put(task)
+        else:
+            # submit the task to rp
+            logger.debug("put {0} to rp-TMGR".format(rp_tid))
+            self.tmgr.submit_tasks(task)
+
+        return self.future_tasks[rp_tid]
+
+    def _get_job_ids(self):
```
benclifford, line 370: this method was changed in PR #1952 (commit 5eb84c0859929110b06e72d4843d45a9633157a2) 
to this:
 +    def _get_block_and_job_ids(self) -> Tuple[List[str], List[Any]]:

but I think you don't need it at all as this class inherits from `NoStatusHandlingExecutor` which does not do any block management.

benclifford:
It would be very good to see this pass the `pytest` test suite in CI, for sustainability purposes.

I'm not sure if it's possible to get a minimal radical pilot install in place during a CI run? (see how for example, work queue is installed for the work queue executor).

eirrgang:
> I'm not sure if it's possible to get a minimal radical pilot install in place during a CI run? (see how for example, work queue is installed for the work queue executor).

For an example GitHub Actions job that installs and tests against a functioning RCT installation, see https://github.com/SCALE-MS/scale-ms/blob/76c2ea1991c48a64236ce780219fb0d6f419b6ac/.github/workflows/tests.yml#L138 and https://github.com/SCALE-MS/scale-ms/blob/76c2ea1991c48a64236ce780219fb0d6f419b6ac/.github/workflows/resource_local.json#L2

parsl/executors/radical/executor.py
AymenFJA: Yes, as sometimes we fail to extract the task type.

parsl/executors/radical/executor.py
AymenFJA: Agreed and removed, thanks.

PR #2529, opened 2022-12-27
Set OpenMP environment variables when imposing affinity

### LINKED ISSUE #2528: Improve thread affinity with OpenMP
I find that I need to add "OMP_NUM_THREADS" before calling an OpenMP executable when running multiple OpenMP applications on a single node to avoid them contending with each other. 

It would be great for HTEx to set the environment variables that control OpenMP's number of threads per process and affinity automatically, rather than me setting them in my job.

- Placing more than one HTEx per node with an MPI launcher and using the MPI launcher's affinity mechanisms. 

**Additional context**
A primer on thread affinity and OpenMP: https://hpc-tutorials.llnl.gov/openmp/ProcessThreadAffinity.pdf

WardLT:
# Description

Sets the thread affinity for subprocesses that use OpenMP launched by a Parsl worker by setting the environment variables.

Fixes #2528. Though, as discussed, we're unsure if this can be done a better way.

## Type of change

- Documentation update

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -619,7 +619,7 @@ def worker(worker_id, pool_id, pool_size, task_queue, result_queue, worker_queue
         logger.info("All processing finished for task {}".format(tid))
 
 
-def start_file_logger(filename,urank, name='parsl', level=logging.DEBUG, format_string=None):
+def start_file_logger(filename, rank, name='parsl', level=logging.DEBUG, format_string=None):
     """Add a stream log handler.
 
     Args:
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -612,7 +619,7 @@ def worker(worker_id, pool_id, pool_size, task_queue, result_queue, worker_queue
         logger.info("All processing finished for task {}".format(tid))
 
 
-def start_file_logger(filename, rank, name='parsl', level=logging.DEBUG, format_string=None):
+def start_file_logger(filename,urank, name='parsl', level=logging.DEBUG, format_string=None):
```
benclifford, line 622: what is this u?

WardLT: A typo that escaped my notice (fat fingered while editing in vim)

benclifford (approved):

PR #2537, opened 2023-01-09
Remove unmanaged/managed executor concept

benclifford:
Unmanaged executors are poorly defined and not tested, and reading the few codepaths involved can be a bit confusing when trying to understand scaling logic.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Good to remove this. We no longer use this 👍

PR #2538, opened 2023-01-09
Remove unused scaling_enabled method from adhoc provider

benclifford:
scaling_enabled usually lives on executors, not on providers, and this method never seems to be invoked.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
👍

PR #2587, opened 2023-02-01
Split main interchange loop into four separate topical methods

benclifford:
## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Alright, I checked through the commits and this is only code reorg into functions. Looks way neater than before!

PR #881, opened 2019-04-15
Add support for configuring prefetch capacity

annawoodard:
This is a squashed cherry pick with a few changes removed
in order to just collect the changes related to prefetch capacity.
It is based on the following commits:

75f9d4eca8ec179564768c74b8c5cd67154ea1d8
5a221f2e8ad581b23e4b4b1ec86d3ee02e41e460
c89f3b2def16023bd08360705e923eeb73815013

PR #2588, opened 2023-02-01
Remove manager to interchange work requests to speedup sequential task distribution

benclifford:
[single-latency.py.txt](https://github.com/Parsl/parsl/files/10582032/single-latency.py.txt)
Prior to this PR, the interchange calculated how many tasks to send to a manager based on the minimum of: how many tasks the manager most recently requested (minus sent tasks); how many tasks the manager has total capacity for, minus the number of tasks in flight.

The most recently requested count is made on an exponential backoff, which continues to backoff when a manager is saturated with long tasks. That means that the requested task count is sometimes not updated for quite a long time.

This PR removes that clause entirely, and the code that delivers that information. Now the interchange relies only on the computed maximum capacity minus the number of tasks in flight - this code was added in PR #881, but alongside the previous "manager requests work" value.

A new task will be dispatched when a result arrives from a manager, freeing up a slot, rather than on a separate free_capacity poll: so new task dispatch is now much more event driven.

On my laptop on `master` prior to this PR, the attached demonstrator code shows that when tasks execute quickly, tasks begin executing just over 1 or just over 2 poll_periods after submission; but when tasks take a long time and a manager is saturated, backoff of poll periods in the manager means that the manager remains idle for sometimes 9 seconds before starting to execute a task.

After this PR, the same code shows tasks executing in just over one poll period in both situations.
I have tried this with a range of poll periods from 10 to 500ms.

With smaller poll periods, 1 .. 3 ms, I have noticed that the slow task saturated case takes a few ms longer for task dispatch compared to the fast task path. I have not investigated why that happens, and I don't intend to for this PR.

@WardLT noticed this behaviour in May 2022, and it is still present at time of writing this PR.

## Type of change

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -393,7 +393,6 @@ class Interchange(object):
                     # We set up an entry only if registration works correctly
                     self._ready_managers[manager_id] = {'last_heartbeat': time.time(),
                                                         'idle_since': time.time(),
-                                                        'free_capacity': 0,
                                                         'block_id': None,
                                                         'max_capacity': 0,
                                                         'worker_count': 0,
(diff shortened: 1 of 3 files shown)


yadudoc (approved):
I think these changes are correct. I've read this over a few times to understand this better, and here's what I've got so far, @benclifford correct me if I'm wrong:
```
* Manager is added to interesting_managers only 
    * upon registration
    * receiving a result
* Manager is removed if
    * real_capacity drops to 0 (when tasks are dispatched)
    * manager is marked as not active
These ops happen in the same thread, so race-conditions are not a concern.
```

benclifford:
While profiling this, I've noticed that there is a situation where task performance is less. There's a race condition  in the sequential task case when finishing one task, looping back to user space, submitting a new task, which means that because the poll is not interrupted, that new task will take almost exactly one poll period to dispatch.

Previously this could have been lower, because the frequent task requests by managers would make the poll end before the poll period.

I'm not super concerned about this at the moment: this PR moves performance in this situation from a big range of times, some of them awful, to being almost exactly the poll period for each task.

PR #2605, opened 2023-02-22
Add option to use Work Queue coprocess feature

benclifford:
This allows persistent python worker processes to exist, so that the cost of starting a python process and making application imports is paid once per worker, rather than once per app invocation.

This model looks a bit more like the high throughput executor (with a work queue worker = an HTEX manager, and a work queue parsl coprocess = an HTEX worker process)

To activate, specify `coprocess=True` to the `WorkQueueExecutor` initialiser.

The parsl coprocess script is machine generated and checked into version control. The source for this is `parsl/executors/workqueue/parsl_coprocess_stub.py` and the command to turn the stub into the coprocess script is:
`poncho_package_serverize --src parsl/executors/workqueue/parsl_coprocess_stub.py --function run_parsl_task --dest parsl/executors/workqueue/parsl_coprocess.py --version work_queue`


benclifford:
this PR uncovered a monitoring bug that should be fixed in PR #2631

PR #2609, opened 2023-02-24
Adding broad write perms to allow gh actions to make tag

yadudoc:
# Description

This PR updates the permission block from `contents: read` to `write-all` which is the broadest permission statement.
The plan is to confirm to first confirm with this PR that the current tag writing failure can be fixed by the write permission grant and then refine it to the specific grant needed just to write tags.

Fixes # (issue)

## Type of change

### DIFF
--- a/.github/workflows/python-publish-to-testpypi.yml
+++ b/.github/workflows/python-publish-to-testpypi.yml
@@ -19,8 +19,7 @@ on:
         required: false
         default: ''
 
-permissions:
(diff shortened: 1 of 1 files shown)


benclifford:
does this still need read permission too?

yadudoc:
Good question. Let me update that perms to explicitly allow both.

benclifford (approved):

PR #2610, opened 2023-02-26
Upgrade flake8 to 5.0.4

benclifford:
Although a later flake8 is available (6.0.0) it requires Python >=3.8

This PR also fixes flake8 errors which appear after the upgrade, which are to do with missing whitespace when using C-style bracketing for if/not/assert.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -139,7 +139,7 @@ class MonitoringHub(RepresentationMixin):
         self._dfk_channel = None  # type: Any
 
         if _db_manager_excepts:
-            raise(_db_manager_excepts)
+            raise _db_manager_excepts
 
         self.client_address = client_address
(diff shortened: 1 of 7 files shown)


benclifford:
there's a dependency conflict between sphinx and flake8 5.0.4 which requires an older importlib-meta:

```
The conflict is caused by:
    The user requested importlib-metadata<5
    flake8 5.0.4 depends on importlib-metadata<4.3 and >=1.1.0; python_version < "3.8"
    sphinx 4.5.0 depends on importlib-metadata>=4.4; python_version < "3.10"
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

Closing this. We can try upgrading to 6.0.0 when we've dropped support for python 3.7

PR #2619, opened 2023-03-08
Loosen pandas version constraint so it can install from wheels on both python 3.7 and python 3.11

benclifford:
# Description

There is no pandas wheel available <1.4 for Python 3.11 so installing on Python 3.11 results in a long (eg 15 minute) addition to the build process as pandas is built locally.

This PR will attempt to make parsl able to use both pandas <1.4 (for continued Python 3.7) support and a later version (for which a Python 3.11 wheel is available).

See PR #2461 for Python 3.11 upgrade information

## Type of change

- Code maintentance/cleanup


benclifford:
This is passing in CI, but I have some memory that recent pandas broke something (in visualization?). So I want to investigate that a bit further before merging this.

PR #2625, opened 2023-03-13
Set ZMQ receive timeouts in the interchange to prevent blocking threads

yadudoc:
# Description

Without the `RCVTIMEO` the `task_puller` thread and `_command_server` on the interchange block indefinitely and do not break when the kill_event is set. Since the recv_pyobj call is in c-land it has a tendency to ignore signals.

This is rolling back changes from this PR: https://github.com/Parsl/parsl/pull/2438
While we do not wait for the kill event, we still need the thread exit from the ZMQ context where it seems to ignore signals. 

Currently, this is blocking/causing hangs on funcX tests wherever we use HTEX.

## Type of change

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -138,12 +138,14 @@ class Interchange(object):
         self.context = zmq.Context()
         self.task_incoming = self.context.socket(zmq.DEALER)
         self.task_incoming.set_hwm(0)
+        self.task_incoming.RCVTIMEO = 1000
         self.task_incoming.connect("tcp://{}:{}".format(client_address, client_ports[0]))
         self.results_outgoing = self.context.socket(zmq.DEALER)
         self.results_outgoing.set_hwm(0)
         self.results_outgoing.connect("tcp://{}:{}".format(client_address, client_ports[1]))
 
         self.command_channel = self.context.socket(zmq.REP)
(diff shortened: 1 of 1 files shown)


benclifford:
can you add some context/details for reproducing the problem? this is exiting fine with signals in other cases, I think, so I'd like to understand what is different in your test setup.

benclifford:
I talked to Kevin who is working on funcX for more context about this.

My understanding is this is to address an issue with the test suite hanging which appears in https://github.com/funcx-faas/funcX/pull/1057

I have replicated what I think is the same hang on my laptop, and investigated it.
However this PR does not fix that hang on my laptop.

The hang on my laptop comes from other parts of the funcX test suite installing a different handler for SIGTERM, and then the htex interchange inheriting that handler. This is bad behaviour by both the funcx test suite and the htex interchange. I opened an issue for the htex side of things - https://github.com/Parsl/parsl/issues/2628

If this current PR #2625 *does* fix things for you, I'm interested in further investigating. If it turns out that it was because you also had a signal handler override hacked into your interchange, outside of this PR, then probably this PR can be closed without merge.

benclifford (changes requested):
.

benclifford:
changing this to draft because I think it can probably be closed but this can wait for yadu to look at later

benclifford:
closing this unmerged -- I talked with @yadudoc and it looks like already-merged #2629 fixes this as I hoped.

Issue #2343, opened 2022-07-14
Threads and multiprocessing cannot be intermingled without deadlocks


benclifford:
**Describe the bug**
Threads and `multiprocessing` cannot be intermingled. Parsl makes heavy use of both threads and `multiprocessing`.

If one thread in a process holds a lock, while another thread forks a new process, then that new process will start with the lock held, and with nothing to ever unlock it. Any attempt in the new process to acquire that lock will hang forever.

I have observed this in practice in a couple of situations, and issue #1467 is a note that caution needs to be taken in a third situation.

The first, most serious, case that I have observed in practice both in DESC, on perlmutter at NERSC and CSD3 at Cambridge, is a deadlock in glibc's `nsswitch` lock in `nsswitch.c`. In this case, in some circumstances when starting up, the WorkQueueExecutor hangs performing a name lookup.

I added debug statements in nsswitch.c around locks taken in `_nss_lookup_function` that show that this is because `_nss_lookup_function` cannot aquire a lock in the forked work queue process, due to it being taken by another parsl component in the parent process at fork time.

The second case I have observed occured when adding log statements around startup. In that case, I did not commit those log statements to master parsl, and moved onto other things.

**To Reproduce**
There is a fairly convoluted reproduction of the DESC WorkQueuExecutor hang on perlmutter documented in a NERSC ticket, produced by Jim Chiang.

**Expected behavior**

Parsl components should not hang.

It might be possible to ensure that there are no threads (or at least threads that take any locks) beyond the main thread during a parsl startup stage where it would then be safe to fork processes; with any necessary threads launched in all components after all components have had a chance to fork.

This would mean that components could not be started up individually and ad-hoc, but that the startup of all components would need to be managed by parsl in two stages.

This separation into fork/thread stages might not be possible - I haven't looked at the code with that in mind yet.

benclifford:
In the short term, I have applied a bodge `time.sleep(15)` at a relevant place to peturb a race condition caused by this issue experienced on perlmutter (and hopefully), in the `desc` branch.

benclifford:
related to this, #2628 and #2629 show some other fork-related parent-process-inherited bad behaviour, this time with signal handlers

benclifford:
cpython has this issue about moving the default to spawn rather than fork, which is related to this.
https://github.com/python/cpython/issues/84559

benclifford:
Places where multiprocess is used at the moment in parsl:

htex: interchange started - forced to be a ForkProcess

[...]

PR #2650, opened 2023-04-14
Deprecate unsafe fork multiprocess in process_worker_pool

benclifford:
This is part of issue #2343, to remove/reduce the use of the unsafe fork multiprocessing start method throughout Parsl.

## Type of change


WardLT (approved):
Looks good to me. I do like spawn as a default over fork

PR #2675, opened 2023-05-02
Support Python 3.12

benclifford:
This PR makes Parsl support Python 3.12.

* The pandas dependency for visualization is loosened to allow more recent pandas version.

* The CI python build step is upgraded, and Python 3.12 is added

**Original description:**

Python 3.12 beta 1 is expected soon. Usually with a new python version, there are a handful of things that break or change. This branch is to discover those, ideally before Python 3.12.0.

See https://peps.python.org/pep-0693/ for Python 3.12 release timetime

## Type of change

- Code maintentance/cleanup


benclifford:
right now this build is failing for pyzmq and greenlet, both in compiling some C code that references the cpython api

benclifford:
pandas (used by monitoring viz and some tests) doesn't build because `distutils` is gone in python 3.12 and it seems to want to build that way.

i removed it (so expect monitoring viz and those tests to fail, but this is too see how far along the build can get.

back to failing with pyzmq building greenlets C API stuff I encountered before

benclifford:
for greenlets, see issue/likely fix here: https://github.com/python-greenlet/greenlet/issues/357

benclifford:
flake8 6.0.0 installs ok, but it is very confused about source code structure - for example, it's seems to be regarding multiline docstrings as a single extra-long line, which is then not permitted due to line length linting.

and many blank line at the end of file errors when there is no blank line there.

benclifford:
numpy <=1.25 does not build on python 3.12rc1 because it still uses distutils. numpy has some details on fixing this here https://github.com/numpy/numpy/issues/23808

khk-globus (approved):

PR #2678, opened 2023-05-03
Fix decorator typing to wrap_errors using ParamSpec

benclifford:
This is a feature of typing that has become available since this code was written.

## Type of change

### DIFF
--- a/parsl/app/errors.py
+++ b/parsl/app/errors.py
@@ -1,6 +1,6 @@
 """Exceptions raised by Apps."""
 from functools import wraps
-from typing import Callable, List, Union, Any, TypeVar, Optional
+from typing import Callable, List, Optional, ParamSpec, TypeVar, Union
 from types import TracebackType
 import logging
 from tblib import Traceback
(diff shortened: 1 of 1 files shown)


benclifford:
This is failing due to a conflict between `dill` and ParamSpec:

```
... long trace inside dill ...
  File "/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/typing.py", line 399, in __init_subclass__
    raise TypeError("Cannot subclass special typing classes")
    ^^^^^^^^^^^^^^^^^
```

This doesn't happen in a simple test serialising a function that has been annotated with a ParamSpec, so I need to dig in deeper to understand what's going on.

khk-globus (commented):
Ah, misclicked and ... where did my comment go?  Sigh.  I had not intended to Approve but rather Comment.  The change _looks_ good, but was going to inquire about the status of the failure you had identified two weeks ago.

benclifford:
this is, mysteriously, no longer giving the mysterious type error.

benclifford:
@khk-globus I guess without being able to recreate the test failure before that I did not understand, I'll merge this. I'm a bit suspicious though...

benclifford:
I'm now seeing that dill ` annot subclass special typing classes` error on my laptop (which is interesting, because I previously did not see it, when making the PR!)

benclifford:
I can recreate this error by forcing typing_extensions == 4.5.0, and bisecting against the typing extensions repo, I see that this error goes away with this PR: https://github.com/python/typing_extensions/pull/162

So I'll make a new PR to have a lower bound for the first typing-extensions release that includes that PR, version 4.6.0

PR #2685, opened 2023-05-10
Remove __main__ blocks from test_*

benclifford:
These tests are all run through pytest, rather than executing as a main process, and the __main__ blocks are often broken which causes problems for static analysis with mypy, and with human understanding of how a test is meant to work.

Previous PRs have removed __main__ blocks from tests individually, but this PR does the entire test_* tree.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
That was a quick one.  Good call.

PR #2686, opened 2023-05-10
Switch configuration executor List type to Sequence

benclifford:
The list of executors in configuration is intended to be read only. Using Sequence for this makes that more explicit, as Sequences cannot be modified.

This is part of a move towards using Sequence rather than List in many places (although the runtime type remains a list) because Sequence has better behaviour with subclasses of its parameter type: it is covariant, rather than invariant.

This PR rearranges the validation logic a bit so that there is no longer an arbitrary property setter for the executor property - it was always a bug to replace the executor list in a config once parsl has been configured, and removing this property setter enforces that more strongly.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/config.py
+++ b/parsl/config.py
@@ -1,7 +1,7 @@
 import logging
 import typeguard
 
-from typing import Callable, List, Optional, Sequence, Union
+from typing import Callable, Optional, Sequence, Union
 from typing_extensions import Literal
 
 from parsl.utils import RepresentationMixin
@@ -20,8 +20,8 @@ class Config(RepresentationMixin):
 
     Parameters
     ----------
-    executors : list of ParslExecutor, optional
-        List of `ParslExecutor` instances to use for executing tasks.
+    executors : sequence of ParslExecutor, optional
+        List (or other sequence) of `ParslExecutor` instances to use for executing tasks.
         Default is [:class:`~parsl.executors.threads.ThreadPoolExecutor()`].
     app_cache : bool, optional
         Enable app caching. Default is True.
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -1096,6 +1096,9 @@ def _create_remote_dirs_over_channel(self, provider: ExecutionProvider, channel:
 
     def add_executors(self, executors):
         for executor in executors:
+            if executor.label in self.executors:
+                raise ConfigurationError('Executor must have unique label, but {executor.label} is duplicated.')
+
```
khk-globus, line 1101: f-string?

parsl/config.py
```diff
@@ -92,9 +91,11 @@ def __init__(self,
                  monitoring: Optional[MonitoringHub] = None,
                  usage_tracking: bool = False,
                  initialize_logging: bool = True) -> None:
+        self._executors: Sequence[ParslExecutor]
         if executors is None:
-            executors = [ThreadPoolExecutor()]
-        self.executors = executors
+            self._executors = [ThreadPoolExecutor()]
+        else:
+            self._executors = executors
```
khk-globus, line 98: Other than the typing and privatization, I don't see the need for the logic change.  The only change of substance is the underscore prefix; mypy should figure out the typing from the function signature with:
```python
if executors is None:
    executors = (ThreadPoolExecutor(),)  # still in keeping with signature
self._executors = executors  # Now set, per signature; no explicit typing necessary
```

benclifford:
this was harder than i expected - i might get back round to it another time but it's not super high priority (as the inactivity shows)

PR #2695, opened 2023-05-15
Eliminate FlowControl class

benclifford:
Recent PRs have rearranged the FlowControl class so that all that is left is delegation to the Timer and JobStatusPoller classes. This PR removes that final bit of delegation and uses JobStatusPoller and Timer directly.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Nice, this makes sense.

PR #2698, opened 2023-05-15
Rename design and rationale docs as historical

benclifford:
This better reflects the place this document has in the current ecosystem.

## Type of change

- Documentation update


docs/devguide/design.rst
```diff
@@ -1,9 +1,4 @@
-Design and Rationale
-
-
-
-Swift vs Parsl
---------------
+Historical: Swift vs Parsl
```
khk-globus, line 1: Looks like we forgot to keep the heading moniker.

benclifford: added

khk-globus (approved):

PR #2704, opened 2023-05-15
Remove disabled but also not-useful flow control test

benclifford:
This test was disabled but before it was disable, the comments suggested that it tested that running a test would get it to test that a block was somehow controlled the flow control. This isn't what was happening - it tested a task could run, using the block started by init_blocks=1 of the test config, and the test made no assertions about the behaviour of blocks or flow control.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Hah; it does seem a _basic_ test (haha!) of a test is "does it include an `assert` statement?!"

PR #2711, opened 2023-05-15
Allow executors specification to be a single string

benclifford:
This PR also prohibits executors from being labelled 'all', which is backwards incompatible (although probably minor). This prohibition is because the specification string 'all' for all executors now overlaps the executor label namespace, and so ambiguity would be possible.

This PR is intended to address minor syntactic noise in the common case of a single executor being specified:

Before:
@python_app(executors=['SuperComputer1'])

After:
@python_app(executors='SuperComputer1')

This PR also adds test cases around the syntax for executors.

## Type of change


parsl/tests/test_python_apps/test_executor_selector.py
```diff
@@ -0,0 +1,45 @@
...
+@pytest.mark.local
+def test_executor_invalid() -> None:
+    with pytest.raises(ValueError):
+        app_executor_invalid().result()
+
+
+@parsl.python_app(executors='all')
+def app_executor_all():
+    return 10
+
```
khk-globus, line 41: I might add one more that _fails_ if given `executors=['all']`.

khk-globus (approved):
Looks, on the whole, simple enough.  One suggestion is to add an additional test to verify that while `'all'` is valid, `['all']` is not.  Or similarly, `['valid1', 'valid2', 'all', 'valid3']` should fail too, right?

But not a deal breaker.

benclifford:
abandoned, not interesting enough to pick up, closing

PR #2712, opened 2023-05-16
Remove DFK parameter from strategy to facilitate use without DFK

benclifford:
For example, the funcX project to use scaling with their use of the HighThroughputExecutor without a DFK.

This PR makes a behavioural change: executors from the DFK are no longer added in the strategy initialiser. The DFK already adds a executors via JobStatusPoller: prior to this PR, this meant executors were added to the strategy twice, but after this PR, this means that executors are only added once - by the job status poller.

## Type of change


yadudoc (approved):
Looks good to me :)

PR #2715, opened 2023-05-19
Move an idempotent loop out of another loop in serialization

benclifford:
This inner loop populates a table of data serializers, which should happen at the same level as populating the table of code serializers, rather than as an inner loop of code serializer population.

This inner loop is idempotent, so the behaviour should mostly be unchanged by this PR, but in the obscure situation that someone modifies serialisation so that it has no code serializers, only data serializers, then before this PR, I think those data serializers would not be visible to users, as the inner loop would never be executed.

This was introduced by PR #2467 which moved away from using a class singleton to using the module as the singleton.

## Type of change


khk-globus (approved):
Oof; subtle find.

Issue #2681, opened 2023-05-04
KeyError raised when visualizing workflow graph with tasks in 'running_ended' state


mxk62:
**Describe the bug**

Clicking "View workflow DAG -- colored by task states" on workflow summary page results in an "Internal Server Error" if there are tasks in `running_ended` state. When run in the debug mode, `parsl-visualize` tool shows that the error is caused by **KeyError** exception raised by `workflow_dag_plot()` function. See the traceback below for more details:

```
Traceback (most recent call last):
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 2551, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 2531, in wsgi_app
    response = self.handle_exception(e)
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 2528, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/mxk/.conda/envs/devenv/lib/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/mxk/lsst/git/parsl/parsl/monitoring/visualization/views.py", line 146, in workflow_dag_details
    workflow_dag_plot=workflow_dag_plot(df_tasks, group_by_apps=group_by_apps))
  File "/home/mxk/lsst/git/parsl/parsl/monitoring/visualization/plots/default/workflow_plots.py", line 259, in workflow_dag_plot
    index, _ = groups_list[name]
KeyError: 'running_ended'
```

**To Reproduce**

1. Setup Parsl 2023.05.01 with Python 3.10 on a desktop PC.
2. Start a Parsl workflow using HTEX with the monitoring enabled.
3. Start the web dashboard using `parsl-visualize` tool.
4. Open the dashboard in a web browser and select the running workflow to get to the workflow summary page.
[...]

PR #2722, opened 2023-05-30
Add test that visualization coloring tables color each known state

benclifford:
This is a regression test for PR #2708, issue #2681, that will also detect other new states added to the states enumeration without being added to the coloring tables.

In order to be able to test this way, this PR moves the two coloring tables to be module scope, instead of local scope inside a method.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Love it.  Thanks for that additional "did we make a change and remember to test it?" test.

PR #2729, opened 2023-06-02
Add version constraints on typing-extensions

benclifford:
Parsl PR #2678 breaks with typing-extensions==4.5.0. See that PR for some further information. This PR adds a lower bound to be later than 4.5.0, and also a major version upper bound to work with typing-extensions Semantic Versioning.

## Type of change


khk-globus (approved):
Subtle find; nice.

PR #2744, opened 2023-06-14
Upgrade flake8 after PR #2737

benclifford:
This should have been done in PR #2737 which made fixes, but the actual upgrade was omitted by mistake.

This PR also removes a version bound on importlib-metadata that isn't needed by more recent versions of flake8.

## Type of change

- Code maintentance/cleanup


benclifford:
this fails with python 3.7 as recent flake8 requires python 3.8. Python 3.7 is end of life in a couple of weeks, so I'll wait for Parsl to drop Python 3.7 support on that day, and then try this PR again...

benclifford:
PR #2749 drops python 3.7, and after that, this PR should be OK.

khk-globus (approved):

PR #2745, opened 2023-06-17
Update more of Globus data provider to Globus Transfer v3 API

KiwiFac3:
# Description

Some Globus functions have different arguments that was previously indicated. I changed those function to have the correct parameters and did the necessary changes to the code to maintain it functionality.

Fixes # This change should fix the Globus API changes 

## Type of change


kylechard (approved):
Tested with various files, sizes, errors. Looks good. One thing to consider in the future is what we should do with failures. Here, Parsl will wait forever if the transfer has an unrecoverable error. In the future we might want to detect different failure conditions and at notify users.

PR #2746, opened 2023-06-19
Mark performance chapter of userguide as historical

benclifford:
This contains performance comparisons of a very old htex (<= June 2020) against two executors which are no longer supported: extreme scale and ipp (IPyParallel).

## Type of change

- Documentation update
- Code maintentance/cleanup


docs/userguide/performance.rst
```diff
@@ -1,7 +1,7 @@
 .. _label-performance:
 
-Performance and Scalability
-===========================
+Historical: Performance and Scalability
+=======================================
```
khk-globus, line 5: Is there any value in giving a sentence or two of context of why it's historical?  When I'm new to a subject or project, I'm sometimes frustrated to see "historical" because the now-old way of doing things often matches my naive understanding of the data structures or expected workflow.

benclifford: I added a justification note here.

khk-globus (approved):

PR #2748, opened 2023-06-19
Rework BadLauncher exception

benclifford:
This PR:

 * adds type annotations to the initializer - from benc-mypy branch

 * uses str instead of repr for human readable text:

   repr which should print machine readable "expression-like" representations, and str is used for the human-readable description in exception messages. Prior to this PR, this exception was rendered as:

       File "<stdin>", line 1, in <module>
       parsl.launchers.errors.BadLauncher: (None, 'This is the error text')

   using the default str, and the human readable text only appears in contexts where expression-like repr output was expected.

   After this PR, the change to means that this exception is rendered as: parsl.launchers.errors.BadLauncher: Bad Launcher provided: None Reason: This is the error text

   Crossreference other str/repr PRs such as: a423955f4a9e03cf6986a6e21d285cf46fa3bc88, PR #1966

 * adds TypeError as a superclass, as this exception class is only ever used to indicate a type error

 * removes the reason field, which only ever contained text indicating a type error, and put that text into the exception implementation. This goes alongside the above change to indicate that this class is only ever a type error.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Nice.

PR #2757, opened 2023-06-22
Fix bad punctuation in AWSProvider.instance_type docstring

benclifford:
## Type of change

- Documentation update


khk-globus (approved):
Weird mistake.

PR #2763, opened 2023-06-25
Add notes justifying why historical sections of docs have been made historical

benclifford:
This is intended to help new readers understand the context of why these sections of been kept, rather than deleted.

## Type of change

- Documentation update


khk-globus (approved):
Nice; thank you.

PR #2764, opened 2023-06-26
Cleanup python_apps/test_futures

khk-globus:
- Use `tmp_path` fixture -- no post-test detritus, please
- Reduce delay; any delay will do, so reduce unnecessary wait
- Remove `print()` clutter
- Remove pytest warnings by not returning a value from test functions
- Remove some redundancies
  - store then return ==> return
  - pytest _already_ shares pertinent data, don't dup in assert message

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
I just went over these changes with @khk-globus. We suspect `local_threads.config` was being imported from before we switched to using `conftest.py`'s `load_dfk_session` bits to load a global DFK.

PR #2770, opened 2023-06-27
Assert that the third byte of a serializer ID must be \n

benclifford:
SerializerBase.chomp() fails to parse the serializer ID when this byte is something else, so deserialization will fail with such serializers with a confusing error:

parsl.executors.errors.DeserializationError: Failed to deserialize return objects. Reason:Received exception, but handling also threw an exception: Buffer does not start with parsl.serialize identifier:b'01X'

This assertion pushes that error closer to initialization.

## Type of change

- Code maintentance/cleanup


parsl/serialize/base.py
```diff
@@ -22,6 +22,7 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
         super().__init_subclass__(**kwargs)
 
         assert len(cls._identifier) == 3
+        assert cls._identifier[2] == 10  # \n in decimal
```
khk-globus, line 25: Does anyone ever run with `python -O` ?  (Do we care?)

Meaning that use of `assert` is removed when Python's optimizer is in play.  If still wanting the assertion, need to use the Exception mechanism:
```python
if len(cls._identifier) != 3:
    raise AssertionError(...)
if cls._identifier[2] != 10:  # \n in decimal
    raise AssertionError(...)
```

benclifford: no one uses `-O`, as far as I know, and in my brief testing with it before I didn't see speedup so I haven't been pushing it.

right now these asserts are to discover programming errors by "internal" parsl code, rather than someone in a "plugin author" role, which I think a plain `assert` is right for.

Later on, I'm hoping to propose a user-exposed serializer configuration/plugin API based on experiences with proxystore and funcx, and I would imagine that would in part get rid of the awkward requirements asserted here entirely - it was experimenting with ideas for that API that highlighted this particular new secret rule that is asserted here to me.

PR #2771, opened 2023-06-27
Remove or reduce some test sleeps

khk-globus:
- Make some test code more Pythonic (e.g., generators, list comprehension)
- Remove a number of unnecessary sleep calls (or vastly reduce)
- Teach some tests to `assert`
- Remove more `print()`s

## Type of change
- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_python_apps/test_fail.py
+++ b/parsl/tests/test_python_apps/test_fail.py
@@ -1,89 +1,59 @@
-import argparse
+import pytest
 
-import parsl
 from parsl.app.app import python_app
-from parsl.tests.configs.local_threads import fresh_config as local_config
+from parsl.dataflow.errors import DependencyError
+
+
+class ManufacturedTestFailure(Exception):
+    pass
 
 
 @python_app
-def sleep_fail(sleep_dur, sleep_rand_max, fail_prob, inputs=[]):
-    import time
+def random_fail(fail_prob: float, inputs=()):
     import random
-
-    s = sleep_dur + random.randint(-sleep_rand_max, sleep_rand_max)
-
-    time.sleep(s)
-    raise Exception("App failure")
+    if random.random() < fail_prob:
+        raise ManufacturedTestFailure("App failure")
 
 
-def test_no_deps(numtasks=2):
+def test_no_deps():
     """Test basic error handling, with no dependent failures
     """
+    futs = [random_fail(1), random_fail(0), random_fail(0)]
 
-    fus = []
-    for i in range(0, numtasks):
-
-        fu = sleep_fail(0.1, 0, .8)
-        fus.extend([fu])
-
-    count = 0
-    for fu in fus:
+    for f in futs:
         try:
-            fu.result()
-        except Exception as e:
-            print("Caught exception : ", "*" * 20)
-            print(e)
-            print("*" * 20)
-            count += 1
-
-    print("Caught failures of  {0}/{1}".format(count, len(fus)))
+            f.result()
+        except ManufacturedTestFailure:
+            pass
 
 
-def test_fail_sequence(numtasks=2):
+@pytest.mark.parametrize("fail_probs", ((1, 0), (0, 1)))
+def test_fail_sequence(fail_probs):
     """Test failure in a sequence of dependencies
 
     App1 -> App2 ... -> AppN
     """
 
-    sleep_dur = 0.1
-    fail_prob = 0.4
+    t1_fail_prob, t2_fail_prob = fail_probs
+    t1 = random_fail(fail_prob=t1_fail_prob)
+    t2 = random_fail(fail_prob=t2_fail_prob, inputs=[t1])
+    t_final = random_fail(fail_prob=0, inputs=[t2])
 
-    fus = {0: None}
-    for i in range(0, numtasks):
-        print("Chaining {0} to {1}".format(i + 1, fus[i]))
-        fus[i + 1] = sleep_fail(sleep_dur, 0, fail_prob, inputs=[fus[i]])
-
-    # time.sleep(numtasks*sleep_dur)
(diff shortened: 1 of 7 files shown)


parsl/tests/test_python_apps/test_garbage_collect.py
```diff
@@ -17,14 +17,13 @@ def test_garbage_collect():
     """
     x = slow_double(slow_double(10))
 
-    if x.done() is False:
-        assert parsl.dfk().tasks[x.tid]['app_fu'] == x, "Tasks table should have app_fu ref before done"
+    assert parsl.dfk().tasks[x.tid]['app_fu'] == x, "Tasks table should have app_fu ref before done"
```
benclifford, line 20: theres a slight race condition here that the x.done() is trying to gloss over: x could have executed and been garbage collected by the time this line is reached, if there's sufficient delay here... (that race condition also existed in different form with the previous `if x.done() is False` test...)

there's no guarantee apart from hoping that time.sleep lasts long enough that this assertion really does happen before the x future is done.

what might be better for this test is running in-process using the thread local executor, and using an Event to control the completion of the test app

khk-globus: :face_palm: This was precisely the snippet I shared with you in Slack, but then realized would not work for anything other than threads (and have thus deleted the shared snippet!).

It sounds like you're suggesting to only strictly run the test in a local configuration then?

parsl/tests/test_python_apps/test_fail.py
```diff
@@ -1,89 +1,62 @@
...
+    import random
+    if random.random() < fail_prob:
+        raise ManufacturedTestFailure("App failure")
 
 
-def test_no_deps(numtasks=2):
+def test_no_deps(ntasks=3):
     """Test basic error handling, with no dependent failures
     """
+    futs = [random_fail(.8) for i in range(ntasks)]
```
benclifford, line 19: I don't see why this test needs to be randomised at all: this test seems to be hoping that some app invocations will fail in this range (otherwise, no failure handling is exercise) and so now I question why this should be done on a random basis rather than a deterministic list of (for example) two successful apps and one non-successful one.

parsl/tests/test_python_apps/test_garbage_collect.py
```diff
@@ -1,30 +1,36 @@
...
-    x.result()
+    evt.set()
+    assert x.result() == 10 * 4
     if parsl.dfk().checkpoint_mode is not None:
         # We explicit call checkpoint if checkpoint_mode is enabled covering
         # cases like manual/periodic where checkpointing may be deferred.
         parsl.dfk().checkpoint()
 
-    time.sleep(0.2)  # Give enough time for task wipes to work
+    time.sleep(0.01)  # Give enough time for task wipes to work
```
benclifford, line 35: ideally there would be no time delay here, but some post-app completion happens asynchronously with telling the user workflow that a task has completed. tagging https://github.com/Parsl/parsl/issues/1279 which discusses this further.

parsl/tests/test_python_apps/test_join.py
```diff
@@ -1,17 +1,24 @@
...
+_inner_app_called = False
+
 
 @python_app(cache=True)
 def inner_app():
-    time.sleep(1)
+    global _inner_app_called
+    if _inner_app_called:
+        raise _InnerAppCalledMoreThanOnce()
+    _inner_app_called = True
```
benclifford, line 21: probably a tight race condition here if two worker threads launch inner app very close together in time?

parsl/tests/test_python_apps/test_overview.py
```diff
@@ -42,19 +36,9 @@ def test_1(N=10):
...
-        mapped_results.append(x)
+    # Create a list of integers, then apply *app* function to each
+    items = range(N)
+    mapped_results = list(map(double, items))
 
     total = app_sum(inputs=mapped_results)
-
-    assert total.result() != sum(items), "Sum is wrong {0} != {1}".format(
-        total.result(), sum(items))
+    assert total.result() != sum(items)
```
benclifford, line 44: maybe this assertion should assert that total.result() is a specific value, not that it is not a specific value?

benclifford: (this same remark applies to the code before too, so I'm not going to hold up merging this PR because of it)

benclifford (changes requested):
my main concern is the introduction of randomization in test_fail

parsl/tests/test_python_apps/test_fail.py
khk-globus: Och, good point.  I think I was trying to match the intent of the original `random`.  But not seeing that I did a good job.  All that needs be done is removing the `.sleep()` call.

parsl/tests/test_python_apps/test_join.py
khk-globus: *nod; now that I look at the tests, I don't see that they care about the `cache=True` aspect at all.  Removing that logic entirely in lieu of another test that specifically tests the caching.

parsl/tests/test_python_apps/test_overview.py
khk-globus: Heh; had just removed the message because it wasn't providing any new information.  I must have had Covid brain last week.  Thanks for that eye.

khk-globus:
> my main concern is the introduction of randomization in test_fail

The main change is in removing the `.sleep()` call that was randomizing _when_ calls would fail.  But these tests weren't verifying anything to do with time&nbsp;&mdash;&nbsp;like that the app fails as soon as possible.  That might be an additional test, but I'm trying for these set of changes to strictly remove the `.sleep()` calls where I can.

benclifford (approved):

Issue #1875, opened 2020-09-26
Lost ability to run tests in arbitrary directory


hategan:
**Describe the bug**
I used to be able to run tests in a specific test directory (so that it doesn't pollute my main parsl directory with hundreds of output files).

This doesn't seem to work any more due to 

```
FileNotFoundError: [Errno 2] No such file or directory: 'parsl/tests/callables_helper.py'
```

It looks like this was introduced by #1852. I'd recommend something like this:

```
helper_path = pathlib.Path(__file__).parent / "callables_helper.py"
spec = importlib.util.spec_from_file_location("dynamically_loaded_module", helper_path)
```

**To Reproduce**
in parsl main dir:
* mkdir tmp
* cd tmp
* pytest ../parsl --config ../parsl/configs/local_threads.py

**Expected behavior**
I'd expect tests that would succeed when pytest is invoked from the main parsl dir to also succeed when invoked as above

PR #2773, opened 2023-06-27
Cleanup some test file detritus

khk-globus:
- Use the well-known `tmp_path` fixture.
- Remove duplicated tests

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_data/test_file_apps.py
+++ b/parsl/tests/test_data/test_file_apps.py
@@ -1,38 +1,33 @@
-import os
-
 import pytest
 
-import parsl
 from parsl.app.app import bash_app
 from parsl.data_provider.files import File
-from parsl.tests.configs.local_threads import config
 
 
 @bash_app
-def cat(inputs=[], outputs=[], stdout=None, stderr=None):
-    infiles = ' '.join([i.filepath for i in inputs])
-    return """echo {i}
-    cat {i} &> {o}
-    """.format(i=infiles, o=outputs[0])
+def cat(inputs=(), outputs=(), stdout=None, stderr=None):
+    infiles = " ".join(i.filepath for i in inputs)
+    return f"cat {infiles} &> {outputs[0]}"
 
 
-@pytest.mark.usefixtures('setup_data')
 @pytest.mark.staging_required
-def test_files():
-
-    if os.path.exists('cat_out.txt'):
-        os.remove('cat_out.txt')
-
-    fs = [File('data/' + f) for f in os.listdir('data')]
-    x = cat(inputs=fs, outputs=[File('cat_out.txt')],
-            stdout='f_app.out', stderr='f_app.err')
+def test_files(setup_data):
+    fs = sorted(str(setup_data / f) for f in setup_data.iterdir())
+    fs = list(map(File, fs))
+    x = cat(
+        inputs=fs,
+        outputs=[File(str(setup_data / "cat_out.txt"))],
+        stdout=str(setup_data / "f_app.out"),
+        stderr=str(setup_data / "f_app.err"),
+    )
+    x.result()
     d_x = x.outputs[0]
-    print(x.result())
-    print(d_x, type(d_x))
+    data = open(d_x.filepath).read().strip()
+    assert "1\n2" == data, "Per setup_data fixture"
 
 
 @bash_app
-def increment(inputs=[], outputs=[], stdout=None, stderr=None):
+def increment(inputs=(), outputs=(), stdout=None, stderr=None):
     # Place double braces to avoid python complaining about missing keys for {item = $1}
     return """
     x=$(cat {i})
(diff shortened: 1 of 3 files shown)


parsl/tests/test_data/test_file_apps.py
```diff
@@ -1,74 +1,60 @@
...
 
 @bash_app
-def cat(inputs=[], outputs=[], stdout=None, stderr=None):
-    infiles = ' '.join([i.filepath for i in inputs])
-    return """echo {i}
-    cat {i} &> {o}
-    """.format(i=infiles, o=outputs[0])
+def cat(inputs=(), outputs=(), stdout=None, stderr=None):
+    infiles = " ".join(i.filepath for i in inputs)
+    return f"cat {infiles} &> {outputs[0]}\n"
```
benclifford, line 10: i think this should be fine without that final \n - there isn't a requirement that bash_apps put a \n on the end of the command they return to execute

parsl/tests/test_data/test_file_apps.py
```diff
@@ -1,74 +1,60 @@
...
-
-    if os.path.exists('cat_out.txt'):
-        os.remove('cat_out.txt')
-
-    fs = [File('data/' + f) for f in os.listdir('data')]
-    x = cat(inputs=fs, outputs=[File('cat_out.txt')],
-            stdout='f_app.out', stderr='f_app.err')
+def test_files(setup_data):
+    fs = sorted(str(setup_data / f) for f in setup_data.iterdir())
+    fs = list(map(File, fs))
```
benclifford, line 16: does this need to be forced into being a list? or can it be left as a generator like object?
(I don't actually know the answer because the history of inputs= is that it was a list, and more recently my type checking work has pushed a lot of this stuff to be arbitrary `Sequence` objects... but all of that has been working on the assumption that what is supplied would be value-like (in the sense that you can observe multiple times... which you can't do with generator-like objects).
It's something to probably keep an eye on anywhere generator-like stuff gets passed into parsl

parsl/tests/test_data/test_file_apps.py
```diff
@@ -1,74 +1,60 @@
...
-    print(x.result())
-    print(d_x, type(d_x))
+    data = open(d_x.filepath).read().strip()
+    assert "1\n2" == data, "Per setup_data fixture"
 
 
 @bash_app
-def increment(inputs=[], outputs=[], stdout=None, stderr=None):
+def increment(inputs=(), outputs=(), stdout=None, stderr=None):
     # Place double braces to avoid python complaining about missing keys for {item = $1}
```
benclifford, line 31: I think this comment can go away too? I don't see any double braces

parsl/tests/test_data/test_file_staging.py
```diff
@@ -1,88 +1,25 @@
...
-@bash_app
-def increment(inputs=[], outputs=[], stdout=None, stderr=None):
-    # Place double braces to avoid python complaining about missing keys for {item = $1}
-    return """
-    x=$(cat {i})
-    echo $(($x+1)) > {o}
-    """.format(i=inputs[0], o=outputs[0])
+def cat(inputs=(), outputs=(), stdout=None, stderr=None):
+    infiles = " ".join(i.filepath for i in inputs)
+    return f"cat {infiles} &> {outputs[0]}\n"
```
benclifford, line 10: no \n here - it shouldn't harm anything though

parsl/tests/test_data/test_file_staging.py
```diff
@@ -1,88 +1,25 @@
...
-        fu = increment(inputs=[prev],  # Depend on the future from previous call
-                       # Name the file to be created here
-                       outputs=[File("test{0}.txt".format(i))],
-                       stdout="incr{0}.out".format(i),
-                       stderr="incr{0}.err".format(i))
-        [prev] = fu.outputs
-        futs[i] = prev
-        print(prev.filepath)
+    fpath.write_text("Hello World")
+    f = cat(inputs=[File(str(fpath))], outputs=[File(str(opath))])
```
benclifford, line 20: it would make sense for File() to accept a pathlike-object, rather than fpath being cast to a string, I think... apparently it isn't, though. (the original implementation predates the pathlike interface which was introduced in Python 3.6, although time has moved on enough that pathlike exists in all supported Python versions now...)

no need for a change here, this is a just rambling.

benclifford: I noted this in an issue https://github.com/Parsl/parsl/issues/2785

benclifford (approved):

benclifford:
crossref issue https://github.com/Parsl/parsl/issues/1875 which talks about running tests without dumping files into undesirable places

benclifford:
I made a bunch of comments on this PR, but they don't need resolving before this PR gets merged: maybe they're just for interest, maybe for another PR.

parsl/tests/test_data/test_file_apps.py
khk-globus: Removed in a follow-up PR.

parsl/tests/test_data/test_file_staging.py
khk-globus: Addressed in a follow-up PR.

parsl/tests/test_data/test_file_staging.py
khk-globus: Upon merge, address via [PR #2806 "Teach File to be PathLike aware"](https://github.com/Parsl/parsl/pull/2806).

parsl/tests/test_data/test_file_apps.py
khk-globus: I don't know.  When I first approached the set of test changes I was making these past two weeks, my onus was on keeping the spirit of what came before and "just" removing the painful `time.sleep()` calls.  In this case, "what came before" was:

```python
fs = [File('data/' + f) for f in os.listdir('data')]
```

which is definitely a list.  I'm happy to consider future Pythonic-focused changes, but I'm new enough to Parsl's code base that I'm currently being tepid.

Going forward, I'll keep your question in mind.

PR #2777, opened 2023-06-28
Show top-10 slow tests for each run

khk-globus:
`pytest` has a handy builtin, `--durations <num>`.

## Type of change

- Code maintenance/cleanup


Makefile
```diff
@@ -84,7 +84,7 @@ site_test:
 
 .PHONY: perf_test
 perf_test:
-	parsl-perf --time 5 --config parsl/tests/configs/local_threads.py
+	parsl-perf --time 5 --config parsl/tests/configs/local_threads.py --durations 10
```
benclifford, line 87: this is not pytest

khk-globus: Och, of course.  "Going too fast."

PR #2778, opened 2023-06-28
Reduce depth to work with WQ for delayed tests

khk-globus:
Based on analysis by BenC with WorkQueue, noted that the parametrization of this test hits WQ's weakspot: depth.  So, reduce it's depth for our tests. 

## Type of change

- Code maintenance/cleanup


benclifford:
I tested this on my laptop work queue setup: test_type5 takes 57s with this PR, 91s with `HEAD^` = b91aa49718a39c712529b2983354cc054ca65fe2

PR #2779, opened 2023-06-28
Turn on Work Queue coprocess mode

benclifford:
This should probably be the preferred way of using the work queue executor, as it can speed things up incredibly on shared filesystem supercomputers, and even on my laptop can reduce WQ pytest run time from around 550s to 350s.

## Type of change

?


khk-globus (approved):

benclifford:
this hung in CI trying to run WQ tests... further investigation needed so converting to draft

benclifford:
in CI runinfo logs:

003$ cat ./submit_scripts/parsl.WorkQueueExecutor.block-0.1688148747.9381468.sh.err
sh: 1: parsl/executors/workqueue//parsl_coprocess.py: not found

so this looks like a packaging/install issue of some kind

benclifford:
subsequent fiddling has shown there are some subtleties with the PATH mechanism set up by the Makefile... i'll make a separate PR for this first

benclifford:
see PR #2804 for build PATH simplifications which should make coprocesses work

PR #2784, opened 2023-07-01
Parse serialization headers only once

benclifford:
Prior to this PR, the serialization header was parsed twice during deserialization, and in two different ways, leading to an awkward header format: the header must be exactly three bytes long for one (length based) parser, and the third byte must be a \n for the other (new-line based) parser.

This PR removes the length based parser, instead allowing an arbitrary length \n-terminated header.

Backwards/forwards compatibility:
The parsl serialization wire format is not a user exposed interface, as parsl requires the same version of parsl to be installed at all locations in a parsl deployment.

This PR does not change any on the wire byte sequences when used with the two existing de*serializers, but opens the opportunity for future implementations to use more than two bytes for identifiers.

Performance:
I made a basic benchmark of serializing and deserializing a few thousand integers (deliberately using a simple object like `int` so that the object processing would be quite small, allowing changes in header time more chance to show up). My main concern with this PR is that I didn't make things noticeably worse, not to improve performance.

After this PR, a serialization->deserialization round trip is about 200ns faster (1165ns before vs 965ns after), so I am happy that this PR does not] damage performance.

Future:
This is part of ongoing work to introduce user pluggable de*serializers.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/serialize/facade.py
+++ b/parsl/serialize/facade.py
@@ -10,7 +10,6 @@ logger = logging.getLogger(__name__)
 """ Instantiate the appropriate classes
 """
 headers = list(METHODS_MAP_CODE.keys()) + list(METHODS_MAP_DATA.keys())
-header_size = len(headers[0])
 
 methods_for_code = {}
 methods_for_data = {}
@@ -70,7 +69,7 @@ def serialize(obj: Any, buffer_threshold: int = int(1e6)) -> bytes:
     if callable(obj):
         for method in methods_for_code.values():
             try:
-                result = method.serialize(obj)
+                result = method._identifier + b'\n' + method.serialize(obj)
             except Exception as e:
                 result = e
                 continue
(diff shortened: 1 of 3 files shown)


parsl/serialize/facade.py
```diff
@@ -79,7 +76,7 @@ def serialize(obj: Any, buffer_threshold: int = int(1e6)) -> bytes:
     else:
         for method in methods_for_data.values():
             try:
-                result = method.serialize(obj)
+                result = method._identifier + b'\n' + method.serialize(obj)
             except Exception as e:
                 result = e
                 continue
```
khk-globus, line 82: As an aside, these two branches are soooo-close.  An opportunity to DRY:

```python
serde_methods = methods_for_code if callable(obj) else methods_for_data
for method in serde_methods.values():
    try:
        result = method._identifier + b'\n' + method.serialize(obj)
        break
    except Exception as e:
        result = e
        continue
```

khk-globus (approved):

PR #2798, opened 2023-07-04
Explain how to use functions from libraries

WardLT:
# Description

The documentation does not explain that you can create Apps outside of the decorator-based mechanism. 
It's a convenient route when you just want to make a "Parsl version" of a function already in a module,
and create Parsl functions programatically.

## Type of change

- Documentation update

## Context

Was reviewing code from a user and saw functions similar to:

```python

@python_app
def f(x): 
   from m import f
   return f(x)
```

I suggested they replace with

```python
from m import f
f = python_app(f)
```

to make the code more succinct.

It wasn't apparent in the docs that you could do this. I know you can because I'm familiar with function serialization.

### DIFF
--- a/docs/userguide/apps.rst
+++ b/docs/userguide/apps.rst
@@ -53,7 +53,6 @@ The following alternative formulation is valid Parsl.
 
 .. code-block:: python
 
-       import random
        factor = 5
 
        @python_app
@@ -71,7 +70,9 @@ In this case, Parsl will establish a dependency between the two apps and will no
 execute the dependent app until all dependent futures are resolved.
 Further detail is provided in :ref:`label-futures`.
 
-A Python app may also act upon files. In order to make Parsl aware of these files, they must be specified by using the ``inputs`` and/or ``outputs`` keyword arguments, as in following code snippet, which copies the contents of one file (``in.txt``) to another (``out.txt``).
+A Python app may also act upon files. In order to make Parsl aware of these files,
+they must be specified by using the ``inputs`` and/or ``outputs`` keyword arguments,
+as in following code snippet, which copies the contents of one file (``in.txt``) to another (``out.txt``).
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/userguide/apps.rst
```diff
@@ -112,7 +141,7 @@ Limitations
 There are some limitations on the Python functions that can be converted to apps:
 
 1. Functions should act only on defined input arguments. That is, they should not use script-level or global variables.
-2. Functions must explicitly import any required modules.
+2. Functions must explicitly import any required modules if they are defined in script which starts Parsl.
```
benclifford, line 144: i'm fairly certain there's more subtlety here...

docs/userguide/apps.rst
```diff
@@ -98,6 +99,34 @@ Any Parsl app (a Python function decorated with the ``@python_app`` or ``@bash_a
...
+
+.. code-block:: python
+
+    from module import function
+    function_app = python_app(function, executors='all')
+
+``function_app`` will act as Parsl App function of ``function``.
+
+It is also possible to create wrapped versions of functions, such as ones with pinned arguments.
+Parsl just requires first calling :meth:`~functools.update_wrapped` with the wrapped function
```
benclifford, line 116: what happens if you don't do this?

benclifford (approved):

docs/userguide/apps.rst
WardLT: How so? 

I'm fairly sure it works this way for functions defined in other files. The  main Parsl process will serialize a reference to the location of the function, and the worker will import those libraries as it loads in the function.

That said, my knowledge of Parsl's function serialization is empirical. I don't understand the underlying codebase, just how it has worked for me in practice.

docs/userguide/apps.rst
WardLT: I should test to be sure. I think we get errors when a function lacks a name, but might be confused with those errors originating from a code which uses Parsl and not Parsl itself.

docs/userguide/apps.rst
benclifford: pickle does that, but dill does more complex stuff (including changing behaviour depending on where on the filesystem your Other Files are located...) that is hard to keep in my head

PR #2800, opened 2023-07-05
Remove callables-as-Launchers for better type checking

benclifford:
from benc-mypy branch

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Good call; nice find.

PR #2804, opened 2023-07-06
Simplify install PATH/PYTHONPATH in Makefile

benclifford:
Prior to this PR, PATH and PYTHONPATH were set to point into the github checkout of parsl.

Both of these paths were set to relative paths, rather than absolute paths.

This caused a specific problem with workqueue coprocesses, which uses $PATH to get a canonical path for the parsl coprocess worker, but (because of the relative path in $PATH) gets a relative canonical path (parsl/executors/...) which then fails to work when invoked from other working directories.

This PR removes that PATH and most PYTHONPATH setting, and relies on Parsl being installed by `pip` into existing system paths.

This PR also corrects one target (htex) so that it actually installs Parsl as needed (and as happens for several other build targets).

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Och; when pulling specific lines from `Makefile` for local testing, the `PYTHONPATH=. ...` had been bothering me.  Thanks for this!

PR #2805, opened 2023-07-06
Add a readthedocs config, to fix bad python version and because it will soon be mandatory

benclifford:
Prior to this PR, readthedocs was using Python 3.7 by default to build the docs. This became unsupported in parsl in PR #2749 and broken in Parsl in PR #2744.

This PR adds a configuration file to use Python 3.11, the latest supported version of Parsl, to build on readthedocs.

This configuration file will become mandatory in readthedocs in September 2023, according to their own documentation, so this file would need adding soon anyway.

See https://readthedocs.org/projects/parsl/builds/21218011/ for evidence that this builds correctly.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,22 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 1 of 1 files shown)


WardLT (approved):
Maybe add in the https://docs.readthedocs.io/en/stable/config-file/v2.html#formats for PDF? Otherwise, LGTM

Issue #2785, opened 2023-07-01
File() should accept a Pathlike


benclifford:
**Is your feature request related to a problem? Please describe.**
When you have a local filename, you can write:

```
File("/home/benc/myfile")
```

to get a `File` object that refers to that file on the local filesystem.

When you have a `PathLike` object instead of a `str`, you cannot pass that into `File`.

Instead you need to turn it into a string like this:

```
File(str(my_path))
```

This is a bit ugly - the `str` call could be eliminated by allowing the `PathLike` to be passed in directly.

I think this might be a simple change at the start of `File.__init__`

**Additional context**
I noticed this during review of PR #2773

PR #2806, opened 2023-07-06
Teach File to be PathLike aware

khk-globus:
This is a follow-up commit after a couple of review comments in [PR #2773](https://github.com/Parsl/parsl/pull/2773) (509e8e2761d7) and the ensuing issue.

Fixes: #2785

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/data_provider/files.py
+++ b/parsl/data_provider/files.py
@@ -5,11 +5,11 @@ to transfer the file as well as to give the appropriate filepath depending
 on where (client-side, remote-side, intermediary-side) the File.filepath is
 being called from.
 """
-
 import os
+
 import typeguard
 import logging
-from typing import Optional
+from typing import Optional, Union
 from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


parsl/data_provider/files.py
```diff
@@ -28,17 +28,18 @@ class File:
...
+           - url (string or PathLike) : url of the file e.g.
               - 'input.txt'
+              - pathlib.Path('input.txt')
               - 'file:///scratch/proj101/input.txt'
               - 'globus://go#ep1/~/data/input.txt'
               - 'globus://ddb59aef-6d04-11e5-ba46-22000b92c6ec/home/johndoe/data/input.txt'
         """
-        self.url = url
+        self.url = str(url)
         parsed_url = urlparse(self.url)
```
benclifford, line 43: There's an __fspath__ dunder-method on PathLike... That can return either bytes or str. And is right next to some functions that encode things to the filesystem encoding, stc
I suspect deeper investigation is an awful can of worms.

benclifford (approved):

PR #2808, opened 2023-07-07
Reduce test_scale_out delay

### LINKED ISSUE #1885: Possible race condition in htex scale down

khk-globus:
Locally, I was consistently seeing this test take 56s.  This commit reduces that to 12s locally.

There are two main changes:

- Bind to 127.0.0.1, rather than to all live interfaces.  Some of the delay was from the network stack

- Replace the littered `time.sleep()` calls with an "Event" based approach (polled files on the filesystem)

The other addition of note is `try_assert()`.  This my approximation of JUnit's `assertTimeout` or QT's `QTRY_VERIFY`: continually try a test condition for a timeout period.

I'm not able to discern at this time if this robustly resolves #1885.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -1,12 +1,16 @@
 import importlib.util
+import itertools
 import logging
 import os
 import pathlib
+import time
+import types
 import signal
 import sys
 import tempfile
 import threading
 import traceback
+import typing as t
 from datetime import datetime
 from glob import glob
 from itertools import chain
@@ -290,3 +294,67 @@ def pytest_ignore_collect(path):
         return True
     else:
         return False
+
+
+def create_traceback(start: int = 0) -> t.Optional[types.TracebackType]:
+    """
+    Dynamically create a traceback.
+
+    Builds a traceback from the top of the stack (the currently executing frame) on
+    down to the root frame.  Optionally, use start to build from an earlier stack
+    frame.
+
+    N.B. uses `sys._getframe`, which I only know to exist in CPython.
+    """
+    tb = None
+    for depth in itertools.count(start + 1, 1):
+        try:
+            frame = sys._getframe(depth)
+            tb = types.TracebackType(tb, frame, frame.f_lasti, frame.f_lineno)
+        except ValueError:
+            break
+    return tb
+
+
+@pytest.fixture
+def try_assert():
+    def _impl(
+        test_func: t.Callable[[], bool],
+        fail_msg: str = "",
+        timeout_ms: float = 5000,
+        attempts: int = 0,
+        check_period_ms: int = 20,
+    ):
+        tb = create_traceback(start=1)
+        timeout_s = abs(timeout_ms) / 1000.0
+        check_period_s = abs(check_period_ms) / 1000.0
+        if attempts > 0:
+            for _attempt_no in range(attempts):
+                if test_func():
+                    return
+                time.sleep(check_period_s)
+            else:
+                att_fail = (
+                    f"\n  (Still failing after attempt limit [{attempts}], testing"
+                    f" every {check_period_ms}ms)"
+                )
+                exc = AssertionError(f"{str(fail_msg)}{att_fail}".strip())
+                raise exc.with_traceback(tb)
+
+        elif timeout_s > 0:
+            end = time.monotonic() + timeout_s
+            while time.monotonic() < end:
+                if test_func():
+                    return
+                time.sleep(check_period_s)
+            att_fail = (
+                f"\n  (Still failing after timeout [{timeout_ms}ms], with attempts "
+                f"every {check_period_ms}ms)"
+            )
+            exc = AssertionError(f"{str(fail_msg)}{att_fail}".strip())
+            raise exc.with_traceback(tb)
+
+        else:
+            raise AssertionError("Bad assert call: no attempts or timeout period")
+
+    yield _impl
--- a/parsl/tests/test_scaling/test_scale_down.py
+++ b/parsl/tests/test_scaling/test_scale_down.py
@@ -1,80 +1,97 @@
 import logging
-import parsl
-import pytest
 import time
-from parsl import python_app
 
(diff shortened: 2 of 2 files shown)


parsl/tests/test_scaling/test_scale_down.py
```diff
@@ -1,80 +1,97 @@
...
+    futs = [waiting_app(i, inputs=inputs) for i in range(ntasks)]
 
-    logger.info("asserting 5 managers")
-    assert len(dfk.executors['htex_local'].connected_managers()) == 5, "Expected 5 managers after some time"
+    while ready_path.read_text().count("\n") < _max_blocks:
+        time.sleep(0.5)
 
-    logger.info("waiting for all futures to complete")
-    [x.result() for x in fus]
+    assert len(dfk.executors['htex_local'].connected_managers()) == 5
```
benclifford, line 83: that explicit 5 should probably be _max_blocks

parsl/tests/test_scaling/test_scale_down.py
```diff
@@ -1,80 +1,97 @@
...
-    time.sleep(25)
+    assert dfk.executors['htex_local'].outstanding == 0
 
-    logger.info("asserting 2 managers remain")
-    assert len(dfk.executors['htex_local'].connected_managers()) == 2, "Expected 2 managers when no tasks, lower bound by min_blocks"
+    def assert_kernel():
+        return len(dfk.executors['htex_local'].connected_managers()) == _min_blocks
 
-    logger.info("test passed")
+    try_assert(
```
benclifford, line 93: some part of me says `try_assert` adds a lot of complexity that might be a relatively small while loop?

parsl/tests/test_scaling/test_scale_down.py
```diff
@@ -1,80 +1,97 @@
...
+        assert_kernel,
+        fail_msg="Expected 2 managers when no tasks, lower bound by min_blocks",
```
benclifford, line 95: 2 should not be explicit

parsl/tests/test_scaling/test_scale_down.py
khk-globus: Bah, of course.  Got halfway there ...

parsl/tests/test_scaling/test_scale_down.py
khk-globus: Yeah, I concur on this point.  However, my thinking was that surely this won't be the only place where we need a "give it some settling time" type assertion.

As discussed out-of-band, neither of us has a particular strong global insight into all of the tests, so agreed to leave it around for now.  If it ends up being used only by this test, we can remove it at some point in the indeterminate future.

parsl/tests/conftest.py
```diff
@@ -290,3 +294,67 @@ def pytest_ignore_collect(path):
...
+                time.sleep(check_period_s)
+            else:
+                att_fail = (
+                    f"\n  (Still failing after attempt limit [{attempts}], testing"
+                    f" every {check_period_ms}ms)"
+                )
+                exc = AssertionError(f"{str(fail_msg)}{att_fail}".strip())
+                raise exc.with_traceback(tb)
+
+        elif timeout_s > 0:
```
benclifford, line 344: this if/elif structure means that timeout_ms will be ignored if attempts !=0, right?

khk-globus: That is correct; an intentional choice.

Issue #1606, opened 2020-03-14
RuntimeError: dictionary changed size during iteration


ravihansa3000:
**Describe the bug**
A RuntimeError exception is thrown (stack trace below) while executing Parsl app. 

```
2020-03-14 14:38:50 parsl.dataflow.memoization:159 [DEBUG]  Ignoring these kwargs for checkpointing: []
Traceback (most recent call last):
  File "/home/users/nus/csipav/Documents/GDC-Pipeline.git/run.py", line 187, in <module>
2020-03-14 14:38:50 parsl.dataflow.memoization:202 [DEBUG]  Task 1 has memoization hash 6888eb84e8d55e06d97d5fb5d0b61eb6
2020-03-14 14:38:50 parsl.dataflow.memoization:210 [INFO]  Task 1 had no result in cache
    main(sys.argv[1:])
  File "/home/users/nus/csipav/Documents/GDC-Pipeline.git/run.py", line 183, in main
2020-03-14 14:38:50 parsl.executors.high_throughput.executor:529 [DEBUG]  Pushing function <function MonitoringHub.monitor_wrapper.<locals>.wrapped at 0x2aaab86a5200> to queue with args (<function split_vcf_by_chromosome at 0x2aaab84929e0>, "{'MuSE': '/home/users/nus/csipav/anaconda3/envs/gdc/bin/MuSE', 'ace2sam': '/home/users/nus/csipav/an...", "<<class 'parsl.data_provider.files.File'> at 0x2aaab88eec90 url=/home/users/nus/csipav/scratch/Parsl...", "'/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Pipeline-Output/OV_1/TCGA-13-0910/1/annotations/...", "'/home/users/nus/csipav/scratch/Parsl-workflows/GDC-Pipeline-Output/OV_1/TCGA-13-0910/1/strelka2-ana...")
[...]

PR #2811, opened 2023-07-12
Remove test_1606 sleep

khk-globus:
This tests internal DFK data structures; only necessary to run locally.  This means we can use `threading.Event()` to enact this test, rather than sleeping the day away.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -161,9 +161,9 @@ def load_dfk_local_module(request, pytestconfig):
     parsl.load. It should be a Callable that returns a parsl Config object.
 
     If local_setup and/or local_teardown are callables (such as functions) in
-    the test module, they they will be invoked before/after the tests. This
-    can be used to perform more interesting DFK initialisation not possible
-    with local_config.
+    the test module, they will be invoked before/after the tests. This can
+    be used to perform more interesting DFK initialisation not possible with
+    local_config.
     """
 
     config = pytestconfig.getoption('config')[0]
--- a/parsl/tests/test_regression/test_1606_wait_for_current_tasks.py
+++ b/parsl/tests/test_regression/test_1606_wait_for_current_tasks.py
@@ -1,17 +1,42 @@
+import threading
+import time
+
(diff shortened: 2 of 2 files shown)


benclifford:
I think this might not be testing a regression of #1606 any more.

The problem addressed in #1606 is that wait_for_current_tasks was accessing the dfk's self.tasks structure in a way that didn't work if you removed elements from it over time: previously a list of keys was generated, and then later on the code tried to get those keys from the tasks dictionary.

This doesn't work when records are removed from the self.tasks dictionary (which was previously grow-only, until #1543 introduced code to tidy that up, with hilarious consequences)

So the 1606 regression test specifically wants tests to complete while the main thread is *inside* the wait for current tasks call - I think the event-driven rearrangements here make that not so true any more.

benclifford:
The above ordering issues aside, I think it's right that this becomes a local config test -- the specific issue this is regression testing only needs tasks to end inside the wait call, and shouldn't care how/where those tasks end otherwise.

khk-globus:
> I think this might not be testing a regression of https://github.com/Parsl/parsl/issues/1606 any more.

> The problem addressed in #1606 is that wait_for_current_tasks was accessing the dfk's self.tasks structure in a way that didn't work if you removed elements from it over time: previously a list of keys was generated, and then later on the code tried to get those keys from the tasks dictionary.

Discussed out-of-band: I think what's here is actually correct.  The subtle detail is the brunt of the work happens _inside_ of `test_kernel()` which is run in a separate thread.  Thus, the order of events is:

1. `test_kernel` starts, and _main thread waits_
1. `test_kernel` creates 3 apps
1. `test_kernel` then signals that the main thread may continue (`may_wait.set()`)
1. `test_kernel` then waits for a 1/100 second, allowing main thread to call `wait_for_current_tasks()`
1. `test_kernel` then sets the relevant events to let the apps continue.

Thus, I believe this successfully tests the condition that was fixed in 583b846812be908f40c0d11220aee8b948a52517.

PR #2817, opened 2023-07-12
Check *all* child process for liveness every HB

khk-globus:
Prior to this, a dead process in a pool of N processes would take, on average, (N * heartbeat) / 2 seconds to be restarted.

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
Changes looks correct. Nice bug fix!

PR #2822, opened 2023-07-14
Reduce apptimeout test delays

khk-globus:
No sense in waiting extra time.

## Type of change

- Code maintenance/cleanup


benclifford (approved):

benclifford:
related test idea for another time: check (somehow) that the timed-out process got killed. From a look at the code I don't see anything that does that... bash_app just stops waiting and reports an exception, without any further management of the launched process...

Issue #69, opened 2018-01-16
DataFuture __str__ is not helpful


yadudoc:
The datafuture output that we have now isn't useful. It has the format : 
<DataFuture at 0x7f71e9bfe668 state=pending>
- [ ]  state does not update
- [ ] at <hex address> isn't helpful, instead it could be path/url ?

Reported by @danielskatz

PR #2823, opened 2023-07-14
Remove useless test

khk-globus:
[edited, per [BenC's comment](https://github.com/Parsl/parsl/pull/2823#issuecomment-1639818371)]

These tests, such as they are, do not appear to be related to #69.  We already exercise the code base in exactly the same way throughout a number of other tests, so just remove this file entirely.

## Type of change

- Code maintenance/cleanup


benclifford:
this file seems to have very little to do with issue #69 - I'd rather just delete most (all?) of it...

khk-globus:
Works for me.  Hang tight.

PR #2825, opened 2023-07-18
Allow BlockProviderExecutor to be configured with no provider

benclifford:
This is already implemented throughout the code, but the BlockProviderExecutor type annotations require a provider to be specified.

This PR loosens that type check.

## Type of change


parsl/executors/status_handling.py
```diff
@@ -44,7 +44,7 @@ class BlockProviderExecutor(ParslExecutor):
     I'm not sure of use cases for switchability at the moment beyond "yes or no"
     """
     def __init__(self, *,
-                 provider: ExecutionProvider,
+                 provider: Optional[ExecutionProvider],
                  block_error_handler: bool):
```
khk-globus, line 48: If this is optional per provider, do the individual providers need an additional warning check when it is provided and not necessary, or not provided and required?

khk-globus (approved):

Issue #2589, opened 2023-02-03
HTEX to support binding to specific interfaces


yadudoc:
Currently the `HighThroughputExecutor` has the `interchange` bind to all interfaces (for eg here: https://github.com/Parsl/parsl/blob/master/parsl/executors/high_throughput/interchange.py#L170) followed by the `managers` attempting to connect to it over the network using a list of addresses. This mechanism allows the `managers` to be launched without having to know the specific interfaces accessible from the internal network of clusters simplifying configuration and user effort. However, binding to all interfaces, esp public interfaces has gotten complaints from TACC admins.

**Describe the solution you'd like**
It would be nice to keep both solutions, ie use all addresses by default and when an interface is specified by the user, `HTEX(network_interface="ib0"` bind to that specifically and pass along the address attached to that interface to the managers.

**Additional context**
This is a priority feature for funcX to switch over to using Parsl executors.

PR #2828, opened 2023-07-18
Update HTEX.Interchange to listen on a single interface

yadudoc:
# Description

This PR updates the `HTEX.Interchange` to listen for connections from the managers only on a specific interface rather than the current default of binding to all interfaces. Binding to all interfaces is generally frowned upon on login nodes, when all we need is to allow connections from the internal network.

Here are the changes:
* Pass `HighThroughputExecutor.address` to `Interchange.interchange_address`
* Interchange will bind only to the `interchange_address` if it is specified instead of the binding to zmq:`*` or `0.0.0.0` which is the current default.
* Adding tests for the Interchange

Please note that configs which specify `HTEX(address="localhost")` or similar where a non IPv4 address will now fail

Fixes # (issue)

https://github.com/Parsl/parsl/issues/2589

## Type of change

- Documentation update


benclifford:
I think this

> Please note that configs which specify HTEX(address="localhost") or similar where a non IPv4 address will now fail

is already out of our regular configuration recommendations - that's what `parsl.addresses.get_address_by_hostname()` is for - so that should be pretty harmless

parsl/executors/high_throughput/interchange.py
```diff
@@ -89,8 +89,9 @@ def __init__(self,
         client_address : str
              The ip address at which the parsl client can be reached. Default: "127.0.0.1"
 
-        interchange_address : str
```
benclifford, line 92: I'm trying to understand how this parameter was used previously: `git grep` suggests it was not used at all? Is that right?

yadudoc: It was not used at all, but I believe this is how it was meant to be used.

benclifford: then maybe can use a single address object attribute instead of both `self.interchange_address` and `self.listen_address` ?

benclifford:
mypy is faliing, but in interchange `__main__`, which I think is unused functionality from earlier prototyping of interchanges running elsewhere? (I'm also having trouble understanding how other commandline args get passed around in that block of code)

If that block of code really is unused, maybe delete it?

benclifford (changes requested):
i think that interchange_address vs listen_address variable is a duplication that can be tidied?

benclifford (approved):

PR #2833, opened 2023-07-19
Remove duplicate import not detected by flake8

benclifford:
This wasn't detected by flake8, but moving the line above the type for _db_manager_excepts caused flake8 to detect it, so perhaps it is the presence of intervening code between the imports that causes that.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Odd, odd.  Thanks for working through flake8-bugbear in the other PR.

PR #2835, opened 2023-07-19
Remove ExecutorStatus abstract class

benclifford:
This removes the ExecutorStatus abstract class, instead using its only concrete implementation `PollItem` everywhere that ExecutorStatus was previously used.

This requires some circular imports to let type annotations work - some now annotate some things as `PollItem` rather than `ExecutorStatus`. This is done using `from __future__ import annotations` and `import` rather than `from ... import ...`

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Nice!

benclifford:
this is going to need some import loop untangling...

PR #2837, opened 2023-07-20
Consolidate job handling in a new module, parsl.jobs

benclifford:
# Description

Prior to this, a lot of job handling was tangled into the parsl.dataflow module.

There's a growing desire in Globus Compute to use the job handling parts of Parsl as a user-exposed feature, distinct from the data flow parts of Parsl.

This PR supports that new use case by creating a new module, parsl.jobs, for job-related code to live in, to help with ongoing work relaxing the previous requirement that a Dataflow Kernel would always exist to manage parsl components.

This PR only moves code around - it should not cause any change in functionality.

This is a breaking change for any externally implemented providers - they would need the same (hopefully superficial) import fixups that this PR does to providers in parsl/providers/

(this is probably a breaking change for Globus Compute making out-of-scope use of Parsl internals, but ... that's out of scope: this PR is part of making that sort of stuff exposed to users properly)

## Type of change

- Code maintentance/cleanup


parsl/jobs/states.py
```diff
@@ -0,0 +1,116 @@
...
+            # can happen for synthetic job failures
+            return None
+        try:
+            with open(path, 'r') as f:
+                f.seek(0, os.SEEK_END)
+                size = f.tell()
+                f.seek(0, os.SEEK_SET)
+                if size > JobStatus.SUMMARY_TRUNCATION_THRESHOLD:
+                    half_threshold = int(JobStatus.SUMMARY_TRUNCATION_THRESHOLD / 2)
+                    head = f.read(half_threshold)
```
khk-globus, line 106: Could also use the `//` operator:
```python
half_threshold = JobStatus.SUMMARY_TRUNCATION_THRESHOLD // 2
```

parsl/jobs/states.py
```diff
@@ -0,0 +1,116 @@
...
+                    f.seek(size - half_threshold, os.SEEK_SET)
+                    tail = f.read(half_threshold)
+                    return head + '\n...\n' + tail
+                else:
+                    f.seek(0, os.SEEK_SET)
+                    return f.read()
+        except FileNotFoundError:
+            # When output is redirected to a file, but the process does not produce any output
+            # bytes, no file is actually created. This handles that case.
+            return None
```
khk-globus, line 116: Is this a loggable/warnable detail?

parsl/executors/base.py
```diff
@@ -3,7 +3,7 @@
 from typing import Any, Callable, Dict, Optional, List
 from typing_extensions import Literal, Self
 
-from parsl.providers.base import JobStatus
+from parsl.jobs.states import JobStatus
```
khk-globus, line 6: Similar to my previous comment (oddly located below!  Hooray for temporal placement!), this doesn't appear to be utilized except during type checking.  You might consider putting it in the same if-condition:
```python
if t.TYPE_CHECKING:
    from parsl.jobs.job_error_handler import JobErrorHandler
    from parsl.jobs.states import JobStatus
```

parsl/providers/base.py
```diff
@@ -1,124 +1,13 @@
-import os
 from abc import ABCMeta, abstractmethod, abstractproperty
-from enum import IntEnum
 import logging
 from typing import Any, Dict, List, Optional
 
 from parsl.channels.base import Channel
+from parsl.jobs.states import JobStatus
```
khk-globus, line 6: Similar suggestion as for `executor.base`: this doesn't appear to be utilized in this file, so can discourage imports of this class from this module by wrapping it in the type checking conditional:
```python
if t.TYPE_CHECKING:
    from parsl.jobs.states import JobStatus
```

khk-globus (approved):
A couple of minor comments which you may address if you wish, but no blockers.  Nice observation and tackling of this structure.

benclifford:
re `if TYPE_CHECKING`: that conditional import breaks with any runtime typechecking (see typeguard, especially version 4 onwards which is much more aggressive about what it manages to typecheck than previous versions). So although it might not break in this case right now, it's an idiom I'd like to avoid. https://peps.python.org/pep-0649/ discusses that more.

benclifford:
most of the other comments seem reasonable for fiddling with in another PR - I'll probably not get round to that though

parsl/jobs/states.py
```diff
@@ -0,0 +1,116 @@
...
+    PENDING = 1
+    RUNNING = 2
+    CANCELLED = 3
+    COMPLETED = 4
+    FAILED = 5
+    TIMEOUT = 6
+    HELD = 7
+
+    def __str__(self) -> str:
+        return self.__class__.__name__ + "." + self.name
```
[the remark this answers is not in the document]
benclifford: This is a bit different from the repr changes that follow on from #1966 style repr tidyups:

repr on this class is still the default, which gives something like this (on Python 3.11):

```
>>> repr(p.JobState.COMPLETED)
'<JobState.COMPLETED: 4>'
```

but this `__str__` was addred in PR #2653 specifically because the str behaviour changed from Python 3.10 to Python 3.11 to rendering an IntEnum as the integer value, rather than as the name of the state.

This makes sense for the use-case it was intended for (when tightening an already existing `int` field into an `IntEnum`) but in the Parsl case, I wanted to preserve the loose human representation being the state name (which is the case in Python <=3.10)

For example, when used in log messages, it is nicer to represent the state to a human reader as `"COMPLETED"` rather than as `4`.

parsl/jobs/states.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+TERMINAL_STATES = [JobState.CANCELLED, JobState.COMPLETED, JobState.FAILED,
+                   JobState.TIMEOUT]
```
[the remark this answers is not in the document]
benclifford: no this should never change

parsl/jobs/states.py
benclifford: The comment suggests its a legitimate case but maybe it's interesting to log it

PR #2842, opened 2023-07-21
Allow user pluggable serializers and add a demonstration ProxyStore serializer

benclifford:
This PR rearranges the deserializer identifier handling so as to dynamically load unknown serializers. These should be identified by a module and class name (and that identifier will be generated automatically).

New serializers, implementing SerializerBase, can be implemented and registered for either the `code` or `data` paths in serialization using the existing `register_method_for_*` functions introduced in #2831 

This PR also introduces an example serializer using this API: it extends dill to use ProxyStore on certain objects determined by a user specified policy.

The API for reconfiguring serializers is rather awkward, as evidenced by one of the ProxyStore test case, but this PR does not attempt to make this any nicer.

## Type of change


parsl/serialize/base.py
```diff
@@ -10,18 +11,24 @@ class SerializerBase:
...
-
+        """
+         Compute the identifier that will be used to indicate in byte streams
+￼        that this class should be used for deserialization.
+￼
+￼        Serializers that use identifiers that don't align with the way this is
+         computed (such as the default concretes.py implementations) should
+         override this property with their own identifier.
+￼
         Returns
```
khk-globus, line 24: From the "not a big deal" camp, there's an extra space indent in this docstring.

parsl/serialize/concretes.py
```diff
@@ -17,7 +17,7 @@ class PickleSerializer(SerializerBase):
     * closures, generators and coroutines
     """
 
-    _identifier = b'01'
+    identifier = b'01'
```
khk-globus, line 21: If we've defined the property in the `SerializerBase`, what's the need for these class properties in the subclasses anymore?

parsl/serialize/base.py
```diff
@@ -10,18 +11,24 @@ class SerializerBase:
...
         -------
         identifier : bytes
         """
-        return self._identifier
+        t = type(self)
+        m = bytes(t.__module__, encoding="utf-8")
+        c = bytes(t.__name__, encoding="utf-8")
+        return m + b' ' + c
```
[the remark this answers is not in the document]
benclifford: I think in the cases where `__qualname__` and `__name__` differ, this is often going to break: the module + name needs to be something importable.

One case that could work, a class defined directly inside another class without any enclosing function/closure, I would have to poke at the import side of things to see how that would work.

parsl/serialize/errors.py
```diff
@@ -0,0 +1,35 @@
...
+    """Failure to serialize task objects.
+    """
+
+    def __init__(self, fname: str) -> None:
+        self.fname = fname
+        self.troubleshooting = "https://parsl.readthedocs.io/en/latest/faq.html#addressing-serializationerror"
+
+    def __str__(self) -> str:
+        return f"Failed to serialize objects for an invocation of function {self.fname}. Refer {self.troubleshooting}"
+
```
khk-globus, line 25: I like this pointer to the documentation.  A couple of questions:
- Is there a way to test that this link is valid?  As in, during CI?  (And I _don't_ mean reaching out to rtd.io in every run.)
- I'm wondering if we might do something similar for the deserialization side.  I'm not seeing a way currently because not all deserialization errors manifest so cleanly or directly, of course.

parsl/serialize/concretes.py
benclifford: some wire compatiblity (though I am not particularly fussed there) and some serialised object length: (47 bytes header vs 2 bytes header)

khk-globus: *nod; so the base class property is meant as a _fallback_, then?  I dig it.

parsl/serialize/errors.py
benclifford: this is actually imported from a different PR #2839 (so this will disappear from this current PR if/when that is merged), and *that* PR inherits it from the original import location... 

That link doesn't even try to do any kind of versioning seriousness, so as soon as that stops being frequently asked, the link will break for the entire installed userbase... I can imagine various over-complicated uses of metaclasses or subclass hooks or etc that are URL+documentation aware... but that seems like a lot of work for an abandoned project.

These two errors are awkwardly asymmetrical in what they report, in part from a half hearted attempt previously to include reference links in some error messages but not others.

parsl/serialize/concretes.py
benclifford: in general a user defined serialiser *should* use that base class impl and not define their own identifier: if you define a different identifier explicitly which does not fit the importable/locatable syntax, you also take on responsibility for ensuring anyone who might deserialise also knows what to do with this identifier. So I guess I'm imagining that only being used if someone really wants to hard code identifiers and is modifying the parsl source code.

PR #2846, opened 2023-07-30
Reduce monitoring filesystem radio log volume

### LINKED ISSUE #2845: monitoring_filesystem_radio log fills rapidly when using MonitoringHub
**Describe the bug**
When I enable the `MonitoringHub` to monitor tasks, it appears to start the filesystem radio, which then proceeds to log a message every second noting that it has started. The repeated log message is here:

https://github.com/Parsl/parsl/blob/70b9a2cd663087bc2197f633681a0e5f23ad68e1/parsl/monitoring/monitoring.py#L329-L346

which seems to be called unconditionally when the `MonitoringHub` is started:

https://github.com/Parsl/parsl/blob/70b9a2cd663087bc2197f633681a0e5f23ad68e1/parsl/monitoring/monitoring.py#L215-L219

**To Reproduce**
1. Setup Parsl 2023.07.24 with Python 3.11 on cluster
2. Run a test script with the HTEX executor with `MonitoringHub` enabled
3. Wait a few minutes
4. Check size of `runinfo/__run_number__/monitoring_filesystem_radio.log`, and see it increasing every second

It's not a huge problem, especially if runs are short, but for long runs, it means that I have to do manual cleanup for a message that doesn't seem to be critical

**Expected behavior**

Remove the log message. In any case, I'm not sure if this is supposed to be started unconditionally since if I follow correctly, using the HTEX uses the HTEX radio, not the filesystem(?). But I'm not so familiar with the logger system in parsl. If you plan to just remove the message, I can open a trivial PR if helpful - as you like.

Thanks!

**Environment**
 - OS: [e.g. ubuntu, centos, MacOS, windows]
 - Python version
 - Parsl version

SL 7, python 3.11, parsl 2023.07.24

**Distributed Environment**
- Where are you running the Parsl script from ? [e.g. Laptop/Workstation, Login node, Compute node]
- Where do you need the workers to run ? [e.g. Same as Parsl script, Compute nodes, Cloud nodes]

benclifford:
This PR moves most log messages in the filesystem radio to debug level, and sets the log level for the monitoring to INFO level. This gives one log message per monitoring message, without per-poll logging. This is more in line with log level styles in Parsl, logging at INFO when something happens.

Output of DEBUG logs is not user-configurable, but I left those log lines there for debug hacking.

Fixes #2845

## Type of change

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -314,7 +314,7 @@ class MonitoringHub(RepresentationMixin):
 def filesystem_receiver(logdir: str, q: "queue.Queue[AddressedMonitoringMessage]", run_dir: str) -> None:
     logger = start_file_logger("{}/monitoring_filesystem_radio.log".format(logdir),
                                name="monitoring_filesystem_radio",
-                               level=logging.DEBUG)
+                               level=logging.INFO)
 
     logger.info("Starting filesystem radio receiver")
     setproctitle("parsl: monitoring filesystem receiver")
@@ -327,7 +327,7 @@ def filesystem_receiver(logdir: str, q: "queue.Queue[AddressedMonitoringMessage]
     os.makedirs(new_dir, exist_ok=True)
 
     while True:  # this loop will end on process termination
-        logger.info("Start filesystem radio receiver loop")
+        logger.debug("Start filesystem radio receiver loop")
 
         # iterate over files in new_dir
         for filename in os.listdir(new_dir):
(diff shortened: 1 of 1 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -327,7 +327,7 @@ def filesystem_receiver(logdir: str, q: "queue.Queue[AddressedMonitoringMessage]
     os.makedirs(new_dir, exist_ok=True)
 
     while True:  # this loop will end on process termination
-        logger.info("Start filesystem radio receiver loop")
+        logger.debug("Start filesystem radio receiver loop")
```
khk-globus, line 331: Without any dynamic data, this log line looks to me to be of negligible value.  Yes, the loop is iterating, but that can also be generally inferred by lack of traceback or other log messages.  What about removing it entirely, moving it out of the loop (just before the `while: True`), or adding something to it?  Perhaps the number of files the iteration would process?

khk-globus (approved):
Have made a suggestion inline, but it's not a blocker.

PR #2847, opened 2023-07-31
Adding typing info for HTEX

yadudoc:
# Description

This PR only adds some typing info for these methods/properties on the `HighThroughputExecutor`:

* `HTEX.outstanding`
* `HTEX.connected_workers`
* `HTEX.connected_managers`

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -1,3 +1,4 @@
+import typing
 from concurrent.futures import Future
 import typeguard
 import logging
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
One minor suggestion, but not a blocker.

parsl/executors/high_throughput/executor.py
```diff
@@ -523,16 +524,23 @@ def hold_worker(self, worker_id: str) -> None:
...
+        """Returns the count of workers across all connected managers"""
         workers = self.command_client.run("WORKERS")
         return workers
 
-    def connected_managers(self):
+    def connected_managers(self) -> List[Dict[str, typing.Any]]:
+        """Returns a list of dicts one for each connected managers.
+        The dict contains info on manager(str:manager_id), block_id,
+        worker_count, tasks(int), idle_durations(float), active(bool)
+        """
```
benclifford, line 543: This structure is frustratingly close to `ManagerRecord` in parsl/executors/high_throughput/manager_record, but not quite - ManagerRecords are transformed here: https://github.com/Parsl/parsl/blob/ce7e8f6282b1a87b56e1fd6b9d55d5a6d5f5f921/parsl/executors/high_throughput/interchange.py#L297

This docstring could be made into a very similar looking TypedDict, though, I think, which could give stronger checking of accesses to the dictionary

PR #2849, opened 2023-08-01
TaskVineExecutor: add serverless execution mode for tasks

tphung3:
# Description

This PR is a follow-up to PR https://github.com/Parsl/parsl/pull/2809 with the following changes:
- Add support for serverless execution mode for tasks. Users can now declare serverless tasks like this:
```
@python_app
def f(x, parsl_resource_specification={'exec_mode': 'serverless'}):
    ...
```
Under the hood, a library task is deployed to each worker and forks itself to execute an actual task/function call.
- Add support for full conda environment replication on worker nodes for both serverless and regular tasks. Given a conda environment, TaskVine will distribute and cache it efficiently, so a shared filesystem is not required. The benefit mainly lies in the assumption that the conda environment can be huge in size and thousands of tasks asking the shared filesystem for the same environment directory at the same time is not very performant. TaskVine on the other hand keeps track of where files are at and has an implicit throttle mechanism to data distribution so things don't go haywire.
- Add support for no conda environment replication, provided that environment is properly setup on worker nodes. Turning on `shared_fs` flag and add some setup code to `init_command` would turn off almost all TaskVine data-aware magic and the workflow run will mostly use the shared filesystem instead (e.g., `init_command="conda run -p path/to/conda"`). This serves as the baseline behavior for TaskVineExecutor.
- Refactor `executor.py` into {`executor.py`, `manager.py`, and `factory.py`} as they are different components.
- Change API: `use_factory` to `worker_launch_method`.
- Some variable renaming and additional logging.

Potential improvements: 
- Caching of functions and large arguments on both the manager and worker sides. Support for function caching is relatively easy and the benefit is that a function doesn't have to be serialized/deserialized every time it's called. Support for argument caching is more tricky and may require some help from Parsl (e.g., some annotations of immutable/sharable/cacheable data from users so some additional APIs are needed - data-intensive parsl?).
- App chaining: If apps are chained like this: `f(g(x)).result()`, then the best case is for the result of g(x) to be cached on the worker and f will be sent there, pick up the result, and then execute. Current implementation will send g(x) back to the manager, then send g(x) out again with f, which is inefficient. Implementing app chaining/imtermediate result caching needs some help from Parsl as well, as of now the dfk implicitly gives executors a stream of tasks instead of a graph.
- Performance improvement: policy to automate object caching and resource allocation for function calls.
- Anything else?

Fixes # (issue)
N/A

## Type of change
- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/taskvine/exec_parsl_function.py
+++ b/parsl/executors/taskvine/exec_parsl_function.py
@@ -149,21 +149,8 @@ def execute_function(namespace, function_code, result_name):
     return result
 
 
-def run():
+def run(map_file, function_file, argument_file, result_file):
     try:
-        # parse the four required command line arguments:
-        # map_file: contains a pickled dictionary to map original names to
-        #           names at the execution site.
-        # function_file: contains the pickled parsl function to execute.
-        # argument_file: contains the pickled arguments to the function call.
-        # result_file: any output (including exceptions) will be written to
-        #              this file.
-        try:
-            (map_file, function_file, argument_file, result_file) = sys.argv[1:]
-        except ValueError:
-            print("Usage:\n\t{} function argument result mapping\n".format(sys.argv[0]))
-            raise
-
         try:
             (namespace, function_code, result_name) = load_function(map_file, function_file, argument_file)
         except Exception:
@@ -189,4 +176,16 @@ def run():
 
 
 if __name__ == "__main__":
-    run() 
+    # parse the four required command line arguments:
+    # map_file: contains a pickled dictionary to map original names to
+    #           names at the execution site.
+    # function_file: contains the pickled parsl function to execute.
+    # argument_file: contains the pickled arguments to the function call.
+    # result_file: any output (including exceptions) will be written to
+    #              this file.
+    try:
+        (map_file, function_file, argument_file, result_file) = sys.argv[1:]
+    except ValueError:
+        print("Usage:\n\t{} function argument result mapping\n".format(sys.argv[0]))
+        raise
+    run(map_file, function_file, argument_file, result_file) 
--- a/parsl/executors/taskvine/exec_parsl_function.py
+++ b/parsl/executors/taskvine/exec_parsl_function.py
@@ -165,7 +165,9 @@ def run(map_file, function_file, argument_file, result_file):
     except Exception:
         traceback.print_exc()
         result = RemoteExceptionWrapper(*sys.exc_info())
-
+    
+    with open('/tmp/test_exec_parsl.tmp', 'w') as f:
+        f.write('debug' + str(result))
     # Write out function result to the result file
     try:
         dump_result_to_file(result_file, result)
--- a/parsl/executors/taskvine/exec_parsl_function.py
+++ b/parsl/executors/taskvine/exec_parsl_function.py
@@ -82,7 +82,7 @@ def unpack_object(serialized_obj, user_namespace):
     from parsl.serialize import deserialize
     obj = deserialize(serialized_obj)
     return obj
-    
+
 
 def encode_function(user_namespace, fn, fn_name, fn_args, fn_kwargs):
     """ Register the given function to the given namespace."""
@@ -165,9 +165,7 @@ def run(map_file, function_file, argument_file, result_file):
     except Exception:
         traceback.print_exc()
         result = RemoteExceptionWrapper(*sys.exc_info())
-    
-    with open('/tmp/test_exec_parsl.tmp', 'w') as f:
-        f.write('debug' + str(result))
+
     # Write out function result to the result file
     try:
         dump_result_to_file(result_file, result)
@@ -190,4 +188,4 @@ if __name__ == "__main__":
     except ValueError:
         print("Usage:\n\t{} function argument result mapping\n".format(sys.argv[0]))
         raise
-    run(map_file, function_file, argument_file, result_file) 
+    run(map_file, function_file, argument_file, result_file)
--- a/parsl/executors/taskvine/exec_parsl_function.py
+++ b/parsl/executors/taskvine/exec_parsl_function.py
@@ -79,25 +79,10 @@ def remap_all_files(mapping, fn_args, fn_kwargs):
 
 
 def unpack_object(serialized_obj, user_namespace):
-    #from parsl.serialize import unpack_apply_message
     from parsl.serialize import deserialize
-    #obj = unpack_apply_message(serialized_obj, user_namespace, copy=False)
     obj = deserialize(serialized_obj)
     return obj
     
-
-#def unpack_function(function_info, user_namespace):
-#    """ Unpack a function according to its encoding scheme."""
-#    return unpack_byte_code_function(function_info, user_namespace)
-#
-#
-#def unpack_byte_code_function(function_info, user_namespace):
-#    """ Returns a function object, a default name, positional arguments, and keyword arguments
-#    for a function."""
-#    from parsl.serialize import unpack_apply_message
-#    func, args, kwargs = unpack_apply_message(function_info["byte code"], user_namespace, copy=False)
-#    return (func, 'parsl_function_name', args, kwargs)
-
 
 def encode_function(user_namespace, fn, fn_name, fn_args, fn_kwargs):
     """ Register the given function to the given namespace."""
@@ -146,7 +131,6 @@ def load_function(map_file, function_file, argument_file):
     fn_args = args_dict['args']
     fn_kwargs = args_dict['kwargs']
     fn_name = 'parsl_tmp_func_name'
-    #(fn, fn_name, fn_args, fn_kwargs) = unpack_function(function_info, user_ns)
 
     mapping = load_pickled_file(map_file)
     remap_all_files(mapping, fn_args, fn_kwargs)
@@ -159,7 +143,6 @@ def load_function(map_file, function_file, argument_file):
 def execute_function(namespace, function_code, result_name):
     # On executing the function inside the namespace, its result will be in a
     # variable named result_name.
-    #raise Exception(f'function_code: {function_code}, namespace: {namespace}')
     exec(function_code, namespace, namespace)
     result = namespace.get(result_name)
 
--- a/parsl/executors/taskvine/exec_parsl_function.py
+++ b/parsl/executors/taskvine/exec_parsl_function.py
@@ -78,17 +78,25 @@ def remap_all_files(mapping, fn_args, fn_kwargs):
             remap_location(mapping, maybe_file)
 
 
-def unpack_function(function_info, user_namespace):
-    """ Unpack a function according to its encoding scheme."""
-    return unpack_byte_code_function(function_info, user_namespace)
+def unpack_object(serialized_obj, user_namespace):
+    #from parsl.serialize import unpack_apply_message
+    from parsl.serialize import deserialize
+    #obj = unpack_apply_message(serialized_obj, user_namespace, copy=False)
+    obj = deserialize(serialized_obj)
+    return obj
+    
 
-
-def unpack_byte_code_function(function_info, user_namespace):
-    """ Returns a function object, a default name, positional arguments, and keyword arguments
-    for a function."""
-    from parsl.serialize import unpack_apply_message
-    func, args, kwargs = unpack_apply_message(function_info["byte code"], user_namespace, copy=False)
-    return (func, 'parsl_function_name', args, kwargs)
+#def unpack_function(function_info, user_namespace):
+#    """ Unpack a function according to its encoding scheme."""
+#    return unpack_byte_code_function(function_info, user_namespace)
+#
+#
+#def unpack_byte_code_function(function_info, user_namespace):
+#    """ Returns a function object, a default name, positional arguments, and keyword arguments
+#    for a function."""
+#    from parsl.serialize import unpack_apply_message
+#    func, args, kwargs = unpack_apply_message(function_info["byte code"], user_namespace, copy=False)
+#    return (func, 'parsl_function_name', args, kwargs)
 
 
 def encode_function(user_namespace, fn, fn_name, fn_args, fn_kwargs):
@@ -119,7 +127,7 @@ def encode_byte_code_function(user_namespace, fn, fn_name, args_name, kwargs_nam
     return code
 
 
-def load_function(map_file, function_file):
+def load_function(map_file, function_file, argument_file):
     # Decodes the function and its file arguments to be executed into
     # function_code, and updates a user namespace with the function name and
     # the variable named result_name. When the function is executed, its result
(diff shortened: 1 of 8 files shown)


parsl/executors/taskvine/exec_parsl_function.py
```diff
@@ -78,17 +78,10 @@ def remap_all_files(mapping, fn_args, fn_kwargs):
...
-    return unpack_byte_code_function(function_info, user_namespace)
-
-
-def unpack_byte_code_function(function_info, user_namespace):
-    """ Returns a function object, a default name, positional arguments, and keyword arguments
-    for a function."""
-    from parsl.serialize import unpack_apply_message
-    func, args, kwargs = unpack_apply_message(function_info["byte code"], user_namespace, copy=False)
-    return (func, 'parsl_function_name', args, kwargs)
+def unpack_object(serialized_obj, user_namespace):
```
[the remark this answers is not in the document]
benclifford: (there are some other places in the parsl serialization code where user_ns is passed around where I think it is unnecessary too - probably historical tangling of deserialisation vs execution)

parsl/executors/taskvine/executor.py
```diff
@@ -88,10 +71,11 @@ class TaskVineExecutor(BlockProviderExecutor, putils.RepresentationMixin):
...
             Default is "TaskVineExecutor".
 
-        use_factory: bool
-            Choose to whether use either the Parsl provider or
-            TaskVine factory to scale workers.
-            Default is False.
+        worker_launch_method: str
+            Choose to use Parsl provider, TaskVine factory, or
+            manual user-provided workers to scale workers.
+            Options are among {'provider', 'factory', 'manual'}.
```
benclifford, line 77: theres a type annotation that's stricter than `str` for this kind of list of strings, that would look something like:

`Union[Literal['provider'], Literal['factory'], Literal['manual']]`

which I think would mean `typeguard` would detect invalid values in this parameter.

parsl/executors/taskvine/executor.py
```diff
@@ -114,80 +98,90 @@ class TaskVineExecutor(BlockProviderExecutor, putils.RepresentationMixin):
...
                  storage_access: Optional[List[Staging]] = None):
 
-        # If TaskVine factory is used, disable the Parsl provider
-        if use_factory:
+        # Set worker launch option for this executor
+        if worker_launch_method == 'factory' or worker_launch_method == 'manual':
             provider = None
+        elif worker_launch_method != 'provider':
+            raise ExecutorError(self, "Worker launch option '{worker_launch_method}' \
+                                       is not supported.")
```
benclifford, line 112: i find this style of `if` statement a bit awkward to reason about in my head, about are all the cases of worker_launch_method covered here, and prefer the style that is:

if a = X or a = Y:
  some thing
elif a = Z:
  some other thing which might be `pass`
else:
  raise failure

that's something of a personal preference though

parsl/executors/taskvine/executor.py
```diff
@@ -242,53 +242,53 @@ def start(self):
...
 
         # Begin work
-        self.submit_process.start()
+        self._submit_process.start()
 
         # Run worker scaler either with Parsl provider or TaskVine factory
-        if self.use_factory:
-            self.factory_process.start()
+        if self.worker_launch_method == 'factory':
+            self._factory_process.start()
```
benclifford, line 272: is this `if` statement still ok when worker_launch_method is something other than `factory` or `provider`? maybe this doesn't need to call `initialize_scaling` with `manual` mode?

parsl/executors/taskvine/factory.py
```diff
@@ -0,0 +1,64 @@
...
+        factory.batch_options = factory_config.batch_options
+
+    # setup factory context and sleep for a second in every loop to
+    # avoid wasting CPU
+    with factory:
+        while not should_stop.value:
+            time.sleep(1)
+
+    logger.debug("Exiting TaskVine factory process")
+    return 0
```
benclifford, line 64: i think this return doesn't do anything?

parsl/executors/taskvine/manager.py
```diff
@@ -0,0 +1,468 @@
...
+                    # different from setup of regular tasks.
+                    # If shared_fs is True, then no environment preparation is done.
+                    # Only the core serverless code is created.
+                    poncho_env_path = _prepare_environment_serverless(manager_config, env_cache_dir, poncho_create_script)
+
+                    # Don't automatically add environment so manager can declare and cache the vine file associated with the environment file
+                    add_env = False
+                    serverless_lib = m.create_library_from_functions('common-parsl-taskvine-lib',
+                                                                     run_parsl_function,
+                                                                     poncho_env=poncho_env_path,
```
benclifford, line 247: I get an error here:

```
  File "/home/benc/parsl/src/parsl/parsl/executors/taskvine/manager.py", line 245, in _taskvine_submit_wait
    serverless_lib = m.create_library_from_functions('common-parsl-taskvine-lib',
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Manager.create_library_from_functions() got an unexpected keyword argument 'poncho_env'
```

I think that's because this parameter was recently introduced (https://github.com/cooperative-computing-lab/cctools/pull/3427 18 hours ago) and so this mode will only work with a future release of ndcctools?

Maybe it would be useful to document that somewhere as a minimum version requirement.

benclifford:
I installed a recent cctools, bc0891de5719d5c4fd89692d76b3650d535144d3 and tried serverless mode with parsl-perf, by modifying taskvine_ex.py to set shared_fs = True, and invoking:

```
$ parsl-perf --config parsl/tests/configs/taskvine_ex.py --resources "{'exec_mode':'serverless'}"
```

This hangs after submitting tasks:

```
==== Iteration 1 ====
Will run 10 tasks to target 120 seconds runtime
Submitting tasks / invoking apps
All 10 tasks submitted ... waiting for completion
Submission took 0.006 seconds = 1693.163 tasks/second
```

I see three vine_worker processes running. 

There are a lot of vine-factory-* directories in here: `runinfo/TaskVineExecutor` and it isn't clear how to correlate them with a parsl runs in a nice way - it feels to me like some of these directories are per-run and belong in the relevant parsl per-run directory, because this is, I think, otherwise incredibly awkward for parsl users.

In the most recent factory directory, I see this:

```
$ cat vine-factory-kfvebapy/error.log 
2023/08/02 09:29:53.19 vine_worker[14477] fatal: Unable to get information from coprocess
Killed
```

In this transaction log 2023-08-02T092952/vine-logs/transactions, I see tasks go to RUNNING state...


```
1690968592552600 14469 MANAGER 14469 START 0
1690968592628957 14469 TASK 1 WAITING app FIRST_RESOURCES 1 {}
1690968592629268 14469 TASK 2 WAITING app FIRST_RESOURCES 1 {}
1690968592629556 14469 TASK 3 WAITING app FIRST_RESOURCES 1 {}
1690968592629816 14469 TASK 4 WAITING app FIRST_RESOURCES 1 {}
1690968592630065 14469 TASK 5 WAITING app FIRST_RESOURCES 1 {}
1690968592630301 14469 TASK 6 WAITING app FIRST_RESOURCES 1 {}
1690968592630550 14469 TASK 7 WAITING app FIRST_RESOURCES 1 {}
1690968592630804 14469 TASK 8 WAITING app FIRST_RESOURCES 1 {}
1690968592631025 14469 TASK 9 WAITING app FIRST_RESOURCES 1 {}
1690968592631274 14469 TASK 10 WAITING app FIRST_RESOURCES 1 {}
1690968592902469 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 CONNECTION 127.0.0.1:52278
1690968592902863 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 TRANSFER INPUT library_code.py 3973 86 1690968592902774
1690968592903188 14469 TASK 11 RUNNING worker-7a639c999393ead9d4dc90583d0622b6  FIRST_RESOURCES {"memory":[0,"MB"],"disk":[0,"MB"],"gpus":[0,"gpus"],"cores":[0,"cores"]}
1690968592903195 14469 LIBRARY 11 SENT worker-7a639c999393ead9d4dc90583d0622b6
1690968592904817 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 TRANSFER INPUT function 2115 65 1690968592904749
1690968592904891 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 TRANSFER INPUT argument 113 48 1690968592904841
1690968592904971 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 TRANSFER INPUT map 5 56 1690968592904913
1690968592905027 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 TRANSFER INPUT infile 71 35 1690968592904990
1690968592905322 14469 TASK 9 RUNNING worker-7a639c999393ead9d4dc90583d0622b6  FIRST_RESOURCES {"time_commit_start":[1690968592.904732,"s"],"time_commit_end":[1690968592.905269,"s"],"time_input_mgr":[0.000537,"s"],"size_input_mgr":[0.002197265625,"MB"],"memory":[15708,"MB"],"disk":[2500,"MB"],"gpus":[0,"gpus"],"cores":[4,"cores"]}
1690968592905370 14469 WORKER worker-7a639c999393ead9d4dc90583d0622b6 RESOURCES {"memory":[15708,"MB"],"disk":[2500,"MB"],"cores":[4,"cores"]}
```

benclifford:
Most of the comments on this review are things that don't need fixing for this PR to be merged - they're just comments that you might be interested in for future work.

However, I would like to be able to see serverless work on my laptop before merging.

parsl/executors/taskvine/exec_parsl_function.py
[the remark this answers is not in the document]
khk-globus: Assuming `user_namespace` is, in fact, not needed, this could be reduced by moving the import out of the function (to the module), then:
```python
def unpack_object(serialized_obj):
    return deserialize(serialized_obj)
```

At which point, there's no need for the extra call in the stack.  Just alias the function:
```python
unpack_object = deserialize
```

Alternatively, could alias at import time (with the other module imports):
```python
from parsl.serialize import deserialize as unpack_object
```

Or even just use the `deserialize` method directly, elsewhere (e.g., in `load_function`).

benclifford:
actually I'm now getting a similar error from the WQ tests:  `	STDERR: 2023/08/02 13:59:58.25 work_queue_worker[65405] fatal: Unable to get information from coprocess` so I wonder if I have misinstalled something, or if something changed in cctools recently.

parsl/executors/taskvine/exec_parsl_function.py
```diff
@@ -188,3 +173,19 @@ def execute_function(namespace, function_code, result_name):
...
+    # function_file: contains the pickled parsl function to execute.
+    # argument_file: contains the pickled arguments to the function call.
+    # result_file: any output (including exceptions) will be written to
+    #              this file.
+    try:
+        (map_file, function_file, argument_file, result_file) = sys.argv[1:]
+    except ValueError:
+        print("Usage:\n\t{} function argument result mapping\n".format(sys.argv[0]))
+        raise
+    run(map_file, function_file, argument_file, result_file)
```
khk-globus, line 191: You might consider wrapping this comment into a `usage()` function, so that this information can be shared with the hapless typo-ing user on the command line.  Consider:
```python
def usage(exc: Optional[Exception] = None):
    print("Usage: {} <mapping_path> <function_path> <args_path> <result_path>", file=sys.stderr)
    print("\n[Brief description of what this script does]", file=sys.stderr)
    print("  - mapping_path - path to a file containing a pickled dictionary ...", file=sys.stderr)
    print("  - function_path - ...", file=sys.stderr)
    
    if exc:
        print(f"\n({type(exc).__name}) {exc}", file=sys.stderr)

    sys.exit(1)  # non-zero exit code to represent "not successful" to the OS
```

Then can utilize the usage in place of the raise:
```python
try:
    map_file, function_file, argument_file, result_file = sys.argv[1:]
except ValueError as err:
    usage(err)  # won't return, per sys.exit
run(map_file, ...)
```

khk-globus (commented):
Ben's left some good comments, so I'll just add this one (optional) request for typing.  For a number of the functions, I would have appreciated typing while reading the code.  Typing helps not only mypy find errors, but also other devs (like me!) understand the intent and follow the threads of data throughout.

For example:
```diff
-def run_parsl_function(map_file, function_file, argument_file, result_file):
+def run_parsl_function(
+    map_file: os.PathLike,
+    function_file: os.PathLike,
+    argument_file: os.PathLike,
+    result_file: os.PathLike
+):
```

benclifford:
>  I wonder if I have misinstalled something

It turns out I have misinstalled something - will try again

benclifford:
I'm getting a core dump on vine_worker if I run with serverless mode, with the below gdb backtrace. Running in `exec_mode=regular`, this runs ok for me.

I'll try to do a build as you describe and see how that works for me.

```
(gdb) 
#0  __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex.S:79
#1  0x000055a77387ce66 in hash_string (s=0x55a700000000 <error: Cannot access memory at address 0x55a700000000>)
    at hash_table.c:386
#2  0x000055a77387c7a8 in hash_table_insert (h=0x55a77546f2a0, 
    key=0x55a700000000 <error: Cannot access memory at address 0x55a700000000>, value=0x1) at hash_table.c:162
#3  0x000055a77384bbd1 in start_process (p=0x55a775497180, manager=0x55a775472fc0) at vine_worker.c:531
#4  0x000055a77384eab8 in work_for_manager (manager=0x55a775472fc0) at vine_worker.c:1336
#5  0x000055a77384f705 in serve_manager_by_hostport (host=0x55a77546f810 "127.0.0.1", port=9000, verify_project=0x0, 
    use_ssl=0) at vine_worker.c:1625
#6  0x000055a77384f7d8 in serve_manager_by_hostport_list (manager_addresses=0x55a77546f7a0, use_ssl=0)
    at vine_worker.c:1649
#7  0x000055a773851330 in main (argc=3, argv=0x7ffffdbe9948) at vine_worker.c:2243
```

benclifford:
this segfault above is resolved in more recent cctools github - see https://github.com/cooperative-computing-lab/cctools/issues/3434 for more discussion.

benclifford:
A user interface issue here: in the parsl+wq coprocess model, use of coprocesses is defined at the executor level for all tasks executed through that executor.

The task vine executor requires that option to be set on all tasks, rather than as part of the configuration of the executor.

This stems, I think, from earlier plans to make only certain functions run in serverless mode, and otherwise run using the spawn-a-python-process mode.

But, as this PR is implemented, I think it would be perfectly reasonable for a user to want to run all tasks through this serverless mode - like in parsl+wq - and probably a more likely situation than wanting to have some tasks running with this PR's serverless mode and some tasks running as fresh python processes.

... and based on my experience with wq+coprocesses, I think there's a strong argument for that being the default mode for this executor (or in other words, what is deficient about the way serverless mode works that would push a user back to separate python process mode?)

benclifford:
@tphung3 this is merged but please consider all the comments that are still open.

tphung3:
I'm working on a paper and will address them soon. Thanks!

PR #2853, opened 2023-08-04
Split and remove JobErrorHandler

benclifford:
Before this PR: The JobErrorHandler class had no state and was used effectively as a repository of stateless library functions, which are called from two distinct pieces of code.

This PR removes the JobErrorHandler class, moving the code into two locations:

`JobErrorHandler.run` (and its helper method `__check_irrecoverable_executor`) are called by `JobStatusPoller` to give each executor a chance to run its own error handler. This code is moved into JobStatusPoller.

`simple_error_handler` (and its helper methods `count_jobs` and `get_error`) are called by the `BlockProviderExecutor` implementation of error handling. They are moved into a new file, `simple_error_handler.py` directly inside that module (rather than inside a class).

After these changes, instances of `JobErrorHandler` are no longer used after being initialised - so the now-empty JobErrorHandler and its containing module are deleted.

This PR is not intended to change any behaviour.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/base.py
+++ b/parsl/executors/base.py
@@ -167,14 +167,12 @@ class ParslExecutor(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def handle_errors(self, error_handler: "parsl.jobs.job_error_handler.JobErrorHandler",
-                      status: Dict[str, JobStatus]) -> None:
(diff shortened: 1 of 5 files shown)


yadudoc (approved):
Makes sense.

PR #2855, opened 2023-08-07
Remove NoStatusHandlingExecutor

benclifford:
Executors which want to use status handling should subclass the BlockProviderExecutor, formerly (PR #2071) known as the StatusHanldingExecutor.

Code which performs status handling should only operate on BlockProviderExecutor instances - the status handling code shouldn't do anything to other ParslExecutor instances.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
Nice!

PR #2857, opened 2023-08-10
Switch type comparisons from == to is, and upgrade flake8, in preparation for Python 3.12

benclifford:
Upcoming Python 3.12 (at least in python 3.12-rc1) requires a newer version of flake8 as flake8 6.0.0 gets confused by things happening inside format strings with Python 3.12-rc1.

This PR makes that upgrade.

Flake8 6.1.0 introduces a new rule to use is-equality, rather than ==-equality, when comparing types, and so this PR fixes violations of that rule too. In theory this shouldn't change behaviour.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Simple enough change; just reran the failed test as it appears to have timed out.

PR #2858, opened 2023-08-10
Add user-pluggable block error handling and a new sliding-window error handler

yadudoc:
# Description

This PR aims to address the limited capabilities of the current `simple_error_handler` at stopping the Parsl runtime when there are repeated failures. The current system only fails if all jobs fail, which is only indicative of configuration errors or problem with the batch scheduler. This PR adds new behavior that updates the existing `block_error_handler` bool variable to take a custom error handler. In addition there's a new `block_error_threshold` that can be used to configure these callbacks.

```
from parsl.jobs.simple_error_handler import simple_error_handler, windowed_error_handler

config = Config(executors=[
         HighThroughputExecutor(
                ...
                block_error_handler = <simple_error_handler / windowed_error_handler>,
                block_error_threshold)
  ]
)
```
 
Fixes # (issue)

## Type of change

- Code maintentance/cleanup


benclifford:
Three of the four CI runs hung in the `--config local` tests at different points (and the fourth was subsequently auto-cancelled), with the last commit b963e39b21061b18b4e270487f06273962f3f72e

Those tests are quite hangy (I think from race conditions to do with threads and zmq and multiprocessing) so it's not unusual for a `--config local` to hang. For three to hang in one run seems suspicious, though.

parsl/executors/high_throughput/executor.py
```diff
@@ -212,7 +214,7 @@ def __init__(self,
                  poll_period: int = 10,
                  address_probe_timeout: Optional[int] = None,
                  worker_logdir_root: Optional[str] = None,
-                 block_error_handler: bool = True):
+                 block_error_handler: Union[bool, Callable[[ParslExecutor, Dict[str, JobStatus]], None]] = False):
```
benclifford, line 217: don't change the default value here?

parsl/executors/status_handling.py
```diff
@@ -163,12 +161,8 @@ def handle_errors(self, status: Dict[str, JobStatus]) -> None:
         """
         if not self.block_error_handler:
             return
-        init_blocks = 3
-        if hasattr(self.provider, 'init_blocks'):
-            init_blocks = self.provider.init_blocks
-        if init_blocks < 1:
-            init_blocks = 1
-        error_handler.simple_error_handler(self, status, init_blocks)
+        assert callable(self.block_error_handler)
```
benclifford, line 164: with block error handler set to true, as is allowed and should be the default, this assert fails at runtime:

```
jobs.job_status_poller:118 _run_error_handlers INFO: BENC: about to enter handle errors
1691736331.298730 2023-08-11 06:45:31 MainProcess-9203 JobStatusPoller-Timer-Thread-140360055682320-140360045541056 parsl.utils:349 make_callback ERROR: Callback threw an exception - logging and proceeding anyway
  File "/home/benc/parsl/src/parsl/parsl/utils.py", line 347, in make_callback
    self.callback(*self.cb_args)
  File "/home/benc/parsl/src/parsl/parsl/jobs/job_status_poller.py", line 113, in poll
    self._run_error_handlers(self._poll_items)
  File "/home/benc/parsl/src/parsl/parsl/jobs/job_status_poller.py", line 119, in _run_error_handlers
    es.executor.handle_errors(es.status)
  File "/home/benc/parsl/src/parsl/parsl/executors/status_handling.py", line 164, in handle_errors
    assert callable(self.block_error_handler)
AssertionError
```

parsl/executors/status_handling.py
```diff
@@ -163,12 +161,8 @@ def handle_errors(self, status: Dict[str, JobStatus]) -> None:
         """
         if not self.block_error_handler:
             return
-        init_blocks = 3
-        if hasattr(self.provider, 'init_blocks'):
```
benclifford, line 167: removing this changes the behaviour of the simple error handler, right? so that it will always wait for the first three blocks, rather than the init_blocks number? If we care about this PR adding in new stuff, not changing old behaviour, then this test could move into simple_error_handler, because it has access to executor.provider.init_blocks there.

parsl/jobs/simple_error_handler.py
```diff
@@ -6,12 +6,20 @@
...
 
-def simple_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int):
+def simple_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int = 3):
     (total_jobs, failed_jobs) = _count_jobs(status)
     if total_jobs >= threshold and failed_jobs == total_jobs:
         executor.set_bad_state_and_fail_all(_get_error(status))
 
 
+def windowed_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int = 3):
+    sorted_status = [(key, status[key]) for key in sorted(status)]
```
benclifford, line 16: this sort is a bit suspicious if `status` is keyed by strings, not ints, as the type signature says... does a failing sequence of blocks 10,11,12 get ignored here because later/earlier blocks 7,8,9 were successful?

parsl/tests/test_scaling/test_block_error_handler.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.mark.local
+def test_block_error_handler_mock():
+    handler_mock = Mock()
+    htex = HighThroughputExecutor(block_error_handler=handler_mock)
+    assert htex.block_error_handler is handler_mock
+
+    bad_state_mock = Mock()
+    htex.set_bad_state_and_fail_all = bad_state_mock
```
benclifford, line 33: does this mock do anything in this test?

benclifford:
this is pretty much how I imagined this would be implemented, after a conversation with Yadu elsewhere

parsl/executors/status_handling.py
benclifford: without that assert, mypy statically finds this:

```
parsl/executors/status_handling.py:164: error: "Literal[True]" not callable  [misc]
```

but mypy's interpretation of asserts about types is "I, the programmer, am telling you, mypy, a fact that you cannot figure out for yourself", rather than "I, the programmer, am asking you to statically validate this assert".

benclifford:
I was also wondering about, because this is a user pluggable interface, if that code needs more protective error handling around the call - however, it looks like the timer code already catches, reports, and continues to time, like this:

``` 1691736331.298730 2023-08-11 06:45:31 MainProcess-9203 JobStatusPoller-Timer-Thread-140360055682320-140360045541056 parsl.utils:349 make_callback ERROR: Callback threw an exception - logging and proceeding anyway
```

and i think that's probably enough error handing around the error handler.

parsl/jobs/simple_error_handler.py
```diff
@@ -6,12 +6,20 @@
...
 
 
-def simple_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int):
+def simple_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int = 3):
     (total_jobs, failed_jobs) = _count_jobs(status)
     if total_jobs >= threshold and failed_jobs == total_jobs:
         executor.set_bad_state_and_fail_all(_get_error(status))
 
 
+def windowed_error_handler(executor: status_handling.BlockProviderExecutor, status: Dict[str, JobStatus], threshold: int = 3):
```
benclifford, line 15: probably rename this file to error_handlers.py or something like that, now its more than simple_error_handler.

parsl/executors/high_throughput/executor.py
yadudoc: Fixed.

parsl/executors/status_handling.py
yadudoc: Fixed.

parsl/executors/status_handling.py
yadudoc: Fixed.

parsl/jobs/simple_error_handler.py
yadudoc: Yep, there was a bug here, I've updated the sorted call to take a lambda that assumes the key is a simple integer, which should work for our ClusterProviders.

yadudoc:
@benclifford I believe I've addressed all your comments. Let me know if you see any further issues. I see that tests were also passing earlier.

parsl/jobs/simple_error_handler.py
benclifford: that's suspicious in a different way - I guess it doesn't make sense for block IDs to be anything other than sequential integers, and maybe the rest of the codebase should be changed to not use `str` for block IDs (not in this PR).

benclifford (approved):

PR #2867, opened 2023-08-11
Rework _get_error in simple_error_handler to give more information

benclifford:
Before (in pytest output, with htex launch_cmd="/bin/false"):

```
Exception: 	STDOUT: Found cores : 8
Launching worker: 1
```

After:

```
parsl.jobs.errors.TooManyJobFailuresError: Error 1:
	EXIT CODE: 1
	STDOUT: Found cores : 8
Launching worker: 1
```

* Fix counting so it counts even when js.message is None

* Add exit code

* Use a new error subclass rather than Exception

* Pluralise messages in for no-messages string and remove brackets

  '[No error message received]' vs 'No error messages received'

# Description

Fixes # (issue)

## Type of change

- Documentation update
- Code maintentance/cleanup

### DIFF
--- a/parsl/jobs/errors.py
+++ b/parsl/jobs/errors.py
@@ -1,0 +1,7 @@
+from parsl.errors import ParslError
+
+
+class TooManyJobFailuresError(ParslError):
+    """Indicates that executor is shut down because of too many block failures.
+    """
+    pass
--- a/parsl/jobs/simple_error_handler.py
+++ b/parsl/jobs/simple_error_handler.py
@@ -4,6 +4,7 @@ from typing import Dict
 
 import parsl.executors.status_handling as status_handling
 from parsl.jobs.states import JobStatus, JobState
+from parsl.jobs.errors import TooManyJobFailuresError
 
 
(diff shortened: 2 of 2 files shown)


parsl/jobs/error_handlers.py
```diff
@@ -42,18 +43,25 @@ def _get_error(status: Dict[str, JobStatus]) -> Exception:
...
+        if js.exit_code is not None:
+            err = err + f"\tEXIT CODE: {js.exit_code}\n"
+
         stdout = js.stdout_summary
         if stdout:
             err = err + "\tSTDOUT: {}\n".format(stdout)
+
         stderr = js.stderr_summary
         if stderr:
             err = err + "\tSTDERR: {}\n".format(stderr)
```
khk-globus, line 61: What's the likelihood of catastrophic failure?  Would it behoove the user for us to use `std(err|out)_summary` instead?

benclifford: not sure what you mean

khk-globus: What I mean _now_ is that I was apparently blind earlier.  Now I clearly see:

```python
stdout = js.stdout_summary
...
stderr =  js.stderr_summary
```
I missed that first go-round.

PR #2881, opened 2023-09-17
Remove unreachable code when app_fu is None

benclifford:
The task record entry app_fu is never None:
it is either missing (during construction) or a Future.

So, this None check will never fire.

Discovered by mypy 0.5.1 warn_unreachable

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Indeed; nice find.  (Is that mypy **1** .5.1?)  Heh.

PR #2884, opened 2023-09-17
Tighten MissingOutputs file type

benclifford:
Outputs must be Files - although early on in Parsl, it was possible to specify them as string pathnames. This now-removed string-or-File dual annotation is probably a holdover from that.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/app/errors.py
+++ b/parsl/app/errors.py
@@ -64,10 +64,9 @@ class MissingOutputs(ParslError):
 
     Contains:
     reason(string)
-    outputs(List of strings/files..)
+    outputs(List of files)
     """
-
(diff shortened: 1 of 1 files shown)


parsl/app/errors.py
```diff
@@ -64,10 +64,9 @@ class MissingOutputs(ParslError):
 
     Contains:
     reason(string)
-    outputs(List of strings/files..)
+    outputs(List of files)
```
khk-globus, line 67: I suppose the typing will make it plain, but might consider `Files` or `List of File objects`

khk-globus (approved):

PR #2885, opened 2023-09-18
Move scale_{in,out} from ParslExecutor to BlockProviderExecutor

benclifford:
Scale in/out behaviour is only for BlockProviderExecutors. Other fairly recent PRs have been consolidating that behaviour, and this PR is another one in that direction.

The non-BlockProviderExecutors (threads and flux) had meaningless stub methods to make ABCMeta allow them to be instantiated. This PR removes those stubs.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
That's a good call; I wonder if there's any appetite for more hierarchy.  I don't have a sense of it at this time ...

PR #2886, opened 2023-09-18
Remove execute_wait stdout, stderr optionality

benclifford:
No in-codebase channels return None for stdout or stderr, and some cluster providers (at least pbspro and slurm) do not handle the case when stdout is None, instead assuming it will be a str.

This PR makes those assumptions concrete.

This PR will introduce API breakage for any externally implemented channels which return None here, although in practice because of the above, it's likely that use of those channels woudl already be broken.

However, I am not aware of any externally implemented channels.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Huh; definitely seems an odd choice for an initial implementation.  Thanks for this.

PR #2887, opened 2023-09-18
Remove three dead assignments from local channel

benclifford:
These initial values will never be used and in the stdout/err case, can give the impression that stdout, stderr might be returned as None. There is no code path which uses these initial assignments as the function will raise an exception in the case that they are not all re-assigned later on.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Oof.  Obvious cleanup.  Thanks!

Issue #2873, opened 2023-08-24
Make "Must first load config" into exception class


benclifford:
**Is your feature request related to a problem? Please describe.**
This:

```
parsl/dataflow/dflow.py:            raise RuntimeError('Must first load config')
```

should be its own exception class so that code trying to detect uninitialized parsl can detect it based on class, rather than on matching strings.

tagging @yadudoc who had some interest in error structuring.

**Additional context**
https://github.com/cytomining/CytoTable/pull/93

d33bs:
Thank you for the changes related to addressing this issue! The newly added `ConfigurationError` for cases where Parsl configuration has not yet been loaded provides greater specificity but requires parsing the Exception message in order to make use of it (using either `str(ConfigurationException)` or `ConfigurationException.args[0]` to filter the specific message). I'd like to suggest an additional Parsl-specific Exception (perhaps `NoConfigurationError`, with a parent Exception of `ConfigurationException`) to help further differentiate specific cases where Parsl configuration has not yet been loaded. See [this PR](https://github.com/cytomining/CytoTable/pull/109) for an example Parsl implementation of where this could make a difference.

benclifford:
@d33bs have a look at https://github.com/Parsl/parsl/pull/2890

d33bs:
Thank you @benclifford , I took a look and thought that looked great for this!

benclifford:
PRs #2878 and #2890 resolve this

PR #2890, opened 2023-09-20
Switch empty dfk() error from ConfigurationError to NoDataFlowKernelError

benclifford:
This is in support of
https://github.com/Parsl/parsl/issues/2873#issuecomment-1726540856

This will break anyone relying on particular exception classes coming from this call.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):

d33bs (commented):
Thank you @benclifford, this looks great! 👍

PR #2894, opened 2023-09-21
TaskVineExecutor: Improve serverless

tphung3:
# Description

This PR brings changes that addresses concerns in PR #2849. 
- Added and edited comments and debug messages so they are accurate or more meaningful.
- Fixed the issue where function and arguments are serialized twice. Now they are serialized once.
- Added stronger type hints and checking in several places.
- Added a default function execution mode option in the Executor interface to support for applications that want to exclusively run functions in either mode. Users can still override this global option by specifying the execution mode at the function level via `parsl_resource_specification`.
- Use `multiprocessing.Event` to signal the stopping condition instead of `Value`.
- Other minor fixes.

Fixes # (issue)

## Type of change

## Further comments
This PR works with the newest commit from CCTools, and each worker only runs one function at a time. More improvements are coming.


tphung3:
This PR depends on CCTools latest commit so let's make it a draft PR for now.

parsl/executors/taskvine/exec_parsl_function.py
```diff
@@ -115,26 +113,23 @@ def encode_byte_code_function(user_namespace, fn, fn_name, args_name, kwargs_nam
...
+    args_dict = unpack_object_from_file(argument_file)
     fn_args = args_dict['args']
     fn_kwargs = args_dict['kwargs']
     fn_name = 'parsl_tmp_func_name'
 
-    mapping = load_pickled_file(map_file)
+    mapping = unpack_object_from_file(map_file)
     remap_all_files(mapping, fn_args, fn_kwargs)
 
+    # Create the namespace to isolate the function execution.
```
benclifford, line 129: this block isn't creating a new namespace, I think? it's using and updating the local variables namespace of the current function invocation (if I correctly understand what `locals()` returns).

This is something htex and the workqueue executor do as well, and I wonder if they should all actually be starting with an empty dictionary, `{}`, instead of `locals()`.

benclifford (approved):

benclifford:
On my laptop, I get around 100 tasks per second with serverless mode turned on, with the below config. 

This compares against (variable) 300 tasks/second using parsl/tests/configs/workqueue_ex.py.

I haven't investigated any further into what is actually going on to see where time is being spent.

```
from parsl.executors.taskvine import TaskVineExecutor
from parsl.executors.taskvine import TaskVineManagerConfig

from parsl.data_provider.http import HTTPInTaskStaging
from parsl.data_provider.ftp import FTPInTaskStaging
from parsl.data_provider.file_noop import NoOpFileStaging

def fresh_config():
    return Config(executors=[TaskVineExecutor(manager_config=TaskVineManagerConfig(port=9000, shared_fs=True),
                                              worker_launch_method='factory',
                                              function_exec_mode='serverless',
                                              storage_access=[FTPInTaskStaging(), HTTPInTaskStaging(), NoOpFileStaging()])])
```

benclifford:
I think the relevant latest cctools is released now, so this can be merged, right? @tphung3

tphung3:
Yes, I can add minor fixes in later in a separate PR. @benclifford

PR #2898, opened 2023-10-04
Do not return a value from a pytest case, to fix a deprecation warning

benclifford:
This fixes this pytest warning:

  PytestReturnNotNoneWarning: Expected None, but
  test_bash_apps/test_error_codes.py::test_div_0 returned True, which
  will be an error in a future version of pytest.
  Did you mean to use `assert` instead of `return`?

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Hey!  One more down!

PR #2906, opened 2023-10-12
Add `__getattr__` support to PR 2904

Andrew-S-Rosen:
# Description

@benclifford: This adds `__getattr__` support to #2904. If all goes as it does locally for me, then these tests should pass.

I also added tests for `__getitem__` with `list` and `tuple` results.

Edit: Looks like `parsl/tests/test_python_apps/test_lifted.py` tests passed, but I'm not sure what the other failures are from.


benclifford:
I was thinking about how maybe this should try to avoid methods beginning with `__` or even just `_` to avoid accidentally implementing screwed up versions of various protocols.

Andrew-S-Rosen:
I do think some sort of guardrails could be useful. In addition, should we also explicitly avoid methods that are already part of the `AppFuture`, perhaps?

benclifford:
I think `__getattr__` will only get called for attributes that don't exist in some more fundamental way - that's how .result still works (assuming it does)

Andrew-S-Rosen:
Ah, of course. Yes yes! You are correct 👍

Issue #2903, opened 2023-10-09
Readthedocs build has been broken for several months


benclifford:

benclifford:
This started failing at this commit https://github.com/Parsl/parsl/commit/41357c696b62fa73ba87ef1fbbec2e8ce08e2dbb (PR #2809)

benclifford:
Possibly because the taskvine module is not importable without ndcctools installed:

```
>>> import parsl.executors.taskvine
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/benc/parsl/src/parsl/parsl/executors/taskvine/__init__.py", line 1, in <module>
    from parsl.executors.taskvine.executor import TaskVineExecutor
  File "/home/benc/parsl/src/parsl/parsl/executors/taskvine/executor.py", line 35, in <module>
    from parsl.executors.taskvine.manager_config import TaskVineManagerConfig
  File "/home/benc/parsl/src/parsl/parsl/executors/taskvine/manager_config.py", line 5, in <module>
    from ndcctools.taskvine.cvine import VINE_DEFAULT_PORT
ModuleNotFoundError: No module named 'ndcctools'
```

although this is not the actual error text appearing in readthedocs.

Although the taskvine executor.py file avoids importing ndcctools unprotected, looks like manager_config.py does not avoid it.

cc @tphung3

benclifford:
I made a draft PR #2907 that hacks out all the ndcctools import to validate that ndcctools is what is breaking thsi.

I also turned on readthedocs reporting status for all pull requests into github - so all PRs now should get a green/red readthedocs test line.

@tphung3 could you look at making taskvine not fail on import when ndcctools is not around?

benclifford:
The immediate failure has been resolved by @tphung3 in PR #2909 - thanks!

To help stop this happen in future, readthedocs now builds each PR and reports status into GitHub. At the moment this is not a "mandatory pass" test, so is only indicative to an attentive human, but it could be made mandatory.

PR #2909, opened 2023-10-13
TaskVineExecutor: Fix Issue 2903

tphung3:
# Description

This PR fixes the issue in #2903.

Fixes # (issue): 2903

## Type of change


benclifford:
I think you need to do something to factory.py too -- I *think* that if you can write `import parsl.executors.taskvine` into a Python REPL (without cctools installed) and have that not raise an error, then that's also what will fix readthedocs.

PR #2910, opened 2023-10-14
Chain exceptions on WorkQueueTaskFailure when a cause is available

benclifford:
This will add exception chaining when an underlying exception is available, so that reporting of a WorkQueueTaskError will include before the WorkQueueTaskError. This means, in the case of a FileNotFoundError, that the filename that could not be found will now be reported.

```
 FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/parsl/parsl/runinfo/003/WorkQueueExecutor/function_data/0011/result'

parsl/executors/workqueue/executor.py:1059: FileNotFoundError

```

### DIFF
--- a/parsl/executors/workqueue/errors.py
+++ b/parsl/executors/workqueue/errors.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from parsl.errors import ParslError
 from parsl.app.errors import AppException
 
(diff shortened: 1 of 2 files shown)


benclifford:
@dthain maybe you have someone to review this

See issue #2914 for the motivating issue that I am debugging.

benclifford:
I discussed this with @tphung3 and I think he's ok with me merging it

PR #2911, opened 2023-10-14
Separate monitoring core and visualization dependencies

benclifford:
This reflects a difference in code maturity and maintenance between the two: monitoring core is much more maintained and stable than visualization at the moment.

Work with LSST has repeatedly encountered dependency problems caused by the visualization dependencies, even though those dependencies are not need by LSST, and this separation allows the visualization dependencies to be skipped.

Breaking change:
If you are using visualization you will now need to:

  pip install parsl[monitoring,visualization]

to get all of the dependencies previously installed by [monitoring]

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ wqex_local_test: $(CCTOOLS_INSTALL)  ## run all tests with workqueue_ex config
 
 .PHONY: config_local_test
 config_local_test:
-	pip3 install ".[monitoring,proxystore]"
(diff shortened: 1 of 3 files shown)


Andrew-S-Rosen:
Thank you for this. It will be a big help, as I've run into some annoying version conflicts with the visualization dependencies even if I just wanted to use monitoring features.

Issue #2908, opened 2023-10-13
Work Queue (& Task Vine?) vs complicated return values


benclifford:
**Describe the bug**
if a python app returns an object that is too complex for pickle to serialize, this breaks.

i) pickle in exec_parsl_function.py cannot serialize the return value

ii) I think (but I haven't confirmed) that passing the result value along a multiprocessing.Queue in parsl/executors/workqueue/executor.py also uses `pickle` to get the object between processes, and so this is likely to fail too.

For comparison, the HighThroughputExecutor doesn't hit these problems:

i) it uses parsl.serialize, rather than pickle, to send back results
ii) deserialization happens in the main process, so that the deserialized object is never sent along a multiprocessing queue

cc @tphung3 @dthain 

**To Reproduce**

In PRs #2904, #2906 , there is a test case that fails with Work Queue, when trying to return a dataclass definition.

**Expected behavior**
More serializable result handling

dthain:
@tphung what do you think?

benclifford:
(serialization is something I'm actively working on so I'm not expecting Cooperative Computing Lab people to do anything here - and it's also sufficiently obscure that this isn't super urgent)

tphung3:
The issue makes sense to me. TaskVineExecutor seems to have the same problem. A quick and correct fix would be to use serialize library from Parsl as mentioned by @benclifford. I think this fix doesn't get in the way of your work in serialization right @benclifford?

benclifford:
I think point i) will be addressed by using `parsl.serialize`. I think the 2nd point is a bit more complicated - I haven't fully debugged what is happening there so I don't know if it is even true or not.

tphung3:
If mq.Queue uses pickle to transfer data (I'm pretty sure it does), then the unpickle operation can happen in the main process in the collector thread. The workqueue process then only sends back a notification to the collector thread so it knows to unpickle results from files. Unpickle is then replaced by `deserialize` method in parsl.

I think this would clear all corner cases right? @benclifford

benclifford:
Yeah I think that will work.

In htex there's a thread that does the deserialisation but it lives in the same process, rather than being a separate process like in WQExecutor

benclifford:
This is fixed by PR #2908 and the test that discovered this is enabled by PR #2928

PR #2916, opened 2023-10-16
Avoid reserializing results into multiprocessing in Work Queue and Task Vine executors

tphung3:
# Description

This PR fixes issue #2908 

Fixes # 2908

## Type of change


tphung3:
@benclifford I think this fixes issue 2908.

benclifford:
@tphung3 I just re-enabled the test that demonstrates issue #2908 (from PR #2904) - I think that should pass now with your changes.

parsl/executors/workqueue/executor.py
```diff
@@ -727,11 +727,18 @@ def _collect_work_queue_results(self):
...
+                    except Exception as e:
+                        logger.debug(f'Cannot load result from result file {task_report.result_file}. Exception: {e}')
+                        future.set_exception(WorkQueueTaskFailure('Cannot load result from result file', None))
+                    else:
+                        future.set_result(result)
                 else:
                     # If there are no results, then the task failed according to one of
                     # work queue modes, such as resource exhaustion.
-                    future.set_exception(WorkQueueTaskFailure(task_report.reason, task_report.result))
+                    future.set_exception(WorkQueueTaskFailure(task_report.reason, None))
```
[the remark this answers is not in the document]
tphung3: Sure I think good error messages are always beneficial. Do you want to merge #2910 when it is ready, I'll update this PR then.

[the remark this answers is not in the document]
benclifford: If you are happy with #2910 I'll merge it now, and then you can work on the conflicts that will arise with this PR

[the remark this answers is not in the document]
benclifford: @tphung3 that's merged now and conflicts ready for you to resolve

tphung3:
The CI failure doesn't seem to have anything to do with the proposed changes. @benclifford Can this be a transient error?

benclifford:
@tphung3 yeah there is a hang that happens fairly often in local tests that I don't think is easy to fix because it's coming from some awkward architectural choices years ago -- right now the answer is "suck it up and hit retry", and I hope to address what I think is the problem in the next year.

I'll hit retry.

parsl/executors/workqueue/executor.py
```diff
@@ -729,14 +729,25 @@ def _collect_work_queue_results(self):
...
+                # If result_received, then there's a result file. The object inside the file
+                # may be a valid result or an exception caused within the function invocation.
+                # Otherwise there's no result file, implying errors from WorkQueue.
                 if task_report.result_received:
-                    future.set_result(task_report.result)
+                    try:
+                        with open(task_report.result_file, 'rb') as f_in:
+                            result = deserialize(f_in.read())
+                    except Exception as e:
+                        logger.debug(f'Cannot load result from result file {task_report.result_file}. Exception: {e}')
```
benclifford, line 740: probably log this as an error not a debug?

parsl/executors/workqueue/executor.py
```diff
@@ -729,14 +729,25 @@ def _collect_work_queue_results(self):
...
+                            future.set_exception(result)
+                        else:
+                            future.set_result(result)
                 else:
                     # If there are no results, then the task failed according to one of
                     # work queue modes, such as resource exhaustion.
-                    ex = WorkQueueTaskFailure(task_report.reason, task_report.result)
-                    if task_report.result is not None:
-                        ex.__cause__ = task_report.result
+                    ex = WorkQueueTaskFailure(task_report.reason, Exception(task_report.reason))
```
benclifford, line 750: I think task_report.result is an exception object that has its own stack trace and more detailed reason for failure:

I feel like that exception object should be chained on using `__cause__` to give more detailed stack traces that preserve the information in the original exception, based on the principle that this code (like most code) is not in the right place to discard exception information.

Likewise for a bunch of other places where there's a task failure exception in an exception context (eg line 741, and the equivalents in taskvine)

tphung3: In the line you referred to there's no result file transferred back meaning that it's an internal WQ error. The exception chaining happens in line 743 where a result file comes back and it contains an exception. The executor sets the original exception then. I'll wrap the exception into a WQ Task Failure to make it clearer.

benclifford: right, but if there is *ever* an exception *anywhere*, I don't want it ever discarded apart from at the very outer fringe of parsl

tphung3: I think the new change adds function chaining whenever it is possible for WQ. Am I understanding you correctly @benclifford ?

benclifford: @tphung3 yes looks good for Work Queue - maybe Task Vine should have the same behaviour, though? This bit of the code is basically the same in both implementations. Other than that, this is ready to merge I think.

benclifford:
@tphung3 a note on force-pushing to a collaborative branch: looks like your force push here undid the test re-enabling I made, which I didn't notice https://github.com/Parsl/parsl/compare/321901aba999465681b0fe4ca5cd1437a38bc9cb..dbb25270b3dc0edec3b37b8ea16deecebfa6b4b8

tphung3:
My bad. Should have pulled the master merges first.

PR #2917, opened 2023-10-16
Remove unused import (from `ruff . --fix`)

Andrew-S-Rosen:
# Description

I applied [ruff](https://github.com/astral-sh/ruff) autofixes (i.e. `ruff . --fix`) to the source code, which found... precisely one unused import. 😅 

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/workqueue/parsl_coprocess.py
+++ b/parsl/executors/workqueue/parsl_coprocess.py
@@ -2,7 +2,6 @@
 
 import sys
 from parsl.app.errors import RemoteExceptionWrapper
-import parsl.executors.workqueue.exec_parsl_function as epf
 
 import socket
 import json


benclifford:
i feel like this removed a lot more than it should have

Andrew-S-Rosen:
@benclifford: There should only be one commit. You probably saw an older commit where I also ran `ruff` on the test suite, which wreaked havoc on things.

benclifford:
the existing tooling usually catches unused imports - it's interesting that it didn't catch this one

benclifford:
I realised why the one file this PR ended up modifying wasn't picked up by flake8: it's because this file isn't source code, but instead output code generated by `poncho_package_serverize` and committed into the Parsl tree (see PR #2605).

It makes less sense to edit machine generated object code, even though it happens to be in Python syntax - this isn't a file for humans to be editing, and changes to that file should be via changing how `poncho_package_serverize` creates that file, or by editing the source file `parsl_coprocess_stub.py`.

In a more pure build system, this file wouldn't exist in version control and would be built from the stub file at compile/build time - but that isn't what's happening, so this file ends up as a violation of "don't commit object code into version control" instead.

Andrew-S-Rosen:
Ah, well now that clears things up!

Issue #2915, opened 2023-10-15
error in parallel documentation


jkitchin:
This example does not work as written:

https://github.com/Parsl/parsl/blob/11509cba9c3f462bf4d56eceee4e06e0479d52f6/docs/userguide/workflow.rst?plain=1#L142

I think it should be either:

```
with ParslPoolExecutor(config) as exec:
    outputs = exec.map(generate, range(1, 5))
```

or 

```
with ParslPoolExecutor(config) as pool:
    outputs = pool.map(generate, range(1, 5))
```

The example also does not work for me on a Mac unless I change the import to `from parsl.configs.local_threads import config`. For the given import `from parsl.configs.htex_local import config` I get this error:

```
      3 from parsl.configs.htex_local import config
----> 5 with ParslPoolExecutor(config) as exec:
      6     outputs = exec.map(generate, range(1, 5))
      8 for result in outputs:

File ~/anaconda3/lib/python3.10/site-packages/parsl/concurrent/__init__.py:31, in ParslPoolExecutor.__init__(self, config)
     25 """Create the executor
     26 
     27 Args:
     28     config: Configuration for the Parsl Data Flow Kernel (DFK)
     29 """
     30 self._config = config
---> 31 self.dfk = DataFlowKernel(self._config)
     32 self._app_cache: Dict[Callable, PythonApp] = {}

File ~/anaconda3/lib/python3.10/site-packages/typeguard/__init__.py:1033, in typechecked.<locals>.wrapper(*args, **kwargs)
   1031 memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
   1032 check_argument_types(memo)
-> 1033 retval = func(*args, **kwargs)
   1034 try:
   1035     check_return_type(retval, memo)

File ~/anaconda3/lib/python3.10/site-packages/parsl/dataflow/dflow.py:185, in DataFlowKernel.__init__(self, config)
    183 self.data_manager = DataManager(self)
    184 parsl_internal_executor = ThreadPoolExecutor(max_threads=config.internal_tasks_max_threads, label='_parsl_internal')
--> 185 self.add_executors(config.executors)
    186 self.add_executors([parsl_internal_executor])
    188 if self.checkpoint_mode == "periodic":

File ~/anaconda3/lib/python3.10/site-packages/parsl/dataflow/dflow.py:1122, in DataFlowKernel.add_executors(self, executors)
   1119             self._create_remote_dirs_over_channel(executor.provider, executor.provider.channel)
[...]

Issue #2871, opened 2023-08-22
Reloading default HighThroughputExecutor after `parsl.clear()` results in `KeyError: 'block_id'` exception


d33bs:
**Describe the bug**

Thanks for the great work on Parsl - it's an awesome tool! I noticed when attempting to load a HighThroughputExecutor configuration after using `parsl.clear()` an exception occurs which states: `KeyError: 'block_id'`. This doesn't seem to happen when using ThreadPoolExecutor. I'm unsure if this is expected behavior so I wanted to raise a bug just in case it helps diagnose or address this issue.

**To Reproduce**
Run the following Python snippet:
```python
import parsl
from parsl.config import Config
from parsl.executors import HighThroughputExecutor

local_htex = Config(
    executors=[HighThroughputExecutor()],
)
parsl.load(local_htex)
parsl.clear()
parsl.load(local_htex)
````

**Expected behavior**
I'd expect something like what is witnessed with the ThreadPoolExecutor from the following example. If this isn't possible or is unreasonable to expect, I'd expect documentation or an upfront message from the logs about workarounds.

```python
import parsl
from parsl.config import Config
from parsl.executors import ThreadPoolExecutor

local_threaded = Config(
    executors=[ThreadPoolExecutor()],
)
parsl.load(local_threaded)
parsl.clear()
parsl.load(local_threaded)
```

**Environment**
- OS: MacOS
- Python: 3.9
- Parsl: 2023.08.21

**Distributed Environment**
- Where are you running the Parsl script from ? Laptop/Workstation
- Where do you need the workers to run ? Same as Parsl script

benclifford:
one thing in your example that sticks out to me is that you can't use the same `Config` object twice.

In the case of the thread pool executor, this probably doesn't matter, but it's likely to be more of a problem with htex.

There's a style for re-loading "the same" config that is used in (for example) the test suite and some applications that I work on that that looks something like this:

```
def fresh_config():
    return Config(executors=[HighThroughputExecutor()])

parsl.load(fresh_config())
parsl.clear()
parsl.load(fresh_config())
```

Probably we should make this more enforced - it wasn't part of the initial design of how Config was meant to work, but it's turned out since then that's a property of Config objects (and the subobjects contained inside a Config).

Can you try something like the above and see how things go?

benclifford:
[...]

PR #2919, opened 2023-10-16
Improve documentation about configuration and workflows

### LINKED ISSUE #2918: Docs: Further clarification on how serialization impacts where imports need to be
Jotting this down from Slack so I/we don't forget:

Me: 

> The current docs as-is are kind of confusing with regards to the need to put import statements in PythonApp definitions. there's some pretty strong-worded statements about always needing to do so, but that's not the case (e.g. if you're making a package that will be installable). i'll give this some thought too. I think Logan Ward and I discussed this at one point
 
@WardLT:

> Yea, we really do need to clarify how serialization works to explain that

### LINKED ISSUE #2925: Allowing for `inputs` and `outputs` to take immutable types
In looking at the Parsl docs, I see that in the [Passing Data section](https://parsl.readthedocs.io/en/stable/1-parsl-introduction.html#Passing-data) of the BashApps, there are various functions defined with mutable kwargs, e.g. `inputs=[]`. Generally, this is [undesirable](https://docs.python-guide.org/writing/gotchas/#:~:text=Python's%20default%20arguments%20are%20evaluated,to%20the%20function%20as%20well) and the Parsl docs should probably avoid promoting questionable programming practices.

In discussing this with @svandenhaute [here](https://github.com/molmod/psiflow/issues/14), it seems that it's possible that Parsl is expecting type `list` for the inputs and outputs with regards to file passing. If that's the reason for the above behavior, perhaps there might be a way to modify things such that if `None` is supplied, e.g. `inputs=None`, then Parsl assumes this is equivalent to `{}`.

Is my understanding of this correct? I have never used this feature. @WardLT, this is in the section of the docs you're updating so I figured I'd tag you on this one.

WardLT:
# Description

Clarifies two problems users found in the documentation

- [x] #2915: Decorators are not needed for `ParslPoolExecutor.map`
- [x] #2871 : Configs may not be re-used
- [x] Fixes #2918 : Explain when functions need imports
- [x] Fixes #2925 : Explain mutable types are undesirable

## Type of change

- Documentation update


docs/userguide/configuring.rst
```diff
@@ -48,13 +48,46 @@ used by the infiniband interface with ``address_by_interface('ib0')``
...
+
+.. code-block:: python
+
+    from parsl.configs.htex_local import config
+    import parsl
+
+    parsl.load(config)
+
+The ``load`` statement can happen after Apps are defined and must occur before tasks are started.
+
```
khk-globus, line 66: Grammar wise, suggest changing the conjunction to `but`:
```diff
- The ``load`` statement can happen after Apps are defined and must occur before tasks are started.
+ The ``load`` statement can happen after Apps are defined but must occur before tasks are started.
```

khk-globus (approved):
LGTM.

Andrew-S-Rosen:
Thanks for this. One additional thing this helps with is that there was previously no mention of `parsl.clear()` in the docs.

docs/userguide/configuring.rst
WardLT: Done :)

WardLT:
@Andrew-S-Rosen , do the new explanations about imports make sense to you?

docs/userguide/apps.rst
```diff
@@ -83,6 +83,65 @@ as in following code snippet, which copies the contents of one file (``in.txt``)
 
        echo(inputs=[in.txt], outputs=[out.txt])
 
+
+Imports and Global Variables
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Parsl apps have access to less information from the script that defined then
```
Andrew-S-Rosen, line 90: "that defined then" --> "that defined them"

Andrew-S-Rosen (commented):
@WardLT: This looks good to me! The point about the type-hints is especially subtle, so I'm glad that was highlighted.

There is some overlap thematically with the [section above](https://github.com/Parsl/parsl/pull/2919/files#diff-a1c5fe7e74358853afcdd3346452a8cfc62a68fd7f091f5186866356347fd8a0R34-R39). It's not a problem, but I just wanted to bring that up as well.

WardLT:
Thanks! I'll fix my grammar and de-duplicate some. Better to keep the docs shorter

WardLT:
There are actually a fair amount of duplication in this section. I'm going to clean it out while it's on my attention

docs/userguide/apps.rst
```diff
@@ -3,107 +3,100 @@
...
-A Parsl app is defined by annotating a Python function with a decorator: 
-the ``@python_app`` decorator for a **Python app**, the ``@bash_app`` decorator for a **Bash app**,
-and the ``@join_app`` decorator for a **Join app**.
+Apps can be one of three types of functions, each with their own type of decorator
 
-Python apps encapsulate pure Python code, while Bash apps wrap calls to external applications and scripts,
-and Join apps allow composition of other apps to form sub-workflows.
+- ``@python_app``: Most Python functions
+- ``@bash_app``: A python function which returns a command line program to execute
+- ``@join_app``: A function which returns launches other apps
```
Andrew-S-Rosen, line 15: "which returns launches" --> "which returns one or more Apps"

docs/userguide/apps.rst
```diff
@@ -3,107 +3,100 @@
...
 
-As a Parsl Python app is executed asynchronously, and potentially remotely, the function
-cannot assume access to shared program state. For example, it must explicitly import any 
-required modules and cannot refer to variables used outside the function. 
-Thus while the following code fragment is valid Python, it is not valid Parsl, 
-as the ``bad_double()`` function requires the `random` module and refers to the external 
-variable ``factor``.
 
-.. code-block:: python
+Python Apps run Python functions. The code inside a function marked by ``@python_app`` will run on a remote system.
```
Andrew-S-Rosen, line 32: "will run on a remote system" --- of course, in this example things are local. Might be better to say something like "is what will be executed either locally or on the remote system"

docs/userguide/apps.rst
```diff
@@ -3,107 +3,100 @@
...
+    for ch, co in good_global('parsl', 'a').result().items():
+        global_var[ch] += co
+
+
+Functions from Modules
+++++++++++++++++++++++
 
-Serializing Functions from Libraries
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The above rules do not apply for functions defined in Python modules.
```
Andrew-S-Rosen, line 96: Recommended change:

The above rules assume that the user is running the example code from a standalone script or Jupyter Notebook.

Functions that are defined in an installed Python module do not need to abide by these guidelines, as they are sent to workers differently than functions defined locally within a script.

docs/userguide/apps.rst
```diff
@@ -3,107 +3,100 @@
...
+Functions in installed modules are sent to workers differently than functions defined in a script.
 
-Parsl can create Apps directly from functions defined in Python modules.
-Supply the function as an argument to ``python_app`` rather than creating a new function which is decorated.
+Supply a module function as an argument to ``python_app`` rather than creating a new function which is decorated.
```
Andrew-S-Rosen, line 99: I mean, there is technically nothing stopping someone from defining a new function. It's just silly.

Suggestion:

To transform an imported function into an App, you do not need to create a new function to decorate. Rather, you can do the following:

docs/userguide/apps.rst
```diff
@@ -125,10 +118,67 @@ to include attributes from the original function (e.g., its name).
     my_max_app = python_app(my_max)
```
Andrew-S-Rosen, line 118: This is equivalent to doing:

.. code-block:: python

    @python_app
    my_max_app(*args, **kwargs):
        return my_max(*args, **kwargs)

docs/userguide/apps.rst
```diff
@@ -125,10 +118,67 @@ to include attributes from the original function (e.g., its name).
...
+
+  See the section on `Futures <futures.html>`_ for more details.
+
+
+Learn more about the types of data allowed in `the data section <data.html>`_.
+
+Special Keyword Arguments
++++++++++++++++++++++++++
+
+Parsl apps can use the following special reserved keyword arguments:
```
Andrew-S-Rosen, line 167: This can use an example because the uninitiated reader may not be familiar that you are referring to passing arguments _to the decorator_ rather than the underlying function.

Andrew-S-Rosen (commented):
Thanks for doing this, @WardLT! Here are some minor suggestions. Feel free to accept, reject, or do whatever with it.

docs/userguide/apps.rst
WardLT: Great point.

docs/userguide/apps.rst
WardLT: That does read much better, thanks.

docs/userguide/apps.rst
WardLT: Good call. I changed the wording to reflect that it isn't necessary to define an own function

docs/userguide/apps.rst
WardLT: Absolutely, and I also a added a section which differentiates "function" and "decorator" kwargs.

docs/userguide/apps.rst
WardLT: Very true. Nice way of tying it to something people would already think of

docs/userguide/configuring.rst
```diff
@@ -48,13 +48,46 @@ used by the infiniband interface with ``address_by_interface('ib0')``
...
+The :class:`~parsl.config.Config` object may not be used again after loaded.
+Consider a configuration function if the application will shut down and re-launch the DFK.
+
+.. code-block:: python
+
+    from parsl.config import Config
+    import parsl
+
+    def make_config() -> Config:
```
benclifford, line 75: in tests and a few other places this is called `fresh_config()`

Issue #2713, opened 2023-05-16
Flux + parsl testing


benclifford:
**Is your feature request related to a problem? Please describe.**

The current CI test suite does not test parsl+flux. That means changes to internals of parsl are not tested against flux. This is especially bothering me as I am working on some fairly serious rearrangements of parsl internals and I think it's very likely to break if not tested.

@jameshcorbett and maybe @vsoch do you have any opinions on this?

**Describe the solution you'd like**

Flux tested against each PR and reporting into github (somewhere: existing parsl CI or somewhere else that reports back to github)

**Describe alternatives you've considered**
Manual testing. That's extremely fragile.

vsoch:
Yeah! I’ve implemented flux testing for several tools, and you can use a fairly simple setup to run tests with flux start. Here is one example: https://github.com/reframe-hpc/reframe/blob/develop/.github/workflows/test-flux.yaml

mercybassey:
Hi, @benclifford I'd like to work on this issue.

benclifford:
@mercybassey ok whats going on in this issue:

We have Parsl in this github repository. Parsl executes tasks using components called "executors" - and there are several different executors. Some of them are listed here: https://parsl.readthedocs.io/en/stable/reference.html#executors

So @jameshcorbett and @vsoch added an executor that lets Parsl execute tasks through the software that they work on, Flux. You can read more about that here https://github.com/flux-framework and https://computing.llnl.gov/projects/flux-building-framework-resource-management

One thing that we've been trying to do more of over the past few years is have more automatic testing of as much as we can, so that for every PR we can have some confidence that we didn't accidentally break something.

For some of our executors we have testing: for example, we test install Work Queue on every test run and then test execution with the Work Queue executor - that's driven by a Makefile that does the install and then runs `pytest` configured to use Flux: https://github.com/Parsl/parsl/blob/86a05dd2b33e5f54d05f93c5bcabf5f8642c90f8/Makefile#L72

However, we don't have any testing like this at all: so I can believe parsl+flux worked it was first contributed by @jameshcorbett but there isn't any further testing that changes in Parsl haven't broken it since then.

So this issue is about testing the integration of Parsl and Flux to satisfy us that it still works, and in the future keeps working.

Some ideas how you could do that:

@vsoch pasted a link to a GitHub workflow that starts up Flux from a container and then runs a simple test script.
[...]

PR #2923, opened 2023-10-19
RADICAL-Pilot Integration with Parsl

AymenFJA:
# Description

Summary: This PR integrates RADICAL-Pilot workload management and runtime system with Parsl as an executor.

The executor offers the following capabilities:

1. MPI execution of heterogeneous tasks:
   -    Single/Multi-node MPI functions that require a private MPI communicator on the runtime.
   -    Single/Multi-node (non)/MPI executables.
2. Heterogeneous task execution of all the above on GPUs and CPUs.

3. Improve the submission mechanism by offering `bulk_mode` submission alongside `stream` submission of tasks from Parsl 
        to the RADICAL-Pilot runtime system.

Fixes: None

Type of change: Additional features.

Can you please guide me on where to put examples that show how the executor can be used?

Thanks all.


AymenFJA:
currently, the tests are complaining about `radical.pilot` not being installed, which makes sense. What is the best approach to fix this? do we need to add `radical.pilot` as a dependency in the `test-requirements`? otherwise no clue here.

benclifford:
@AymenFJA there are a few different approaches to testing that could happen here:

1. Do you want this code tested in the parsl build system on every PR?

  *yes* - see for example, Work Queue and Task Vine which install the cctools package (containing workqueue and taskvine) and the run the default 100-or-so tests against that local install - see this Makefile target: https://github.com/Parsl/parsl/blob/65cdfb1c18fc538f856e6440551a4421fd42e1bb/Makefile#L72

  An alternative approach, which I haven't tried but I think would be fine, is if you need a more custom environment in Github Actions, would be to run a whole separate action with that more custom environment testing only radical-pilot + parsl - in that case, you would add some new job into the CI workflow here, alongside the `main-test-suite` job, here: https://github.com/Parsl/parsl/blob/65cdfb1c18fc538f856e6440551a4421fd42e1bb/.github/workflows/ci.yaml#L9

  An even-more-alternative approach is that you run your own tests in your own infrastructure, on every parsl PR, and report results into github - that's what "readthedocs" does now, for example - that feels like a lot more work, but it's something I imagine might be nice if we ever persuade HPC facilities to test parsl automatically.

  *no* - see for example, Flux. Although we've talked a little bit about it here: https://github.com/Parsl/parsl/issues/2713. In that case, you'd need to make your tests not run during parsl CI. Some approaches approaches to that:

i) you can make your tests autodetect that radical isn't installed, and not run - see how flux does this here: https://github.com/Parsl/parsl/blob/65cdfb1c18fc538f856e6440551a4421fd42e1bb/parsl/tests/test_flux.py#L34
I am a bit uncomfortable with this approach because I dislike the general "this test will fail, so we didn't run it" approach to not failing.

ii) you can use the regular test suite, nothing radical specific, using a radical config file - see how the WQ and taskvine tests run this way -- and then do not invoke the pytest that way in CI (with the assumption that whoever is actually testing this would invoke pytest in a suitable environment and act on failures)

iii) If you want radical-specific tests that aren't generalisable across executors, mark them as `pytest.mark.local` and also mark them as requiring radical - see this test that requires a clean network: https://github.com/Parsl/parsl/blob/65cdfb1c18fc538f856e6440551a4421fd42e1bb/parsl/tests/test_staging/test_staging_https.py#L23
Then when invoking in environments where radical is not available (such as the tests in CI) we can add a `not radical` explicit disable, in the same way that a `not cleannet` restriction is added here: https://github.com/Parsl/parsl/blob/65cdfb1c18fc538f856e6440551a4421fd42e1bb/Makefile#L78


That's quite a lot of options - my preference is for doing more testing in CI rather than less.

I'll also note that the documentation build is broken - that's usually because the import of the executor is failing in a plain environment - that same problem happened for Work Queue/Task Vine in recent months, and was fixed here: https://github.com/Parsl/parsl/pull/2909

benclifford:
@AymenFJA  (if its enough to `pip install radical.pilot` to get everything needed - then option 1 above would be add it to test requirements as you mentioned above -- on re-reading I think that's what you're saying?)

AymenFJA:
> @AymenFJA (if its enough to `pip install radical.pilot` to get everything needed - then option 1 above would be add it to test requirements as you mentioned above -- on re-reading I think that's what you're saying?)

@benclifford Yes, that was my first understanding of how to approach this issue. Let me try to add it to the `test-requirements` as this approach seems to be straightforward.

AymenFJA:
@benclifford the test are failing now with an error that is not related to `radical` i think. I also did not see the `radical_test` invoked anywhere in the failed/passed tests logs. Any idea?

benclifford:
@AymenFJA I think your test files not match this rule:

https://docs.pytest.org/en/7.1.x/how-to/usage.html

AymenFJA:
> I think that test failure you saw is indeed unrelated to your PR - I see the python3.10 test case for parsl (not your code) ran fine.

@benclifford thanks. The tests are getting invoked now. At the moment, it seems like the functions are being executed by a different executor than the `RadicalPilotExecutor` executor and fail. These tests should be run exclusively by the `RadicalPilotExecutor`. 

For example, when I test the `radical_tests` on my machine I do `make radical_ex_test` and this triggers the `RadicalPilotExecutor` and it's runtime system to execute all of the specified tests via the `RadicalPilotExecutor`.

benclifford:
There are two sets of test in parsl, intermingled: one where a test specifies its own config, which is activated with `--config local` and identified with `@pytest.mark.local`, and another where the tests should run with every executor and the config is specified on the command line.

So I think you should be trying to run both kinds of tests:

Run all the "every executor" tests with a radical config file - this gives confidence that many parsl features like returning a value, and task ordering will work with the radical executor.

Mark the radical-only tests with `@pytest.mark.local`  and specify the radical config you want in the actual test file - for example this  `test_local_adhoc` test runs in `--config local` mode and tests behaviour with a specific parsl config, not for all of the different configs in the Makefile:

https://github.com/Parsl/parsl/blob/a5aede7e3e0dab3d88896f7f47c7c73ebc6887f4/parsl/tests/sites/test_local_adhoc.py#L4

benclifford:
Separate to getting tests running in CI, I'll tag @yadudoc for review - he has done a bunch of work on parsl+MPI recently.

benclifford:
looks like the last commit is hanging in the radical tests (and github actions is timing out after about an hour, as configured)

```
Wed, 25 Oct 2023 21:08:27 GMT
pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10
Wed, 25 Oct 2023 21:08:28 GMT
============================= test session starts ==============================
Wed, 25 Oct 2023 21:08:28 GMT
platform linux -- Python 3.8.18, pytest-7.4.3, pluggy-1.3.0
Wed, 25 Oct 2023 21:08:28 GMT
Using --random-order-bucket=module
Wed, 25 Oct 2023 21:08:28 GMT
Using --random-order-seed=227914
Wed, 25 Oct 2023 21:08:28 GMT

Wed, 25 Oct 2023 21:08:28 GMT
rootdir: /home/runner/work/parsl/parsl/parsl/tests
Wed, 25 Oct 2023 21:08:28 GMT
configfile: pytest.ini
Wed, 25 Oct 2023 21:08:28 GMT
plugins: cov-4.1.0, random-order-1.1.0, typeguard-2.13.3
Wed, 25 Oct 2023 21:08:28 GMT
collected 9 items / 2 deselected / 7 selected
Wed, 25 Oct 2023 21:08:28 GMT

Wed, 25 Oct 2023 21:50:18 GMT
Error: The operation was canceled.
```

You should be able to see any logfiles that this created by looking in the artifacts directory for that run - in the artifacts tab, for example here: https://github.com/Parsl/parsl/actions/runs/6645997496?pr=2923#artifacts

AymenFJA:
Hey @benclifford thanks for clarifying the tests. I did some changes to separate the MPI test from the general parsl tests (for all executors). The tests are passing on my local machine, I see it takes very lone time here, any idea how to debug this, or at least how to pull the radical sessions from the CI assuming that the radical tests started (although i see that the tests timed out maybe due to exceeding the time limit )

AymenFJA:
Hey @benclifford thanks for clarifying the tests. I did some changes to separate the MPI test from the general parsl tests (for all executors). The tests are passing on my local machine, I see it takes very lone time here, any idea how to debug this, or at least how to pull the radical sessions from the CI assuming that the radical tests started (although i see that the tests timedout maybe due to exceeding the time limit )

benclifford:
Have a look at the test logs and you should see that the last command to run was this one:

```
pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10
```

and if you choose to turn on timestamps (in one of the drop downs in the top right) you should see that nothing happens after that for the rest of the time hour.

If your code is making log files, you should be able to see those log files in the artifacts I mentioned, which might help you debug what is happening there.

benclifford:
on my laptop, this also seems to hang:

```
pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10
===================================================== test session starts ======================================================
platform linux -- Python 3.11.2, pytest-7.4.0, pluggy-0.13.1
Using --random-order-bucket=module
Using --random-order-seed=840798

rootdir: /home/benc/parsl/src/parsl/parsl/tests
configfile: pytest.ini
plugins: random-order-1.1.0, forked-1.6.0, hypothesis-6.82.1, cov-4.0.0, typeguard-2.13.3, xdist-1.26.1
collected 9 items / 2 deselected / 7 selected                                                                                  

parsl/tests/test_callables.py 
```

benclifford:
digging in on my laptop, it looks like it hangs here: I get `BENC: before` but not `BENC: after`

```
$ git diff
diff --git a/parsl/executors/radical/executor.py b/parsl/executors/radical/executor.py
index 89f63bbb39..ed930c6c0f 100644
--- a/parsl/executors/radical/executor.py
+++ b/parsl/executors/radical/executor.py
@@ -179,8 +179,10 @@ class RadicalPilotExecutor(ParslExecutor, RepresentationMixin):
         logger.info("starting RadicalPilotExecutor")
         logger.info('Parsl: {0}'.format(parsl.__version__))
         logger.info('RADICAL pilot: {0}'.format(rp.version))
+        logger.debug("BENC: before")
         self.session = rp.Session(uid=ru.generate_id('rpex.session',
                                                      mode=ru.ID_PRIVATE))
+        logger.debug("BENC: after")
 
```

I'm not sure what further info is available for debugging that call?

AymenFJA:
> Have a look at the test logs and you should see that the last command to run was this one:
> 
> ```
> pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10
> ```
> 
> and if you choose to turn on timestamps (in one of the drop downs in the top right) you should see that nothing happens after that for the rest of the time hour.
> 
> If your code is making log files, you should be able to see those log files in the artifacts I mentioned, which might help you debug what is happening there.


> pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10

It should take a while to prepare the pilot environment, but it should not hang forever, at least, this happens when I run the command you provided locally. 

I should see something like this: 
```shell
aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$ ls
CITATION.cff        LICENSE      codemeta.json    mypy.ini          rpex.session.surfacebook.aymen.019656.0000  test-requirements.txt
CITATION.md         MANIFEST.in  docs             parsl             runinfo
CODE_OF_CONDUCT.md  Makefile     interchange.log  requirements.txt  setup.py
CONTRIBUTING.rst    README.rst   mypy-stubs       rpex.cfg          tag_and_release.sh
aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$ c^C
aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$
```

and then it passes: 

```shell
(parsl_test_suite) aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$ pytest parsl/tests/test_callables.py -k "not cleannet and not dynamically_loaded_module and not test_check_importlib_file_function" --config parsl/tests/configs/local_radical.py --random-order --durations 10
=================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.10, pytest-7.4.2, pluggy-1.3.0
Using --random-order-bucket=module
Using --random-order-seed=926372

rootdir: /home/aymen/RPEX/NEW_PARSL/parsl/parsl/tests
configfile: pytest.ini
plugins: cov-4.1.0, random-order-1.1.0, typeguard-2.13.3
collected 9 items / 2 deselected / 7 selected

parsl/tests/test_callables.py .......                                                                                                                                                [100%]

=================================================================================== slowest 10 durations ===================================================================================
98.03s setup    test_callables.py::test_check_import_module_function
8.64s call     test_callables.py::test_check_import_module_function
6.18s teardown test_callables.py::test_check_import_module_function_partial
3.16s call     test_callables.py::test_check_import_module_function_partial
3.12s call     test_callables.py::test_check_importlib_module_function
3.02s call     test_callables.py::test_check_this_module_function_partial
2.96s call     test_callables.py::test_check_this_module_function
2.87s call     test_callables.py::test_check_importlib_module_partial
1.03s call     test_callables.py::test_check_base_app

(1 durations < 0.005s hidden.  Use -vv to show these durations.)
======================================================================= 7 passed, 2 deselected in 129.06s (0:02:09) ========================================================================
(parsl_test_suite) aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$
```

Radical should generate a session: `rpex.session.surfacebook.aymen.019656.0000` that has all of the logs files or at least the error.
I do not see this session in the artifacts of the workflow after it fails here.

benclifford:
eg. maybe set `path: .`

AymenFJA:
> digging in on my laptop, it looks like it hangs here: I get `BENC: before` but not `BENC: after`
> 
> ```
> $ git diff
> diff --git a/parsl/executors/radical/executor.py b/parsl/executors/radical/executor.py
> index 89f63bbb39..ed930c6c0f 100644
> --- a/parsl/executors/radical/executor.py
> +++ b/parsl/executors/radical/executor.py
> @@ -179,8 +179,10 @@ class RadicalPilotExecutor(ParslExecutor, RepresentationMixin):
>          logger.info("starting RadicalPilotExecutor")
>          logger.info('Parsl: {0}'.format(parsl.__version__))
>          logger.info('RADICAL pilot: {0}'.format(rp.version))
> +        logger.debug("BENC: before")
>          self.session = rp.Session(uid=ru.generate_id('rpex.session',
>                                                       mode=ru.ID_PRIVATE))
> +        logger.debug("BENC: after")
>  
> ```
> 
> I'm not sure what further info is available for debugging that call?

Do you see something like `rpex.session.xxxx.xxxx`? if so you can check the logs inside that folder.

AymenFJA:
> github actions are configured to archive whats in the runinfo/ directory. On my laptop it looks like those rpex.session.* files appear in the working directory, not in runinfo.
> 
> It would probably be better if the relevant files get put in the runinfo directory, but in the meantime, you can edit this on your branch for debugging purposes to grab the whole tree instead of just the runinfo subdirectory:
> 
> https://github.com/Parsl/parsl/blob/cc52e976a6dd705db52d155025c9971e76c1a3a1/.github/workflows/ci.yaml#L77

That makes a lot of sense. Allow me to set the `session path` in the executor. Thanks for noticing this.

benclifford:
I see rpex.session.* and about 112 files in there. After a few minutes I see a bit more progress: 

```
DEBUG    parsl.executors.radical.executor:executor.py:382 submit bulk: 1
```

I'll leave it to churn a bit longer and see if it makes any more progress. In the meantime, it would be good if you can push on getting more debuggability in CI, because more debuggability there will lead to more debuggability for end users, I think.

AymenFJA:
> I see rpex.session.* and about 112 files in there. After a few minutes I see a bit more progress:
> 
> ```
> DEBUG    parsl.executors.radical.executor:executor.py:382 submit bulk: 1
> ```
> 
> I'll leave it to churn a bit longer and see if it makes any more progress. In the meantime, it would be good if you can push on getting more debuggability in CI, because more debuggability there will lead to more debuggability for end users, I think.

@benclifford, the sessions of Radical are direct in the `runinfo`. The problem now is that the ci test can not archive the runinfo: 
`Error: The log was not found. It may have been deleted based on retention settings.`

benclifford:
well that's frustrating - let me see what I can do about that, because the whole point of capturing runinfo is for exactly this sort of situation

AymenFJA:
> well that's frustrating - let me see what I can do about that, because the whole point of capturing runinfo is for exactly this sort of situation

Agreed and thanks.

benclifford:
I just pushed a modification to the CI workflow which might help here... let's see what happens.

benclifford:
@AymenFJA I'll be offline for some time, but I hope that the timestamps I pushed up will get you useful files, or at least point you in the right direction for further timestamp fiddling.

AymenFJA:
> @AymenFJA I'll be offline for some time, but I hope that the timestamps I pushed up will get you useful files, or at least point you in the right direction for further timestamp fiddling.

Thank you, @benclifford.

AymenFJA:
@benclifford From the logs that I just pulled, it seems like there is no MPI launch method (MPIEXEC or MPIRUN) that exists on the test machine. Is there any way to install it when we prepare the test machine?

benclifford:
Our build system used to install MPI, when we had a different MPI-based executor, the extreme scale executor.

This was the script:

https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/parsl/executors/extreme_scale/install-mpi.sh

and then it was invoked previously like this:

https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/Makefile#L77

It looks like you already found a different way that seems ok too, in your commit 7d3eeac

benclifford:
@AymenFJA your most recent tests look like they passed the radical stage, and then failed in the testing of parsl-visualize because I deleted a bunch of stuff needed for that test in https://github.com/Parsl/parsl/pull/2923/commits/96d45cc739a404d6faeabc9a5e20bb76d937e48e

I can revert those changes now I think, so that this branch runs all the tests it did before.

benclifford:
@AymenFJA ok the tests move onto the `--config local` stage and there is a failure of `parsl/tests/test_radical/test_mpi_funcs.py` there, with an extremely long exception message.

benclifford:
environment-wise, a lot of parsl work assumes that the execute-side environment will be the same as the submit side, or at least that you're going to put some effort into managing the differences somehow outside of Parsl - work queue and task vine, for example, have the ability to create a remote conda environment; some work has used other containers.

So there's a topic of "how to manage your remote environment" that we mostly avoid by saying "start with a shared filesystem" that I personally think is an interesting direction ripe for development.

AymenFJA:
> environment-wise, a lot of parsl work assumes that the execute-side environment will be the same as the submit side, or at least that you're going to put some effort into managing the differences somehow outside of Parsl - work queue and task vine, for example, have the ability to create a remote conda environment; some other work has used containers.
> 
> So there's a topic of "how to manage your remote environment" that we mostly avoid by saying "start with a shared filesystem" that I personally think is an interesting direction ripe for development.

Good point. While Radical totally assumes that there are two different environments:

1. Client (local user or login nod, etc..)
2. Agent (Remote resources such as compute nodes)

It also supports using existing environments such as python env and conda (including the client side env) if we instruct Radical to use it and it is possible also via the RPEX executor.

AymenFJA:
@benclifford  Ok, this is passing now. Should I follow the TaskVine approach by making RPEX importable only by explicitly doing:

```python
from parsl.executors.radical.executor import RadicalPilotExecutor
```

to fix the `docs`?

benclifford:
Yeah something like how taskvine or the other conditional/protected import stuff (monitoring, I think?) is probably how to fix the docs.

benclifford:
ok cool, now that tests are passing, hopefully @yadudoc can look at the MPI parts of this and how it aligns with other MPI stuff that is going on, and I'll actually look at the code rather than grumbling about test cases failing.

yadudoc (commented):
Hi @AymenFJA here are some comments from a partial review. I haven't covered the executor itself yet.

parsl/tests/test_radical/test_mpi_funcs.py
```diff
@@ -0,0 +1,23 @@
...
+    return msg
+
+
+apps = []
+
+
+@pytest.mark.local
+def test_radical_mpi(n=10):
+    for i in range(n):
+        t = test_mpi_func(None, msg='mpi.func.%06d' % i, sleep=1, ranks=2)
```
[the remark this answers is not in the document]
AymenFJA: Yes, I changed the tests accordingly to reflect this.

parsl/tests/test_radical/test_mpi_funcs.py
```diff
@@ -0,0 +1,23 @@
...
+
+from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
+
+
+@parsl.python_app
+def test_mpi_func(comm, msg, sleep, ranks):
+    import time
+    msg = 'hello %d/%d: %s' % (comm.rank, comm.size, msg)
+    time.sleep(sleep)
+    return msg
```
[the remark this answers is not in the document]
AymenFJA: Yes, this is also changed now in the tests, and we assert the returned `comm_size` with the specified number of ranks.

parsl/tests/test_radical/test_mpi_funcs.py
```diff
@@ -0,0 +1,23 @@
+import parsl
+import pytest
+
+from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
+
+
+@parsl.python_app
+def test_mpi_func(comm, msg, sleep, ranks):
```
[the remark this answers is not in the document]
AymenFJA: So 2 options are here: 

1. We can set the `comm` as an argument but it must be the first one in the function definition/signature
2. We can set the `comm` as a keyword argument and it can be `None` anywhere.

This is how Radical Pilot identifies the `comm`

parsl/tests/configs/local_radical.py
```diff
@@ -0,0 +1,23 @@
...
+rpex_cfg.pilot_env_setup.append(parsl_src)
+
+
+def fresh_config():
+
+    return Config(
+            executors=[
+                RadicalPilotExecutor(
+                    label='RPEXBulk',
+                    rpex_cfg=rpex_cfg.get_cfg_file(),
```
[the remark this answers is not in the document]
AymenFJA: This setup comes from RP's (Radical Pilot) capabilities to run both functions and executables MPI/non-MPI concurrently. Here, we need to tell Radical Pilot the masters/workers resource specification somehow. The resource config (`rpex_cfg`) mainly specified the masters/workers and their setup per node and their numbers, etc.

parsl/executors/radical/rpex_resources.py
```diff
@@ -0,0 +1,135 @@
...
+
+    worker_gpus_per_node : int
+        The number of GPUs a worker will operate on per node.
+        Default is 0.
+
+    worker_cores_per_node : int
+        The number of CPU cores a worker will operate on per node.
+        Default is 4.
+
+    masters_per_node : int
```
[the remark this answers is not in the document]
AymenFJA: This is fixed now, and it is switched to `cores_per_master` instead.

parsl/tests/test_radical/test_mpi_funcs.py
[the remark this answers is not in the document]
benclifford: I feel MPI based apps are going to have different semantics compared to regular python_apps (see also how bash_apps have different semantics to python_apps) - so I don't think we need to be too scared about adding in new parameters that only make sense in the context of a multi-rank MPI app.

Other bits of parsl use magic parameters in a few ways, without much consistency:

* `inputs` is a magic list that will get file *input* staging done on all of its members
* `outputs` is a magic list that will get file *output* staging done on all of its members
* `parsl_resource_specification` lets an app invoker make special requests for how it is executed (for example, amount of memory to budget when using the Work Queue Executor)

What little consistency there is suggests that `comm` should be a keyword argument, option 2 above, and that you don't need to specify it when invoking - instead it will be silently inserted by an MPI-aware executor.

`test_mpi_func(msg='mpi.func.%06d' % i, sleep=1, ranks=2)`

rather than 

` test_mpi_func(None, msg='mpi.func.%06d' % i, sleep=1, ranks=2, comm=None)`

I'm not sure how this overlaps with ideas that Yadu has had about multirank python apps: so far in his prototype (#2905 ) we mostly talked about command-line based stuff which looks, I think, quite different.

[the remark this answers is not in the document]
yadudoc: That makes sense. If the executor passes in a communicator, and the function makes use of it, then it couldn't possibly run on any of the other executors currently supported. A new `mpi_python_app` that's separate from `python_app` might be more appropriate. This would be required since a single invocation of this app with ranks>1 is expected to launch multiple instances of the function, and we'd need to think through existing options like `inputs`, `outputs`, and `stdout` in this context. 

For now, setting `comm` to be a special kwargs should be good enough to be consistent. 

For now, to get this feature across the finish line,

AymenFJA:
@benclifford one thing I stumbled on while trying to fix the tests and include more is:
Radical tends to execute bash apps (executables) like this (i.e. using double quotation)
```
$RP_PROF rank_start
/bin/bash "-c" "$((5/0))"
RP_RET=$?
$RP_PROF rank_stop
```

which results in exit code: 2

while if you do:
```
/bin/bash '-c' '$((5/0))'
```

results in exit code: 1

This fails in here: https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/tests/test_bash_apps/test_error_codes.py#L48 as the tests expect exit code 1
How to align this? any idea?

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,447 @@
...
+                    if BASH in task_type:
+                        task_type = BASH
+                        func = func.args[0]
+                    else:
+                        task_type = PYTHON
+
+                except Exception:
+                    logger.exception('unwrap failed')
+
+                    return func, args, task_type
```
benclifford, line 311: maybe this should raise an exception (perhaps the existing exception) here rather than return?

benclifford:
@AymenFJA the executor-level semantics for running a parsl task are basically "here's a python `callable`, go run it". That Python callable isn't necessarily the function that has been decorated by a bash_app or python_app decorator, but instead there can be various wrappers around that.

The most obvious wrapper (and where you are hitting a problem with your implementation here...

https://github.com/Parsl/parsl/pull/2923/files#diff-69a18fbee447c4ccac0849af38f44d7dd7858503c75d6c32e7b27933597fda49R269

... is that bash_app pretty means "put this python wrapper `remote_side_bash_executor`  around this other python callable", https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/app/bash.py#L13

Other wrappers that get put around including the monitoring wrapper, when resource monitoring is turn on (which spawns a monitoring process along side the task), and some file staging options (for example, to perform a worker-side file staging before executing the actual task code).

The naive answer is "don't do this," or "if you want to unwrap the Python callable stack and simulate it being called elsewhere, you need to simulate it properly".

The more complicated answer is, let's try to understand why this needs to happen here: why do you need to try to figure out and launch an executable differently than Parsl already does? (for example, what breaks when you run everything as a submitted Python callable rather than trying to untangle it?)

AymenFJA:
The tests are failing as we still do not support inputs/ouptuts on the executor level, I am trying to wrap my head around how to do so at least for the `bash_apps`

benclifford:
> The tests are failing as we still do not support inputs/ouptuts on the executor level, I am trying to wrap my head around how to do so at least for the bash_apps

@AymenFJA the work queue executor maps parsl input/output files to work queue's own specification - that might be interesting to read: https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/executors/workqueue/executor.py#L452

basically it sweeps over the argument lists in the same pattern as the regular Parsl staging code to pick out files that should be staged in/out, and then registers them with the work-queue-task that is being built from the parsl task.

AymenFJA:
> > The tests are failing as we still do not support inputs/ouptuts on the executor level, I am trying to wrap my head around how to do so at least for the bash_apps
> 
> @AymenFJA the work queue executor maps parsl input/output files to work queue's own specification - that might be interesting to read:
> 
> https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/executors/workqueue/executor.py#L452
> 
> basically it sweeps over the argument lists in the same pattern as the regular Parsl staging code to pick out files that should be staged in/out, and then registers them with the work-queue-task that is being built from the parsl task.

Thanks, I aligned the inputs outputs with Radical `staging inputs/outputs`

AymenFJA:
> > The tests are failing as we still do not support inputs/ouptuts on the executor level, I am trying to wrap my head around how to do so at least for the bash_apps
> 
> @AymenFJA the work queue executor maps parsl input/output files to work queue's own specification - that might be interesting to read:
> 
> https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/executors/workqueue/executor.py#L452
> 
> basically it sweeps over the argument lists in the same pattern as the regular Parsl staging code to pick out files that should be staged in/out, and then registers them with the work-queue-task that is being built from the parsl task.

Thanks, I aligned the inputs outputs with Radical `staging inputs/outputs`

> @AymenFJA the executor-level semantics for running a parsl task are basically "here's a python `callable`, go run it". That Python callable isn't necessarily the function that has been decorated by a bash_app or python_app decorator, but instead there can be various wrappers around that.
> 
> The most obvious wrapper (and where you are hitting a problem with your implementation here...
> 
> https://github.com/Parsl/parsl/pull/2923/files#diff-69a18fbee447c4ccac0849af38f44d7dd7858503c75d6c32e7b27933597fda49R269
> 
> ... is that bash_app pretty means "put this python wrapper `remote_side_bash_executor` around this other python callable",
> 
> https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/app/bash.py#L13
> 
> Other wrappers that get put around including the monitoring wrapper, when resource monitoring is turn on (which spawns a monitoring process along side the task), and some file staging options (for example, to perform a worker-side file staging before executing the actual task code).
> 
> The naive answer is "don't do this," or "if you want to unwrap the Python callable stack and simulate it being called elsewhere, you need to simulate it properly".
> 
> The more complicated answer is, let's try to understand why this needs to happen here: why do you need to try to figure out and launch an executable differently than Parsl already does? (for example, what breaks when you run everything as a submitted Python callable rather than trying to untangle it?)

Well, we 

> @AymenFJA the executor-level semantics for running a parsl task are basically "here's a python `callable`, go run it". That Python callable isn't necessarily the function that has been decorated by a bash_app or python_app decorator, but instead there can be various wrappers around that.
> 
> The most obvious wrapper (and where you are hitting a problem with your implementation here...
> 
> https://github.com/Parsl/parsl/pull/2923/files#diff-69a18fbee447c4ccac0849af38f44d7dd7858503c75d6c32e7b27933597fda49R269
> 
> ... is that bash_app pretty means "put this python wrapper `remote_side_bash_executor` around this other python callable",
> 
> https://github.com/Parsl/parsl/blob/f8b18366269c308138e2eeadbc00a4dceb63bf01/parsl/app/bash.py#L13
> 
> Other wrappers that get put around including the monitoring wrapper, when resource monitoring is turn on (which spawns a monitoring process along side the task), and some file staging options (for example, to perform a worker-side file staging before executing the actual task code).
> 
> The naive answer is "don't do this," or "if you want to unwrap the Python callable stack and simulate it being called elsewhere, you need to simulate it properly".
> 
> The more complicated answer is, let's try to understand why this needs to happen here: why do you need to try to figure out and launch an executable differently than Parsl already does? (for example, what breaks when you run everything as a submitted Python callable rather than trying to untangle it?)

Thanks, Ben. We always differentiate between `executables` in Parsl lingo `bash_app` and `functions` `python_app` and the way we create them, process them, and execute them. For executables, we tend to create a launcher script and a wrapper script (around the executable and its arguments) that can be invoked by the launcher script. We do this mainly for MPI capabilities to control the number of RANKS and others and capabilities like: `pre_execution`, `pre_launching`, and many more. Now, from my perspective, I need to extract that executable command from the `bash_app`, which is fundamentally a Python function, in order to let Radical process it appropriately. I do this in memory and I do not create a serialized `parsl_function_file` for a very obvious reason: `performance` as if we execute millions of these tasks then we will definitely nail down the file system of any machine.  Now that is my perspective on how and why I do it in this way. I think at the same time @andre-merzky has a very good reason too. Please let me know if I misunderstood your question/suggestion. Thanks.

parsl/tests/test_radical/test_mpi_funcs.py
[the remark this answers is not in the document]
AymenFJA: This is now addressed here: https://github.com/Parsl/parsl/pull/2923/commits/0e4c709fb833650bc74b78f0ae00409fe56f5745

parsl/executors/radical/executor.py
AymenFJA: > maybe this should raise an exception (perhaps the existing exception) here rather than return?

addressed here: https://github.com/Parsl/parsl/pull/2923/commits/9fe38da162aa0d2f6fb4981fb1bfc4f565cf3b30

andre-merzky:
> why do you need to try to figure out and launch an executable differently than Parsl already does?

I agree with @AymenFJA's answer here.  To rephrase in my own words though:  RP attempts to provide the same semantic behavior for different task types: executables, function calls, shell commands, code snippets (which are `eval`'ed) etc.  That behavior includes `pre_exec`, `post_exec`, but also profiling probes and logging granularity.  Unwrapping the callables makes that simple for us, as wrapping around those is then easy and we can, for example, execute the `pre_exec` commands in the same `popen` call w/o needing to spawn additional processes.  

But we don't *need* to do this.  @benclifford : are you merely curious, or also concerned about the added overhead?  If the latter, we can try to run both options and do some profiling.  I would suggest to do that outside of this PR though.

benclifford:
> But we don't need to do this. @benclifford : are you merely curious, or also concerned about the added overhead?

What I'm basically concerned about is that the right stuff gets executed - i.e. correctness, rather than performance/overhead - that's a programming language person vs HPC tension (we don't care what we run as long as we run it fast vs we don't care how slow it runs as long as the correct thing gets run)

So what I'm claiming is the executor in this PR will not correctly execute all the different Python callables that Parsl might throw at it - just the ones that happen to fit the structure that this code expects:

https://github.com/Parsl/parsl/pull/2923/files#diff-69a18fbee447c4ccac0849af38f44d7dd7858503c75d6c32e7b27933597fda49R270

You could say "we'll only accept explicitly written out apps, and we won't any 'wrapper style' Parsl functionality such as monitoring and remote-side file staging" as a limitation of that executor - and I think that's probably fine for as long as you are clear about it your own minds and in documentation to users.   [From the parsl perspective, I'll probably play devil's-power-user and make sure we have enough test cases to demonstrate what those users do which would be easy to disable for radical) because I think we might not have enough test coverage there.]

Way way more "interesting for Parsl, because it is completely new functionality for Parsl is how this sort of "the callable is not the app the user wrote" interacts with MPI-based stuff:

The current implementation of #2905 only works with command-line based MPI, and does not attempt to address python-function MPI (eg. the test_mpi_funcs.py testcase in this radical #2923 PR)

In that #2905 PR,  it's up to the user to explicitly specify where the launcher gets put in a user command-line - see this example, rather than the executor trying to get it in there:

https://github.com/Parsl/parsl/pull/2905/files#diff-ff4d4ac1c264520bd747fac9ecc03216b0715048e24b3df881872c80069158ceR27

The #2905 syntax isn't something concrete: it's out there as a strawman for us to experiment with users, and I probably prefer a style where that doesn't need to be added into the command-line explicitly.  PR #2905 is implemented that way at the moment, I think, because it's the easiest way to hook into how the high throughput executor does things, without having to do the lame unwrapping that I am so displeased with above.

Maybe this leads to some change in the core parsl executor API (beyond the scope of this PR #2923): I do want you to run the entire Python function to get wrapping behaviour, but both in the case of #2905-with-nicer-syntax and this PR #2923, I want the executor to get to decide how any bash-style commandline gets executed rather than parsl supplying a Popen call in a wrapper.

And so I don't have a "do it this way" answer for this bit - which maybe leads to keeping both #2905 and #2923 behaviour, accepting that they're different and accepting the limitations both these PRs have, and working on some better Parsl core interface after some experience.

benclifford:
This PR disables some tests, and I thought I would go through and keep disabling more tests that didn't pass to get a better feel for which parsl features aren't supported. Here's the command line I ended up with:

```
 LOCAL_SANDBOX=True pytest parsl/tests/ -k "not cleannet and not issue363 and not dynamically_loaded_module and not test_check_importlib_file_function and not test_apptimeout and not test_parallel_for and not test_file_staging and not test_error_codes and not test_bash_memoization and not test_memo_stdout and not test_increment and not test_bad_stdout_specs and not test_files and not test_output_chain_filenames and not test_outputs and not test_workflow4 and not test_fail and not test_python_walltime and not test_custom_exception and not test_fut_case_2 and not test_fut_case_4 and not test_name_error and not test_import_error and not test_error and not test_two_errors and not test_one_error_one_result and not test_lifted and not test_226" --config parsl/tests/configs/local_radical.py  --durations 10
```

I haven't investigated each disablement individually, but from briefly looking at the errors that came past, I think the main causes are broadly: handling of stdout not being right; handling of output files not being right; exceptions not being returned properly from the user application; something not working with imports.

AymenFJA:
> This PR disables some tests, and I thought I would go through and keep disabling more tests that didn't pass to get a better feel for which parsl features aren't supported. Here's the command line I ended up with:
> 
> ```
>  LOCAL_SANDBOX=True pytest parsl/tests/ -k "not cleannet and not issue363 and not dynamically_loaded_module and not test_check_importlib_file_function and not test_apptimeout and not test_parallel_for and not test_file_staging and not test_error_codes and not test_bash_memoization and not test_memo_stdout and not test_increment and not test_bad_stdout_specs and not test_files and not test_output_chain_filenames and not test_outputs and not test_workflow4 and not test_fail and not test_python_walltime and not test_custom_exception and not test_fut_case_2 and not test_fut_case_4 and not test_name_error and not test_import_error and not test_error and not test_two_errors and not test_one_error_one_result and not test_lifted and not test_226" --config parsl/tests/configs/local_radical.py  --durations 10
> ```
> 
> I haven't investigated each disablement individually, but from briefly looking at the errors that came past, I think the main causes are broadly: handling of stdout not being right; handling of output files not being right; exceptions not being returned properly from the user application; something not working with imports.

@benclifford thanks. I just committed fixes for staging in/out and stdout/err and more. We still have some tests that are failing for example: 
```python
def test_returns_a_class_instance():
    # precondition
    assert returns_a_class_instance().result().q == "A"

    # test of commuting . and result()
    assert returns_a_class_instance().q.result() == "A"


def test_returns_a_class_instance_no_underscores():
    # test that _underscore attribute references are not lifted
    f = returns_a_class_instance()
    with pytest.raises(AttributeError):
        f._nosuchattribute.result()
    f.exception()  # wait for f to complete before the test ends


def test_returns_a_class():

    # precondition that returns_a_class behaves
    # correctly
    assert returns_a_class().result().a == "X"

    # check that the deferred __getattr__ functionality works,
    # allowing [] to be used on an AppFuture
    assert returns_a_class().a.result() == "X"

    # when the result is not indexable, a sensible error should
    # appear in the appropriate future
```
And the reason is that we do not serialize the results of the function. I am preparing a fix for this. 

I agree with you in the end, Radical will have some limitations that we can describe in the documentation, but at the moment, balancing between performance and things that are running correctly is important for us. 

I will eventually filter what runs and what doesn't in the Makefile.

andre-merzky:
> What I'm basically concerned about

Thanks for expanding on this, @benclifford, much appreciated.  

> So what I'm claiming is the executor in this PR will not correctly execute all the different Python callables that Parsl might throw at it - just the ones that happen to fit the structure that this code expects:

That's a fair point.  What we could do w/o to much trouble is to fall back to what you suggest (just run the damn callable) in case the structure is different than expected).

Your point about the Parsl semantics we are stripping away (you mention monitoring and remote staging) is also fair.  


> And so I don't have a "do it this way" answer for this bit - which maybe leads to keeping both https://github.com/Parsl/parsl/pull/2905 and https://github.com/Parsl/parsl/pull/2923 behavior, accepting that they're different and accepting the limitations both these PRs have, and working on some better Parsl core interface after some experience.

Understood, and again, thanks for the details.  Once you guys get some clarity of what way a better core interface would look like, we can of course iterate on the RCT integration.

benclifford:
>  Once you guys get some clarity of what way a better core interface would look like, we can of course iterate on the RCT integration.

The main point of intrigue for me at the moment is:

The current parsl bash implementation wants to run a thread of execution which (inside all the wrappers) happens to do a popen, then block for that popen-ed executable to complete, and then continue (out through any enclosing wrappers).

It's awkward to get MPI launching in there: in the #2905 MPI implementation, what happens is that an environment variable is set that means "if you prefix this onto your command-line, then it means you'll get run as a multi-node MPI command" - and so that's how the High Throughput Executor injects eg. `mpirun` onto the front of command lines.

There are two other MPI executor prototypes: this #2923 and Dragon (slide 4 of https://parsl-project.org/parslfest/2023/mendygral-parsldragon.pdf) - these executors don't want to interact with command lines in that way, I think, because I think both Radical and Dragon do their own "launcher" behaviour without using a commandline prefix.

So I want to hack around with @yadudoc on ideas here: for example, can we keep a wrapper that you can execute, but then the executor (whichever executor) gets to choose how the MPI command line gets executed rather than always using Popen?

andre-merzky:
> The main point of intrigue for me at the moment is:

@benclifford :  sounds like a good direction to me!  How much of this has impact on this specific PR?  My hope would be that we can close this PR before too long, i.e., without waiting for the steps you proposed.

benclifford:
@andre-merzky I don't think this PR needs to wait for any MPI specific work - that's something that I think is best driven by some kind of user experience (for all three of this PR and #2905 and the Dragon work)

But I do care a lot about all the disabled tests, because I feel like that translates pretty directly into weird user support interactions like this:

> Q: when I run tasks, i get this strange error
> A: radical pilot parsl executor cannot run parsl tasks

which makes Parsl look bad.

I think this...

> What we could do w/o to much trouble is to fall back to what you suggest (just run the damn callable) in case the structure is different than expected).

... probably addresses most or all of those test failures/user experiences, though.

andre-merzky:
understood and agreed.  Thanks @benclifford

parsl/tests/configs/local_radical.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+from parsl.config import Config
+from parsl.executors.radical import RadicalPilotExecutor
+from parsl.executors.radical import ResourceConfig
+
+# This is temporary; once everything is merged, we will use Parsl instead of
+# this fork.
+parsl_src = "pip install git+https://github.com/AymenFJA/parsl.git"
```
benclifford, line 9: I think this open issue (to do with cloud execution) probably applies here: https://github.com/Parsl/parsl/issues/1137

We moved away from "grid/cloud" style execution, but as isolated environments that aren't grid/cloud (eg this, and containers) become more prominent it's something that's interesting again.

benclifford: a rough thought I've had about this before:

parsl has a `get_version` which tries to figure out some coherent version by looking at itself: are you running from a git repo? what's your version string from parsl.version (as mutated by the parsl release script)?

Maybe that's interesting and not too hard to adapt into function which can give an "install me!" `pip` command - that would fit in this line of code but also make sense for general users (including outside of radical, for example in containers, for other parsl non-radical users?)

Do you think that would work in this case for radical? If so, I'll try to code something up in a separate PR.

benclifford: (eg you'd write:    `parsl_src = parsl.install_command_for_this_version()`)

AymenFJA:
Hey @benclifford sorry for the delay, I was overcommitted. So I just pushed changes that fixed most of the tests except:
```
test_apptimeout and  test_python_walltime
```

Which I will try to dig deep into it to fix them as well (hopefully). So far, everything works as expected. I still do not like how we dump the `bash_app` into a file, and am willing to improve it at some point or switch to another execution mechanism. This approach still gives Parsl all of the advantages of RP heterogeneous resource allocation, scheduling, and execution.

The archive_info will fail mostly, due to the large number of logs that we generate on both `client` and `agent` side.

parsl/tests/configs/local_radical.py
AymenFJA: > (eg you'd write: `parsl_src = parsl.install_command_for_this_version()`)

Well that can be helpful for RP partially. This is also related to how Parsl CI is doing the tests. Parsl CI tests do not create a virtual env and performs the installation directly on the native Python version of the VM. If we have a virtual env created before starting the tests and pointing RP to it, it will lead to:

1. RP will be able to use that env and cut short the bootstrapping time to 90% less.
2. We will not need to specify `parsl_src = "pip install git+https://github.com/AymenFJA/parsl.git"` at all.
3. This will reduce the produced logs on the agent side dramatically and eventually will help to reduce the `artifacts` to way less size and collection time.

AymenFJA:
At the moment, the tests are failing because of the `workqueue` executor.

benclifford:
> The archive_info will fail mostly, due to the large number of logs that we generate on both client and agent side.

OK, I am a fan of preserving all logs, but if it's exploding the github archiver, maybe the right thing to do is delete specifically the radical logs?

benclifford:
I see this passing in CI with the limited set of tests you've restricted and modulo the log archiving. But I went to run on my laptop and I get failures in some of the error handling tests: this is on deserialization of exceptions, I think, which is a bit fragile in Python in my experience.

```
parsl/tests/test_error_handling/test_fail.py F

_________________________________ test_simple __________________________________

self = RadicalPilotExecutor(
    4,
    30,
    'local.localhost',
    access_schema='local', 
    bulk_mode=True, 
    gpus=...ition=None, 
    project=None, 
    rpex_cfg='/home/benc/parsl/src/parsl/runinfo/011/rpex.cfg', 
    working_dir=None
)
task = <Task object, uid task.000045>, state = 'FAILED'

    def task_state_cb(self, task, state):
        """
        Update the state of Parsl Future tasks
        Based on RP task state callbacks.
        """
        if not task.uid.startswith('master'):
            parsl_task = self.future_tasks[task.uid]
    
            if state == rp.DONE:
                if task.description['mode'] in [rp.TASK_EXEC,
                                                rp.TASK_EXECUTABLE]:
                    parsl_task.set_result(int(task.exit_code))
                else:
                    # we do not support MPI function output
                    # serialization. TODO: To be fixed soon.
                    if not task.description.get('use_mpi'):
                        result = deserialize(eval(task.return_value))
                        parsl_task.set_result(result)
                    else:
                        parsl_task.set_result(task.return_value)
    
            elif state == rp.CANCELED:
                parsl_task.cancel()
    
            elif state == rp.FAILED:
                if task.description['mode'] in [rp.TASK_EXEC,
                                                rp.TASK_EXECUTABLE]:
                    # for some reason RP sometimes report a
                    # task with exit code 0 as FAILED
                    if task.exit_code == 0:
                        parsl_task.set_result(int(task.exit_code))
                    else:
                        parsl_task.set_exception(BashExitFailure(task.name,  # noqa: F405
                                                                 task.exit_code))
                else:
                    if task.exception:
                        try:
>                           s = rp.utils.deserialize_bson(task.exception)

parsl/executors/radical/executor.py:199: 

obj = 'gASVrAkAAAAAAACMEHBhcnNsLmFwcC5lcnJvcnOUjBZSZW1vdGVFeGNlcHRpb25XcmFwcGVylJOU\nKYGUfZQojAZlX3R5cGWUjAhidWlsdGluc5SMClZ...gaaFBoHGhRdWgeaCApgZR9lChoI2hUaCVoVWgnSwBoKEsAaCkpaCpLAGgrSwBoLEtA\naC1LAHViaC5LCHViaD5LCHVidWJ1YnVijAVjYXVzZZROdWIu\n'

    def deserialize_bson(obj):
    
>       return pickle.loads(codecs.decode(obj.encode(), "base64"))
E       TypeError: unpickle_exception() takes 4 positional arguments but 7 were given

../../virtualenv-3.11/lib/python3.11/site-packages/radical/pilot/utils/serializer.py:101: TypeError

During handling of the above exception, another exception occurred:

    def test_simple():
        with pytest.raises(ValueError):
            fut = always_fail()
>           fut.result()

parsl/tests/test_error_handling/test_fail.py:14: 
/usr/local/lib/python3.11/concurrent/futures/_base.py:456: in result
    return self.__get_result()
/usr/local/lib/python3.11/concurrent/futures/_base.py:401: in __get_result
    raise self._exception
parsl/dataflow/dflow.py:301: in handle_exec_update
parsl/dataflow/dflow.py:571: in _unwrap_remote_exception_wrapper
    result = future.result()
/usr/local/lib/python3.11/concurrent/futures/_base.py:449: in result
    return self.__get_result()

self = None

    def __get_result(self):
        if self._exception:
            try:
>               raise self._exception
E               parsl.serialize.errors.DeserializationError: Failed to deserialize objects. Reason: Received exception, but handling also threw an exception: unpickle_exception() takes 4 positional arguments but 7 were given

/usr/local/lib/python3.11/concurrent/futures/_base.py:401: DeserializationError
``

**Comment by benclifford** [comment]
ok, the previous error comes from incompatible tblibs on the submitting and executing workers (1.7.0 on my submit side, and 3.0.0 - that latest - on the worker side). So this comes from a worker side that is improperly built (i.e. it isn't serialization-compatible with the submit side).

That's a general complication in remote environment construction though, distinct from actually getting tasks into radical.

You mentioned here about using a single environment both sides for testing, and that's almost definitely a better solution here: it doesn't solve "how do you get the right environment remotely?" generally but it punts the problem onto the user. https://github.com/Parsl/parsl/pull/2923#discussion_r1388518781

**Comment by benclifford** [comment]
(similar serialization error, pandas 1.3.5 vs latest pandas - likely the same "punt it onto the user" solution)

**Comment by benclifford** [comment]
@AymenFJA in cd83126 and 48935f2 on this branch, I made the CI build inside a virtualenv, in the `.venv` directory - is that what you need to reuse the submit side environment?

**Comment by AymenFJA** [comment]
> @AymenFJA in [cd83126](https://github.com/Parsl/parsl/commit/cd831266d9d0470f9abb5ae3e07ab0f29275a1b6) and [48935f2](https://github.com/Parsl/parsl/commit/48935f2041d1c9e3bdbe4a28c8b4b9641d5b2120) on this branch, I made the CI build inside a virtualenv, in the `.venv` directory - is that what you need to reuse the submit side environment?

Thanks, Ben. Yes, if the Python env is activated and the env path is in the `PATH,` then RP would be able to pick it up immediately. My only concern is that it won't work with the MPI test, as we need `mpi4py` on the submit side. That put us in front of 2 choices:

1. Let RP create a Python env only for the MPI test, and it will take a while to build the mpi4py (and we need to deal with the logs generated by the env on the submitted side)
2. Or we can add `mpi4py` to the test env as a dependency while it's getting created (I prefer this one, IMHO). 

I made the changes and am happy to commit accordingly.

The call is yours. Thanks.

**Comment by benclifford** [comment]
I think it's fine to install mpi4py as a dependency in the submit-side environment, so as to stay away from problems with the environments diverging.

I looked at how mpi4py used to be installed in parsl (eg in version 1.0.0 of Parsl) and it was installed like this - so that if you installed with 'pip install parsl[extreme_scale]' to get the Extreme Scale Executor (which is now removed), you would get mpi4py:

https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/setup.py#L24

But if this is only needed for test requirements, rather than for the parsl radical pilot executor in general use, then it would be better to add it to test-requirements.txt.

**Comment by benclifford** [comment]
Note that because of the way GitHub Actions works, only the `$PATH` environment variable is carried along from the virtualenv activation to other buld steps. If you need more environt variables carried over, you should be able to see from 48935f2 how to do that.

**Comment by AymenFJA** [comment]
> I think it's fine to install mpi4py as a dependency in the submit-side environment, so as to stay away from problems with the environments diverging.
> 
> I looked at how mpi4py used to be installed in parsl (eg in version 1.0.0 of Parsl) and it was installed like this - so that if you installed with 'pip install parsl[extreme_scale]' to get the Extreme Scale Executor (which is now removed), you would get mpi4py:
> 
> https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/setup.py#L24
> 
> But if this is only needed for test requirements, rather than for the parsl radical pilot executor in general use, then it would be better to add it to test-requirements.txt.

Yes, so far this is only for the test requirements. We assume if the use wants to use MPI capabilities on the functions level then the user would: 

1. Instruct RP to create a Python env with `mpi4py`
or
2. If not RP (default behavior) will use an existing env and assume that it has `mpi4py` already installed in it.

**Comment by AymenFJA** [comment]
> I think it's fine to install mpi4py as a dependency in the submit-side environment, so as to stay away from problems with the environments diverging.
> 
> I looked at how mpi4py used to be installed in parsl (eg in version 1.0.0 of Parsl) and it was installed like this - so that if you installed with 'pip install parsl[extreme_scale]' to get the Extreme Scale Executor (which is now removed), you would get mpi4py:
> 
> https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/setup.py#L24
> 
> But if this is only needed for test requirements, rather than for the parsl radical pilot executor in general use, then it would be better to add it to test-requirements.txt.

Yes, so far this is only for the test requirements. We assume if the use wants to use MPI capabilities on the functions level then the user would: 

1. Instruct RP to create a Python env with `mpi4py`
or
2. If not RP (default behavior) will use an existing env and assumes that it has `mpi4py` already installed in it.

> I think it's fine to install mpi4py as a dependency in the submit-side environment, so as to stay away from problems with the environments diverging.
> 
> I looked at how mpi4py used to be installed in parsl (eg in version 1.0.0 of Parsl) and it was installed like this - so that if you installed with 'pip install parsl[extreme_scale]' to get the Extreme Scale Executor (which is now removed), you would get mpi4py:
> 
> https://github.com/Parsl/parsl/blob/993d6fd428ccb16d61bdf77a5577dc3c461b92ec/setup.py#L24
> 
> But if this is only needed for test requirements, rather than for the parsl radical pilot executor in general use, then it would be better to add it to test-requirements.txt.

That makes sense, I just committed to the changes let's see if something goes wrong in the venv or not.

**Comment by benclifford** [comment]
Last test run ended with this:

```
Fri, 10 Nov 2023 18:49:22 GMT
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Fri, 10 Nov 2023 18:49:22 GMT
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/capture.py:570: KeyboardInterrupt
Fri, 10 Nov 2023 18:49:22 GMT
(to show a full traceback on KeyboardInterrupt use --full-trace)
```

which I've encountered elsewhere with code sending an unexpected SIGINT, I think... (at least, that's where KeyboardInterrupt comes from)

**Comment by AymenFJA** [comment]
> Last test run ended with this:
> 
> ```
> Fri, 10 Nov 2023 18:49:22 GMT
> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
> Fri, 10 Nov 2023 18:49:22 GMT
> /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/capture.py:570: KeyboardInterrupt
> Fri, 10 Nov 2023 18:49:22 GMT
> (to show a full traceback on KeyboardInterrupt use --full-trace)
> ```
> 
> which I've encountered elsewhere with code sending an unexpected SIGINT, I think... (at least, that's where KeyboardInterrupt comes from)

Well, this is RP terminating after a failure and it shows as  `keyboard interrupt`. 

The tests failed because:
```
pmgr.0000.log:1699642161.909 : pmgr.0000            : 9395  : 139891631449856 : ERROR    : [Callback]: pilot 'pilot.0000' failed (exit)
pmgr_launching.0000.log:1699642161.905 : pmgr_launching.0000  : 9585  : 140229306472192 : ERROR    : virtenv_mode is local, no local env found
pmgr_launching.0000.log:1699642161.906 : pmgr_launching.0000  : 9585  : 140229306472192 : ERROR    : bulk launch failed
```

Which tells me that RP could not find the `.venv` that was created in the tests for one reason, because we need to activate it in each job in the CI. That also means we need to make sure that in that env the `test-requirements` and `requirements`  are installed in `.venv`. 

How should we approach this issue @benclifford in the tests from your perspective please?

**Comment by benclifford** [comment]
I'll have a play around a bit more with the venv setup in CI, but not today - you can fiddle too. Maybe doing a proper venv activate at the start of each `run:` block will work better.

For failures in the rest of parsl, we've tried to not kill the user's workflow process (that is `pytest` in the CI case) but instead flag all submitted task futures as failed, not accept any more tasks, and allow the user's workflow process to continue with the users own error handling - in the (superclass of the) High Throughput Executor: https://github.com/Parsl/parsl/blob/70b53eb7bbf28cda6288133071dd3a1b6e0d6fa3/parsl/executors/status_handling.py#L136

**Comment by benclifford** [comment]
@AymenFJA I fiddled with virtualenv initialisation a bit more, and it looks like the (non-disabled) tests are passing now

File: parsl/executors/radical/executor.py [hunk 8]
```diff
@@ -0,0 +1,583 @@
...
+        self.tmgr.submit_tasks(tds)
+
+        # prepare or use the current env for the agent/pilot side environment
+        if cfg.pilot_env_mode != 'client':
+            logger.info("creating {0} environment for the executor".format(cfg.pilot_env.name))
+            pilot.prepare_env(env_name=cfg.pilot_env.name,
+                              env_spec=cfg.pilot_env.as_dict())
+        else:
+            client_env = sys.prefix.split
+            logger.info("reusing ({0}) environment for the executor".format(client_env))
```
benclifford, line 285: This looks weird on my laptop:

```
INFO     parsl.executors.radical.executor:executor.py:285 reusing (<built-in method split of str object at 0x7fb63c646c10>) environment for the executor
```

benclifford:
I had a brief mess around with the timeout test on my laptop, including changing some of the timings in that test for larger values, and I couldn't see anything immediately wrong on the parsl side.

I don't know enough about the radical pilot logs to see if the executable for that test really is taking more than 0.2 seconds to run.

parsl/executors/radical/executor.py
AymenFJA: sorry I fixed it: this should be like this now:
```
 parsl.executors.radical.executor:executor.py:285 reusing (/home/aymen/.venv) environment for the executor
```

AymenFJA:
> I had a brief mess around with the timeout test on my laptop, including changing some of the timings in that test for larger values, and I couldn't see anything immediately wrong on the parsl side.
> 
> I don't know enough about the radical pilot logs to see if the executable for that test really is taking more than 0.2 seconds to run.

I will dig deeper into the `bash_app` timeout. The reason the tests are not working (for the timeout) is that we do not execute the `bash_apps` and wrap them as Parsl does. I am investigating an approach like this:

1. Mark the `bash_apps` as `rp.PROC` and that way we can use Parsl bash_app execution approach to execute a single-line or multi-line bash_apps.
2. I will introduce another decorator only for executables I.e. executing for example: `python example.py` or `./c_app`. This will allow to not abandon RP's capabilities on executing multi-node MPI executables if the user does not want to use a `python_app` or `bash_app`

benclifford:
From the way this PR sets a timeout on the task description based on the parsl app kwargs, I was imagining that that would cause radical pilot to enforce a timeout though... what's happening there?

AymenFJA:
> From the way this PR sets a timeout on the task description based on the parsl app kwargs, I was imagining that that would cause radical pilot to enforce a timeout though... what's happening there?

If I understand your question correctly. Yes, on the executables level, RP sets a timeout, and I quote, "Any timeout larger than 0 will result in the task process being killed after the specified number of seconds. The task will then end up in `CANCELED` state."

But it does not raise an exception, which is what Parsl tests expect for example, `AppTimeout`

benclifford:
ah ok. So what state does a cancelled task end up back in the parsl side of things?

AymenFJA:
> ah ok. So what state does a cancelled task end up back in the parsl side of things?

We set it as `future.cancel()` check this line: 

https://github.com/AymenFJA/parsl/blob/65ff8f4ac2c1afeeaf24f74dc40a2ac8df8e56db/parsl/executors/radical/executor.py#L184

We can move forward with this PR and I can open another feature PR with `timeout` workaround, what do you think?

benclifford:
yeah I think this is fine to merge now - I'm going to pull out the CI changes I made because that has potential to affect other things so I would like it to be a PR; once that is merged, I'll merge this #2923.

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,583 @@
...
+from parsl.executors.base import ParslExecutor
+from parsl.app.errors import RemoteExceptionWrapper
+from parsl.serialize import pack_apply_message, deserialize
+from parsl.serialize.errors import SerializationError, DeserializationError
+
+RPEX = 'RPEX'
+BASH = 'bash'
+PYTHON = 'python'
+
+os.environ["RADICAL_REPORT"] = "False"
```
benclifford, line 36: I'm a bit uncomfortable with screwing with the environment just because someone did "import parsl" and then didn't use radical pilot at all: this is something that affects non-radical parsl users.

parsl/executors/radical/executor.py
```diff
@@ -0,0 +1,583 @@
...
+        return self.future_tasks[rp_tid]
+
+    def shutdown(self, hub=True, targets='all', block=False):
+        """Shutdown the executor, including all RADICAL-Pilot components."""
+        logger.info("RadicalPilotExecutor shutdown")
+        self.session.close(download=True)
+
+        return True
+
+    @property
```
benclifford, line 575: Since the start of development of this executor, there have been some changes to the scaling API: I think it's safe to remove all three of `scaling_enabled`, `scale_in` and `scale_out`: scaling_enabled was removed in PR #2545 because scaling behaviour now comes from subclassing `BlockProviderExecutor` and `scale_in`, `scale_out` are now only needed for `BlockProviderExecutor` subclasses - this work was intended to remove the need for all these stubs on executors which will manage their own workers (such as the radical pilot executor)

parsl/executors/radical/executor.py
andre-merzky: That should have been addressed by us.  A [`radical.utils` PR](https://github.com/radical-cybertools/radical.utils/pull/395) now changes the default behavior.

AymenFJA:
> The most recent change I made to how radical is installed (which only installs `radical.pilot` quite late in the CI, right before radical tests) suggests that there's a problem for non-radical users if radical is not installed - I think this will break things for all non-RADICAL parsl users.
> 
> This was masked by installing `radical.pilot` right at the start of the test run (and that's part of the argument for installing things later rather than earlier)
> 
> ```
> pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/local_threads.py --random-order --durations 10
> ============================= test session starts ==============================
> platform linux -- Python 3.9.18, pytest-7.4.3, pluggy-1.3.0
> Using --random-order-bucket=module
> Using --random-order-seed=818270
> 
> rootdir: /home/runner/work/parsl/parsl/parsl/tests
> configfile: pytest.ini
> plugins: typeguard-2.13.3, cov-4.1.0, random-order-1.1.0
> collected 239 items / 1 error / 1 skipped
> 
> ==================================== ERRORS ====================================
> _______________ ERROR collecting test_radical/test_mpi_funcs.py ________________
> ImportError while importing test module '/home/runner/work/parsl/parsl/parsl/tests/test_radical/test_mpi_funcs.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> /opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/importlib/__init__.py:127: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> parsl/tests/test_radical/test_mpi_funcs.py:4: in <module>
>     from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
> parsl/tests/configs/local_radical_mpi.py:4: in <module>
>     from parsl.executors.radical import RadicalPilotExecutor
> parsl/executors/radical/__init__.py:1: in <module>
>     from parsl.executors.radical.executor import RadicalPilotExecutor
> parsl/executors/radical/executor.py:14: in <module>
>     import radical.pilot as rp
> E   ModuleNotFoundError: No module named 'radical'
> =========================== short test summary info ============================
> SKIPPED [1] parsl/tests/configs/ec2_single_node.py:30: 'public_ip' not configured in user_opts.py
> !!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
> ========================= 1 skipped, 1 error in 0.50s ==========================
> make: *** [Makefile:52: local_thread_test] Error 1
> ```

I have the same behavior on my machine now. How should we approach this? I honestly have no clue how to fix this. I assumed that isolating RP from the Parsl import (explicit import) would prevent this issue, but apparently not on the test level, at least for tests that are only for the RADICAL executor.

benclifford:
@AymenFJA I'll have a look at what's happening with imports.

AymenFJA:
> @AymenFJA I'll have a look at what's happening with imports.

Well this worked for me: 

```python
@pytest.mark.local
def test_radical_mpi(n=7):
    from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
    # rank size should be > 1 for the
    # radical runtime system to run this function in MPI env
    for i in range(2, n):
        t = test_mpi_func(msg='mpi.func.%06d' % i, sleep=1, ranks=i, comm=None)
        apps.append(t)
    assert [len(app.result()) for app in apps] == list(range(2, n))
```
instead of : 

```python
from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
@pytest.mark.local
def test_radical_mpi(n=7):
    from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
    # rank size should be > 1 for the
    # radical runtime system to run this function in MPI env
    for i in range(2, n):
        t = test_mpi_func(msg='mpi.func.%06d' % i, sleep=1, ranks=i, comm=None)
        apps.append(t)
    assert [len(app.result()) for app in apps] == list(range(2, n))
```
```
(test_rpex_final) aymen@surfacebook:~/RPEX/NEW_PARSL/parsl$ pytest parsl/tests/test_radical/test_mpi_funcs.py --config parsl/tests/configs/local_radical_mpi.py
=================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.10, pytest-7.4.3, pluggy-1.3.0
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
rootdir: /home/aymen/RPEX/NEW_PARSL/parsl/parsl/tests
configfile: pytest.ini
plugins: cov-4.1.0, random-order-1.1.0, typeguard-2.13.3
collected 1 item

parsl/tests/test_radical/test_mpi_funcs.py s                                                                                                                                         [100%]

===================================================================================== warnings summary =====================================================================================
parsl/app/python.py:49
  /home/aymen/RPEX/NEW_PARSL/parsl/parsl/app/python.py:49: PytestCollectionWarning: cannot collect 'test_mpi_func' because it is not a function.
    def __call__(self, *args, **kwargs):

test_radical/test_mpi_funcs.py::test_radical_mpi
  /home/aymen/ve/test_rpex_final/lib/python3.8/site-packages/radical/utils/atfork/stdlib_fixer.py:61: UserWarning: Import `radical` modules before `logging` to avoid the application to deadlock on `fork()`!
    warnings.warn(msg)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================= short test summary info ==================================================================================
SKIPPED [1] parsl/tests/test_radical/test_mpi_funcs.py:17: intended for local config
============================================================================= 1 skipped, 2 warnings in 22.98s ==============================================================================
```

Is this a sufficient approach in your opinion? It isolates this specific RADICAL executor test to only when the `pytest` function gets invoked.

AymenFJA:
> @AymenFJA I'll have a look at what's happening with imports.

Sorry for the noise. I tried this, it does allows other executors to pass but it actually ignores the test: 

```python
@pytest.mark.local
def test_radical_mpi(n=7):
    from parsl.tests.configs.local_radical_mpi import fresh_config as local_config
    # rank size should be > 1 for the
    # radical runtime system to run this function in MPI env
    for i in range(2, n):
        t = test_mpi_func(msg='mpi.func.%06d' % i, sleep=1, ranks=i, comm=None)
        apps.append(t)
    assert [len(app.result()) for app in apps] == list(range(2, n))
```

PR #2927, opened 2023-10-26
Make pytest fail when tasks are running after a test

benclifford:
When pressing ctrl-c in pytest, prior to this PR, it would hang as wait_for_task_completion would be called in the unwinding of the fixture stack. However, because ctrl-c was pressed, tasks wouldn't be expected to all complete.

This PR changes that to assert that all tasks are completed, rather than *waiting* for all tasks to complete.

A non-finished task now gives a test error - which is arguably better anyway because it more aggressively flushes out tests that do not perform a complete shutdown.

This means that pressing ctrl-C in a pytest leads to an assertion error; when previously it led to a hang.

One recently introduced test is fixed to conform to this new requirement.

This is part of safe-shutdown work.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -243,7 +243,7 @@ def setup_data(tmpd_cwd):
 
 
 @pytest.fixture(autouse=True, scope='function')
-def wait_for_task_completion(pytestconfig):
+def assert_no_outstanding_tasks(pytestconfig):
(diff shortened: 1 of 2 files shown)


khk-globus (approved):
Och; I'm very much in favor of this change.  Well-explained rationale as well.  Thank you for this.

PR #2930, opened 2023-10-26
Move task_launch_lock into initial TaskRecord

benclifford:
There's no reason for it to be omitted and added later, and this style is more defensive.

## Type of change

- Code maintentance/cleanup


khk-globus (approved):
Huh; that is an odd original choice.  Maybe the initial change wasn't aware of the data structure above?  Or was there initially the potential of bailing before making the Lock object?  Oh well.

benclifford:
@khk-globus more that that code is traditionally quite chaotic and you could add whatever you wanted into the task dictionary whenever you wanted... it's become more centralised in recent years

PR #2951, opened 2023-11-11
Update PR template based on my recent experiences

benclifford:
This PR adds a "mandatory" breaking changes section.

Rather than the previous assumption that bug fixes and new features would not break anything, this now assumes that all code changes might break something and encourages the pull-requester to think about this more seriously.

This PR also expands "Documentation update" to refer to anything involving human readable text.

## Type of change

- Documentation update


khk-globus (approved):
I like this; I may borrow it, in one form or another.  Thanks for the intentional thought-process behind this.

PR #2952, opened 2023-11-13
Install into virtualenv inside CI

benclifford:
This is motivated by upcoming PR #2923 which adds a RADICAL-Pilot executor: that executor is able to re-use a submit side virtualenv, but not able to re-use `--user` level installs.

This should not change any other behaviour of the CI: packages will be installed at different paths, but anything that was reliant on those paths being a specific values is suspicious.

## Type of change

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,17 +30,25 @@ jobs:
         # this is to make the workqueue binary installer happy
         sudo apt-get install -y libpython3.5
 
+    - name: setup virtual env
+      run: |
+        make virtualenv
+        source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Approving for the CI aspect, but I wonder if we might want to more heavily consider `tox` for some of these actions, both at the human CLI and also here in CI-land.  For one, I know that I sometimes install arbitrary libraries or tools in my dev sandbox while playing around or developing, but those shouldn't make it into the actual product or be available for tests.  [Tox might make that more clear](https://christophergs.com/python/2020/04/12/python-tox-why-use-it-and-tutorial/).

Issue #2415, opened 2022-08-19
Harmonize resource allocation specification between WorkQueue and RCT


WardLT:
Two of the execution providers for Parsl, WQ and RCT, support defining resource requirements for task but do so in different ways.

WQ looks for a single `parsl_resource_specification` keyword (see [docs](https://parsl.readthedocs.io/en/1.2.0/userguide/configuring.html#heterogeneous-resources)) that contains all of the components of the specifications and RCT expects a different keyword argument for each component (see [code](https://github.com/radical-cybertools/radical.pilot/blob/174d44e50be680ada9e63dfa1b55cf98021e78db/src/radical/pilot/agent/executing/parsl_rp.py#L252)).

We should combine both how the resources are passed (single kwargs vs many) and the terms used in each. 

Tagging the RCT team - @AymenFJA @mturilli
Tagging the WQ team - @btovar @tphung3

AymenFJA:
@WardLT , we used to ignore the `resource_specification` that is coming from `Parsl_dfk` as our executor is based on the `HTEX` executor that did not support resource specifications on the task-level. We should have a look again, and it should not be a problem to translate whatever resource specifications coming from Parsl on the task-level to RP. Keep in mind RP has more granular resource requirements on the task-level that need to be satisfied before launching the tasks.

benclifford:
Based on my experience adding `parsl_resource_specification` into htex as a prototype for prioritised queueing, I'm mostly of the opinion that where parsl is right now:

* resource specifications should all happen in a `parsl_resource_specification` parameter - so that it is the only thing that needs to change based on environment
* the contents of the resource specification dictionary can be executor specific: each executor is different enough that it doesn't make sense (to me) to force things into a standardised schema. (executors being different is a feature, not a bug)
* if, informally, you see resource attributes in the WQ executor which correspond to RCT attributes, then reuse the same name if it makes sense

benclifford:
(as a further example of why we shouldn't standardise on a fixed schema, a parsl executor submitting into funcX would likely need resource specifications of endpoint ID and container ID for each task, which doesn't make sense at all for Work Queue)

benclifford:
@AymenFJA I guess we should revisit this issue in light of recently merged radical pilot executor

AymenFJA:
> @AymenFJA I guess we should revisit this issue in light of recently merged radical pilot executor

[...]

PR #2962, opened 2023-11-16
Fix/issue 2415 for RadicalPilotExecutor 

AymenFJA:
# Description

This PR:

- Fix issue #2415 and adapt `parsl_resource_specifications` as the only way to set task resources.
- Fix a bug when setting an exception for MPI functions.
- Cleanup


parsl/executors/radical/executor.py
```diff
@@ -200,22 +204,25 @@ def task_state_cb(self, task, state):
...
+                                else:
+                                    raise ValueError("Unknown exception-like type received: {}".format(type(s)))
+                            except Exception as e:
+                                parsl_task.set_exception(
+                                    DeserializationError("Received exception, but handling also threw an exception: {}".format(e)))
+                        else:
+                            parsl_task.set_exception(eval(task.exception))
                     else:
-                        parsl_task.set_exception('Unknow failure for this task')
+                        parsl_task.set_exception('Task failed for unknow reason')
```
benclifford, line 225: should be `unknown` with an extra N?

parsl/executors/radical/executor.py
```diff
@@ -361,14 +363,25 @@ def unwrap(self, func, args):
...
         """
         Convert parsl function to RADICAL-Pilot rp.TaskDescription
         """
 
         task = rp.TaskDescription()
         task.name = func.__name__
         task.timeout = kwargs.get('walltime', 0.0)
+
+        if parsl_resource_specification and isinstance(parsl_resource_specification, dict):
+            logger.debug('mapping parsl resource sepcfications >> rp resource sepcfications')
```
benclifford, line 376: specifications

parsl/executors/radical/executor.py
```diff
@@ -361,14 +363,25 @@ def unwrap(self, func, args):
...
+            for key, val in parsl_resource_specification.items():
+                if key not in task.as_dict():
+                    key = PARSL_RP_RESOURCE_MAP.get(key, None)
+                    if not key:
+                        logger.warning('ignoring task {0} as it is not supported by RP'.format(key))
```
benclifford, line 381: this is ignoring a resource specification key, not a task?

AymenFJA: This will only ignore the key.

benclifford: maybe log message is better: ignoring key 'memory'...

PR #2965, opened 2023-11-20
Add logging around pipes to help debug ZMQ-related hangs

benclifford:
This kind of logging in the desc branch has revealed a lot of ZMQ-related hanging behaviour, usually when the other end of the connection is missing. This logging can help flush out those issues as users experience them.

# Changed Behaviour

More debug log load

## Type of change


khk-globus (approved):
Eww.  I don't like it, but if we're not robust, we clearly need _something_ to help on this front.

PR #2969, opened 2023-11-21
Introduced Safer FTP Alternative: Added `FTP_TLS`

fazledyn-or:
# Description
While triaging your project, our bug fixing tool generated the following message(s)-
> In file: [ftp.py](https://github.com/Parsl/parsl/blob/master/parsl/data_provider/ftp.py#L60), method: `_ftp_stage_in`, a clear-text protocol such as FTP, Telnet or SMTP is used. These protocols transfer data without any encryption, which expose applications to a large range of risks. iCR suggested that data should be transferred over only secure transport channels.

# Changed Behaviour
- Added `FTP_TLS` support for `ftp.py`.

Which means there's option to use FTP with TLS for `in_task_transfer_wrapper` and `_ftp_stage_in` methods. 

## Type of change
- New feature

## Previously Found & Fixed
- https://www.github.com/conan-io/conan/pull/15137
- https://www.github.com/jschneier/django-storages/pull/1320
- https://www.github.com/geopython/pycsw/pull/918

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/parsl/data_provider/ftp.py
+++ b/parsl/data_provider/ftp.py
@@ -50,15 +50,20 @@ class FTPInTaskStaging(Staging, RepresentationMixin):
         return in_task_transfer_wrapper(f, file, working_dir)
 
 
-def in_task_transfer_wrapper(func, file, working_dir):
+def in_task_transfer_wrapper(func, file, working_dir, secure_ftp=False):
     def wrapper(*args, **kwargs):
         import ftplib
         if working_dir:
             os.makedirs(working_dir, exist_ok=True)
 
         with open(file.local_path, 'wb') as f:
-            ftp = ftplib.FTP(file.netloc)
-            ftp.login()
+            if secure_ftp:
+                ftp = ftplib.FTP_TLS(file.netloc)
+                ftp.login()
+                ftp.prot_p()
+            else:
+                ftp = ftplib.FTP(file.netloc)
+                ftp.login()
             ftp.cwd(os.path.dirname(file.path))
             ftp.retrbinary('RETR {}'.format(file.filename), f.write)
             ftp.quit()
@@ -68,13 +73,18 @@ def in_task_transfer_wrapper(func, file, working_dir):
     return wrapper
 
 
-def _ftp_stage_in(working_dir, parent_fut=None, outputs=[], _parsl_staging_inhibit=True):
+def _ftp_stage_in(working_dir, parent_fut=None, outputs=[], _parsl_staging_inhibit=True, secure_ftp=False):
     file = outputs[0]
     if working_dir:
         os.makedirs(working_dir, exist_ok=True)
     with open(file.local_path, 'wb') as f:
-        ftp = ftplib.FTP(file.netloc)
-        ftp.login()
+        if secure_ftp:
+            ftp = ftplib.FTP_TLS(file.netloc)
+            ftp.login()
+            ftp.prot_p()
+        else:
+            ftp = ftplib.FTP(file.netloc)
+            ftp.login()
         ftp.cwd(os.path.dirname(file.path))
         ftp.retrbinary('RETR {}'.format(file.filename), f.write)
         ftp.quit()


parsl/data_provider/ftp.py
```diff
@@ -50,15 +50,20 @@ def replace_task(self, dm, executor, file, f):
         return in_task_transfer_wrapper(f, file, working_dir)
 
 
-def in_task_transfer_wrapper(func, file, working_dir):
+def in_task_transfer_wrapper(func, file, working_dir, secure_ftp=False):
```
benclifford, line 53: This new parameter, and the corresponding one added to _ftp_stage_in, are not exposed to the end user anywhere - so I think this new code path won't ever be taken?

benclifford (changes requested):

parsl/data_provider/ftp.py
fazledyn-or: I've added the FTP_TLS functionality in case users ever want to use the `secure_ftp` mode. The method is called by `replace_task`. If anyone's interested to use the secure mode, they can set the parameter value to `True`.

benclifford: Right, the method is called by `replace_task` without setting that parameter either way -- just using the default. So there isn't any user expressible workflow/configuration that will set this parameter to true.

fazledyn-or: Then, I should probably add parameter `secure_ftp` to `replace_task` method which will call the `in_task_transfer_wrapper` method using the parameter. Sounds okay?

benclifford: I think you could probably expose this as an option for users by giving FTPSeparateTaskStaging and FTPInTaskStaging init methods to make them configurable, like the GlobusStaging provider (or like other configurable Parsl components in general).

fazledyn-or: Thanks for the idea! Sounds great!

benclifford: If you store it into `self.secure_ftp` in `__init__` you'll be able to pass it into the two functions you already changed, without having to pass it via replace_task.

fazledyn-or: I was able to change it for `FTPInTaskStaging` but not sure how I'd do it for `FTPSeparateTaskStaging` and `_ftp_stage_in`. Can you please help?

benclifford: I don't have time to help with this right now - ping me in exactly 48 hours and I should have a bit more time.

fazledyn-or: @benclifford  Pinging you according to your last message.

benclifford: ok so where are you getting stuck? what code did you try so far?

fazledyn-or: I'm stuck on class "FTPSeparateTaskStaging" since it calls the ‘_ftp_stage_in_app‘ method which calls ‘_ftp_stage_in‘ method by passing it.

Not sure how I'd pass the parameter 'secure_ftp'.

benclifford: have a look at how the `working_dir` parameter gets passed in

fazledyn-or: Thanks! It should be good to go now.

benclifford:
ok this looks good. I'm waiting to find (or set up) an ftp server to test this against.

PR #2971, opened 2023-11-21
Remove unused STOP message exit path in htex worker pool

benclifford:
# Changed Behaviour

This code path seems to be unused, so should not change anything.

## Type of change

- Code maintentance/cleanup


yadudoc (approved):
👍  Looks like this code has been obsolete for a while.

PR #2973, opened 2023-11-21
Remove defaults for Manager init, forcing keywords to be specified, and perhaps revealing a bug that start_method is unused

benclifford:
# Description

This PR removes theoretically-always-unused argument defaults in the process worker pool manager, and forces arguments to be supplied as kwargs.

These defaults have two downsides:

i) users reading the code might believe the defaults will be used as end-Parsl-user defaults. However, generally there will be up to two other layers of defaults that will override the defaults removed by this PR: defaults specified in the argparser of process_worker_pool; and defaults specified in the constructor to the HighThroughputExecutor object.

These defaults, then, are misleading, as they are defaults only in the presence of a coding error such as ii) below.

ii) in the presence of a coding error adding a new parameter, where the new parameter is not wired all the way through, rather than raise an error about a missing parameter, the otherwise-unused default is used. This was a practical problem demonstrated by the threads/spawn/fork start method parameter introduced in #2433 and removed as broken in #2980.

Switching to mandatory kwargs can help with accidentally misarranged parameters in the construction of a Manager object, a similar but different coding error to ii) above.

# Changed Behaviour

none exposed to user. a class of programming error should be eliminated.

## Type of change

- Code maintentance/cleanup


benclifford:
@WardLT tagging you here because start method code came from you...

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -48,22 +48,22 @@ class Manager:
...
+                 mem_per_worker,
+                 max_workers,
+                 prefetch_capacity,
+                 uid,
+                 block_id,
+                 heartbeat_threshold,
+                 heartbeat_period,
+                 poll_period,
+                 cpu_affinity,
+                 available_accelerators: Sequence[str]):
```
rjmello, line 66: We may want to set defaults in the argparser for these to maintain consistency with the other parameters and avoid breaking any workflows that call `process_worker_pool.py` directly.

benclifford: I don't think removing defaults in this thing breaks any invocations that work prior to this PR: for example, Manager had a task port default of `"50097"` - but default that isn't accessible: the arg parser doesn't let process_worker_pool even get that far.

```
$ process_worker_pool.py 
usage: process_worker_pool.py [-h] [-d] [-a ADDRESSES] [-l LOGDIR] [-u UID] [-b BLOCK_ID] [-c CORES_PER_WORKER]
                              [-m MEM_PER_WORKER] -t TASK_PORT [--max_workers MAX_WORKERS] [-p PREFETCH_CAPACITY]
                              [--hb_period HB_PERIOD] [--hb_threshold HB_THRESHOLD]
                              [--address_probe_timeout ADDRESS_PROBE_TIMEOUT] [--poll POLL] -r RESULT_PORT
                              [--cpu-affinity {none,block,alternating,block-reverse}]
                              [--available-accelerators [AVAILABLE_ACCELERATORS ...]]
process_worker_pool.py: error: the following arguments are required: -t/--task_port, -r/--result_port
```

I'd really like to, in a different PR, remove more defaults from the arg parser. 

At least one default is broken there:

```
$ process_worker_pool.py -t 10000 -r 10001
  File "/home/benc/parsl/virtualenv-3.11/bin/process_worker_pool.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/process_worker_pool.py", line 765, in <module>
    logger.info("Accelerators: {}".format(" ".join(args.available_accelerators)))
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: can only join an iterable
```

and I think the cpu_affinity default of `None` will break things if actually used (although I haven't tested that).

Some of the defaults this PR removes from Manager don't make sense as defaults: for example, default port numbers for task and result connections, which need to be coordinated with the corresponding interchange rather than defaulted.

rjmello: That works for me. :rocket:

rjmello (approved):

PR #2974, opened 2023-11-21
Standardize HTEX worker process initiation

### LINKED ISSUE #2972: HTEX worker watchdog is not working
**Describe the bug**
The HTEX `Manager.worker_watchdog` method creates a worker process object when a worker fails, but [does not start](https://github.com/Parsl/parsl/blob/741293a76c2ad114b48bcc927701258b7ba7f355/parsl/executors/high_throughput/process_worker_pool.py#L401-L411) the process.

**To Reproduce**
1. Run a function that kills the worker
```python
@python_app
def kill_worker():
    import sys
    sys.exit(2)
```
2. Observe that no new worker process starts.

**Expected behavior**
The worker watchdog should recognize that the worker is dead and start a new process.

**Environment**
 - Parsl version: 2023.11.20

rjmello:
## Description

The HTEX worker watchdog was missing a worker argument and not starting any processes. This PR standardizes the initiation of worker processes to ensure consistency and fix the watchdog.

I also expanded the worker failure tests to cover restarts by the worker watchdog. 

Fixes #2972

## Type of change

- Bug fix
- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -398,15 +398,7 @@ class Manager:
                     except KeyError:
                         logger.info("Worker {} was not busy when it died".format(worker_id))
 
-                    p = self.mpProcess(target=worker, args=(worker_id,
-                                                            self.uid,
-                                                            self.worker_count,
-                                                            self.pending_task_queue,
-                                                            self.pending_result_queue,
-                                                            self.ready_worker_queue,
-                                                            self._tasks_in_progress,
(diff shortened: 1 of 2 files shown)


benclifford:
There's a potential race condition when launching new workers using the `fork` start method, because the manager starts threads and then makes forks afterwards. Right now, that `fork` method (due to another bug) is not accessible to users, but if/when that gets fixed, this race condition has the chance to appear.

Fortunately, `fork` is set to go away, and apparently none of our users have even noticed they aren't using it.

See #2343 for more context on fork vs threads.

PR #2975, opened 2023-11-22
Use usual runinfo for ParslExecutor tests

benclifford:
I don't think there is any need for this test to put the run directory somewhere different, vs the other tests which all use the default runinfo/ directory.

Putting the logs in a different location means they are not available in CI runinfo/ archives.

# Changed Behaviour

This should not change anything for end users because it is a change to the test suite.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/tests/sites/test_concurrent.py
+++ b/parsl/tests/sites/test_concurrent.py
@@ -9,7 +9,7 @@ def f(x):
     return x + 1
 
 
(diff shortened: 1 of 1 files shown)


WardLT (approved):
LGTM. I like the tmpdir so pytest deletes my old runinfo files for me, but see the advantage of keeping the tests simpler.

PR #2983, opened 2023-11-30
Use spawn for process worker pool

rjmello:
## Description

To avoid issues with forked processes and threads (see #2343), we will use spawn calls to start HTEX workers.

- Created spawn and fork multiprocessing context objects that we can use throughout the codebase to increase clarity on how we start processes.

- Migrated away from using the `MacSafeQueue` in the process worker pool since it was only needed to get the sizes of two queues for a single debug log statement.

- Replaced `ready_worker_queue` with `ready_worker_count`. The former was a multiprocessing `Queue` that we only used to track the number of workers that are ready to receive tasks. The latter is a multiprocessing `Value`, which is a more appropriate data structure.

- Created an HTEX manager thread to transfer monitoring messages from a managed queue, which we can utilize across processes started in fork and spawn contexts, to the results queue. We separate the queues so that the results queue does not rely on a manager process, which causes slower queue operations.

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -10,7 +10,7 @@ import pickle
 import time
 import queue
 import uuid
-from typing import Sequence, Optional
+from typing import Sequence, Optional, Union
 
 import zmq
 import math
@@ -18,7 +18,7 @@ import json
 import psutil
 import multiprocessing
 from multiprocessing.managers import DictProxy
-from multiprocessing.sharedctypes import SynchronizedBase
+from multiprocessing.sharedctypes import Synchronized
 
 from parsl.process_loggers import wrap_with_logs
 
@@ -174,9 +174,9 @@ class Manager:
         if mem_per_worker and mem_per_worker > 0:
             mem_slots = math.floor(available_mem_on_node / mem_per_worker)
 
-        self.worker_count = min(max_workers,
-                                mem_slots,
-                                math.floor(cores_on_node / cores_per_worker))
+        self.worker_count: int = min(max_workers,
+                                     mem_slots,
+                                     math.floor(cores_on_node / cores_per_worker))
 
         self.pending_task_queue = SpawnContext.Queue()
         self.pending_result_queue = SpawnContext.Queue()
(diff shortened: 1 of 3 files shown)


benclifford:
ok, I think this looks good. I'd like some other person (eg @yadudoc or @khk-globus) to have a look at this too though.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -381,14 +386,36 @@ def worker_watchdog(self, kill_event: threading.Event):
...
+        We separate the queues so that the result queue does not rely on a manager
+        process, which adds overhead that causes slower queue operations but enables
+        use across processes started in fork and spawn contexts.
+
+        We transfer the messages to the result queue to reuse the ZMQ connection between
+        the manager and the interchange.
+        """
+        logger.debug("Starting monitoring messages thread")
+
+        while not kill_event.is_set():
```
benclifford, line 402: I think this loop isn't guaranteed to keep looping if there are no monitoring messages after the kill event is set (to cause monitoring_queue.get() to exit) and so setting kill_event won't always cause a termination? (eg. if there is no monitoring turned on).

So then this thread is reliant on something external killing the process worker pool, not the kill event?

```
$ pytest parsl/tests/ --config parsl/tests/configs/htex_local.py
[...]
$ grep 'Monitoring-Hand' runinfo/005/ -r
runinfo/005/htex_local/block-0/41bfdd691502/manager.log:2023-12-06 15:10:00.662 parsl:400 5911 Monitoring-Handler [DEBUG]  Starting monitoring messages thread
runinfo/005/htex_local/block-0/41bfdd691502/manager.log:2023-12-06 15:10:00.662 parsl:403 5911 Monitoring-Handler [DEBUG]  Starting monitor_queue.get
```

rjmello: Good catch. I've addressed this by adding a timeout to the `monitoring_queue.get()` call.

yadudoc (approved):
Nice work! Just ran the tests on my Mac and everything passes.

There are two pieces that I looked closely at: 
`ready_worker_count`: There's locking only on the writer side so this looks pretty safe. If the worker fails-hangs on queue.get, then this could get whacked, but at that point the workers are mostly shutting down so this shouldn't be an issue.
 `handle_monitoring_messages`: I see that the hang possibility if there are no monitoring messages is handled with a timeout.

PR #2984, opened 2023-12-04
Fix blocking Work Queue executor send/retrieve loop

### LINKED ISSUE #2801: investigate task-end-to-task-end delays in Work Queue
**Describe the bug**
Performance is poor in the work queue executor when running sequential tasks: the time from one task ending to the next task starting is around 1 second. I'm not sure if this is inside the parsl wq executor code, or inside core wq.

**To Reproduce**
submit two tasks, one which depends on the other. observe a 1 second delay between task 1 completing on the worker and task 2 starting on the worker.

**Expected behavior**
This could be much faster.

**Environment**
my laptop

cc @khk-globus

colinthomas-z80:
# Description
The Work Queue executor was attempting to wait for tasks when there were none in the queue, causing the timeout period to be observed in the task throughput. 

Now the executor re-checks whether the queue is empty before attempting to wait on another task retrieval.

# Fixes

#2801 

- Bug fix

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -959,43 +959,42 @@ def _work_queue_submit_wait(*,
 
         # If the queue is not empty wait on the WorkQueue queue for a task
         task_found = True
-        if not q.empty():
-            while task_found and not should_stop.value:
-                # Obtain the task from the queue
-                t = q.wait(1)
-                if t is None:
-                    task_found = False
-                    continue
-                # When a task is found:
-                executor_task_id = t.tag
-                logger.debug("Completed Work Queue task {}, executor task {}".format(t.id, t.tag))
-                result_file = result_file_of_task_id.pop(t.tag)
+        while not q.empty() and task_found and not should_stop.value:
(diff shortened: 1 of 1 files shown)


colinthomas-z80:
@benclifford Is there anything further you'd like from this? If all is good I can make an equivalent taskvine PR or add on here

benclifford:
@colinthomas-z80 hi - I'm not waiting on anything from you but I don't think any more PR reviews will happen in 2023.

benclifford:
This drops the runtime of the pytest suite on my laptop from 7 minutes to 10 seconds.

I also used `git diff -r origin/master --color-moved --color-moved-ws=allow-indentation-change` to check that all the indentation changes didn't change anything else in that main block.

benclifford (approved):

benclifford:
I think this change means that the Work Queue Executor submit/wait process will now use 100% CPU because there is nothing stopping it from hard looping checking both submit and result queues continuously. That hard loop comes from the loop trying to do two things and maybe would be fixed by splitting this loop into two loops that can then each perform their own long wait without interfering with the other half.

PR #2986, opened 2023-12-08
Add slurmprovider options

tylern4:
# Description

Adds optional parameters to `SlurmProvider` for choosing the `queue` and `constraint` slurm options. This makes the `SlurmProvider` easier to work on NERSC systems which uses these options for job submissions. Queue is used to choose the jobs priority ([regular, debug, realtime, etc.](https://docs.nersc.gov/jobs/policy/#qos-limits-and-charges)) and constraint to choose between cpu and gpu nodes on Perlmutter.

# Changed Behaviour

Current users would have had to place the options in `scheduler_options` which is unchanged. Users now have the option to adopt the new parameters in their code, or continue to use `scheduler_options`.

## Type of change

- New feature

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -47,6 +47,10 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
     account : str
         Slurm account to which to charge resources used by the job. If unspecified or ``None``, the job will use the
         user's default account.
+    queue : str
+        Slurm queue to place job in. If unspecified or ``None``, no queue slurm directive will be specified.
+    constraint : str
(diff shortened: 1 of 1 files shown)


parsl/providers/slurm/slurm.py
```diff
@@ -133,6 +139,11 @@ def __init__(self,
             self.scheduler_options += "#SBATCH --partition={}\n".format(partition)
         if account:
             self.scheduler_options += "#SBATCH --account={}\n".format(account)
+        if queue:
```
benclifford, line 142: Maybe this parameter be called `qos` so to match the SLURM vocabulary? (eg. compare to sites that use `partition` as their queue abstraction, not `qos`: https://help.jasmin.ac.uk/article/4890-how-to-submit-a-job-to-slurm)

PR #2989, opened 2023-12-09
Add more info to DependencyError for non-task Futures

benclifford:
In some situations, a dependency in a dependency error does not have a task ID (becuase it is not an AppFuture) (case 1), or has a task ID from a different DFK (case 2)

Previously:

Case 1: the dependency error listed a task ID as `None`.

Case 2: the dependency error listed a task ID as if it was that task in the same DFK as the dependent task.

In both those cases now, the error message renders those "foreign" Futures using repr.

This PR also tidies up list formatting to remove []-brackets around the list of dependencies.

Now:

Dependency failure for task 0 with failed dependencies from <Future at 0x7f8ca335ef90 state=finished raised ValueError>

Previously:

Dependency failure for task 0 with failed dependencies from [None]

# Changed Behaviour

The task ID entry of `DependencyError.dependency_exceptions_tids` attributed was previously an `int` or `None`, even though it is annotated as a str. It is now a str. Anyone relying on that programmatically will have to change their code.

## Type of change

### DIFF
--- a/parsl/dataflow/errors.py
+++ b/parsl/dataflow/errors.py
@@ -36,7 +36,9 @@ class DependencyError(DataFlowException):
        in a dependency.
 
     Args:
-         - dependent_exceptions_tids: List of dependency task IDs which failed
+         - dependent_exceptions_tids: List of exceptions and identifiers for
+           dependencies which failed. The identifier might be a task ID or
+           the repr of a non-DFK Future.
          - task_id: Task ID of the task that failed because of the dependency error
     """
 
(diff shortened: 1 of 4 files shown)


khk-globus (approved):

parsl/dataflow/errors.py
```diff
@@ -45,8 +47,9 @@ def __init__(self, dependent_exceptions_tids: Sequence[Tuple[Exception, str]], t
         self.task_id = task_id
 
     def __str__(self) -> str:
-        dep_tids = [tid for (exception, tid) in self.dependent_exceptions_tids]
-        return "Dependency failure for task {} with failed dependencies from tasks {}".format(self.task_id, dep_tids)
+        dep_ids = [tid for (exception, tid) in self.dependent_exceptions_tids]
+        deps = ", ".join(dep_ids)
+        return f"Dependency failure for task {self.task_id} with failed dependencies from {deps}"
```
khk-globus, line 52: - Aren't using `exception`, so stylistically consider the underscore prefix
- Not that "it matters," but that could be a generator instead; no need for the interim data structure:
    ```python
    dep_ids = (tid for _exc, tid in self.dependent_exception_tids)
    deps = ", ".join()
    ```
    or just:
    ```python
    deps = ", ".join(tid for _exc, tid in self.dependent_exception_tids)
    ```

khk-globus (approved):

Issue #2072, opened 2021-05-17
"Affinity does not work if there are more workers than cores" seems to cause hang


benclifford:
**Describe the bug**
I'm running parsl `--config local` tests in a docker container with a single CPU allocated to that container via `docker run --cpuset-cpus=0`

The `parsl/tests/sites/test_affinity.py` test hangs. When I dig into the htex worker logs I find this error. This error should be properly propagated to the submit side in some way rather than resulting in a hang.

```
2021-05-17 13:04:35.266 worker_log:505 Rank:1 [INFO]  Worker 1 started
2021-05-17 13:04:35.266 worker_log:507 Rank:1 [DEBUG]  Debug logging enabled
2021-05-17 13:04:35.266 worker_log:31 Rank:1 [ERROR]  Exceptional ending for worker on thread MainThread
Traceback (most recent call last):
  File "/home/benc/parsl/src/parsl/parsl/process_loggers.py", line 27, in wrapped
    r = func(*args, **kwargs)
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/process_worker_pool.py", line 514, in worker
    assert cores_per_worker > 0, "Affinity does not work if there are more workers than cores"
AssertionError: Affinity does not work if there are more workers than cores
```

**To Reproduce**
Run the above test in a docker container with --cpuset-cpus=0 set, on a machine that has more than 1 core.

**Expected behavior**
The error should propagate to the submit side and the test should fail (more generally - the workflow user should be presented with this error rather than a hang)

**Environment**
python 3.9.2, parsl master bf83c06af78ac0e4243fa2b6686d3b99110cf01c

benclifford:
@yadudoc and I had some discussion offline about the block error handling for the cases where the process worker pool starts, but then fails deeper into startup.

I think our talk was more in the context of misconfigured networking (eg. `process_worker_pool` start but cannot make a network connection), but I realise now this #2072 misconfiguration is also exactly that failure case.

PR #2990, opened 2023-12-12
New method for assigning cpu affinity

cms21:
# Description

This provides a new method for setting cpu affinity via a method similar to setting thread binding in mpiexec calls.

In the config object, users can now pass a string of this type:
`cpu_affinity="list:<worker1_threads>:<worker2_threads>:<worker3_threads>"`

Each worker's threads can be specified as a comma separated list or a hyphenated range:
`thread1,thread2,thread3`
or
`thread_start-thread_end`

An example for Sunspot/Aurora:
`cpu_affinity="list:0-7,104-111:8-15,112-119:16-23,120-127:24-31,128-135:32-39,136-143:40-47,144-151:52-59,156-163:60-67,164-171:68-75,172-179:76-83,180-187:84-91,188-195:92-99,196-203"`

This example assigns 16 threads each to 12 workers. Note that there are threads that are skipped; this is because the total number of hardware threads is not evenly divisible by the number of GPUs.

This will give a user more flexibility in setting cpu affinity and also allows them to leave idle certain threads if they choose (currently required by all cpu-affinity strategies in parsl).

If the number of thread "ranks" (colon separated thread lists/ranges) doesn't match the total number of workers an exception will be raised.

# Changed Behaviour

No current behavior is changed

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -725,7 +725,14 @@ if __name__ == "__main__":
                         help="Poll period used in milliseconds")
     parser.add_argument("-r", "--result_port", required=True,
                         help="REQUIRED: Result port for posting results to the interchange")
-    parser.add_argument("--cpu-affinity", type=str, choices=["none", "block", "alternating", "block-reverse"],
+
+    def strategyorlist(s):
+        if s in ["none", "block", "alternating", "block-reverse"]:
+            return s
+        if s[0:4] == "list":
+            return s
+    
+    parser.add_argument("--cpu-affinity", type=strategyorlist, #choices=["none", "block", "alternating", "block-reverse"],
                         help="Whether/how workers should control CPU affinity.")
     parser.add_argument("--available-accelerators", type=str, nargs="*",
                         help="Names of available accelerators")
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -576,7 +576,7 @@ def worker(
     # If desired, set process affinity
     if cpu_affinity != "none":
         # Count the number of cores per worker
-        avail_cores = sorted(os.sched_getaffinity(0))  # Get the available processors
+        avail_cores = sorted(os.sched_getaffinity(0))  # Get the available threads
         cores_per_worker = len(avail_cores) // pool_size
         assert cores_per_worker > 0, "Affinity does not work if there are more workers than cores"
 
@@ -588,6 +588,23 @@ def worker(
             my_cores = avail_cores[cores_per_worker * cpu_worker_id:cores_per_worker * (cpu_worker_id + 1)]
         elif cpu_affinity == "alternating":
             my_cores = avail_cores[worker_id::pool_size]
+        elif cpu_affinity[0:4] == "list":
+            thread_ranks = cpu_affinity.split(":")[1:]
+            if len(thread_ranks) != pool_size:
+                raise ValueError("Affinity list {} has wrong number of thread ranks".format(cpu_affinity))
+            threads = thread_ranks[worker_id]
+            thread_list = threads.split(",")
+            my_cores = []
+            for tl in thread_list:
+                thread_range = tl.split("-")
+                if len(thread_range) == 1:
+                    my_cores.append(int(thread_range[0]))
+                elif len(thread_range) == 2:
+                    start_thread = int(thread_range[0])
+                    end_thread = int(thread_range[1]) + 1
+                    my_cores += list(range(start_thread,end_thread))
+                else:
+                    raise ValueError("Affinity list formatting is not expected {}".format(cpu_affinity))
         else:
             raise ValueError("Affinity strategy {} is not supported".format(cpu_affinity))
 
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -732,7 +732,7 @@ if __name__ == "__main__":
         if s[0:4] == "list":
             return s
     
-    parser.add_argument("--cpu-affinity", type=strategyorlist, #choices=["none", "block", "alternating", "block-reverse"],
+    parser.add_argument("--cpu-affinity", type=strategyorlist,
                         help="Whether/how workers should control CPU affinity.")
     parser.add_argument("--available-accelerators", type=str, nargs="*",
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -576,7 +576,7 @@ def worker(
     # If desired, set process affinity
     if cpu_affinity != "none":
         # Count the number of cores per worker
-        avail_cores = sorted(os.sched_getaffinity(0))  # Get the available processors
+        avail_cores = sorted(os.sched_getaffinity(0))  # Get the available threads
```
benclifford, line 579: I feel like this vocabulary should be made more consistent: cores/threads/processors are used apparently interchangeably(?) here, where one word might do better: mostly this code uses `cores`

benclifford:
This looks good. Given the fragility of htex process worker pool changes in the past (for example, spawn method and worker restarts both didn't work at all), I'd like to figure out how to test this more automatically - I guess that would need at least a two core system so tests could play around with pinning?

benclifford:
Other random bits:

CI is broken because of code formatting (`flake8 parsl/`).

The description of the PR is pretty good documentation and it would be good to paste it pretty much as-is into the main documentation somewhere - for example the "Multi-Threaded Applications" section of docs/userguide/configuring.rst, or the cpu_affinity docstring in parsl/executors/high_throughput/executor.py

parsl/executors/high_throughput/process_worker_pool.py
cms21: This call gets the available hardware threads on the node, not the available physical cores, despite the variable name.  It's unfortunate that `cores` is used as it is in the code because for machines that support multiple hardware threads per core (e.g. many hpc systems) this makes the code confusing.  I don't work with this code often enough to remember what Parsl is doing here always and every time I look at it I have to remind myself what's going on.  Some kind of comment explaining this would be helpful here.

benclifford:
This is failing with type checking error:
```
parsl/executors/high_throughput/process_worker_pool.py:595: error: Value of type "str | bool" is not indexable  [index]
```
which is saying, basically, what if someone passed in a bool here, not a string...

because part of the existing code says you are allowed to do that, even though in practice that doesn't happen and I think is probably broken.

I'll make a PR to remove that code, and then this type error should go away.

cms21:
Okay, thanks.  I also realized that I unintentionally did away with the argparse check, so I tried to fix that.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -590,6 +590,23 @@ def worker(
             my_cores = avail_cores[cores_per_worker * cpu_worker_id:cores_per_worker * (cpu_worker_id + 1)]
         elif cpu_affinity == "alternating":
             my_cores = avail_cores[worker_id::pool_size]
+        elif cpu_affinity[0:4] == "list":
+            thread_ranks = cpu_affinity.split(":")[1:]
+            if len(thread_ranks) != pool_size:
+                raise ValueError("Affinity list {} has wrong number of thread ranks".format(cpu_affinity))
```
benclifford, line 596: This value error (and I guess the other value errors in the affinity code around here) doesn't result in a very good user failure: the worker dies and the manager automatically restarts it (because of worker watchdogs, #2974), forever.

For a workflow user that looks like a hang.

This isn't something that is introduced by this PR, because there were already value errors in this code block that give the same behaviour.

This is basically issue #2072 I think, but I'm crossreferencing it here because I just encountered it, and I think it's more likely to happen with users giving much more complicated inputs here.

benclifford:
@cms21 I added a rough test here, that checks that when htex is launched with a single worker / single core affinity, then the worker sees that same single core affinity.

benclifford:
@cms21 or @WardLT can you have a look at the test I added and check it makes sense. Aside from that, I'm fine with merging the rest of this PR.

cms21:
Is there a reason that the test is just for one worker/thread?

cms21:
Hey, is there anything I can help with so that we can wrap up this PR?

parsl/executors/high_throughput/process_worker_pool.py
WardLT: Wow, that's a gnarly bug. Maybe out-of-scope for this PR?

parsl/tests/test_htex/test_cpu_affinity_explicit.py
```diff
@@ -0,0 +1,52 @@
...
+    affinity = f"list:{single_core}"
+
+    logger.debug(f"Will test with affinity for one worker, one core: {affinity}")
+
+    config = fresh_config()
+    config.executors[0].cpu_affinity = affinity
+    config.executors[0].max_workers = 1
+
+    logger.debug(f"config: {config}")
+    # TODO: is there a `with` style for this, to properly deal with exceptions?
```
WardLT, line 40: Not really, though that would be nice. I usually have my `parsl.load` in a pytest fixture. You could also use the [ParslPoolExecutor](https://parsl.readthedocs.io/en/stable/userguide/workflow.html#parallel-workflows-with-loops)'s, which has a similar effect.

That said, I have no objections to how you do it here.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -729,7 +746,17 @@ def start_file_logger(filename, rank, name='parsl', level=logging.DEBUG, format_
...
-    parser.add_argument("--cpu-affinity", type=str, choices=["none", "block", "alternating", "block-reverse"],
+
+    def strategyorlist(s: str):
+        allowed_strategies = ["none", "block", "alternating", "block-reverse"]
+        if s in allowed_strategies:
+            return s
+        elif s[0:4] == "list":
+            return s
+        else:
+            raise argparse.ArgumentTypeError("cpu-affinity must be one of {} or a list format".format(allowed_strategies))
```
WardLT, line 757: Awesome! I did not know this way of writing type checking!

WardLT (approved):
I think this is ready to merge. 

It's a good feature and the documentation for existing methods for CPU affinity is even better.

PS. #2072 is going to keep me up at night now, but I don't think we should fix it here.

parsl/executors/high_throughput/process_worker_pool.py
benclifford: yes probably out of scope for this PR - but users will encounter it more so you should be aware of it more than usual as you encourage users to use this feature.

PR #2991, opened 2023-12-12
Tidy types around CPU affinity

benclifford:
Prior to this PR:

In practice, the --cpu-affinity parameter was always specified, and when a user does not specify any value in the HighThroughputExecutor, the default value of `'none'` will be passed in to this parameter. So in practice, throughout process_worker_pool, all cpu_affinity variables will be a string, 'none' by default, or as specified by users.

The commandline args code allows --cpu-affinity to be omitted. This results in cpu-affinity having the value `None`. This is not handled properly by the worker code, which assumes that one of the explicit string values will be supplied, and will (I think - I haven't tested) raise:

> raise ValueError("Affinity strategy {} is not supported".format(cpu_affinity))

in this case.

The type annotations for the worker allow cpu_affinity to be either a bool or a string, and the default value of cpu_affinity in Manager.__init__ (which is never used, because it is always overridden by the commandline argument value, even when no commandline argument is explicitly supplied) is a bool, False. This was probably intended to mean "no cpu affinity", which is what is actually implemented by the string `'none'`.

Passing a bool value True or False here also looks like it is not implemented in the worker and will (I think - I haven't tested) also raise:

> raise ValueError("Affinity strategy {} is not supported".format(cpu_affinity))

This PR gets rid of the above broken values:

`None` - by requiring --cpu-affinity to be always supplied. Due to the above described handling of None, that commandline argument is already required, but that requirement is not caught at the argparser level.

bools - by tightening the type annotation of the worker, and removing one of the unused default values on Manager. This parameter was already required, due to the above described handling of bools, but that requirement is not caught at the Python invocation level.

This should leave this value only accepting strings.

# Changed Behaviour

I think this codepath only affects codepaths that result in errors.

# Fixes

Type error that @cms21 encountered implementing #2990

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -10,7 +10,7 @@ import pickle
 import time
 import queue
 import uuid
-from typing import Sequence, Optional, Union
+from typing import Sequence, Optional
 
 import zmq
 import math
@@ -62,7 +62,7 @@ class Manager:
                  heartbeat_threshold=120,
                  heartbeat_period=30,
                  poll_period=10,
-                 cpu_affinity=False,
+                 cpu_affinity=False, # BAD - should go away in PR 2973.
                  available_accelerators: Sequence[str] = ()):
         """
         Parameters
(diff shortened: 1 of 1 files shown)


benclifford:
draft until after #2973

rjmello (approved):

PR #2993, opened 2023-12-13
Update `address_by_hostname` to return IP address

yadudoc:
# Description

Previously address_by_hostname returned the human-readable hostname returned by platform.node(). A common usage: `HTEX(address=address_by_hostname())` caused issues following PR#2828 which required address to be a valid IP address. This PR updates address_by_hostname to return an IP address instead. This could potentially fail, because the IP address resolved on a host need not be the reachable from the workers. 

Tested on Perlmutter.

# Changed Behaviour

User workflows that use `HighThroughputExecutor(address=address_by_hostname())` will now work on machines
where hostname is set. 

## Type of change

- Bug fix


yadudoc:
@benclifford Probably worth testing on a couple of machines for DESC.

benclifford:
I'm somewhat uncomfortable with this PR changing the behaviour of other components (monitoring, work queue) in order to work around a change in htex (PR #2828) - so I put off merging this until after the long Christmas vacation. I'll merge it now people should have more time to deal with unanticipated breakage. (eg. change in resolution location could change the resolved IP address because hostname resolution is not globally consistent)

benclifford (approved):

Issue #2994, opened 2023-12-13
Cannot Instantiate SlurmProvider in parsl-2023.12.11


Oddant1:
**Describe the bug**
When I attempt to instantiate a SlurmProvider in parsl-2023.12.11 I get the following error:
`AttributeError: class SlurmProvider uses qos in the constructor, but does not define it as an attribute`

**To Reproduce**
1. Install parsl-2023.12.11
2. Attempt to instantiate SlurmProvider
3. See error

**Expected behavior**
The provider should be instantiated

**Environment**
 - OS: Seen on Ubuntu and MacOS github workers and Linux Mint 21.2
 - Python version: 3.8 and 3.11
 - Parsl version: 2023.12.11

**Distributed Environment**
- I'm not actually running a script at all just trying to instantiate the provider/a config with the provider

benclifford:
Probably introduced by #2986 - maybe @tylern4 can look at this.

Basically for RepresentationMixin, each constructor argument needs to be stored on `self` so that it can be used to create `repr` output later.

This gets checked at instantiation, not statically, so it's more likely to happen for classes we don't test in CI.

Oddant1:
Awesome, thank you @benclifford and @tylern4.

PR #2995, opened 2023-12-14
Add a regression test for broken SlurmProvider repr(), issue #2994

benclifford:
This test checks that repr runs ok for SlurmProvider. That's a method that one might expect would always work, but RepresentationMixin has been a source of exceptions in the past (although with #2996, it would hopefully be less aggressive).

It might be useful to make this sort of repr test for more classes that are not otherwise tested.

This test does not need a slurm installation, as it does not test any interactions with slurm itself.


khk-globus (approved):
Love it.

PR #2996, opened 2023-12-14
Make RepresentationMixin validation less aggressive

benclifford:
# Description

Prior to this PR:
RepresentationMixin fails when `repr()` is called if an implementing class does not store all of the parameters used in its __init__ method.

This leads to a fairly common user-exposed error not caught in CI in Parsl components which are not tested in CI: primarily the provider classes. This immediate work, for example, is motivated by issue #2994 introduced in PR #2986 which is in release `2023.12.11`.

This PR changes the behaviour of repr so that it will by default use the string `<unrecorded>` instead of raising an error at runtime in this situation.

For testing/validation purposes, the validation code is re-enabled when running pytests, so that other pytests (such as PR #2995) can be implemented to fail rather than silently omit parameters.

# Changed Behaviour

When a class in future incorrectly implements the RepresentationMixin protocol, users will now be presented with that parameter listed as undefined, rather than getting an exception. For most users, that is a better way of doing things.

For developers, RepresentationMixin will now only detect omitted parameters when that behaviour is enabled, for example inside the test suite. That is likely to lead to an *increase* in this sort of coding error (traded off against this sort of error being less breaking).

## Type of change

- New feature


parsl/tests/test_utils/test_representation_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+    p1 = "parameter 1"
+    p2 = "the second parameter"
+
+    try:
+        # repr should not raise an exception
+        r = repr(BadRepr(p1, p2))
+        # parameter 2 should be found in the representation, but not
+        # parameter 1
+        assert p1 not in r
+        assert p2 in r
```
khk-globus, line 58: May I suggest subclassing instead of hoping/praying that the test puts things back the way they should be?  Consider:

```python
class _TestRepresentationMixin(RepresentationMixin):
    _validate_repr = False

class BadRepr(_TestRepresentationMixin):
    def __init__(self, x, y):
        self.y = y

def test_repr_bad_unvalidated():
    # now no need for the finally.
    p1 = "parameter 1"
    p2 = "the second parameter"

    # repr should not raise an exception
    r = repr(BadRepr(p1, p2))
    assert p2 in r
    assert p1 not in r
```

benclifford: @khk-globus see latest commit a71b812

khk-globus (approved):
Woot!

PR #2997, opened 2023-12-14
Fix missing qos and constraint attributes that broke SlurmProvider in #2986

### LINKED ISSUE #2994: Cannot Instantiate SlurmProvider in parsl-2023.12.11

tylern4:
# Description

Adds new options for `SlurmProvider` as attributes to fix #2994

# Fixes

Fixes # 2994

## Type of change

- Bug fix

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -132,6 +132,8 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         self.exclusive = exclusive
         self.move_files = move_files
         self.account = account
+        self.qos = qos
+        self.constraint = constraint
         self.scheduler_options = scheduler_options + '\n'
         if exclusive:
             self.scheduler_options += "#SBATCH --exclusive\n"


benclifford:
I'll test this out on perlmutter now.

benclifford:
OK. I tried this out in a fresh install on perlmutter, in two configurations:

first a minor variant of the configuration given on https://docs.nersc.gov/jobs/workflow/parsl/ which passes `debug` as the first positional argument, which ends up going into the `partition` argument.

Secondly, I deleted that positional argument, and the scheduler options parameter (which set `-c cpu`), and instead set new parameters from PR #2986:  qos=`debug` and constraint=`cpu`.

Both of these tests ran OK on perlmutter.

Also, at the top of parsl log, I checked that SlurmProvider renders in the initial configuration log message.

PR #2998, opened 2023-12-19
Use consistent "job" terminology in SubmitError, not "task"

benclifford:
In Parsl generally, the thing launched by a provider is a job, not a task; and a task is the sort of thing that is scheduled by the DataFlowKernel, distinct from the provider mechanism.

# Changed Behaviour

Changed error message text.

External components which create a SubmitError using named arguments would need to change. I am not aware of any such code.

## Type of change

- Code maintentance/cleanup


rjmello (approved):

parsl/providers/errors.py
```diff
@@ -46,18 +46,18 @@ def __str__(self):
...
 
 class SubmitException(ExecutionProviderException):
-    '''Raised by the submit() method of a provider if there is an error in launching a task.
+    '''Raised by the submit() method of a provider if there is an error in launching a job.
     '''
 
-    def __init__(self, task_name, message, stdout=None, stderr=None):
-        self.task_name = task_name
+    def __init__(self, job_name, message, stdout=None, stderr=None):
+        self.job_name = job_name
```
khk-globus, line 53: This is effectively an API change.  Any sense in (for a time) adding a property for `.task_name`?  Something like:
```python
@property
def task_name(self) -> str:  # *is* it a string?  No typing in __init__ signature ...
    log.warning("[...deprecated; use .job_name instead...]")  # remove after now() + 6mo
    return self.job_name
```

I'm also fine "just doing it," but a suggestion.

khk-globus (approved):

parsl/tests/test_providers/test_submiterror_deprecation.py
```diff
@@ -0,0 +1,19 @@
...
+    j = "the_name"
+
+    ex = SubmitException(j, "m")
+
+    # the new behaviour
+    assert ex.job_name == j
+
+    # the old behaviour
+    with pytest.deprecated_call():
+        assert ex.task_name == j
```
khk-globus, line 19: Did not know about `pytest.deprecated_call()`.  Nice.

parsl/providers/errors.py
benclifford: i added something like this in a subsequent commit

parsl/tests/test_providers/test_submiterror_deprecation.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.mark.local
+def test_submit_execption_task_name_deprecation():
+    """This tests the deprecation warning of task_name in SubmitException
+    """
+    j = "the_name"
+
+    ex = SubmitException(j, "m")
+
+    # the new behaviour
+    assert ex.job_name == j
```
khk-globus, line 15: Were you so inclined, I like to use `randomstring()` (a fixture) for tests like this.  Sort of a "keeping honest" promise.  Here's an [example implementation](https://github.com/funcx-faas/funcX/blob/main/compute_endpoint/tests/conftest.py#L149-L156) from Globus Compute.  Then, this might look like:

```python
j = f"some_job_{randomstring()}"
```

parsl/tests/test_providers/test_submiterror_deprecation.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+from parsl.providers.errors import SubmitException
+
+
+@pytest.mark.local
+def test_submit_execption_task_name_deprecation():
+    """This tests the deprecation warning of task_name in SubmitException
```
khk-globus, line 8: Small potatoes&nbsp;&mdash;&nbsp;spelling of `exception`:

```diff
- def test_submit_execption_task_name_deprecation():
+ def test_submit_exception_task_name_deprecation():
```

parsl/tests/test_providers/test_submiterror_deprecation.py
benclifford: I'll bring this into Parsl in a separate PR - it looks good to have around

PR #2999, opened 2023-12-19
Remove unused type-check in dfk._count_deps

benclifford:
The list of dependencies is always a sequence of futures, so there is no need to check the type of each element.

# Changed Behaviour

none

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -262,9 +262,8 @@ class DataFlowKernel:
         """
         count = 0
         for dep in depends:
-            if isinstance(dep, Future):
-                if not dep.done():
-                    count += 1
+            if not dep.done():
+                count += 1
 
         return count
 


rjmello (approved):

parsl/dataflow/dflow.py
```diff
@@ -262,9 +262,8 @@ def _count_deps(self, depends: Sequence[Future]) -> int:
         """
         count = 0
         for dep in depends:
-            if isinstance(dep, Future):
-                if not dep.done():
-                    count += 1
+            if not dep.done():
+                count += 1
```
khk-globus, line 266: At that point, could make a simple summation.  Consider:
```python
def _count_deps(self, depends: Sequence[Future]) -> int:
    return sum(1 for d in depends if not d.done())
      # or, if it reads better to you:
    return sum(not d.done() for d in depends)
```

It also occurs to me that this method is effectively a `staticmethod`.  But ... not pertinent (per se) to this PR.

khk-globus (approved):

parsl/dataflow/dflow.py
benclifford: if we're proper yak-shaving: this is only used in one place:

```
parsl/dataflow/dflow.py:            if self._count_deps(task_record['depends']) != 0:
```

so there's no need to iterate over remaining dependencies once any not-done dependency has been observed.

For even more rambling wrt being a static method: `TaskRecord` is a place where having a `self` for context would make sense.

PR #3001, opened 2024-01-02
Mark radical MPI test

khk-globus:
A temporary measure to allow testing locally on dev setups that don't have `radical.pilot` installed and/or working.  Use with:

```console
$ pytest parsl/tests/ -k "not radical" ...
```

For now, this is not _implemented_ anywhere (e.g., not in `Makefile`) so it's strictly a manual usage.  At some point, the tests are due for an overhaul to clean up the testing situation.  I won't enumerate the specifics here, but suffice to say that this is a on-going conversation.

## Type of change

- Code maintentance/cleanup

### DIFF
--- a/parsl/tests/test_radical/test_mpi_funcs.py
+++ b/parsl/tests/test_radical/test_mpi_funcs.py
@@ -17,6 +17,7 @@ apps = []
 
 
 @pytest.mark.local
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
All good to go 👍

PR #3003, opened 2024-01-02
Set default test executor working dir

khk-globus:
In pursuit of no test detritus, set the default working directory of executors and the `dfk` to a test-run-specific temporary directory.  For "conforming" tests, (i.e., those that are aware of the fixtures in `conftest.py` that attempt to make fresh configurations), this enables auto-cleanup.  The rest will follow in piecemeal PRs "as we get there."

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -118,7 +118,7 @@ def pytest_configure(config):
 
 
 @pytest.fixture(autouse=True, scope='session')
-def load_dfk_session(request, pytestconfig):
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
Nice idea to put test logs into a tmp dir :)

benclifford:
I would expect this to move run logs out of the artifacts that github archives, which is a bad thing? But I still see plenty of run logs in the archive artifact I just looked at, so perhaps it doesn't do all of them, which is a bad thing.

PR #3007, opened 2024-01-02
Address pytest warning for checkpoint

khk-globus:
An older test; the only detail of interest for the tests is the checkpoint files, so update accordingly: `assert` the interesting bits (no change), and remove the return.  While here, update test to use the `conftest` infrastructure (`local_config()`).

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
Neat!

PR #3009, opened 2024-01-03
Minor refactor of checkpoint tests

yadudoc:
# Description

Current checkpoint tests an app that generates a random number to infer that subsequent calls are being pulled from the cache. This PR updates the random apps to instead use uuid which comes with better uniqueness guarantees to simplify the tests. 

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Och; much more direct.  Nicely done!

Issue #3013, opened 2024-01-04
test_radical/test_mpi_funcs.py needs unknown test setup


khk-globus:
**Describe the bug**
CI is unable to successfully complete `config_local_test` if `radical_local_test` is not run first.

That is, this appears to fail:
```console
$ make config_local_test
```
It's difficult to tell from the CI logs, but when running locally, this consistently hangs on `test_radical/test_mpi_funcs.py`:

<img src="https://github.com/Parsl/parsl/assets/99676336/911b2ba3-24f3-4300-9889-883f59274151" title="Showing test hang on `test_mpi_funcs.py` when running `make config_local_test` on local console" width="500">

Conversely, if the `radical_local_test` is executed first, then the test passes in CI.
```console
$ make radical_local_test config_local_test
```

Oddly, the `radical_local_test` does _not_ run that particular test, as noted by the `s` in the Pytest output:
```console
$ make radical_local_test
...
parsl/tests/test_radical/test_mpi_funcs.py s                       [ 13%]
...
```

**Expected behavior**
- Each test suite should be independent and not rely on an unstated setup.
  - In particular, this is a blocking issue for parallelizing the tests in CI.
- Specifically here, `config_local_test` should be able to complete on it's own without having to first run `radical_local_test`.

**Environment**
 - OS: Ubuntu22 (local laptop), Ubuntu20 (CI)
 - Python version: confirmed with 3.10 and 3.12
 - Parsl version: master as of this issue date

khk-globus:
If a dev is able to get to this before these timeout, here are two workflow runs that show the issue.

* Working: https://github.com/Parsl/parsl/actions/runs/7400896228 "What breaks it?"
* Failing: https://github.com/Parsl/parsl/actions/runs/7401184228 "Remove radical_test"

Note that the sole difference of interest between those two is in the workflow definition with:

* [working workflow](https://github.com/Parsl/parsl/actions/runs/7400896228/workflow)
* [failing workflow](https://github.com/Parsl/parsl/actions/runs/7401184228/workflow)

```diff
@@ -79,7 +79,7 @@ jobs:
     - name: make ${{ matrix.test-name }}
       run: |
         source .venv/bin/activate
-        make local_thread_test htex_local_test htex_local_alternate_test wqex_local_test vineex_local_test radical_local_test config_local_test
+        make local_thread_test htex_local_test htex_local_alternate_test wqex_local_test vineex_local_test config_local_test
 
     - name: Checking parsl-visualize
[...]

PR #3015, opened 2024-01-04
Split out `make test` into a matrix

khk-globus:
Parallelize the tests by using GH's matrix.  There's no need to execute the tests serially.

This reduces CI build time to ~11m, with the vineex test being the bottleneck. Accordingly, reduce timeout limit from 60m to 15m, with the expectation that if tests take longer than that, the *tests* need to be addressed.  In the semi-rare case of downloads taking too long, the developer in question can run another attempt in GH's CI UI.

N.B. Marking as a draft PR because it will fail until Issue #3013 is addressed.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/taskvine/install-taskvine.sh
+++ b/parsl/executors/taskvine/install-taskvine.sh
@@ -12,7 +12,7 @@ TARBALL="cctools-$CCTOOLS_VERSION-x86_64-ubuntu20.04.tar.gz"
 
 # If stderr is *not* a TTY, then disable progress bar and show HTTP response headers
 [[ ! -t 1 ]] && NO_VERBOSE="--no-verbose" SHOW_HEADERS="-S"
-wget "$NO_VERBOSE" "$SHOW_HEADERS" -O /tmp/cctools.tar.gz "https://github.com/cooperative-computing-lab/cctools/releases/download/release/$CCTOOLS_VERSION/$TARBALL"
+wget -c "$NO_VERBOSE" "$SHOW_HEADERS" -O /tmp/cctools.tar.gz "https://github.com/cooperative-computing-lab/cctools/releases/download/release/$CCTOOLS_VERSION/$TARBALL" || true
 
 mkdir -p /tmp/cctools
 tar -C /tmp/cctools -zxf /tmp/cctools.tar.gz --strip-components=1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,12 +7,54 @@ on:
       - synchronize
 
 jobs:
+  lint:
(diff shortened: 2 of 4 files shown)


parsl/executors/taskvine/install-taskvine.sh
```diff
@@ -12,7 +12,7 @@ TARBALL="cctools-$CCTOOLS_VERSION-x86_64-ubuntu20.04.tar.gz"
 
 # If stderr is *not* a TTY, then disable progress bar and show HTTP response headers
 [[ ! -t 1 ]] && NO_VERBOSE="--no-verbose" SHOW_HEADERS="-S"
-wget "$NO_VERBOSE" "$SHOW_HEADERS" -O /tmp/cctools.tar.gz "https://github.com/cooperative-computing-lab/cctools/releases/download/release/$CCTOOLS_VERSION/$TARBALL"
+wget -c "$NO_VERBOSE" "$SHOW_HEADERS" -O /tmp/cctools.tar.gz "https://github.com/cooperative-computing-lab/cctools/releases/download/release/$CCTOOLS_VERSION/$TARBALL" || true
```
benclifford, line 15: what failure mode are you overriding here with || true?

benclifford:
One thing I've had an eye on for a while, and this PR probably forces the issue, is more mechanical configuration of the compulsory-to-merge CI tests.

Right now that's a manually curated list in github settings, which needs a change every time a Python version is added or removed: if you forget when you remove a Python version from the test matrix, you'll discover that right away by not being able to merge; but when adding a version, you won't.

I've seen this technique mentioned, which has been on my vague-todo-list for a long time: https://github.com/orgs/community/discussions/26733#discussioncomment-3253151

which creates a single test that depends on everything in a matrix.

Maybe that's the right thing to do here, so that github merge requirements can be configured to require a single `allgood` test which depends on the entire much-enlarged matrix.

parsl/executors/taskvine/install-taskvine.sh
khk-globus: Well, egg on my face.  I was going to say: _In the "I use this locally" case, the `-c` parameter appears to return a non-zero if the file is already downloaded._  But going to prove that _now_, with a copied-from-my-terminal interaction, I'm no longer getting the error I thought I was working around:
```console
$ wget -c -O /tmp/cctools.tar.gz "https://github.com/cooperative-computing-lab/cctools/releases/download/release/7.7.2/cctools-7.7.2-x86_64-ubuntu20.04.tar.gz"
    ...
        The file is already fully retrieved; nothing to do.

$ echo $?
0
```

So ... nevermind.  I'll remove this hunk.

Issue #1429, opened 2019-11-08
max_workers is ambiguous


yadudoc:
`max_workers` kwargs option for `HighThroughputExecutor` is ambiguous. Far too many users have asked what the option means, and I think we are better off renaming this as `max_workers_per_node`.

benclifford:
Crossref #1353 

Maybe some phrasing subtlety when using fractional nodes: this parameter doesn't so directly limit the number of workers on a node in the case of two node-fractions being on the same name node.

annawoodard:
(I think this is basically a dupe of #1353). As I mention there, I favor removing this option entirely. Otherwise, `max_workers_per_manager` is slightly more correct with respect to using fractional nodes.

Andrew-S-Rosen:
I am reviving this old discussion because I just spent the last few hours in a Parsl death spiral because of a mismatch between the Parsl documentation's description of `max_workers` and its actual use. My hope is that this comment may help others. Tagging @WardLT because he runs calculations with this type of pattern a lot and surely is well-aware of this "gotcha" already.

The tl;dr --- to the best of my knowledge, `max_workers` is _not_ necessarily the number of workers on a node like [the documentation](https://parsl.readthedocs.io/en/stable/stubs/parsl.executors.HighThroughputExecutor.html) says. That is evident in the scenario below and likely is what @benclifford was alluding to (but I didn't discover this issue report until afterwards).

Here is the hypothetical scenario. We are launching PythonApps where each App is a Python script that calls some MPI-driven executable. The underlying MPI compute task uses a full node's worth of resources. We would like to run many of these PythonApps concurrently on a single Slurm allocation request. So, for instance, it would be a scenario where a single 10-node Slurm allocation would run, at most, 10 concurrent Apps where each App calls an executable that uses a full, dedicated node. For clarity, the Python script is calling `srun`, so everything is done using `SimpleLauncher()`.

If `max_workers` were indeed the number of workers per node, one would assume that it should be set to `max_workers=1` since we want each node to run a single, dedicated PythonApp at a time. However, doing this does not yield any concurrency. To achieve concurrency, one must actually set `max_workers` to be the number of concurrent Apps to run _even though we only want one App per node.

Here is what appears to be a proper configuration for the aforementioned scenario. Note how `max_workers` is not 1.

```python
import parsl
from parsl.config import Config
from parsl.executors import HighThroughputExecutor
from parsl.launchers import SimpleLauncher
from parsl.providers import SlurmProvider

[...]

PR #3016, opened 2024-01-08
Support for running MPI applications with `HighThroughputExecutor`

yadudoc:
# Description

The `HighThroughputExecutor` (HTEX) is designed to follow the pilot job model where nodes are provisioned, agents launched per node, which then execute tasks. However, this limits the executor from launching MPI aware applications which are not limited to single node execution. This PR extends HTEX to do:

1. Support an MPI mode that can be toggled with `enable_mpi_mode: bool` option. In this mode a single manager is launched per batch-job, that in turn keeps track of the provisioned nodes, and can schedule multi-node MPI applications.
2. Support for passing MPI application requirements such as `ranks_per_node` via the special `parsl_resource_specification` kwarg.
3. MPI applications with resource specifications defined are launched with environment variables such as `PARSL_MPI_PREFIX` that can be used to launch the application from within bash_apps.

I've got user documentation here -> https://github.com/Parsl/parsl/blob/mpi_experimental_3/docs/userguide/mpi_apps.rst

Tested on :
1. Polaris@ALCF with CosmicTagger and LAMMPS
2. Perlmutter@NERSC with LAMMPS

## Type of change

- New feature
- Code maintenance/cleanup


parsl/executors/high_throughput/executor.py
```diff
@@ -174,6 +174,16 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
 
     worker_logdir_root : string
         In case of a remote file system, specify the path to where logs will be kept.
+
+    enable_mpi_mode: bool
+        If enabled, MPI launch prefixes will be composed for the batch scheduler based on
+        the nodes available in each batch job and the resource_specification dict passed
+        from the app
+
+    mpi_launcher: str
```
benclifford, line 183: from a UX perspective, this is going to need some more serious documentation about why this is somehow different from the `Launcher` classes specified elsewhere in the htex config.

parsl/executors/high_throughput/executor.py
```diff
@@ -545,10 +567,6 @@ def submit(self, func, resource_specification, *args, **kwargs):
         Returns:
               Future
         """
-        if resource_specification:
```
benclifford, line 548: might be a better user experience to check for unknown keywords in the resource dictionary, rather than ignore this check completely.

parsl/executors/high_throughput/executor.py
```diff
@@ -258,6 +270,11 @@ def __init__(self,
         self.worker_logdir_root = worker_logdir_root
         self.cpu_affinity = cpu_affinity
 
+        self.enable_mpi_mode = enable_mpi_mode
+        assert mpi_launcher in VALID_LAUNCHERS, \
```
benclifford, line 274: I think if you do a `Literal` sequence in the type specification, typeguard might be able to check this

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -0,0 +1,102 @@
+import logging
```
benclifford, line 1: architecturally, this feels to me like a weird reimplementation of the parsl.launchers code. I'm not sure what the best approach though, but two enemy implementations of a launcher API, at least on first glance, feels bad to me.

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,191 @@
...
+from enum import Enum
+from typing import Dict, List
+
+from parsl.multiprocessing import SpawnContext
+from parsl.serialize import (pack_res_spec_apply_message,
+                             unpack_res_spec_apply_message)
+
+logger = logging.getLogger(__name__)
+
+
```
benclifford, line 16: maybe the behaviour in this class belongs with the relevant provider in parsl.providers?

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -181,6 +199,17 @@ def __init__(self, *,
         self.monitoring_queue = self._mp_manager.Queue()
         self.pending_task_queue = SpawnContext.Queue()
         self.pending_result_queue = SpawnContext.Queue()
+        self.task_scheduler: NoopScheduler
```
benclifford, line 202: this signature could look nicer if the naming didn't imply that a task_scheduler always did nothing -- eg a different base class?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,191 @@
...
+    def _return_nodes(self, nodes: List[str]) -> None:
+        for node in nodes:
+            self.nodes_q.put(node)
+        with self._free_node_counter.get_lock():
+            self._free_node_counter.value += len(nodes)  # type: ignore[attr-defined]
+
+    def put_task(self, task_package: dict):
+        user_ns = locals()
+        user_ns.update({"__builtins__": __builtins__})
+        _f, _args, _kwargs, resource_spec = unpack_res_spec_apply_message(
```
benclifford, line 148: I talked with @yadudoc elsewhere a bit about having the resource spec sent separately to the apply message, to avoid deserialization of the potentially heavy apply message here inside the manager code: instead extract it in the `HighThroughputExecutor` in the submitting process, and send it alongside all the task messages.

That could avoid deserializing the function and arguments inside the manager, which I think is quite a serious performance concern.

parsl/executors/radical/executor.py
```diff
@@ -400,8 +400,11 @@ def task_translate(self, tid, func, parsl_resource_specification, args, kwargs):
 
     def _pack_and_apply_message(self, func, args, kwargs):
         try:
-            buffer = pack_apply_message(func, args, kwargs,
-                                        buffer_threshold=1024 * 1024)
+            buffer = pack_res_spec_apply_message(func,
```
benclifford, line 403: It's not clear to me why this doesn't work with the earlier pack_apply_message implementation - the other end is still using the original unpack_apply_message, right?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,191 @@
...
+        pending_task_q: multiprocessing.Queue,
+        pending_result_q: multiprocessing.Queue,
+    ):
+        super().__init__(pending_task_q, pending_result_q)
+        self.scheduler = identify_scheduler()
+        # PriorityQueue is threadsafe
+        self._backlog_queue: queue.PriorityQueue = queue.PriorityQueue()
+        self._map_tasks_to_nodes: Dict[str, List[str]] = {}
+        self.available_nodes = get_nodes_in_batchjob(self.scheduler)
+        self._free_node_counter = SpawnContext.Value("i", len(self.available_nodes))
```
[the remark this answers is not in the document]
benclifford: or even a lock and a regular `List`, which would support `len`!

parsl/executors/radical/executor.py
yadudoc: Radical uses the `execute_task` from the HTEX process_worker_pool which now uses the updated `unpack_res_spec_apply_message` method. I updated both the worker side calls and the executor to match this.

parsl/executors/high_throughput/executor.py
yadudoc: I agree. Have you had a chance to look at the docs? Is there anything that could be added there for clarity?

parsl/executors/high_throughput/executor.py
yadudoc: This check was placed since HTEX did not support `resource_specification`. We are not doing any validation of  the resource_specification now, but I'm hoping to do that in a later round of work.

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: I do not like the duplication either, but it made some sense to separate it given that launchers are designed to launch one manager per node and no more, while mpi_launchers have to be flexible to support a wider range.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: Conceptually yes, in that we are figuring out resources on a per job basis and is very tied to batch-jobs. Right now, this functionality is usable only at by the manager, which is why I placed it here. Might be worth a chat.

parsl/executors/high_throughput/executor.py
benclifford: probably fine to leave it like this then: it's only worth doing if it makes things easier, I think.

parsl/executors/high_throughput/executor.py
yadudoc: From discusssion with @benclifford, we want to do a basic check to confirm that the user supplied spec only contains keys that are supported.

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: Let's plan on reworking this MPI launcher to integrate into the existing launchers. There are some troublesome bits from how the launchers are limited to just launching managers one per node.

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, requested: int, available: int):
+        self.requested = requested
+        self.available = available
+
+    def __str__(self):
+        return f"MPIResourceUnavailable(requested={self.requested} available={self.available}"
+
+
+class NoopScheduler:
```
yadudoc, line 78: Call this a `TaskScheduler` instead

parsl/executors/high_throughput/mpi_resource_management.py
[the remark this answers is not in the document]
yadudoc: I've converted the `nodes_q` to a regular queue since we expect only two different threads to access it. I still think it is nicer to use that `mp.Value` as a thread-safe counter since it comes with its own lock.

docs/userguide/mpi_apps.rst
```diff
@@ -40,116 +73,275 @@ to the desired number of blocks.
...
-If not, the executor may need to first partition nodes into distinct groups for each MPI task.
-Partitioning is necessary if your MPI launcher does not automatically ensure MPI task gets exclusive nodes.
-Resources vary in how the list of available nodes is provided,
-but they typically are found as a "hostfile" referenced in an environment variable (e.g., ``PBS_NODEFILE``).
-We recommend splitting this hostfile for the host block into hostfiles for each worker
-by adding code like the following to your ``worker_init``:
+1. Set ``HighThroughputExecutor(enable_mpi_mode=True)``
+2. Set ``HighThroughputExecutor(mpi_launcher="LAUNCHER")`` to one from ("srun", "aprun", "mpiexec")
+3. Set the ``max_workers`` to the number of MPI Apps you expect to run per scheduler job (block).
+4. Set ``cores_per_worker=1e-6`` to prevent HTEx from reducing the number of workers if you request more workers than cores.
```
WardLT, line 81: Why wouldn't setting MPI mode do this automatically?

docs/userguide/mpi_apps.rst
```diff
@@ -3,33 +3,66 @@ MPI Apps
...
 Configuring the Provider
 ++++++++++++++++++++++++
 
-Parsl must be configured to deploy workers on exactly one node per block.
-This part is simple.
-Instead of defining `a launcher <execution.html#launchers>`_ which will
-place an executor on each node in the block, simply use the :class:`~parsl.launchers.SimpleLauncher`.
+Parsl must be configured to deploy workers on exactly one node per block. This part is
+simple. Instead of defining a launcher which will place an executor on each node in the
+block, simply use the :class:`~parsl.launchers.SimpleLauncher`.
```
WardLT, line 64: Could we not ignore the launcher in MPI mode if it's not being used?

docs/userguide/mpi_apps.rst
```diff
@@ -40,116 +73,275 @@ to the desired number of blocks.
...
-        return f'mpiexec -n {n} --ppn 1 hostname'
+    def echo_hello(n: int, stderr='std.err', stdout='std.out', parsl_resource_specification: Dict):
+        return f'$PARSL_MPI_PREFIX hostname'
+
+    # Alternatively, you could also use the resource_specification to compose a launch
+    # command using env vars set by Parsl from the resource_specification like this:
+    @bash_app
+    def echo_something(n: int, stderr='std.err', stdout='std.out', parsl_resource_specification: Dict):
+        total_ranks = os.environ("")
+        return f'aprun -N $PARSL_RANKS_PER_NODE -n hostname'
```
WardLT, line 173: Where does the `hostname` come from in this example? What are the environment variables which Parsl sets?

docs/userguide/mpi_apps.rst
```diff
@@ -40,116 +73,275 @@ to the desired number of blocks.
...
 
 .. code-block:: python
 
-    @bash_app
-    def echo_hello(n: int, stderr='std.err', stdout='std.out'):
-        return (f'mpiexec -n {n} --ppn 1 '
-                '--hostfile /tmp/hostfiles/local_hostfile.`printf %02d $PARSL_WORKER_RANK` '
-                'hostname')
+    resource_specification = {
+      'NUM_NODES': <int>,        # Number of nodes required for the application instance
```
WardLT, line 185: Why all caps? Could we use a NamedTuple or data class to simplify type checking?

Also, what do you think about adopting the same terms as https://radicalpilot.readthedocs.io/en/stable/apidoc.html#taskdescription

WardLT: Well, it doesn't seem they have a number of number of nodes. They do ranks and ranks_per_node.

docs/userguide/mpi_apps.rst
```diff
@@ -40,116 +73,275 @@ to the desired number of blocks.
...
+      'RANKS_PER_NODE': <int>,   # Number of Ranks / application elements to be launched per node
+    }
 
-.. note::
+    # The above are made available in the worker env vars:
+    # echo $PARSL_NUM_NODES, $PARSL_RANKS_PER_NODE
```
WardLT, line 190: Is there an environment variable for the list of nodes selected by Parsl? I'll need that to use MPI mode when launching deepspeed tasks

docs/userguide/mpi_apps.rst
```diff
@@ -40,116 +73,275 @@ to the desired number of blocks.
...
+parsl workers will provision the requested resources and then compose MPI launch command prefixes
+(Eg: ``mpiexec -n <ranks> -ppn <ranks_per_node> -hosts <node1..nodeN>``). These launch command prefixes are
+shared with the app via environment variables.
 
 .. code-block:: python
 
     @bash_app
-    def echo_hello(n: int, stderr='std.err', stdout='std.out'):
-        return f'mpiexec -n {n} --ppn 1 hostname'
+    def echo_hello(n: int, stderr='std.err', stdout='std.out', parsl_resource_specification: Dict):
```
WardLT, line 165: Do my functions need a "parsl_resource_specificiation" kwarg to be used in MPI mode?

Andrew-S-Rosen: I'd strongly lean towards an implementation that does not require injecting Parsl logic into the underlying functions whenever possible.

docs/userguide/mpi_apps.rst
yadudoc: The intent here is to make the user aware of the fact that by default we will generally not provision more workers than cores on a compute node. I am wary of trying to set an alternate set of preset values for mpi_mode that takes away configurability when we do not know how users will use this.

docs/userguide/mpi_apps.rst
yadudoc: I think this makes sense. I will bring this up in the Parsl user meeting.

docs/userguide/mpi_apps.rst
yadudoc: Good question. The resource_specification tries to match Balsam jobs (https://balsam.readthedocs.io/en/latest/user-guide/jobs/). Once we have better clarity on what options we need we could switch to a data class. I think switching away from all caps is a stylistic choice that I can agree with.

docs/userguide/mpi_apps.rst
yadudoc: Yep, I've added additional info on the env vars we support, in this case `PARSL_MPI_NODELIST` is what you are looking for.

docs/userguide/mpi_apps.rst
yadudoc: Clarified in docs. The idea here was for each rank to print the output from `/bin/hostname`.

docs/userguide/mpi_apps.rst
WardLT: Woo!

docs/userguide/mpi_apps.rst
WardLT: That's a good idea. Overloading a node drastically could lead to performance degradation and annoyed users too.

WardLT: (feel free to resolve this)

docs/userguide/mpi_apps.rst
WardLT: Is the syntax wrong then? Should there not be an integer after `-n`.

docs/userguide/mpi_apps.rst
WardLT: @yadudoc / @benclifford , what do you see as the remedy to this? It seems we go wrong from the beginning, in that the resources are combined with the "app kwargs" starting from the initial call to PythonApp.__call__ but then pulled out and treated separately throughout the Executor logic (e.g., as a separate argument in HTEX.submit). 

Could we not make it so that HTEx only adds the resource spec into the function call if it is present in the signature?

benclifford: There are no python apps involved in this MPI use case.

I think separately there's a reasonable use case that python functions can come from external sources and be decorated directly with `@python_app`; and Parsl could make it so those don't get passed parsl-specific kwargs.

That's out of scope for this MPI PR and I encourage you to write up an issue with the desired semantics separately.

WardLT: Alright, I'm convinced that we should punt on this issue so that we have the MPI logic usable with bash apps. I still think this issue should be resolved, and I'll make an Issue about it

yadudoc: Marking this as resolved with the idea that we'll deal with the `parsl_resource_specification` handling in a separate issue.

docs/userguide/mpi_apps.rst
yadudoc: Fixed. The app looks like this now:

```python
@bash_app
def echo_hostname(n: int, stderr='std.err', stdout='std.out', parsl_resource_specification: Dict):
    total_ranks = os.environ("")
    return f'aprun -N $PARSL_RANKS_PER_NODE -n {total_ranks} /bin/hostname'
```

yadudoc: Fixed.

docs/userguide/mpi_apps.rst
yadudoc: I've updated the docs to specify that only `SingleNodeLauncher` is allowed and have added asserts to the code to only allow that one launcher. `SingleNodeLauncher` is the default launcher so it is nice to support the default case with mpi_mode. If we have a new launcher in the future that could work with the mpi_mode, we can update that assert then.

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class InvalidResourceSpecification(Exception):
+    """Exception raised when Invalid keys are supplied via resource specification"""
+
+    def __init__(self, invalid_keys: Set[str]):
+        self.invalid_keys = invalid_keys
+
+    def __repr__(self):
+        return f"Invalid resource specification options supplied: {self.invalid_keys}"
```
benclifford, line 18: repr shouldn't be a human-targeted string: see #1966 and https://docs.python.org/3/library/functions.html#repr

The default `repr` should be ok for exceptions and this human readable text can go into `__str__`.

parsl/executors/high_throughput/executor.py
```diff
@@ -545,10 +576,7 @@ def submit(self, func, resource_specification, *args, **kwargs):
         Returns:
               Future
         """
-        if resource_specification:
-            logger.error("Ignoring the call specification. "
-                         "Parsl call specification is not supported in HighThroughput Executor.")
-            raise UnsupportedFeatureError('resource specification', 'HighThroughput Executor', None)
+        validate_resource_spec(resource_specification)
```
benclifford, line 579: what happens when this returns false?

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __str__(self):
+        return self.__repr__()
+
+
+def validate_resource_spec(resource_spec: Dict[str, str]) -> bool:
```
benclifford, line 24: this should return `None` with an invalid state being signalled by an exception. *or* it should return bool and not raise exceptions

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+            pass
+
+    if mpi_launcher == "srun":
+        all_prefixes["PARSL_MPI_PREFIX"] = all_prefixes["PARSL_SRUN_PREFIX"]
+    elif mpi_launcher == "aprun":
+        all_prefixes["PARSL_MPI_PREFIX"] = all_prefixes["PARSL_APRUN_PREFIX"]
+    elif mpi_launcher == "mpiexec":
+        all_prefixes["PARSL_MPI_PREFIX"] = all_prefixes["PARSL_MPIEXEC_PREFIX"]
+    else:
```
benclifford, line 131: a different error handling path here is that someone specified an invalid mpi_launcher that is either a typo or something they assumed would work, and that should raise an error rather than silently misconfigure.

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+
+def get_nodes_in_batchjob(scheduler: Scheduler) -> List[str]:
+    """Get nodelist from all supported schedulers"""
+    nodelist = []
+    if scheduler == Scheduler.Slurm:
+        nodelist = get_slurm_hosts_list()
+    elif scheduler == Scheduler.PBS:
+        nodelist = get_pbs_hosts_list()
+    elif scheduler == Scheduler.Cobalt:
+        nodelist = get_cobalt_hosts_list()
```
benclifford, line 56: is it legitimate that an unknown scheduler has no nodes?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+    elif os.environ.get("PBS_NODEFILE"):
+        return Scheduler.PBS
+    elif os.environ.get("COBALT_NODEFILE"):
+        return Scheduler.Cobalt
+    else:
+        return Scheduler.Unknown
+
+
+class MPIResourceUnavailable(Exception):
+    """Raised if there are no free resources available for an MPI request"""
```
benclifford, line 73: does "resource" mean a rank? or a node? or some other abstract resource?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __init__(self, requested: int, available: int):
+        self.requested = requested
+        self.available = available
+
+    def __str__(self):
+        return f"MPIResourceUnavailable(requested={self.requested} available={self.available}"
```
benclifford, line 80: close `)`

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+        self.pending_task_q = pending_task_q
+        self.pending_result_q = pending_result_q
+
+    def put_task(self, task) -> None:
+        return self.pending_task_q.put(task)
+
+    def get_result(self, block: bool, timeout: float):
+        return self.pending_result_q.get(block, timeout=timeout)
+
+    def put_result(self, result) -> None:
```
benclifford, line 103: why is put_result, the far end of the communication abstracted here but not get_task?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+    def put_task(self, task_package: dict):
+        """Schedule task if resources are available otherwise backlog the task"""
+        user_ns = locals()
+        user_ns.update({"__builtins__": __builtins__})
+        _f, _args, _kwargs, resource_spec = unpack_res_spec_apply_message(
+            task_package["buffer"], user_ns, copy=False
+        )
+
+        nodes_needed = resource_spec.get("NUM_NODES")
+        if nodes_needed:
```
benclifford, line 182: whats the behaviour if you don't request any nodes with NUM_NODES? no nodes get allocated and the task gets to run anyway? and hopefully doesn't do a $PARSL_MPI_PREFIX invocation?

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -0,0 +1,222 @@
...
+        self.pending_task_q.put(task_package)
+
+    def _schedule_backlog_tasks(self):
+        """Attempt to schedule backlogged tasks"""
+        try:
+            _nodes_requested, task_package = self._backlog_queue.get(block=False)
+            self.put_task(task_package)
+        except queue.Empty:
+            return
+        except MPIResourceUnavailable:
```
benclifford, line 204: does this `except` need to happen here? does self.put_task already handle rescheduling the task into the backlog again?

parsl/executors/high_throughput/executor.py
yadudoc: `HTEX.submit` will fail with `InvalidResourceSpecification`, which I think is reasonable.

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: Fair enough, I've updated this method.

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: We assert that the mpi_launcher is a valid type here: https://github.com/Parsl/parsl/blob/mpi_experimental_3/parsl/executors/high_throughput/executor.py#L279

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: I think this will likely lead to a hang, there are a few solutions I can think of: 
1) Check if the provider is supported (cons: if we fail to get nodelist on a supported provider worker will hang) 
2) Push empty resource info to the worker and add a check there so that the task can then fail and bubble back through the result channel.

Either way, I think we should do this fix separately.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: In this case it is always nodes. I'll rename `s/resource/node` since that would be clearer.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: Tasks when scheduled are placed into the `pending_task_q` which is a`mp.Queue` from which the workers will get the task. The task scheduler abstraction didn't make much sense from the worker side, so I didn't implement the worker side primitive.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: No nodes get allocated, task gets to run, but you still get a `$PARSL_MPI_PREFIX`.  I'd assume that it is a user-bug if they put in `$PARSL_MPI_PREFIX` in their app and miss the resource_specification, and you'd get a broken invocation that will error out.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: Yes, the logic goes like this:
```
1. If the queue is empty -> return
```

benclifford: I'm suggesting that self.put_task will never raise an MPIResourceUnavailable exception so this except block will never fire.

Alternatively, if self.put_task can raise an MPIResourceUnavailable exception, then this exception should be protected against in the invocation `self.task_scheduler.put_task` in `Manager.pull_tasks` otherwise it will cause the Manager.pull_tasks loop to exit when resources are unavailable.

parsl/executors/high_throughput/mpi_resource_management.py
benclifford: put_result probably doesn't make sense to abstract on the worker side then?

parsl/executors/high_throughput/mpi_prefix_composer.py
benclifford: so this code path is never reached (or should never be reached) and a more robust behaviour might be to raise RuntimeError("impossible behaviour") rather than put in dead code.

parsl/executors/high_throughput/mpi_resource_management.py
benclifford: so i guess you get to run up to maximum workers worth of tasks like this and then they queue like any other task?

parsl/executors/high_throughput/mpi_resource_management.py
benclifford: do a RuntimeError("this is impossible")

parsl/executors/high_throughput/mpi_resource_management.py
benclifford: (delete this dead code exception block)

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: Fixed

benclifford:
I was going to merge this now but I have heard that there is secret interesting feedback to still be evaluated, so I will defer that merge.

christopherwharrop-noaa:
Kyle Chard asked me to add feedback from my testing.  Here are the highlights:

1. This PR branch appears to break the FluxExecutor (Traceback provided below)
2. I was able to get my test suite to pass using the new MPI mode
    a. The "srun" launcher worked flawlessly after removing configs needed for use with FluxExecutor
    b. The "mpiexec" launch also ran the code fine, but warnings were given because Intel MPI prefers the "srun" launcher.  This is a local system issue, not a Parsl issue.  This kind of issue would not be possible for Parsl to work around I think, but users should be aware that certain systems and MPI implementations require/prefer certain launchers.
3. There are conflicts in the conventions used for keys in `parsl_resource_specification` that I think should be addressed before this is merged (NOTE the overlapping of `NUM_NODES` and `num_nodes`)
    a. This PR is using: `parsl_resource_specification={"NUM_NODES": 3, "RANKS_PER_NODE": 2}`
    b. The FluxExecutor uses: `parsl_resource_specification={"num_tasks": 6, "num_nodes": 3}`
    c. This might be why this PR breaks FluxExecutor, but I do not know for sure
4. It is somewhat unclear (to me) what the proper setting for `max_workers` should be, how it relates to other parameters, and how it affects performance
    a. Is `max_workers` supposed to be equal to the max number of concurrent MPI apps desired?  
    b. If so, why not make `max_workers` equal to the total number of cores per block?  If you have a block with a total of N cores in it, then you'd never need more than N workers (allowing N concurrent `mpiexec -n 1 ./a.out`) My guess is this could result in bad performance from oversubscribing the node with hundreds of workers running on it?
    c. How does `max_workers` interplay with `cores_per_worker` and `cores_per_node`?
5. In one place in the docs it says to use `cores_per_worker=1e-6` but in another place/example it sets that value to an integer, like `cores_per_worker=8`.  I'm not sure which is correct.

Here is the traceback from the FluxExecutor failure when using this PR:

```
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/view/lib/python3.11/site-packages/parsl/dataflow/dflow.py", line 302, in handle_exec_update
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/view/lib/python3.11/site-packages/parsl/dataflow/dflow.py", line 570, in _unwrap_remote_exception_wrapper
    result = future.result()
             ^^^^^^^^^^^^^^^
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/._view/vh6hrawpm6eqhgv7fjwhsrmmmwkuw5ap/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/._view/vh6hrawpm6eqhgv7fjwhsrmmmwkuw5ap/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/._view/vh6hrawpm6eqhgv7fjwhsrmmmwkuw5ap/lib/python3.11/site-packages/parsl/executors/flux/execute_parsl_task.py", line 30, in main
    returnval = execute_task(fn_buf)
    ^^^^^^^^^^^^^^^^^
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/view/lib/python3.11/site-packages/parsl/executors/high_throughput/process_worker_pool.py", line 544, in execute_task
    f, args, kwargs, resource_spec = unpack_res_spec_apply_message(bufs, user_ns, copy=False)
      ^^^^^^^^^^^^^^^^^
  File "/work/noaa/gsd-hpcs/charrop/hercules/SENA/ExascaleWorkflowSandbox.yadu/install/spack/var/spack/environments/chiltepin/.spack-env/view/lib/python3.11/site-packages/parsl/serialize/facade.py", line 99, in unpack_res_spec_apply_message
    func, args, (kwargs, resource_spec) = unpack_apply_message(packed_buffer, user_ns=user_ns, copy=copy)
      ^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

christopherwharrop-noaa:
One additional comment I have is that I really really like that launcher environment variables are set for use in the bash Apps.  One of the things that has always plagued us is the difficulty in building portable scripts/apps that can work with different MPI launchers.  Being able to specify the launcher in the Config for the resource being configured, and then using `$PARSL_MPI_PREFIX` in the App is really nice.  And it's especially nice that I have the flexibility to add whatever crazy custom MPI launcher options I need for my particular system/application when I need to.

christopherwharrop-noaa:
I don't know if this is helpful or not, but my test suite, which passes with this PR, performs the following basic tests:

1. Run MPI "hello world" across three nodes, two ranks per node, verifies there is a "hello" from exactly two ranks on each of three different nodes.  This test helps ensure the specified number of nodes and ranks per node are being honored.
2. Run two concurrent "calulate pi" toy MPI apps. The first one runs on exactly one node and uses all available cores.  The second one runs on exactly two nodes, using all available cores on each node.  The tests verify that a) the apps actually do run concurrently, b) the apps run on a disjoint set of nodes. This test helps verify correct tracking of which nodes are in use such that available resources do get used for concurrent execution and that resources are not allocated to more than one app at the same time.

yadudoc:
Christ NOAA

1. I’ve got a fix in for the issue with the FluxExecutor, please give the branch a try again when you get a chance.
2. Nice! I’m curious about the error messages you got from `mpiexec`, could you share a paste please? With MPI and how machines are hyper customized, we have to lean on the users to make informed choices on mpi_launchers.
3. I based the special keywords following a discussion we had and it is clear that the special kwargs don’t completely work with WQ, Flux, and Balsam.  We could support for `num_tasks` and `ranks_per_node` if we can reach consensus. I would like to know what you think of these options.
4. Workers do take up memory (~300MB/worker). Which is why the docs recommend setting `max_workers` to the max concurrent MPI tasks you expect to run.
    1. We basically pick min(`cores_per_worker`, `max_workers`, `mem_slots`). This definitely is clunky!
5. I’ll update the docs to clarify that one, cores_per_worker=1e-6 is used so that the # of workers is determined by `max_workers`

yadudoc:
@christopherwharrop-noaa Thank you so much for testing the changes and the detailed feedback. Sounds like the toy PI app could be useful as a simpler proxy than what we have in the docs. Could you share your test scripts, please?

As for getting this supported in `globus-compute`, I'm working on it :) I'll make sure to keep you posted.

christopherwharrop-noaa:
@yadudoc - I will get to all of your requests ASAP.  A quick one first...

When I use the `mpiexec` launcher on my system (Slurm + code compiled with Intel MPI), it does run, but puts some warnings in the output.  (For some odd reason these warnings go to stdout instead of stderr?).  Anyway, I noticed because the extra output causes my tests to fail.  So, the code works (this is true also for the toy MPI Pi application as well), but it complains because of the different launching mechanism.  

I agree this has to be a user issue because there is no way for Parsl to divine out of thin air how every system in the world is configured.  I just wanted to make note of it in case it was something you'd want to mention in documentation or a FAQ or something.

```
(base) [charrop@hercules-login-2 tests]$ more  parsl_htex_mpi_hello_run.out
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
MPI startup(): Warning: I_MPI_PMI_LIBRARY will be ignored since the hydra process manager was found
Hello world from host hercules-03-64, rank 0 out of 6
Hello world from host hercules-03-64, rank 1 out of 6
Hello world from host hercules-04-01, rank 2 out of 6
Hello world from host hercules-04-01, rank 3 out of 6
Hello world from host hercules-04-02, rank 4 out of 6
Hello world from host hercules-04-02, rank 5 out of 6
(base) [charrop@hercules-login-2 tests]$
```

christopherwharrop-noaa:
I also like your proposal for `num_tasks` and `ranks_per_node`.  That seems very straightforward and clear.  I've only been exposed to Flux's options, so I can't comment on how well it meshes with the other executors' method for specifying resources.  That being said, I 100% approve of your proposal.

benclifford:
For @christopherwharrop-noaa 's comments 4,5 about parameters max_workers, etc: broadly these parameters make sense with the "original" mode of htex which tries to do things based on one worker per core, and so you have to set them to trick "original" htex to configure some re-purposed internal pieces in the weird way that "new MPI htex" wants.

I've already been fairly convinced that the MPI executor is different enough in behaviour that the user facing configuration interface (i.e. the `HighThroughputExecutor` constructor/initializer) should be different for MPI mode (for example, an MPIExecutor) - backed by mostly the same code as the existing HighThroughputExecutor, and @christopherwharrop-noaa 's highlighting of the confusingness of these parameters, broadly irrelevant to the user experience of MPI execution, is more fuel for that.

christopherwharrop-noaa:
To further complicate the site-specific issue, I originally had a hard time getting this PR to work because I had forgotten about a special setting that I needed for Flux which I had in my setup.

Flux uses its own custom pmi launching, so it won't work with the usual launchers.  For example, instead of `mpiexec -n 100 ./a.out` you just do `./a.out` and it uses the resources passed in (e.g. `num_tasks=100`) to figure out how to launch it using its own pmi.  

BUT... additionally, Intel MPI sets `$I_MPI_PMI_LIBRARY` in the environment when its module is loaded.  If found, it causes Flux's custom pmi to be bypassed and everything goes out to lunch and it crashes in `MPI_Init()`.  So... when using FluxExecutor with Intel MPI, I have to do `unset I_MPI_PMI_LIBRARY` to get it to work.

I had forgotten about all of that, and so initially I could not get my executable to run correctly with the MPI-enabled htex.  But, removing the `unset`, so that `$I_MPI_PMI_LIBRARY` is correctly defined again for Intel MPI, it all works. 

I know this is a bit of a long story, but it does highlight, I think, the fact that users have to take some responsibility in making sure they have MPI configured correctly for their particular setup (and it can be a bit tedious sometimes).  There's just no way a package like Parsl could ever magically do all that.

christopherwharrop-noaa:
@yadudoc - My tests for this PR can be found [here](https://github.com/NOAA-GSL/ExascaleWorkflowSandbox/blob/feature/mpi-htex/tests/test_parsl_htex_mpi_hello.py).  There is a corresponding test file for flux in the same directory that runs the same tests.

This is currently a bit of a mess.  The code for the `mpi_pi.f90` is in the same directory in the repository and you are welcome to use it if it's useful.  I wrote it because I needed something that actually ran for several seconds so that I could do my concurrent execution tests.

christopherwharrop-noaa:
Commenting on this separately so it doesn't get lost in a lot of words....

I tested your latest and I can confirm that it fixes my FluxExecutor issue.  Thank you!

christopherwharrop-noaa:
Ok... One additional finding.  I said that your latest push fixes my Flux issue.  Well, it DOES fix the issue I reported, and all my MPI apps run with no problems.  But I just noticed that a simple Flux sanity check App is now broken.  I've verified this by re-installing the latest Parsl release and running my tests, then reinstalling your PR version and running the tests.  I did this a couple times to make sure, and it's repeatable.  This failure makes absolutely no sense to me, and I can't imagine what it has to do with this PR.  But, I thought I'd better report it.

My App runs a single Flux command: `flux resource list`.  That's it!  And it fails with this error:

```
flux-resource: ERROR: Unable to connect to Flux: broker socket /run/flux/local was not found
```

I do not know why/where this would be coming from.  Especially since all the Flux Apps that run real MPI executables all work.  Additionally, another sanity check test that runs only one command, `flux pmi barrier` also works!  

Anyway, not sure what is going on yet, but thought I needed to mention it.

christopherwharrop-noaa:
> I've already been fairly convinced that the MPI executor is different enough in behaviour that the user facing configuration interface (i.e. the `HighThroughputExecutor` constructor/initializer) should be different for MPI mode (for example, an MPIExecutor) - backed by mostly the same code as the existing HighThroughputExecutor, and @christopherwharrop-noaa 's highlighting of the confusingness of these parameters, broadly irrelevant to the user experience of MPI execution, is more fuel for that.

I agree that the behavior is different enough to justify a separate MPIExecutor altogether.  Aside from the reasons already mentioned is that HTEX="High Throughput Executor".  MPI applications typically do not fit into the "High Throughput Computing" category; they are more often associated with traditional "High Performance Computing".  The focus and goals of HTC is different than traditional HPC. 
For workflows that are composed of a lot of serial or single-node multi-threaded codes, throughput is what matters most.  But, for workflows running a lot of very large MPI applications, we care more about the raw computational performance of the machine and scalability of the codes and their communications patterns than we do about how many "things" per unit time we can run.  HTC and HPC have different needs.

I would say that MPI execution doesn't fit very well into the HTEX context.  Perhaps putting the "square peg" MPI into the HTEX "round hole" doesn't make sense?

This is my opinion.  I don't know if others share it or not.

WardLT:
I think @christopherwharrop-noaa and whomever he is responding to have a point. How problematic would it be to implement an MPIExecutor? Could it just be a subclass of HTEx with different documentation and some options (e.g., max_workers) fixed or renamed?

benclifford:
> I think @christopherwharrop-noaa and whomever he is responding to have a point.

He's responding to me.

> Could it just be a subclass of HTEx with different documentation and some options (e.g., max_workers) fixed or renamed?

yes, I think that a subclass with a different `__init__` is what I am imagining.

(motivated by more functional/template style reasoning: let's make a "template" that configures all the parameters that are weird here into their special weird values)

christopherwharrop-noaa:
It just occurred to me that a counterargument to my comment above is that the multi-node MPI Apps will create a resource fragmentation problem in the pilot job.  With HTEX, where everything is serial or single-node, you don't have that problem.  But with MPI Apps needing varying sized chunks of resources within the pilot, you now have the potential for wasted idle resources (no queued Apps are small enough to fit into the free resources), and maybe worse, you also have potential for App starvation if you don't use a FIFO method for scheduling the MPI jobs.

If you were to mix the MPI jobs with serial jobs, you'd be able to backfill the "holes".  But now, you're in completely new territory I think for scheduling the Apps, which might be somewhere Parsl developers don't want to go.

You've probably already considered all of that, but I'm now realizing that the added complexity of MPI job support is much more than it appeared on the surface.

benclifford:
@christopherwharrop-noaa 

> If you were to mix the MPI jobs with serial jobs, you'd be able to backfill the "holes". But now, you're in completely new territory I think for scheduling the Apps, which might be somewhere Parsl developers don't want to go.

The model for that right now is submit your serial jobs as "i require 1 node" jobs.

You can't run (safely) run Python jobs, because Python jobs aren't supported by this mode; and you can't run partial node command line jobs this way either.

Architecture-wise that probably requires an entirely different executor architecture, but there are plenty of other people already working on MPI+normal job executors, I think(?), that interface to Parsl - including flux and radical - so I wouldn't particularly expect this to be a priority for Parsl to NIH-reinvent, rather than collaborating with those groups.

WardLT:
I'm also comfortable with a solution which prevents "partial node" and MPI apps running in the block. That does feel like something far from what Parsl does now (one block, one type of worker) and could be something to engage radical or flux for.

Andrew-S-Rosen:
Thanks everyone for the progress on this interesting PR.

Before noticing this PR and associated comments, I spent several hours over the weekend being terribly confused about the subtleties of `max_workers` and its meaning in the context of running MPI-based applications with `HighThroughputExecutor`. A summary of my thoughts on the matter can be found in https://github.com/Parsl/parsl/issues/1429#issuecomment-1937462128. @benclifford left a very helpful comment, ultimately pointing me to this PR, which is quite related to my concerns. I agree with what has been stated elsewhere that the MPI mode should be configured differently, as the `max_workers` is not on a per-node basis (as described in the documentation) and `cores_per_worker` is hacky. This ultimately makes it very difficult for new users trying to understand the different terminology. This PR is an opportunity to do things right.

As I mention in the linked comment, the following is a pretty typical scenario for us:

> We are launching PythonApps where each App is a Python script that calls some MPI-driven executable. The underlying MPI compute task uses a full node's worth of resources. We would like to run many of these PythonApps concurrently on a single Slurm allocation request via the `HighThroughputExecutor`. So, for instance, it would be a scenario where a single 10-node Slurm allocation would run, at most, 10 concurrent Apps where each App calls an executable that uses a full, dedicated node. For clarity, the Python script is calling `srun`, so everything is done using `SimpleLauncher()`.

Similarly, there are variations on the above theme that are important. For instance, have each PythonApp using multiple nodes (e.g. 2 nodes per App, so that's 5 concurrent Apps possible) or, potentially, half a node each (e.g. 32/64 cores per App, so that's 20 concurrent Apps possible). The former is by far more important/common. The latter is not particularly necessary, but one could imagine it being desirable on nodes with many cores. Is the latter the "partial node" scenario you are talking about, @WardLT? If so, I agree that it should be lower priority but could be nice down the line. In any case, we definitely need to ensure the former is fully functional in Parsl with this PR.

@WardLT, have you thought about how common ASE-based workflows like those I sketched above would be handled with this new approach? I assume you have and will defer to your assessment of the matter!

benclifford:
@Andrew-S-Rosen i think what you're doing is pretty much a target for this PR - it would be interesting if you could try it out. However, I think the partial-node allocation isn't something this PR does, because I think it allocates entire nodes only (?).

christopherwharrop-noaa:
Maybe my earlier comments were a bit confusing.  I wasn't trying to recommend sharing of nodes.  In MPI-world, it is often (usually?) a bad idea for two separate jobs to share nodes.  The reason is that their executions can influence each other via cache and memory consumption, not to mention affinity issues.  This can result in unpredictable execution time and can even result in crashes.  A lot of big MPI codes are highly optimized for cache utilization, and very repeatable execution times are expected.  Different sites have different philosophies and policies, but I think it's pretty common to set schedulers to never share nodes between jobs by default.  A common exception would be to allow serial jobs to share nodes.  But, one additional issue when sharing nodes is that the only way to do that safely is if 1) The scheduler knows how much memory each node has, and 2) Each job requests a specific amount of memory (either via a user request or via a default).  If both of those conditions are not met, then you risk oversubscribing the memory on the node, causing multiple jobs to crash.  So far we've only discussed requests for number of nodes and cores per node.  But, if sharing of nodes is an option, you also will need total memory per node, and memory per node when calling the app.  Right?  In certain circumstances, I think it could be very beneficial for a user to allow sharing of nodes, but memory consumption then becomes really important.

Andrew-S-Rosen:
@yadudoc totally agree. The sharing nodes situation is messy and, I would argue, a niche use case anyway. As long as normal, multi-node stuff is handled fine, then that's what matters. I haven't had time to try this out yet, but I look forward to doing so. Thanks for this!

benclifford:
> But, if sharing of nodes is an option, you also will need total memory per node, and memory per node when calling the app. Right? 

htex already does multiple tasks on one node without memory limitations/memory allocation! The assumption there is that the user won't explode their node by using too much memory, without htex becoming involved.

WardLT:
@Andrew-S-Rosen 

> @WardLT, have you thought about how common ASE-based workflows like those I sketched above would be handled with this new approach? I assume you have and will defer to your assessment of the matter!

I have thought about this a lot. The plan for this PR is to ensure that Parsl can at least run `bash_app`s in MPI mode. ASE workflows and the like will require being able to assign resources to Python apps, which we're going to target in a follow-on PR. 

My hope is that we'd be able to configure ASE to use Parsl's wrapper over the MPI launcher (i.e., `PARSL_MPI_PREFIX`) to launch applications from subprocesses. I don't see that being an issue

>  However, I think the partial-node allocation isn't something this PR does, because I think it allocates entire nodes only

Yea, the route we're pursuing treats nodes as indivisible. I typically call non-MPI versions of codes when running on a partial nodes (QE and CP2K have this). Is there a code you're working with that still must use `mpiexec` even if using less than one node?

christopherwharrop-noaa:
@benclifford - I see what you mean about multiple Apps within a single node, and I guess that hadn't really crystallized until now.  I don't want to digress away from the purpose of this PR.  The work being proposed here is really important, and I don't want to introduce tangential issues that slow it down.   It seems like a philosophical design question largely unrelated to what's happening in this PR.  Thank you for helping me appreciate the nuances.

Andrew-S-Rosen:
Thanks @benclifford. I think part of the confusion is that this PR does not seem specifically focused on `PythonApp`s. 

> I don't think this PR is intended that you would make your own srun commands outside of that framework.

We'll get to this after this PR. 😉

benclifford:
@Andrew-S-Rosen 

>  I think part of the confusion is that this PR does not seem specifically focused on PythonApps.

right, an  MPI-configured htex executor from this PR does not meaningfully have a way of running python apps in the traditional sense. (any python apps will run, but not in any MPI-allocated sense, so I think best to not think about it...)

benclifford (approved):
I'll merge this now as a prototype, with some discomfort: this code is clearly prototype and introduces a load of mess for both users and programmers, but MPI in Parsl is fundamentally a mess for everyone because it's at odds with what Parsl is.

The main concerns I have are:
* duplication of the entire batch system interface into a duplicate competing batch system interface
* the user interface for configuring MPI mode probably should be exposed as an MPIExecutor, not through the HighThroughputExecutor initialiser: many fields suddenly get different meanings (such as everything to do with workers, because workers now mean different things)

Easy wins to make it look like you care a little bit about further work:
* implement MPIExecutor as a subclass with a different init method with parameters aimed at MPI but inheriting the class implementation
* get rid of cores_per_worker equals almost-but-not-quite-0 - if you're trying to disable the cores per worker mechanism, make it so it can be disabled.

PR #3018, opened 2024-01-09
Reduce WorkQueue CPU time at expense of more latency in complex submission patterns

colinthomas-z80:
# Description
After #2984 was merged it was observed that the work queue send/retrieve loop consumes CPU when there is no work to be done. This PR suspends execution waiting on the task input queue when both queues are empty. 

## Type of change

- Bug fix
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/workqueue/executor.py
+++ b/parsl/executors/workqueue/executor.py
@@ -849,7 +849,7 @@ def _work_queue_submit_wait(*,
             break
 
         # Submit tasks
-        while task_queue.qsize() > 0 and not should_stop.value:
+        while task_queue.qsize() > 0 or q.empty() and not should_stop.value:
             # Obtain task from task_queue
             try:
                 task = task_queue.get(timeout=1)


benclifford:
does this then introduce an up-to-1 second latency on result processing, if the next thing to happen is a result being returned?

colinthomas-z80:
Yes it appears that would be the case if retrievals became available inside the submit loop. hmm

khk-globus:
> does this then introduce an up-to-1 second latency on result processing, if the next thing to happen is a result being returned?

A "quick & dirty" approach to address this might be to convert the `should_stop` to a `multiprocessing.Event`, move the check to a thread, and use a poison pill methodology.  Consider:

```python
    def __init__():
        ...
        self.should_stop = multiprocessing.Event()
        ...

    ...

    def _work_queue_submit_wait(...):
        ...
        def _watchdog():
            if self.should_stop.wait():
                task_queue.put(None)

        threading.Thread(target=_watchdog, daemon=True).start()

        while True:
            task = task_queue.get()

            if task is None:
                break

            logger.debug("Removing task from queue")
            ...
```
Would clearly need to touch up some other parts of the class&nbsp;&mdash;&nbsp;and I don't know the larger setup to know if this is feasible for you, but an idea based strictly on this changeset.

khk-globus:
Och; going too fast.  I missed that `task_queue` != `q`.  :man_facepalming:

benclifford:
> hmm

@colinthomas-z80 my gut feeling here is this could perhaps be two separate loops: the problem stems from having one loop trying to do two different things (so each thing has to keep iterating in order to check if should do the other thing, rather than being able to block for a longer time)

in a more async-io or unix-select world, you'd be able to block on both of the queues at once in a single loop and deal with whichever has something to do, but it's not immediately obvious if you can do that with the two different queue implementations here.

colinthomas-z80:
@benclifford I was mistaken in the behavior of WorkQueue.empty() in acknowledging the potential 1 second delay. WorkQueue.empty() should only be true if there are no tasks in the system at all, regardless if they are ready to retrieve or not. 

Because of this we will only block on the task submit queue if Work Queue is inactive. This should be good because the only possible subsequent event would be new tasks showing up in the queue we are waiting on.

benclifford:
@colinthomas-z80 

ok. I see that the `pytest` runtime doesn't get worse with this PR, which is good, and the original reason for this work. I'll draw some diagrams for myself to try to convince myself this is (or is not) ok in general... my main concern is: if this loop is not using 100%, then it is mostly inside one or other of the waits... so if something arrives in the *other* queue (that is not being waiting on) during that wait, does that mean that event experience latency before being processed? and if so, what are the circumstances when that would happen? experimentally it doesn't happen inside `pytest` for example (which is good).

colinthomas-z80:
If we are waiting on the submit queue, then we do not have to worry about the retrieve queue, because the only way we initiate a blocking wait on the submit queue is if work queue was empty. 

If we are waiting on the retrieve queue, it is possible the submit queue could receive new tasks, and this delay could be up to 1 second. However waiting on the retrieve queue is necessary to allow work queue to perform its functions. 

The worst case I can think of is if a long running task was sent out, and then short running serialized tasks were dispatched and retrieved. Because of the long running task, each serialized retrieval we would wait one second.

benclifford:
trying to imagine practical situations where that would affect things, I think if you had a workflow that had a bunch of independent dependency chains of tasks:

```
    1 -> 2 -> 3

    7 -> 8 -> 9
    ...
```

that would result in a situation where: there would always be some tasks running in WQ (so there would be latency on WQ starting new tasks) but Parsl's DFK would be also often submitting new tasks (as individual dependencies complete) that would then be subject to that latency.

I think that's something to still bear in mind, but I'll merge this PR because it is at least an improvement CPU-wise in the almost-purely-serialised case that motivated this.

PR #3036, opened 2024-01-30
Mark ZMQ tests as local because they don't need a DFK

benclifford:
This will run these tests with `--config local` instead of once per DFK configuration.

# Changed Behaviour

Nothing user facing. These tests shouldn't be interacting with the surrounding DFK, so the absence of a DFK shouldn't affect test behaviour.

## Type of change

- Code maintenance/cleanup


rjmello (commented):
FYI, I did this as part of https://github.com/Parsl/parsl/pull/3030.

benclifford:
this was done in #3030 - no need to merge this

Issue #3039, opened 2024-01-30
PR #2764 violates some pytest assumptions about sharedness of working files


benclifford:
**Describe the bug**

Prior to PR #2764, working files for tests were generally stored in the current directory. When running on a multi-node system (eg an HPC system), it was implicit that either:

* this working directory would be accessible across all nodes and default file staging would be used

* or perhaps as a not very supported alternative: the mostly abandoned file staging code would be configured to stage files appropriately.

PR #2764 changes the placement of this working files to go into a temporary working directory which by default is not accessible across a shared filesystem. This causes tests running with a DFK configuration that there's a shared filesystem with tests placing their files in that shared filesystem to fail.

Some possible fixes:

i) implement file staging configuration to deal with staging files from worker nodes to non-shared filesystems on the submit node.
ii) use a shared directory instead of the default tmp_path

Non-fixes:

set TMPDIR to a shared filesystem: multiprocessing creates temporary sockets(?) in this directory, and that kind of special file does not work on (for example) perlmutter's shared gpfs home filesystem.

**To Reproduce**

Try running pytests on perlmutter+slurm, for example

**Environment**
parsl 6c1d4df65ea1eba58839b02a5b22b93394ae2bc1 2024.01.29

Issue #3037, opened 2024-01-30
sometimes Interchange will accept a hostname address, in violation of test expectations


benclifford:
**Describe the bug**
In PR #2828, interchange binding behavior was changed with the expectation that it would no longer accept a hostname as a binding address.

This no-hostname behaviour is tested in `parsl/tests/test_htex/test_htex_zmq_binding.py`, `test_interchange_binding_with_non_ipv4_address` which tests that starting the interchange will raise a `ZMQError` when the interchange is passed the address `localhost`.

In an install I just made on Perlmutter, this test fails: the interchange will accept `localhost` as an address, and can successfully run the test suite with that change:

```
--- a/parsl/tests/configs/htex_local_alternate.py
+++ b/parsl/tests/configs/htex_local_alternate.py
@@ -39,7 +39,7 @@ def fresh_config():
     return Config(
         executors=[
             HighThroughputExecutor(
-                address="127.0.0.1",
+                address="localhost",
                 label="htex_Local",
                 working_dir=working_dir,
                 storage_access=[FTPInTaskStaging(), HTTPInTaskStaging(), NoOpFileStaging()],
```

I have tried on my laptop and the same change results in the expected hang, with this reported in interchange.log:

```
  File "/home/benc/parsl/virtualenv-3.11/lib/python3.11/site-packages/zmq/sugar/socket.py", line 498, in bind_to_random_port
    self.bind(f'{addr}:{port}')
  File "/home/benc/parsl/virtualenv-3.11/lib/python3.11/site-packages/zmq/sugar/socket.py", line 302, in bind
    super().bind(addr)
  File "zmq/backend/cython/socket.pyx", line 564, in zmq.backend.cython.socket.Socket.bind
  File "zmq/backend/cython/checkrc.pxd", line 28, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: No such device (addr='tcp://localhost:54144')
```

On perlmutter, it looks like libzmq is coming through conda:

> conda list | grep -i zero
zeromq                    4.3.5                h59595ed_0    conda-forge

/global/homes/b/bxc/.conda/envs/tmp-parsl-slurm-test/lib/libzmq.so.5.2.5

(the .so version is different from the release version in ZMQ, as far as I understand things).

On my laptop dev environment, libzmq is coming via I think a binary install from `pip install pyzmq` - there's no conda involved in that virtualenv:

[...]

PR #3040, opened 2024-01-30
Remove use of tmp_path fixture.

khk-globus:
PR #2764 introduced `tmp_path` without understanding of the shared FS expectation for some tests (i.e., those run on site-specific hosts).  PR #2803 fixed this with `tmpd_cwd` which is a temporary directory created inside of the current working directory&nbsp;&mdash;&nbsp;which is assumed to be on a shared FS.

Fixes: #3039

## Type of change

- Bug fix (in tests)

### DIFF
--- a/parsl/tests/sites/test_concurrent.py
+++ b/parsl/tests/sites/test_concurrent.py
@@ -24,7 +24,7 @@ def make_config():
 
 
 @mark.local
-def test_executor(tmpdir):
+def test_executor():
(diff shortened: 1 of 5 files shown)


benclifford:
This passes on perlmutter, modulo the test failing in #3037 which I deleted in order to see this PR #3040 pass.

benclifford (approved):

PR #3044, opened 2024-01-31
Permit both typeguard 2.x and 4.x

### LINKED ISSUE #3017: parsl vs typeguard 4 doesn't work
**Describe the bug**
multiprocessing.Queue is used as a generic type in "monitoring/monitoring.py"; example case; see [1]
multiprocessing.Queue is not a "type" (at-least in Python 3.11.7)
Quoting pydoc:
"multiprocessing.Queue = Queue(maxsize=0) method of multiprocessing.context.DefaultContext instance
    Returns a queue object"
This is causing code that uses MonitoringHub to fail.

[1]: See: https://github.com/Parsl/parsl/blob/479cf9db614b8f1e954f177ecbbc0ddccfb4e581/parsl/monitoring/monitoring.py#L171

**To Reproduce**
1. Setup Parsl 2024.1.8 with Python 3.11.7
2. Run the following test script:

```
> cat test.py                                                                                                                                                                                                                                      (episim37)
import parsl
from parsl.monitoring.monitoring import MonitoringHub

config = Config(
   executors=[
       HighThroughputExecutor(
           label="local_htex",
           cores_per_worker=1,
           max_workers=4,
       )
   ],
   monitoring=MonitoringHub(
       hub_port=55055,
       monitoring_debug=False,
   ),
   strategy='none'
)

parsl.load(config)
```
3. Run script
```
python test.py
```
```
  File "/home/parantapa/test.py", line 25, in <module>
    parsl.load(config)
  File "/home/parantapa/miniconda3/envs/episim37/lib/python3.11/site-packages/parsl/dataflow/dflow.py", line 1428, in load
    cls._dfk = DataFlowKernel(config)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/parantapa/miniconda3/envs/episim37/lib/python3.11/site-packages/parsl/dataflow/dflow.py", line 116, in __init__
    self.hub_interchange_port = self.monitoring.start(self.run_id, self.run_dir)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/parantapa/miniconda3/envs/episim37/lib/python3.11/site-packages/parsl/monitoring/monitoring.py", line 171, in start
    comm_q: Queue[Union[Tuple[int, int], str]]
            ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'method' object is not subscriptable
```

**Expected behavior**
Code to not crash

**Environment**
 - OS: ArchLinux
 - Python version: 3.11.7
 - Parsl version: 2024.1.8

**Distributed Environment**
- Where are you running the Parsl script from ? [e.g. Laptop/Workstation, Login node, Compute node]
Laptop

- Where do you need the workers to run ? [e.g. Same as Parsl script, Compute nodes, Cloud nodes]
Laptop

benclifford:
This PR loosens some type checking to accomodate both typeguard 2.x and 4.x as large parsl users (eg. LSST DESC) still package version 2.x but other users want the latest - 4.x

In typeguard 4.x, local variables are instrumented for type-checking, which is incompatible with using multiprocessing.Queue: multiprocessing.Queue is not a type as far as typeguard is concerned, so a single use of that, in MonitoringHub, is commented out by this PR.

Typeguard 2.x and 4.x have different type error classes, so explicit type checking of exception classes cannot happen in tests as the typeguard 4.x TypeCheckError class does not exist when typeguard 2.x is installed. This PR makes that type-check of type-checking errors happen using string matching of the name of the type-check error class in test_bash_apps/test_stdout.py

See for example issue #3017

This PR is ugly but attempts to be pragmatic about user packaging needs.

# Changed Behaviour

different exceptions on user type errors

# Fixes

Fixes #3017

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -171,7 +171,10 @@ class MonitoringHub(RepresentationMixin):
         self.logger.debug("Initializing ZMQ Pipes to client")
         self.monitoring_hub_active = True
 
-        comm_q: Queue[Union[Tuple[int, int], str]]
+        # This annotation is incompatible with typeguard 4.x instrumentation
+        # of local variables, because Queue is not a type: it's a method on
+        # multiprocessing.context.DefaultContext.
+        # comm_q: Queue[Union[Tuple[int, int], str]]
         comm_q = SizedQueue(maxsize=10)
 
         self.exception_q: Queue[Tuple[str, str]]
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -23,7 +23,7 @@ from parsl.serialize import deserialize
 
 from parsl.monitoring.message_type import MessageType
 from parsl.monitoring.types import AddressedMonitoringMessage, TaggedMonitoringMessage
-from typing import cast, Any, Callable, Dict, Optional, Sequence, Tuple, Union
+from typing import cast, Any, Callable, Dict, Optional, Sequence, Tuple, Union, TYPE_CHECKING
 
 _db_manager_excepts: Optional[Exception]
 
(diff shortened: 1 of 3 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -171,7 +171,10 @@ def start(self, run_id: str, run_dir: str) -> int:
         self.logger.debug("Initializing ZMQ Pipes to client")
         self.monitoring_hub_active = True
 
-        comm_q: Queue[Union[Tuple[int, int], str]]
+        # This annotation is incompatible with typeguard 4.x instrumentation
+        # of local variables, because Queue is not a type: it's a method on
```
rjmello, line 175: We should be able to use `multiprocessing.queues.Queue` for the type hint.

rjmello (changes requested):

parsl/monitoring/monitoring.py
benclifford: that fixed it partly, but didn't allow use of `[]` subscripts. But I then made a change that changes which type definition is in use depending on whether you are at runtime or inside mypy. which is pretty ugly.

rjmello: It ain't pretty, but it works.

rjmello (approved):

PR #3045, opened 2024-01-31
Remove unused codepath in htex scale_in for explicit blocks

benclifford:
Before this PR, there was a code path where HighThroughputExecutor.scale_in could be invoked with an explicit list of block IDs to scale in, rather than allowing HTEX to choose its own blocks to scale in.

This codepath is not used in the codebase, and I think has never been used since it was introduced in #1563

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -603,7 +603,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
     def workers_per_node(self) -> Union[int, float]:
         return self._workers_per_node
 
-    def scale_in(self, blocks=None, block_ids=[], force=True, max_idletime=None):
+    def scale_in(self, blocks=None, force=True, max_idletime=None):
         """Scale in the number of active blocks by specified amount.
 
         The scale in method here is very rude. It doesn't give the workers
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -627,46 +627,40 @@ def scale_in(self, blocks=None, block_ids=[], force=True, max_idletime=None):
...
-
         Returns
         -------
         List of job_ids marked for termination
         """
-        logger.debug(f"Scale in called, blocks={blocks}, block_ids={block_ids}")
-        if block_ids:
-            block_ids_to_kill = block_ids
+        logger.debug(f"Scale in called, blocks={blocks}")
+        managers = self.connected_managers()
```
khk-globus, line 635: "While you're here," and changing these lines anyway, do you want to continue on the path of not pre-formatting these log messages?  That is:
```python
logger.debug("Scale in called, blocks=%s", blocks)
```
Not a blocker, but a whittling away of that other task ... "while here anyway."

parsl/executors/high_throughput/executor.py
```diff
@@ -627,46 +627,40 @@ def scale_in(self, blocks=None, block_ids=[], force=True, max_idletime=None):
...
+        block_info = {}  # block id -> list( tasks, idle duration )
+        for manager in managers:
```
khk-globus, line 637: Not a change from the original code, but perhaps formalize the typing of `block_info`?  Again, "while here"?

khk-globus (approved):

parsl/executors/high_throughput/executor.py
benclifford: would be an interesting linting rule to write on a lazy sunday afternoon...

PR #2977, opened 2023-11-22
move dir fix parsl_coprocess.py

colinthomas-z80:
# Description

Issue resolving relative task directory

# Fixes

#2914 

## Type of change

- Bug fix

PR #3047, opened 2024-01-31
WQ coprocess: generate parsl_coprocess.py at runtime

### LINKED ISSUE #2914: Newly frequent WorkQueueTaskFailure in CI

colinthomas-z80:
# Description
Generate parsl_coprocess.py at runtime.

A more pragmatic solution for #2914 which was temporarily fixed by #2977, where the generation of this file remained a manual process. Now it will be generated by the current version of ndcctools. 

# Changed Behaviour

Using work queue coprocess will put parsl_coprocess.py in the user's directory at runtime. If there are better suggestions of where to store it I am open to them.

# Fixes

#2914

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -302,7 +302,7 @@ class DataFlowKernel:
             res = self._unwrap_remote_exception_wrapper(future)
 
         except Exception as e:
-            logger.debug("Task {} try {} failed".format(task_id, task_record['try_id']))
+            logger.info(f"Task {task_id} try {task_record['try_id']} failed with exception of type {type(e).__name__}")
             # We keep the history separately, since the future itself could be
             # tossed.
             task_record['fail_history'].append(repr(e))


benclifford:
ok I like this approach better. I had imagined it happening as part of the `pip` install procedure but the way its done in this PR seems fine too, I think.

Directorywise, maybe the right place is inside the particular executor instances working directory (alongside, for example, `"function_data"` ?

eg.
            self.function_data_dir = os.path.join(self.run_dir, self.label, "function_data")

benclifford:
this failed on a couple of CI tests with:

```
parsl.executors.errors.ExecutorError: Executor WorkQueueExecutor failed due to: Workqueue Submit Process is not alive
```

and digging into the downloadable logs:

```
1706718555.814392 2024-01-31 16:29:15 WorkQueue-Submit-Process-7549 MainThread-139731849144128 parsl.process_loggers:31 wrapped ERROR: Exceptional ending for _work_queue_submit_wait on thread MainThread
  File "/home/runner/work/parsl/parsl/parsl/process_loggers.py", line 27, in wrapped
    r = func(*args, **kwargs)
  File "/home/runner/work/parsl/parsl/parsl/executors/workqueue/executor.py", line 846, in _work_queue_submit_wait
    from ndcctools.poncho import package_serverize
  File "/tmp/cctools/lib/python3.8/site-packages/ndcctools/poncho/package_serverize.py", line 9, in <module>
    from ndcctools.poncho import package_create as create
  File "/tmp/cctools/lib/python3.8/site-packages/ndcctools/poncho/package_create.py", line 13, in <module>
    import conda_pack
ModuleNotFoundError: No module named 'conda_pack'
```

which I guess is a dependency problem...

colinthomas-z80:
conda-pack appears to be listed as a dependency in the parsl-workqueue documentation.

Would this be as simple as adding a "pip install conda-pack" in install-workqueue.sh ?

benclifford:
I noticed a problem to be aware of perhaps when upgrading from an earlier parsl: when I re-installed parsl in my dev env from your branch, it did not remove all traces of the old parsl_coprocess.py script, so I ended up with hanging work_queue_worker processes that when I ran them manually showed this:

```$ work_queue_worker --coprocess parsl_coprocess.py parsl-dev-3-11-3926 9000
work_queue_worker: creating workspace /tmp/worker-1000-104
work_queue_worker: using 4 cores, 15704 MB memory, 539 MB disk, 0 gpus
  File "/home/benc/parsl/virtualenv-3.11/bin/parsl_coprocess.py", line 6, in <module>
    with open(__file__) as f:
         ^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/benc/parsl/src/parsl/parsl/executors/workqueue/parsl_coprocess.py'
```

benclifford:
now in CI, I see this:

```
parsl.executors.errors.BadStateException: Executor WorkQueueExecutor failed due to: Error 1:
	EXIT CODE: 143
	STDOUT: Found cores : 4
Launching worker: 1

	STDERR: 2024/01/31 18:42:41.67 work_queue_worker[7954] fatal: could not resolve path `parsl_coprocess.py': No such file or directory
Terminated
```

which I guess is the paths not being set right to point to where this PR puts parsl_coprocess.py. In a shared-fs environment, I'd suggest using a full path here, but I'm not sure what that looks like when WQ is in a non-shared fs environment.

benclifford:
(before this PR, the install script put parsl_coprocess.py in the virtualenv bin directory, which is added to $PATH when you activate the virtualenv)

colinthomas-z80:
Right. So I can copy it into a shared location and it works, but now it is not packaged with the worker environment for non-shared fs runs... hmm

colinthomas-z80:
I do not think there is a good opportunity for us to generate the file at runtime and get it to the worker before it starts. I wish to turn my attention to the install script to see if it may be generated there so it may be included in the venv

benclifford:
> I do not think there is a good opportunity for us to generate the file at runtime and get it to the worker before it starts. I wish to turn my attention to the install script to see if it may be generated there so it may be included in the venv

ok. I don't really have any experience in that sort of Python build/packaging - I had a look a few months ago very briefly with this in mind but I didn't come up with anything. So I'm probably not going to be much help here.

PR #3050, opened 2024-02-02
Simplify curvezmq ctx destructor

rjmello:
## Description

The underlying `zmq.Context` object has its own `__del__` method, which calls `self.destroy()`, so we only need to close the auth thread.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/curvezmq.py
+++ b/parsl/curvezmq.py
@@ -49,9 +49,6 @@ class BaseContext(metaclass=ABCMeta):
         self.cert_dir = cert_dir
         self._ctx = zmq.Context()
 
-    def __del__(self):
-        self.destroy()
-
     @property
     def encrypted(self):
         """Indicates whether encryption is enabled.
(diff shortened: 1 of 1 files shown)


parsl/curvezmq.py
```diff
@@ -136,6 +133,12 @@ def __init__(self, cert_dir: Optional[Union[str, os.PathLike]]) -> None:
         if self.encrypted:
             self.auth_thread = self._start_auth_thread()
 
+    def __del__(self):
+        # Avoid issues in which the auth_thread attr was
+        # previously deleted
+        if getattr(self, "auth_thread", None):
+            self.auth_thread.stop()
+
```
khk-globus, line 141: This isn't on the base class; perhaps just:
```python
if self.auth_thread:
    self.auth_thread.stop()
```
On the other hand, I don't see the need for this method at all.  If it's just for cleanup purposes, then the thread is already stopped when this object is finalized.  If anything, if we want to manually stop the thread (which I don't have a problem with in theory), then I'd rather we implement a proper interface for these classes, such as a `.shutdown()` method, or `__enter__()`/`__exit__()` methods.

rjmello: The use of `getattr` is an extra precaution to avoid warnings ([destructor exceptions are ignored](https://docs.python.org/3/reference/datamodel.html#object.__del__)) in the unlikely event that the `auth_thread` attribute gets deleted.

The method is not strictly necessary, but the underlying `zmq.Context` object has it's own `__del__` method, which closes sockets via `self.destroy()`. We need stop the auth thread before attempting to close sockets.

benclifford:
This is less `__del__` than in #3030 so I am pretty much in favour of it. I don't have my head around the different ways in which this could go wrong, so we should look out for those.

benclifford (approved):

PR #3053, opened 2024-02-06
Better support in HTEX for reporting jobs that fail without connecting

yadudoc:
# Description

When it comes to an executor using a provider like SlurmProvider to provision resources, there are a few different failure modes that are not well supported.

1. Job fails to submit -> Reported as a provider failure
2. Job starts, workers fail to start with an non-zero exit code reported (LocalProvider) -> Reported as a provider failure
3. Job starts, but workers fail to start (bad env/network config) -> Silently ignored
4. Job starts, but terminates later due to reaching walltime -> Not a failure. 

In case3, when the batch system deems a task as completed/failed, the provider may not see further info reported from the queue. In such a situation the provider marks the job as `JobState.COMPLETED` instead of `JobState.FAILED`. However, this leads to a situation where Parsl would repeatedly attempt provisioning jobs when it should shut down.

# Changed Behaviour

This PR introduces a new `JobState.MISSING` which is set by the `HighThroughputExecutor` when it detects that a job reported by the provider to have reached a terminal state did not connect. 

Fixes # (issue)

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -14,7 +14,7 @@ import queue
 import threading
 import json
 
-from typing import cast, Any, Dict, NoReturn, Sequence, Set, Optional, Tuple
+from typing import cast, Any, Dict, NoReturn, Sequence, Set, Optional, Tuple, List
 
 from parsl import curvezmq
 from parsl.utils import setproctitle
@@ -184,6 +184,7 @@ class Interchange:
             self.worker_task_port, self.worker_result_port))
 
         self._ready_managers: Dict[bytes, ManagerRecord] = {}
+        self.connected_block_history: List[str] = []
 
         self.heartbeat_threshold = heartbeat_threshold
 
@@ -285,6 +286,9 @@ class Interchange:
                     for manager in self._ready_managers.values():
                         outstanding += len(manager['tasks'])
                     reply = outstanding
+
+                elif command_req == "CONNECTED_BLOCKS":
+                    reply = self.connected_block_history
 
                 elif command_req == "WORKERS":
                     num_workers = 0
@@ -416,6 +420,7 @@ class Interchange:
                                                         'worker_count': 0,
                                                         'active': True,
                                                         'tasks': []}
+                    self.connected_block_history.append(msg['block_id'])
                 if reg_flag is True:
                     interesting_managers.add(manager_id)
                     logger.info("Adding manager: {!r} to ready queue".format(manager_id))
--- a/parsl/jobs/states.py
+++ b/parsl/jobs/states.py
@@ -39,12 +39,19 @@ class JobState(IntEnum):
     HELD = 7
     """This job is held/suspended in the batch system"""
 
+    MISSING = 8
+    """This job has reached a terminal state without the resources(managers/workers)
+    launched in the job connecting back to the Executor. This state is set by HTEX
+    when it is able to infer that the block failed to start workers for eg due to
+    bad worker environment or network connectivity issues.
+    """
+
     def __str__(self) -> str:
         return self.__class__.__name__ + "." + self.name
 
 
 TERMINAL_STATES = [JobState.CANCELLED, JobState.COMPLETED, JobState.FAILED,
-                   JobState.TIMEOUT]
+                   JobState.TIMEOUT, JobState.MISSING]
 
 
 class JobStatus:
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -547,6 +547,10 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
         """
         return self.command_client.run("MANAGERS")
 
+    def connected_blocks(self) -> List[str]:
+        """List of connected block ids"""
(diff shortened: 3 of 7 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -411,11 +415,12 @@ def process_task_outgoing_incoming(self, interesting_managers: Set[bytes], hub_c
                     # We set up an entry only if registration works correctly
                     self._ready_managers[manager_id] = {'last_heartbeat': time.time(),
                                                         'idle_since': time.time(),
-                                                        'block_id': None,
+                                                        'block_id': msg['block_id'],
```
benclifford, line 418: this already happens on line 423/428: `m.update(msg)` I think? because it copies all the fields across from the registration message into the ready manager record. I don't think the pattern of "initialise everything blank here, then update it 10 lines later" needs to be specially different for the block id compared to the other fields in the manager record.

yadudoc: This change is unrelated to what is going on in this PR anyway, so I'll remove this.

parsl/jobs/states.py
```diff
@@ -39,12 +39,18 @@ class JobState(IntEnum):
     HELD = 7
     """This job is held/suspended in the batch system"""
 
+    MISSING = 8
+    """Scheduler is no longer tracking this job most likely due to it exiting.
```
benclifford, line 43: i feel like this text doesn't adequately describe what `MISSING` is meant to represent. This text also describes jobs that have gone into COMPLETED state because they disappeared from the scheduler and were then set to COMPLETED by the relevant provider, I think... MISSING gives more context to the state beyond it disappearing from the scheduler that is, at least in the htex case, to do with "we know this didn't start properly" and we have a different failure signal that is different from the scheduler.

yadudoc: Updated docstring.

yadudoc:
@benclifford I ran some quick tests on frontera with a bad `worker_init` setting to simulate broken python environment and I see the right exception being set:

```
ERROR    parsl.executors.status_handling:status_handling.py:141 Setting bad state due to exception
parsl.jobs.errors.TooManyJobFailuresError: Error 1:

INFO     parsl.dataflow.dflow:dflow.py:305 Task 0 try 0 failed with exception of type BadStateException
ERROR    parsl.dataflow.dflow:dflow.py:351 Task 0 failed after 0 retry attempts
  File "/home1/02551/yadunand/parsl/parsl/dataflow/dflow.py", line 302, in handle_exec_update
  File "/home1/02551/yadunand/parsl/parsl/dataflow/dflow.py", line 570, in _unwrap_remote_exception_wrapper
    result = future.result()
             ^^^^^^^^^^^^^^^
  File "/home1/02551/yadunand/anaconda3/envs/parsl_py311/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/home1/02551/yadunand/anaconda3/envs/parsl_py311/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
parsl.executors.errors.BadStateException: Executor HTEX failed due to: Error 1:
```

benclifford:
```
ERROR    parsl.executors.status_handling:status_handling.py:141 Setting bad state due to exception
parsl.jobs.errors.TooManyJobFailuresError: Error 1:
```

After that "error <n>:" text there's an opportunity for job status to put in an exception/message. Look at the example messages in this PR https://github.com/Parsl/parsl/pull/3043 which result from provider `submit` errors, for example.

I think this is an obvious place to make this error something useful to the user: is there any captured stderr from the job? There should be in many cases (eg. a Python process starting up and then printing out that it cannot find a module, or the shell printing out that it cannot find Python).

That text is forming in parsl.jobs.error_handlers._get_error, which relies on the error set in the `JobStatus` object. 

The way things are structured now, I think that in-job stdout/stderr is probably awkward to get to, but the error message here could at least describe whats going on in a static message about jobs maybe starting and where to look for more info.

yadudoc:
It looks like there's an issue with Providers not reporting sufficient info with the JobStatus object, I've added a new issue here -> https://github.com/Parsl/parsl/issues/3058. As a temporary fix, I have updated the `htex.status` method to add a message to point users to where the logs are when the exception message is empty. The exception report looks like this now:

```
ERROR    parsl.dataflow.dflow:dflow.py:351 Task 0 failed after 0 retry attempts
  File "/home1/02551/yadunand/parsl/parsl/dataflow/dflow.py", line 302, in handle_exec_update
  File "/home1/02551/yadunand/parsl/parsl/dataflow/dflow.py", line 570, in _unwrap_remote_exception_wrapper
    result = future.result()
             ^^^^^^^^^^^^^^^
  File "/home1/02551/yadunand/anaconda3/envs/parsl_py311/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/home1/02551/yadunand/anaconda3/envs/parsl_py311/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
parsl.executors.errors.BadStateException: Executor HTEX failed due to: Error 1:
```

PR #3054, opened 2024-02-07
Update monitoring task_inputs after input Futures are resolved

benclifford:
The monitoring task table stores a column that should store the inputs argument of a task, after Future resolution.

Prior to this PR:

Although the sending side of the code, DataFlowKernel._send_task_log_info, is correctly invoked to send that information, it first sends an earlier representation of that task where inputs are still unresolved futures - for example:

[<AppFuture at 0x7fca83d22610 state=pending>, <AppFuture at 0x7fca83d7b550 state=pending>, <AppFuture at 0x7fca83d80110 state=pending>]

When a later update is sent, not all columns of the task table are updated; only a specific list of columns that are expected to change. task_inputs was not on that list, so even after future resolution, the monitoring task_inputs column was not updated.

This PR adds task_inputs to the list of columns that are updated on every task update message.

This PR adds a test which if run without this change results in this test failure:

```
            result = connection.execute(text("SELECT task_inputs FROM task"))
            (task_inputs, ) = result.first()
>           assert task_inputs == "[" + repr(TOKEN) + "]"
E           AssertionError: assert '[<Future at ...ate=pending>]' == '[54739]'
E             - [54739]
E             + [<Future at 0x7f946e3f9010 state=pending>]

parsl/tests/test_monitoring/test_incomplete_futures.py:66: AssertionError

```

# Changed Behaviour

Better info in monitoring db for tasks with `inputs=` parameters.

## Type of change

- Bug fix


parsl/tests/test_monitoring/test_incomplete_futures.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if os.path.exists("runinfo/monitoring.db"):
+        logger.info("Monitoring database already exists - deleting")
+        os.remove("runinfo/monitoring.db")
+
+    logger.info("loading parsl")
+    parsl.load(fresh_config())
+
+    # this is arbitrary
+    TOKEN = 54739
```
khk-globus, line 30: If arbitrary, then may I suggest `random.randint()`?

parsl/tests/test_monitoring/test_incomplete_futures.py
```diff
@@ -0,0 +1,68 @@
...
+    with engine.begin() as connection:
+        result = connection.execute(text("SELECT COUNT(*) FROM task"))
+        (task_count, ) = result.first()
+        assert task_count == 1
+
+        result = connection.execute(text("SELECT task_inputs FROM task"))
+        (task_inputs, ) = result.first()
+        assert task_inputs == "[" + repr(TOKEN) + "]"
+
+    logger.info("all done")
```
khk-globus, line 68: No need for this logger call once test is merged to the regular CI.  A dev who's actively working on the test can put it back in while developing, if necessary, but no sense in emitting it every time when no one's watching.

parsl/tests/test_monitoring/test_incomplete_futures.py
```diff
@@ -0,0 +1,68 @@
...
+    assert f2.result() == TOKEN
+
+    # this cleanup gives a barrier that allows the monitoring code to store
+    # everything it has in the database - without this, there's a race
+    # condition that the task will not have arrived in the database yet.
+    # A different approach for this test might be to poll the DB for a few
+    # seconds, with the assumption "data will arrive in the DB within
+    # 30 seconds, but probably much sooner".
+    parsl.dfk().cleanup()
+    parsl.clear()
```
khk-globus, line 56: From a "stop wasting the human's time" perspective, I **much** prefer this "event-based" approach, rather than a timeout approach.

parsl/tests/test_monitoring/test_incomplete_futures.py
```diff
@@ -0,0 +1,68 @@
...
+    # potential race condition:
+    # monitoring will take some time to update the record after submitting the task or completing the Future
+
+    f1 = Future()
+
+    f2 = this_app(inputs=[f1])
+
+    f1.set_result(TOKEN)
+
+    assert f2.result() == TOKEN
```
khk-globus, line 47: Is there any value in checking the intermediate state of the monitoring DB between these statements?

khk-globus (approved):

parsl/tests/test_monitoring/test_incomplete_futures.py
benclifford: it's a bit non-deterministic so I didn't go that way - after line 43, the task row will appear (with an in-progress value) "eventually" in the monitoring database, but it's not synchronous with line 43 completing.

What I would expect is that various records would appear - basically the asserts that are in eg. PR #3056 - but I think it would be some code that has to sit there and poll the DB a few times until the "eventually consistent" DB reflects that the task is launched.

The interesting thing to check for the specific bug in this PR would be to check that the input field does not have `TOKEN` stored in it until after line 45, as a pre-condition.

That's also what could happen instead of shutting down the DFK with `cleanup` a few lines later: don't shut down parsl here, but instead poll the DB until it looks right.

I don't really have any strong feelings about whether it's the right thing to do or not.

PR #3056, opened 2024-02-07
Test app names are stored in monitoring under various decoration patterns

benclifford:
This follows some investigations of app names that turned out to be a bug in a different piece of code, resolved in
https://github.com/LSSTDESC/gen3_workflow/pull/58

## Type of change

- Code maintenance/cleanup


parsl/tests/test_monitoring/test_app_names.py
```diff
@@ -0,0 +1,99 @@
...
+    # this is imported here rather than at module level because
+    # it isn't available in a plain parsl install, so this module
+    # would otherwise fail to import and break even a basic test
+    # run.
+    import sqlalchemy
+
+    if os.path.exists("runinfo/monitoring.db"):
+        logger.info("Monitoring database already exists - deleting")
+        os.remove("runinfo/monitoring.db")
+
```
khk-globus, line 61: We have `tmpd_cwd` now, so please consider using that fixture instead of this piecemeal checking and deleting.

parsl/tests/test_monitoring/test_app_names.py
```diff
@@ -0,0 +1,99 @@
...
+    logger.info("Generating fresh config")
+    c = fresh_config()
+    logger.info("Loading parsl")
+    parsl.load(c)
+
+    logger.info("invoking and waiting for result")
```
khk-globus, line 67: From the PoV of an automated test, what's the value-add of leaving these logger calls here?  At the point the test fails, `pytest` shows the exact line of failure and or AssertionError, no?

parsl/tests/test_monitoring/test_app_names.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # ... that was tried once ...
+        result = connection.execute("SELECT COUNT(*) FROM try")
+        (c, ) = result.first()
+        assert c == 1
+
+        # ... and has the expected name.
+        result = connection.execute("SELECT task_func_name FROM task")
+        (c, ) = result.first()
+        assert c == expected_name
```
khk-globus, line 97: If you wanted, could make this a loop:
```python
for expected, sql_text, assert_msg in (
    (1, "SELECT COUNT(*) FROM workflow"),
    (1, "SELECT COUNT(*) FROM task"),
    (1, "SELECT COUNT(*) FROM try"),
    (expected_name, "SELECT task_func_name FROM task"),
):
    c, *_ = connection.execute(sql_test).first()
    assert c == expected, "Expect one workflow, one task, tried once, with expected name"
```

parsl/tests/test_monitoring/test_app_names.py
```diff
@@ -0,0 +1,99 @@
...
+    logger.info("invoking and waiting for result")
+    app = get_app()
+    assert app().result() == expected_result
+
+    logger.info("cleaning up parsl")
+    parsl.dfk().cleanup()
+    parsl.clear()
+
+    logger.info("checking database content")
+    engine = sqlalchemy.create_engine("sqlite:///runinfo/monitoring.db")
```
khk-globus, line 76: We're starting to use this "magic number" in a number of spots.  Perhaps not for this PR, but would be nice to reference a single definition somewhere.

benclifford: there's a more general factoring possibility I don't have my head round with the style of monitoring tests in general - tests I've written (not all committed) often look roughly like:

i) start some config that has monitoring
 - with a fresh monitoring db so that all thats in there is this test
ii) run a little bit of workflow - different per test
iii) shut down parsl - this is a barrier so that async monitoring information has "definitely" arrived in the DB
iv) make some assertions about that database, different per test (the motivation for starting with a clean DB is that because we started with a clean DB, we can be quite loose in queries, not having to select the particular workflow we just ran, for example)

There's a lot of boilerplate in there around the "run this workflow" "assert what monitoring saw" steps.

parsl/tests/test_monitoring/test_app_names.py
benclifford: fixed in 8d03fabc483b70cd436c57e80ddb7bf9d131aef4

parsl/tests/test_monitoring/test_app_names.py
```diff
@@ -0,0 +1,86 @@
...
+    parsl.dfk().cleanup()
+    parsl.clear()
+
+    engine = sqlalchemy.create_engine(c.monitoring.logging_endpoint)
+    with engine.begin() as connection:
+
+        def count_rows(table: str):
+            result = connection.execute("SELECT COUNT(*) FROM workflow")
+            (c, ) = result.first()
+            return c
```
khk-globus, line 72: Am I missing where `table` is utilized?

benclifford: should be fixed in f5e428013f

khk-globus (approved):
Och; I'm sorry.  Between getting sick and this getting pushed farther down the stack, I obviously forgot about it.

PR #3060, opened 2024-02-08
Fix two bugs and add support for N masters, N workers with load balancing in RadicalPilotExecutor

### LINKED ISSUE #3029: Test failures from Radical pilot executor
**Describe the bug**

I noticed some instability in the radical pilot executor that caused failures in the CI tests for the `mpi_experimental_3` branch. These failures are intermittent so rerunning failed tasks gets past these errors, but it would be good to have these tests or the cause of instability fixed.

Here's a small snippet of the failure from github actions (https://github.com/Parsl/parsl/actions/runs/7479805965/job/20357787240#step:8:2239)

```
.venv/lib/python3.9/site-packages/typeguard/__init__.py:1033: in wrapper
    retval = func(*args, **kwargs)
parsl/dataflow/dflow.py:1428: in load
    cls._dfk = DataFlowKernel(config)
.venv/lib/python3.9/site-packages/typeguard/__init__.py:1033: in wrapper
    retval = func(*args, **kwargs)
parsl/dataflow/dflow.py:186: in __init__
    self.add_executors(config.executors)
parsl/dataflow/dflow.py:1130: in add_executors
    block_ids = executor.start()
parsl/executors/radical/executor.py:244: in start
    pilot = self.pmgr.submit_pilots(pd)
.venv/lib/python3.9/site-packages/radical/pilot/pilot_manager.py:588: in submit_pilots
    pilot = Pilot(pmgr=self, descr=pd)
.venv/lib/python3.9/site-packages/radical/pilot/pilot.py:120: in __init__
    self._endpoint_fs      = self._session._get_endpoint_fs     (pilot)
.venv/lib/python3.9/site-packages/radical/pilot/session.py:1539: in _get_endpoint_fs
    resource_sandbox  = self._get_resource_sandbox(pilot)
.venv/lib/python3.9/site-packages/radical/pilot/session.py:1410: in _get_resource_sandbox
    shell = self.get_js_shell(resource, schema)
.venv/lib/python3.9/site-packages/radical/pilot/session.py:1456: in get_js_shell
    shell = rsup.PTYShell(js_url, self)
.venv/lib/python3.9/site-packages/radical/saga/utils/pty_shell.py:236: in __init__
    self.pty_info   = self.factory.initialize (self.url,    self.session,
.venv/lib/python3.9/site-packages/radical/saga/utils/pty_shell_factory.py:206: in initialize
    self._initialize_pty(info['pty'], info)
.venv/lib/python3.9/site-packages/radical/saga/utils/pty_shell_factory.py:427: in _initialize_pty
    raise ptye.translate_exception (e) from e
.venv/lib/python3.9/site-packages/radical/saga/utils/pty_shell_factory.py:269: in _initialize_pty
    pty_shell.write(" export PROMPT_COMMAND='' PS1='$';"
.venv/lib/python3.9/site-packages/radical/saga/utils/pty_process.py:858: in write
    raise ptye.translate_exception(e, "write failed (%s)" % e) \

self = <radical.saga.utils.pty_process.PTYProcess object at 0x7f9d58bb34f0>
data = " export PROMPT_COMMAND='' PS1='$'; set prompt='$'\n", nolog = False

    def write (self, data, nolog=False) :
        """
        This method will repeatedly attempt to push the given data into the
        child's stdin pipe, until it succeeds to write all data.
        """
    
        with self.rlock :
    
            if not self.alive (recover=False) :
                raise ptye.translate_exception(
                        se.NoSuccess("cannot write to dead process (%s) [%5d]"
                                     % (self.cache[-256:], self.parent_in)))
    
            try :
                log = self._hide_data (data, nolog)
                log =  log.replace ('\n', '\\n')
                log =  log.replace ('\r', '')
                if  len(log) > _DEBUG_MAX :
                    self.logger.debug ("write: [%5d] [%5d] (%s ... %s)"
                                    % (self.parent_in, len(data),
                                       log[:30], log[-30:]))
                else :
                    self.logger.debug ("write: [%5d] [%5d] (%s)"
                                    % (self.parent_in, len(data), log))
    
                # attempt to write forever -- until we succeeed
                while data :
    
                    # check if the pty pipe is ready for data
>                   _, wlist, _ = select.select ([], [self.parent_in], [],
                                                 _POLLDELAY)
E                                                ValueError: filedescriptor out of range in select()

.venv/lib/python3.9/site-packages/radical/saga/utils/pty_process.py:842: ValueError
```

 

**To Reproduce**
1. Rerun tests in CI until you see this failure, this is an intermittent failure

**Expected behavior**
Tests passing without instability

**Environment**
 - Github Actions
 - branch: mpi_experimental_3

It would be nice if someone from the radical team, perhaps @AymenFJA, could take a quick look.

AymenFJA:
# Description

This PR:

- Fix issue #3029
- Fix issue #3070, which is not terminating due to internal component failure.
- Add support to N masters and N workers with task load balancing across the masters.

# Fixes

Fixes #3029, #3070

## Type of change

- Bug fix: #3029
- Bug fix: Termination triggers on internal master and worker failures.
- New feature: N masters and N workers support with load balancing.
- Code maintenance/cleanup


parsl/executors/radical/executor.py
```diff
@@ -186,6 +194,23 @@ def task_state_cb(self, task, state):
...
+        and sets the provided exception as their result, indicating
+        a failure.
+
+        Parameters:
+        - exception: The exception to be set as the result for all
+                     outstanding tasks.
+        """
+        for fut_task in self.future_tasks.values():
+            if not fut_task.done():
+                fut_task.set_exception(exception)
```
benclifford, line 212: maybe there's a very slight race condition here - can fut_task get set by another thread (eg. task completion) in the space between line 211 and line 212? If so, set_exception would raise an exception here. I've seen a few situations like that elsewhere in Parsl.

benclifford:
ok thanks. there are a bunch of changes in this and I'm not super familiar with this bit of the code, so I'll try to use the review as a learning experience. If you're making several changes at once, it can be easier both for review and tracking down later bugs to make several smaller, single focused PRs (a rule of thumb for me is, if you start listing out multiple changes in the PR description then consider splitting it...)

AymenFJA:
> ok thanks. there are a bunch of changes in this and I'm not super familiar with this bit of the code, so I'll try to use the review as a learning experience. If you're making several changes at once, it can be easier both for review and tracking down later bugs to make several smaller, single focused PRs (a rule of thumb for me is, if you start listing out multiple changes in the PR description then consider splitting it...)

@benclifford totally understandable and I agree. In the future, the PRs will be more focused. During this review please let me know if I can help you with anything vague or any questions. Thank you and sorry for the late response..

Makefile
```diff
@@ -10,6 +10,7 @@ export PATH := $(CCTOOLS_INSTALL)/bin/:$(PATH)
 export CCTOOLS_VERSION=7.7.2
 export HYDRA_LAUNCHER=fork
 export OMPI_MCA_rmaps_base_oversubscribe=yes
+export RADICAL_TEMP_SANDBOX=True
```
benclifford, line 13: i think this variable is not interpreted as a Python bool so it's maybe misleading to use `True` here in case someone thinks `False` would do something opposite.

parsl/executors/radical/executor.py
```diff
@@ -202,63 +227,67 @@ def start(self):
                    'resource': self.resource}
 
         if not self.resource or 'local' in self.resource:
-            # move the agent sandbox to the working dir mainly
-            # for debugging purposes. This will allow parsl
-            # to include the agent sandbox with the ci artifacts.
-            if os.environ.get("LOCAL_SANDBOX"):
-                pd_init['sandbox'] = self.run_dir
+            # set the agent dir to /temp as a fix for #3029
```
benclifford, line 230: Is this changed sandbox behaviour the only part that fixes #3029?

If so:

it's moving things into a temporary directory only when the CI tests are run (or in other situations where the `RADICAL_TEMP_SANDBOX` environment variable is set).

in non-test situations, this behaviour is otherwise unchanged, right? If so, does this PR stop that situation occurring for normal users who are running real applications, not our CI?

@AymenFJA

benclifford: I investigated this a bunch more (and have commented on what I see in other issues) to try to understand what was actually happening in the CI environment, because I like to at least have some understanding of how that environment behaves.

I can see why this is happening in the CI (due to entering a strange zone of fd number-space which radical.saga cannot deal with) and therefore why this would happen in a user workflow (when a user's workflow submitting process has opened many - around 1000 - files).

This PR avoid the misbehaving radical.saga code path by having a special non-default option to turn on a local sandbox.

But in https://github.com/radical-cybertools/radical.saga/issues/885 Andrew mentions, effectively, that radical.saga would be better replaced by psij-python - that would have the same effect of removing the misbehaving codepath, in a different way.

So I tried that out in this draft PR: https://github.com/Parsl/parsl/pull/3079 which doesn't actually avoid the misbehaving codepath, as far as I can tell.

But maybe this is a better approach longer term, if radical.saga is effectively deprecated in favour of psij-python?

benclifford:
I just merged the latest parsl `master` into this branch - it needed a couple of manual conflict resolutions due to PR #3421, which re-ordered a lot of import statements - re-ordered import statements are hard for git to merge automatically.

parsl/executors/radical/rpex_resources.py
```diff
@@ -149,12 +151,16 @@ def _get_cfg_file(cls, path=None):
...
         if path:
+            config_path = 'rpex.cfg'
             config_path = path + '/' + config_path
-        with open(config_path, 'w') as f:
-            json.dump(cfg, f, indent=4)
-        return config_path
+            with open(config_path, 'w') as f:
+                json.dump(cfg, f, indent=4)
+        else:
+            config_obj = ru.Config(from_dict=cfg)
```
benclifford, line 165: the import of `ru` here was removed by PR #3454 - I'm not sure on what should be done here to fix that.

Makefile
benclifford: I think everything to do with RADICAL_TEMP_SANDBOX can be removed from this PR now, because I think the problem it was solving was fixed in a different way by moving away from radical SAGA?

benclifford (approved):

Issue #3059, opened 2024-02-07
Draining HTEX workers near end of walltime


benclifford:
**Is your feature request related to a problem? Please describe.**

At present, htex is not aware of worker walltime limits imposed by a batch system: what HTEX sees is that workers are killed, just like they would be killed for other reasons (such as out-of-memory situations).

Those tasks then can be retried by parsl's retry mechanism - sometimes they are retried immediately on the worker that is part-way through being killed and fail again immediately.

It's natural to ask "can you make Parsl not send tasks to workers near the end of their walltime?".

The work queue executor has such a facility: a worker can be told a walltime limit, tasks can be annotated with a walltime, and Work Queue will not schedule tasks onto workers without sufficient remaining walltime.

Some HTEX users would like this sort of behaviour for htex.

What options are available?

In the based htex model, htex does not pay attention to any kind of resource beyond a collection of n x "indiivdual worker available right now".

I think it could be straightforward in that model to implement a less expressive model:
managers can decide they are expiring "soon" and tell the interchange to stop sending tasks - for example, by setting their total worker count to 0 in their registration. Managers could decide in two different ways: they could be told a walltime to expect to live; or the batch system could tell them via a Unix signal that they will be killed soon. This model does not accomodate tasks of varying expected walltime: the manager would decide it's near the end and stop accepting new tasks of any kind. (compare to the Work Queue model where a worker nearing end of life will ramp down the required walltime for new tasks but can continue running short tasks up until the very end).

Unmerged PR #3016 introduces a notion of per-task resource requirement - in that PR, that resource requirement is only a count of required MPI nodes - and so introduces the notion that some tasks cannot be scheduled on what is available now, but might be schedulable later, and that despite that, some other tasks might be schedulable now.

It's possible that this #3016 framework could be extended to other resource types. In the context of this current issue, that would be remaining walltime required by a task (but more generally might include dynamically allocating GPU accelerators rather than having them fixed per task, or number of CPU execution units (cores, hyperthreads, ...)).

**Additional context**
For University of Ghent

benclifford:
crossref https://github.com/Parsl/parsl/issues/3323 which is related work

PR #3063, opened 2024-02-09
Drain workers as walltime expiry nears

benclifford:
# Description

This is an implementation of the first part (the "easy part") of issue #3059. 

It adds a parameter to HighThroughputExecutor specifying a "not quite as long as walltime" parameter, after which time workers will drain themselves: they will continue with existing tasks but ask the interchange to not send any more.

When they are drained, the worker pools will exit immediately.

In some launcher environments, the exit of any one worker causes the batch job to end, killing all the other workers - so I'll make this configurable to not exit (although there won't be any feedback further into the block scaling system to cause that to terminate this block, so it will sit idle until the batch job is over) 

# Changed Behaviour

This is new functionality. If you don't enable it, things should behave the same as before.

## Type of change

- New feature


yadudoc:
I like the general approach in the PR and this is a reasonable fix to avoid scheduling tasks close to walltime if the app walltime is known. For many workflows with a single app type, this would work. However, if there is a mix of short and long tasks, determining the right `drain_period` is not clear. Perhaps this isn't meant to solve the problem of placing tasks on a single batch job instead of spreading evenly so that batch jobs can be scaled down towards the tail end of a campaign and that needs to be handled separately.

svandenhaute:
> When they are completely drain, they won't do anything except sit waiting for the batch system to enforce walltime. It would be possible to make workers exit as they become empty, but there are some subtleties here I think: at least: in some launcher environments, the exit of any one worker causes the batch job to end, killing all the other workers. There might be a race condition between a worker asking for 0 tasks, becoming empty by chance, but then receiving another task (which then fails) - this is more likely in a low workers-per-manager configuration, I think, because a manager will be completely empty at every single task completion.

In which launcher environments is this the case? AFAIK, only a manager exit could ever trigger the batch job to end?
Why would they wait for the batch system to enforce walltime instead of just having the manager  terminate gracefully when all its workers are idle?

benclifford:
@svandenhaute i just pushed some exiting behaviour (currently not configurable) - when a manager is fully drained, it will exit. How that interacts with the enclosing resource manager will vary with resource manager configuration/general behaviour... in a single manager-per-block case, this should be fine in all situations.

i plan on adding configuration for this next.

parsl/executors/high_throughput/executor.py
```diff
@@ -196,6 +197,13 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
         Timeout period to be used by the executor components in milliseconds. Increasing poll_periods
         trades performance for cpu efficiency. Default: 10ms
 
+    drain_period : int
+        Number of seconds after start, after which workers will drain and then exit. Set this to a
+        time that is slightly less than the maximum walltime of batch jobs to trade off not using
+        the full walltime vs tasks being killed as they are executing when the walltime expires.
+        The time set here should be something like the walltime minus some time needed for the
+        batch job to start the workers, minus the expected maximum length of an individual task.
+
```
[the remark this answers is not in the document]
benclifford: I switched to your phrasing

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -312,6 +335,14 @@ def pull_tasks(self, kill_event):
                 self.heartbeat_to_incoming()
                 last_beat = time.time()
 
+            if self.drain_time and time.time() > self.drain_time:
+                logger.info("Requesting drain")
+                self.drain_to_incoming()
+                self.drain_time = None
+                # This will start the pool draining...
+                # Drained exit behaviour does not happen here. It will be
+                # driven by the interchange sending a DRAINED_CODE message.
```
rjmello, line 344: Any thoughts on passing `drain_period` to the interchange and moving this logic there to avoid passing messages back-and-forth?

benclifford: The original structure of the "please drain me now" (workers->interchange) pattern comes from a potential extension that is you don't know a drain period ahead of time, but get told that by your batch system as a signal - so that the "please drain me" message would be requested due to a Mysterious External Event rather than a timer (you tell slurm: "give me a signal a few minutes before my batch job ends").

In the end I didn't implement that but I think I'd like to keep that as an implementation possibility to support a couple of potential use cases: submitting batch jobs with different characteristics (see issue #3113) and supporting pre-emptive queues (where a user submits a batch job with "I don't care how long I run for, just tell me before you terminate me" in exchange for some other valuable resource - for example 50% charging discount at NERSC in the `preempt` queue - https://docs.nersc.gov/jobs/policy/)

I think there has to be a "you are drained, you can exit now" message (interchange->workers) because that's how the worker pool knows it can exit. It can't use a "exit when empty" decision because because a pool will often be empty: for example, a 1 worker pool will be empty between every single task.

rjmello: That works for me.

rjmello (approved):
:guitar:

PR #3066, opened 2024-02-11
Fix address field in TaskVine manager config

benclifford:
Previously it looks like this field was ignored, because it was always (in the absence of a project name) overridden by
parsl.addresses.get_any_address. This PR makes that override only happen when the address field is set to None.

Previously this field had a class level default value of socket.gethostname() That value was never used, because of the above override, and resulted in the hostname of the documentation build host being listed as the default value for this parameter, incorrectly.

This PR removes that default value, allows the address field to be None, and makes the address field be None by default.

Cross-reference this Debian documentation reproducibility issue where the presence of socket.gethostname() breaks binary reproducibility: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1063542

Testing:

Prior to this PR, setting address='10.10.10.10' does not break anything, even though that is not an IP address of my test environment. After this PR, setting that IP address causes taskvine to hang as I would expect.

## Type of change

- Bug fix
- Code maintenance/cleanup


benclifford:
I tried this with different hostnames, the reproducibility bug reported in the above debian issue, and I get what I think is a binary-identical build: the only file that differs is this pickle file, which i think is fine?

$ diff _build/ ~/tmp/b3 -r
Binary files _build/doctrees/environment.pickle and /home/benc/tmp/b3/doctrees/environment.pickle differ

benclifford:
@tphung3 or maybe @colinthomas-z80 can you review this as far as changing taskvine functionality goes?

colinthomas-z80:
I believe the manager_config address only serves to override the address given to the factory for the workers to reach out to. So if an erroneous address is set it would only misguide the workers deployed by the factory.

benclifford:
@colinthomas-z80 is that a positive or negative review of this PR?

colinthomas-z80:
It is a positive change to error on a bad address and I do not see any issues with functionality.

khk-globus (approved):

rjmello (approved):

PR #3069, opened 2024-02-12
Renaming parsl_resource_spefication options for MPI Mode

yadudoc:
# Description

This PR adds the following renaming changes to the `parsl_resource_specification` options defined in the #3016 

* `NUM_NODES` -> `num_nodes`
* `LAUNCHER_OPTIONS` -> `launcher_options`
* `RANKS_PER_NODE` -> `ranks_per_node`
* New kwarg `num_tasks`

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/mpi_prefix_composer.py
+++ b/parsl/executors/high_throughput/mpi_prefix_composer.py
@@ -25,13 +25,19 @@ def validate_resource_spec(resource_spec: Dict[str, str]):
         is invalid (e.g, contains invalid keys)
     """
     user_keys = set(resource_spec.keys())
-    legal_keys = set(("RANKS_PER_NODE",
-                      "NUM_NODES",
-                      "LAUNCHER_OPTIONS",
+    legal_keys = set(("ranks_per_node",
+                      "num_nodes",
+                      "num_ranks",
+                      "launcher_options",
                       ))
     invalid_keys = user_keys - legal_keys
     if invalid_keys:
         raise InvalidResourceSpecification(invalid_keys)
+    if "num_nodes" in resource_spec:
+        if not resource_spec.get("num_ranks") and resource_spec.get("ranks_per_node"):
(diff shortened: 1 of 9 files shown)


parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -25,13 +25,19 @@ def validate_resource_spec(resource_spec: Dict[str, str]):
...
+                      "num_nodes",
+                      "num_tasks",
+                      "launcher_options",
                       ))
     invalid_keys = user_keys - legal_keys
     if invalid_keys:
         raise InvalidResourceSpecification(invalid_keys)
+    if "num_nodes" in resource_spec:
+        if not resource_spec.get("num_tasks") and resource_spec.get("ranks_per_node"):
+            resource_spec["num_tasks"] = resource_spec["num_nodes"] * resource_spec["ranks_per_node"]
```
benclifford, line 38: so there's a conflict of some kind if a user also specifies num_tasks, right?  - which in this case is that the user's num_tasks will be (silently) ignored?

benclifford: (more generally, it's not clear without turning my brain on, and so won't be clear to users immediately, which of the 8 combinations of having/not having these parameters are "sensible" vs "not sensible")

benclifford:
i think this is fine to merge. there's a bit of fuzziness between use of the word "rank" and the word "task" in these parameters (and the associated variables) right?

benclifford (approved):

benclifford:
(probs should sort out those type errors though)

parsl/executors/high_throughput/mpi_prefix_composer.py
yadudoc: That logic is supposed to set `num_tasks` and `ranks_per_node` if one is not set but the other is. I'm only using `num_tasks` because it is the kwarg used by flux. I feel like it is better to either call it `ranks` or `tasks`.

PR #3073, opened 2024-02-14
Refresh docs and types for usage tracking code

benclifford:
## Type of change

- Update to human readable text: Documentation/error messages/comments
- Code maintenance/cleanup

PR #3075, opened 2024-02-14
Correct return docstring of UsageTracker.send_message

benclifford:
## Type of change


benclifford:
this should be part of #3073 - i'll add it there.

PR #3076, opened 2024-02-15
test for looking at radical vs CI behaviour

benclifford:
not for merge, just for messing with CI environment


benclifford:
I don't need this PR any more

PR #3085, opened 2024-02-18
Remove sudo advice for inexperienced people

benclifford:
If you don't understand Python dependency errors, you almost definitely should not be trying to fix them as root.

## Type of change


khk-globus (approved):
Och; quite wise.  Thanks for this.

PR #3087, opened 2024-02-19
Update style of scaling test File usage

benclifford:
In this particular test configuration, this doesn't really matter: all that is being passed around is a path wrapped in an object; no staging happens; and if staging was configured, this test would usually not work... so it's arguable even that File shouldn't be used here at all and just pass around paths.

It is also not necessary to cast an os.PathLike to a str before constructing a File around it.

# Changed Behaviour

Neither of these changes should change the behaviour of this test, but its probably better that they are in the right style in case someone looks at them.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_scaling/test_scale_down.py
+++ b/parsl/tests/test_scaling/test_scale_down.py
@@ -44,16 +44,16 @@ def local_config():
 
 
 @python_app
-def waiting_app(ident: int, inputs=()):
+def waiting_app(ident: int, inputs=(), outputs=()):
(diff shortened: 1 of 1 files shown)


khk-globus (approved):

khk-globus:
*nod; a good changeset with a number of little embedded details.  (How'd you run across this?)

benclifford:
> (How'd you run across this?)

making the same test but for htex partial scaledown

Issue #2532, opened 2023-01-09
Usage tracking does not send end message


benclifford:
**Describe the bug**

The usage tracking code tries to send an end message, in addition to a start message. But only the start message is delivered.

Using `tcpdump -i any -n port 50077` I see this single packet at the start of a run, but not one at the end.

```
14:05:22.540668 wlp0s20f3 Out IP 192.168.2.114.50619 > 52.3.111.203.50077: UDP, length 227
```

I think this is because the usage tracking code sends packets asynchronously, to avoid hangs, but also kills outstanding usage tracking processes immediately after spawning one to send a packet. (so maybe in some situations, an end packet *will* be non-deterministically sent).

**To Reproduce**
Turn on usage tracking; run tcpdump as above. Observe a packet at the start of a run (eg pytest) but not at the end

**Expected behavior**
Packets should be sent at both the start and the end

**Environment**
parsl master 5e8a0e62d7e840e6bdadd7fffc8ea4d09cba4533
my laptop

benclifford:
fixed by #3088

PR #3088, opened 2024-02-19
Allow a few seconds for the end message usage process to work

benclifford:
Before this PR, the end message process was terminated immediately, which mean that end messages (in all of my tests) were not sent.

This PR waits 10 seconds for termination before forcing a (now stronger) kill, which in practice is enough for the end message to be sent.

Reasons for making that timeout any larger are primarily DNS-resolution-time related - the tracking DNS record has a TTL of only a few hundred seconds, so it's likely to need re-resolving at end time in any non-trivial run.

In practice on a small pytest run on my laptop, with DNS cached, this new close procedure takes 20ms.

# Changed Behaviour

More information is delivered to the usage tracking server - the end message contains different information to the start message

# Fixes

Fixes issue #2532 

## Type of change

- Bug fix

### DIFF
--- a/parsl/usage_tracking/usage.py
+++ b/parsl/usage_tracking/usage.py
@@ -32,7 +32,7 @@ def async_process(fn: Callable[P, None]) -> Callable[P, None]:
 
 
 @async_process
-def udp_messenger(domain_name: str, UDP_IP: str, UDP_PORT: int, sock_timeout: int, message: str) -> None:
+def udp_messenger(domain_name: str, UDP_PORT: int, sock_timeout: int, message: str) -> None:
     """Send UDP messages to usage tracker asynchronously
 
     This multiprocessing based messenger was written to overcome the limitations
@@ -40,7 +40,6 @@ def udp_messenger(domain_name: str, UDP_IP: str, UDP_PORT: int, sock_timeout: in
 
     Args:
           - domain_name (str) : Domain name string
-          - UDP_IP (str) : IP address YYY.YYY.YYY.YYY
           - UDP_PORT (int) : UDP port to send out on
           - sock_timeout (int) : Socket timeout
     """
@@ -49,15 +48,7 @@ def udp_messenger(domain_name: str, UDP_IP: str, UDP_PORT: int, sock_timeout: in
     try:
         encoded_message = bytes(message, "utf-8")
 
-        if domain_name:
-            try:
-                UDP_IP = socket.gethostbyname(domain_name)
-            except Exception:
-                # (False, "Domain lookup failed, defaulting to {0}".format(UDP_IP))
-                pass
-
-        if UDP_IP is None:
-            raise Exception("UDP_IP is None")
+        UDP_IP = socket.gethostbyname(domain_name)
 
         if UDP_PORT is None:
             raise Exception("UDP_PORT is None")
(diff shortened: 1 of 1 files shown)


parsl/usage_tracking/usage.py
```diff
@@ -192,7 +192,17 @@ def send_message(self) -> None:
...
+        or won't respond to SIGTERM.
+        """
+        logger.error("BENC CLOSE1")
         for proc in self.procs:
-            proc.terminate()
+            proc.join(timeout=timeout)
+            if proc.is_alive():
+                logger.info("Usage tracking process did not end itself; sending SIGKILL")
+                proc.kill()
+        logger.error("BENC CLOSE2")
```
khk-globus, line 208: > ```python
> logger.error("BENC ...")
> ```
Leftover debugging detail?

khk-globus: Does the `.join(timeout=<value_greater_than_0>)` introduce a noticeable delay anywhere of consequence?  That is, it takes _some_ time for the UDP message to send and the child process to close, and unlike before, we're now waiting inline for that to happen.  (Don't get me wrong, we should absolutely join the child process!)

The first idea that comes to mind is that `.send_message()` should wrap that up automatically, since `udp_messenger` is `@async_process` and has no logic to handle more than one message at a time.  That would similarly elide the need for `.close()` at all.  On the other hand, that flirts with `threading`, which is anathema to `multiprocessing`...

parsl/usage_tracking/usage.py
```diff
@@ -192,7 +192,17 @@ def send_message(self) -> None:
...
 
-    def close(self) -> None:
-        """We terminate (SIGTERM) the processes added to the self.procs list """
+    def close(self, timeout: float = 10.0) -> None:
+        """First give each process one timeout period to finish what it is
+        doing, then kill it (SIGKILL). There's no softer SIGTERM step,
+        because that adds one join period of delay for what is almost
+        definitely either: going to behave broadly the same as to SIGKILL,
+        or won't respond to SIGTERM.
+        """
```
khk-globus, line 201: From a documentation standpoint, I assume the removal of SIGTERM is less about the join period, and more that shy of a stuck network, it won't do anything.  By the time it could be executed, nominally the child process will have already completed.  Alternatively, the "adds one join period of delay" could be handled by:
```python
for proc in self.procs:
    proc.terminate()
    proc.join(timeout=timeout)
    if proc.is_alive():
        proc.kill()
```
Same timeout with or without the `.terminate()` call.

Meanwhile, I think it is likely correct to remove the `.terminate()`, given how the usage_tracker is utilized during (at least) the shutdown of the DFK in `.cleanup()`.

parsl/usage_tracking/usage.py
benclifford: It takes 20ms on my laptop with cached DNS. When DNS is malfunctioning, it will take much longer (the classic "many second delay is always DNS") and that's what the timeout here should bound.

One argument against send message not waiting inline is to avoid that above potential timeout delay at startup, where it's much more noticeable/annoying to users: this model gives the startup message the entire length of the workflow to complete asynchronously before introducing timeout delays.

parsl/usage_tracking/usage.py
khk-globus: Och; I missed that you said exactly that in the description.  Apologies.

parsl/usage_tracking/usage.py
khk-globus: Have talked out of band on this one -- I was suggesting another approach.  Let's not go through with it now, but it's perhaps on the mental radar of N future tasks.

khk-globus (approved):

PR #3091, opened 2024-02-20
SlurmProvider to report stdout/err when job reaches terminal state

yadudoc:
# Description

This PR updates the stdout/err paths specified in the Slurm templates and updates the reporting of `JobStatus` at terminal states to include the stdout/err paths. I have tested these changes on Expanse by setting the address to `127.0.0.1`.

# Changed Behaviour

```
parsl.executors.errors.BadStateException: Executor Expanse failed due to: Error 1:
        Job is marked as MISSING since the workers failed to register to the executor. Check the stdout/stderr logs in the submit_scripts directory for more debug information
        STDOUT: Found cores : 128
Found nodes : 1
0: Failed to find a viable address to connect to interchange. Exiting

        STDERR: srun: error: exp-9-55: task 0: Exited with exit code 5
```

# Fixes

Fixes #3058

## Type of change

Choose which options apply, and delete the ones which do not apply.

- Bug fix

PR #3094, opened 2024-02-21
PBSProvider to report stdout/err paths when jobs fails

yadudoc:
# Description

This PR adds better support for reporting stdout/err summary when jobs reach a terminal state with the `PBSProvider`.
This is the PBSPro version of the fixes that went in with PR #3091.

* Updated PBSProvider to report job stdout/err via JobStatus when job reaches terminal state
* Adding a simple test to confirm that stdout/err paths are included in internal table
* Removing config for Comet@SDSC which is now retired

# Changed Behaviour

When jobs silently fail, for eg, workers are not able to connect to the login node or workers are unable to start due to bad worker environments, the failure information captured in the job stdout/err is summarized in an exception.
Example error from testing on Polaris@ALCF :

```
(gce) rchard@polaris-login-01:~/src/Parsl/tests/failures> python yadu1.py
Launched task.. waiting
  File "/home/rchard/src/Parsl/tests/failures/yadu1.py", line 60, in <module>
    priming()
  File "/home/rchard/src/Parsl/tests/failures/yadu1.py", line 21, in priming
    print(f"Result : {future.result()}")
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/site-packages/parsl/dataflow/dflow.py", line 302, in handle_exec_update
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/site-packages/parsl/dataflow/dflow.py", line 570, in _unwrap_remote_exception_wrapper
    result = future.result()
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/home/rchard/.anaconda3/envs/gce/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
parsl.executors.errors.BadStateException: Executor Polaris-1 failed due to: Error 1:
        STDOUT: Failed to find a viable address to connect to interchange. Exiting
```
## Type of change

- Bug fix

### DIFF
--- a/parsl/providers/pbspro/pbspro.py
+++ b/parsl/providers/pbspro/pbspro.py
@@ -119,13 +119,17 @@ class PBSProProvider(TorqueProvider):
 
                 job_state = job.get('job_state', JobState.UNKNOWN)
                 state = translate_table.get(job_state, JobState.UNKNOWN)
-                self.resources[job_id]['status'] = JobStatus(state)
+                self.resources[job_id]['status'] = JobStatus(state,
+                                                             stdout_path=self.resources[job_id]['job_stdout_path'],
+                                                             stderr_path=self.resources[job_id]['job_stderr_path'])
                 jobs_missing.remove(job_id)
 
         # squeue does not report on jobs that are not running. So we are filling in the
         # blanks for missing jobs, we might lose some information about why the jobs failed.
         for missing_job in jobs_missing:
-            self.resources[missing_job]['status'] = JobStatus(JobState.COMPLETED)
+            self.resources[missing_job]['status'] = JobStatus(JobState.COMPLETED,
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
Looks good; one Slack-shared suggestion is to consider YAGNI with the tests.  Meaning that `MagicMock` is perhaps "overkill" here.  `mock.Mock(spec=)` will work just fine.

PR #3096, opened 2024-02-22
Remove "anonymous" in relation to usage tracking

benclifford:
While this and similar words have been used repeatedly in sellling usage tracking to end users, the user's IP address (regarded as at least possibly identifying information under e.g. GDPR, CCPA and HIPAA) is clearly logged on the receiving side, and for at least that reason it is likely incorrect to be calling this information collection anonymous.

https://github.com/Parsl/parsl_tracking/blob/2c9e11e991e38100126d28b18a57f52a0cc01aa0/tracking_listener.py#L38

Until PR #2666, an easily reversible encoding of username was also transmitted, suggesting a general culture of insincerity or naivety around anonymisation, and the onus should now be on the Parsl team to demonstrate a more serious approach before using this kind of phrasing again.

## Type of change

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,7 @@ To run the Parsl tutorial notebooks you will need to install Jupyter::
 
 Detailed information about setting up Jupyter with Python is available `here <https://jupyter.readthedocs.io/en/latest/install.html>`_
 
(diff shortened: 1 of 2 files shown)


khk-globus (approved):
Seems on target; do we store the IP address beyond the web logs?  The repo-link is private so I can't see the offending line, but getting these docs up to speed with what _actually_ occurs is the correct call.

Meanwhile, I'm left to wonder if we can delineate the difference between what is sent and what we actually hold on to/collect.  Will need more access to that other repo for that conversation.

PR #3097, opened 2024-02-22
Test htex_auto_scale partial scaling-in

benclifford:
The existing scaling-in test parsl/tests/test_scaling/test_scale_down.py only tests full scaling-in which is implemented in a separate code path to partial scaling-in (case 4b vs case 1a in parsl/jobs/strategy.py)

## Type of change

- Code maintenance/cleanup


parsl/tests/test_scaling/test_scale_down_htex_auto_scale.py
```diff
@@ -0,0 +1,151 @@
...
+    fut = waiting_app(0, outputs=outputs, inputs=[File(finish_path)])
+
+    def check_one_block():
+        return len(dfk.executors['htex_local'].connected_managers()) == 1
+
+    try_assert(
+        check_one_block,
+        fail_msg="Expected 1 managers during a single long task",
+        timeout_ms=15000,
+    )
```
khk-globus, line 104: I've mentioned this in Slack, but I think we should do something about these timeouts.  That is, clearly they're needed for the test to pass, but I'm wondering if we can engineer the test setup so that (a) we don't need to up the value to 15s and, more importantly, (b) the test spends ~no time unnecessarily waiting.  That is, we're clearly waiting for (an amalgamation) of some loops internally ... can we short-circuit those loops somehow and still ensure this test is valid and useful?

To put some meat to my displeasure with this, when running this test locally, it took 27s.

(In my experience, the usual avenues for this kind of request are mocking, subclassing, and judicious test setups.)

khk-globus (changes requested):
This is a good test, but I'd like to see if we can speed it up.  As it stands, it takes 27s locally, which is a big proportion of time when the hundreds of other tests and configurations take ~10m as a whole.

parsl/tests/test_scaling/test_scale_down_htex_auto_scale.py
```diff
@@ -0,0 +1,151 @@
...
+    ready_path.touch()
+    inputs = [File(finish_path)]
+    outputs = [File(ready_path)]
+
+    futs = [waiting_app(i, outputs=outputs, inputs=inputs) for i in range(ntasks)]
+
+    while ready_path.read_text().count("\n") < _max_blocks:
+        time.sleep(0.5)
+
+    assert len(dfk.executors['htex_local'].connected_managers()) == _max_blocks
```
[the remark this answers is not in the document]
benclifford: I switched this to try_assert.

On the immediately following connected managers assert, I added a note to future readers who are trying to understand why the assert fails. What's really needed here is to check how many blocks are successfully running at least to the registration stage, but that info isn't so readily available - running one worker per manager and one manager per block gives two different proxies for that.

parsl/tests/test_scaling/test_scale_down_htex_auto_scale.py
```diff
@@ -0,0 +1,151 @@
...
+        return r
+
+    dfk.job_status_poller.callback = hook_cb
+
+    def check_strategy_runs():
+        return strategy_iterated.is_set()
+
+    try_assert(
+        check_strategy_runs,
+        fail_msg="Expected strategy to have run within this period",
```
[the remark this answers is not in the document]
benclifford: yep. leftovers from a more complicated initial implementation

parsl/tests/test_scaling/test_scale_down_htex_auto_scale.py
benclifford: goes down to about 12 seconds if I set the strategy to run 1s, 0.5s, or 0.1s using direct-poking-into-the-strategy-timer

benclifford: (I'm not poking at the strategy code timing deeper than this: this PR is part of a sponsored project to bugfix a single bug in scaling in, not refactor it for testability)

khk-globus (approved):

PR #3106, opened 2024-02-23
`isort` on the test code

yadudoc:
# Description

`isort` **should** only resort import statements and add white spaces to separate different categories of imports.
This PR should not introduce any functional changes. I wanted to get some feedback before sticking isort on the rest of the codebase. If folks like the changes, I'd like to add this to the linter scripts.

# Changed Behaviour

No functional changes.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/__init__.py
+++ b/parsl/tests/__init__.py
@@ -1,5 +1,5 @@
-from parsl.dataflow.memoization import id_for_memo
 from parsl.data_provider.files import File
+from parsl.dataflow.memoization import id_for_memo
 
 
 @id_for_memo.register(File)
--- a/parsl/tests/configs/ad_hoc_cluster_htex.py
+++ b/parsl/tests/configs/ad_hoc_cluster_htex.py
@@ -1,9 +1,9 @@
-from parsl.providers import AdHocProvider
-from parsl.channels import SSHChannel
-from parsl.executors import HighThroughputExecutor
-from parsl.config import Config
-
 from typing import Any, Dict
+
+from parsl.channels import SSHChannel
+from parsl.config import Config
+from parsl.executors import HighThroughputExecutor
+from parsl.providers import AdHocProvider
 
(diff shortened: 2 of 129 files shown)


benclifford:
seems fine though i didn't examine the giant diff in great detail. it would be good to run with --check-only alongside flake8, mypy so that linting-edit-wars don't arise.

yadudoc:
I checked with flake8 and there are no conflicts there. All tests are passing as well.

khk-globus (approved):
I'm personally in favor of this.  It may not be to everyone's precise liking, but it is canonical and removes the human factor of "but my way is better than your way."

In terms of making it not conflict with flake8 and other tools, there is a best-practice ordering for entry into pre-commit (or another linting tool).  I'm happy to add that, or @yadudoc it's all you.  Either way, I'm definitely, definitely in favor of adding at least flake8, pyupgrade, isort, and black into a pre-commit hook.

benclifford (changes requested):
please introduce CI enforcement of this if you're intending to change the house style: this is going to diverge from sortedness within a few commits otherwise

yadudoc:
Fair enough. It sounds like we are agreed that this is a good idea, so here's what I'll do:
1. Replace this PR which only has isort on test code
2. Redo PR with isort on the complete code-base
3. Add isort to pre-commit
4. Add isort to linter checks

benclifford:
> Redo PR with isort on the complete code-base

slightly cynically I have some interest in what kind of import-race-conditions this uncovers...

Andrew-S-Rosen:
If you're adding `black` and `isort`, is there any reason to not just use `ruff format`?

benclifford:
given the current burst of new user/across codebase activity from Outreachy, I think don't do such a code-disrupting change as this for a few more weeks... it will break many people will outstanding changes vs `master`

Issue #3022, opened 2024-01-12
Reduce number of file descriptors used on interchange from manager connections 


benclifford:
**Is your feature request related to a problem? Please describe.**
Right now, the interchange needs two file descriptors per manager (one for sending tasks and heartbeats, and another for sending results).

There are some reasons why it would be interesting to consider using only one:

* about once a year I encounter someone who runs out of file descriptors on their interchange: for example, one user has a restriction of 4096 fds on their submit node. This restricts them to around 2000 workers, but using one fd per manager would double that to around 4000 workers.

* as documented in #3019 sometimes a half-connection can exist where only one of the ports connects successfully. This is awkward to debug

I think this history of these separate connections is due to informal queuing theory analysis of tasks flowing through the system vs results flowing in the other direction through the system, under load. I'm not convinced that the way the code is now that these two flows need two separate TCP connections, given the tight coupling under load between receiving a result and sending a new task. Specifically, I think it should be on the htex developers to properly (with actual theory) justify the need for two sockets.

Issue #2957, opened 2023-11-14
Add support for internal version checks in HTEX


yadudoc:
**Is your feature request related to a problem? Please describe.**
As Globus compute moves to using Parsl's HTEX managers often run different python and parsl versions than the interchange. Currently, this raises a `VersionMismatch` exception since HTEX requires a version match to avoid the several different errors that could creep in from version mismatch. In the GC context, this is less of a concern and an option to disable this check is preferrable.

**Describe the solution you'd like**
A kwarg option for the `HighThroughputExecutor(disable_internal_version_checks: bool = False)` which defaults to the current behavior.

benclifford:
I think disabling that check, rather than putting some effort into protocol work, is an approach that will come back to haunt you :ghost: 

"disable/enable" features look nicer (to me) with fewer negatives, and phrased about whether the check should be made or not, rather than phrased as a change against soon-forgotten built-in behaviour: eg. instead of "disable_internal_version_checks: bool = False", have "check_version_mismatches: bool = True".

yadudoc:
I was going to recommend [Messagepack](https://github.com/funcx-faas/funcx-common/tree/main/src/globus_compute_common/messagepack). I agree that doing this properly now is the right way to go.

benclifford:
I don't think it brings in any additional requirements, but it's also interesting maybe to think about this happening inside Parsl: any of the places where we're interested in multiple environments existing - which is cloud stuff, not so interesting any more, but also WQ and radical which have the ability to build their own environments, and places where people are using containers.  I think that is broadly the same situation as with Globus Compute and doesn't change anything here, but I'm noting it on this issue anyway.

PR #3109, opened 2024-02-28
Make manager->interchange tasks_output protocol always JSON objects

benclifford:
This is to permit imminent changes to this protocol to allow quenching of task delivery as part of work to drain workers near the end of their walltime.

Prior to this PR, the protocol used two formats: the first message in the worker->interchange task_output pipe was a JSON object, the "registration message".

Subsequent messages were not JSON, but an integer encoded as a byte sequence. This integer was historically used for two purposes: as a heartbeat (with value HEARTBEAT_CODE = (2 ** 32) - 1) and as a count of tasks to request from the interchange.

This latter use was removed in PRs #2588 and #3062, leaving only the heartbeat use case.

To allow this channel to be used for richer messages, this PR switches the messages on this stream to always be JSON object, with a 'type' entry: currently 'registration' or 'heartbeat'.

This is in line with the 'result_package' format used on the manager->interchange results_outgoing channel, and so aligns with issue #3022 which talks about unifying those two channels.

Code paths:

This PR removes the reg_flag variable used in registration code, which was redundant and only used to route control flow from a try block into an immediately following code block: that second block is merged upwards.

That try block protected/protects json decoding, and is moved higher up because there is now no need to have a separate path for "expecting JSON" and "expecting encoded int".

Type checking: This PR adds some additional checking for JSON structure to report a more useful error when the message is not a dictionary; this drives mypy to specialise the type of `msg` to dict[Any, Any] which can then not be used to update a ManagerRecord. This PR adds a # type: ignore with commentary to address that.

Testing: while registration is implicitly tested by htex tests successfully running tasks, there isn't any automated testing of manager->interchange heartbeats, so human review should pay special attention there.

# Changed Behaviour

The wire protocol is changed: anyone pretending they can use different versions of Parsl on worker and interchange (#2957) sides will encounter breakage (likely a hang).

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -27,7 +27,6 @@ from parsl.monitoring.message_type import MessageType
 from parsl.process_loggers import wrap_with_logs
 
 
-HEARTBEAT_CODE = (2 ** 32) - 1
 PKL_HEARTBEAT_CODE = pickle.dumps((2 ** 32) - 1)
 
 LOGGER_NAME = "interchange"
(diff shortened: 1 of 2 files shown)


khk-globus (approved):
The changes look solid, and certainly tests pass.  I ran a couple of local one-offs, so "appears to work."  Time will tell if we need to iterate more thoroughly here.

I will mention in passing that I'm pleased to be rid of that weird `reg_flag`.  Has been "smelly" since day one of my interactions with Parsl.  Obviously more changes to come in this section, but this is a good step.

PR #3115, opened 2024-03-01
Remove unused codepath from HighThroughputExecutor scale_in

benclifford:
The codepath would be used when scale_in is called with:

  force=False
  max_idletime=None

and would pick blocks from the list of blocks which are currently idle.

This PR removes that unused codepath, merging the choice of forced/non-forced scale-in and idle time specification into a single parameter that indicates "do not scale in blocks that have not been idle this long".

The two remaining cases from force=True/False, max_idletime=None/number are:

* max_idletime=None, previously: force=True, max_idletime=None

This means that the requested number of blocks should be scaled in, even if the blocks are not idle. The use case for this path is "parsl is shutting down, so we need to tidy up everything. If there are tasks still running, we don't care because terminating in-progress tasks is part of parsl shutdown."

* max_idletime=some time, previousy: force=False, max_idletime=some time

This means that the scaling code has decided to apply downwards pressure on the number of blocks: there are more blocks than needed. However, this pressure should not disrupt already running tasks, and it is less urgent to cancel blocks, because the same call will happen every 5 seconds to keep applying that pressure.

# Changed Behaviour

This is removing an unused code path, so should not change any user-visible behaviour.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -668,7 +668,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
     def workers_per_node(self) -> Union[int, float]:
         return self._workers_per_node
 
-    def scale_in(self, blocks, force=True, max_idletime=None):
+    def scale_in(self, blocks, max_idletime=None):
         """Scale in the number of active blocks by specified amount.
 
         The scale in method here is very rude. It doesn't give the workers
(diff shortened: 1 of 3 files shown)


benclifford:
This PR failed (on Python 3.11) in a scaling test that, although its scaling related, seems unrelated to this PR, but it is a test introduced in `master` very recently - #3097 one commit ago - so there is some possibility that this new test is now uncovering some breakage as it is supposed to, just non-deterministically. I'll make a note in PR #3097 about the details

PR #3116, opened 2024-03-01
Change htex pull_tasks iteration from exponential backoff to calculated time

benclifford:
Before this PR, this loop runs repeatedly, under high load quite often, but slowing down exponentially until reaching once per heartbeat period.

The other things happening in this loop:

* emit a debug log
* check heartbeat related events (motivating upcoming drain work in PR #3063 would like to add an additional event there)

This PR calculates the earliest of the two heartbeat related events and wait in poll until that time.

# Changed Behaviour

This will change:

i) the two heartbeat checks may happen earlier than previously, closer to the calculated time, because this loop will more accurately aim for the heartbeat time. This might uncover bugs or misassumptions elsewhere in Parsl code or users. Related to this, this PR changes time comparisons from > to >= so that events can fire in the same fractional-second as the corresponding computed poll ends.

ii) the debug log message will be output at different times, as the loop will now iterate at different times. The timing of the log message was quite weird before, and is now differently weird.

This PR removes one use of the somewhat arbitrary poll timer htex parameter.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -288,9 +288,17 @@ class Manager:
         last_interchange_contact = time.time()
         task_recv_counter = 0
 
-        poll_timer = self.poll_period
-
         while not kill_event.is_set():
+
+            # This loop will sit inside poller.poll until either a message
+            # arrives or one of these event times is reached. This code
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
That's a nice little improvement.

PR #3117, opened 2024-03-01
Deprecate `max_workers` for `max_workers_per_node`

### LINKED ISSUE #1429: max_workers is ambiguous

rjmello:
## Description

The latter is less ambiguous than the former. We will support both for the time being but raise a deprecation warning for `max_workers`. Also, `max_workers_per_node` will take precedence if both are defined.

Fixes #1429

## Type of change

- Bug fix
- Code maintenance/cleanup


benclifford:
this failed in CI with

```
parsl/tests/test_mpi_apps/test_mpi_mode_enabled.py F

_____________________________ test_simulated_load ______________________________

rounds = 100

    @pytest.mark.local
    def test_simulated_load(rounds: int = 100):
    
        node_choices = (1, 2, 4)
        sleep_choices = (0, 0.01, 0.02, 0.04)
        ranks_per_node = (4, 8)
    
        futures = {}
        for i in range(rounds):
            resource_spec = {
                "num_nodes": random.choice(node_choices),
                "ranks_per_node": random.choice(ranks_per_node),
            }
            future = mock_app(sleep_dur=random.choice(sleep_choices),
                              parsl_resource_specification=resource_spec)
            futures[future] = resource_spec
    
        for future in futures:
>           total_ranks, nodes = future.result(timeout=10)
```

which is probably an unrelated test error - perhaps this test is making too tight an assumption on how fast it can run that quantity of apps - tagging @yadudoc because this is in his MPI prototype.

rjmello:
CI was failing because I missed a couple of updates to test configs that modified `max_workers` after initialization. Should be good to go now.

I also created getter/setter methods for `max_workers` that raise deprecation warnings and serve as a pass-through to `max_workers_per_node`.

benclifford (approved):

PR #3118, opened 2024-03-03
Adding onto #3111: shallow resolution of lists, sets, and tuples

Andrew-S-Rosen:
@benclifford: This PR into #3111 is to augment support for shallow-level `AppFuture` resolution.

A few comments:
1. The newly added tests pass on Python 3.11 for me, but I can't figure out the typing syntax for Python 3.8/3.9. I tried replacing `tuple`, `list`, and `set` with `typing.Tuple`, `typing.List`, and `typing.Set`, but that didn't help matters.
2. I tried to generalize [this line](https://github.com/Parsl/parsl/pull/3118/files#diff-d95e44771f15076748fffdf10ae0a92f25b3aeed06a61c26b4488b0103220b20R44) as being of type `Sequence` or `Iterable` instead of `Union[tuple, list, set]`, but when doing the former, the output of the test has `Path(<mapping>)` objects instead of the expected normal `pathlib.Path("str")` kind of result. So, I stuck with the explicit `Union`.
3. This does not yet address the `dict` type (or, rather, general `Mapping` type). I wanted to solve the above instances first.

### DIFF
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -1,6 +1,5 @@
 from concurrent.futures import Future
 from functools import singledispatch
-from typing import Union
 
 
 @singledispatch
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -35,8 +35,7 @@ def _(iterable: Union[tuple, list, set]):
     # a "deep" traversal would instead recursively call traverse_to_gather
     # here to inspect whatever is inside the sequence
 
-    type_ = type(iterable)
-    return type_([v for v in iterable if isinstance(v, Future)])
+    return [v for v in iterable if isinstance(v, Future)]
 
 
 @traverse_to_unwrap.register
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -30,17 +30,21 @@ def _(fut: Future):
 # Below is an example of shallow traversal of iterables.
 
 
-@traverse_to_gather.register
-def _(iterable: Union[tuple, list, set]):
+@traverse_to_gather.register(tuple)
+@traverse_to_gather.register(list)
+@traverse_to_gather.register(set)
+def _(iterable):
     # a "deep" traversal would instead recursively call traverse_to_gather
     # here to inspect whatever is inside the sequence
 
     return [v for v in iterable if isinstance(v, Future)]
 
 
-@traverse_to_unwrap.register
+@traverse_to_unwrap.register(tuple)
+@traverse_to_unwrap.register(list)
+@traverse_to_unwrap.register(set)
 @singledispatch
-def _(iterable: Union[tuple, list, set]):
+def _(iterable):
     def unwrap(v):
         if isinstance(v, Future):
             assert (
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -1,5 +1,6 @@
 from concurrent.futures import Future
 from functools import singledispatch
+from typing import Union
 
 
 @singledispatch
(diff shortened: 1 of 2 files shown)


parsl/dataflow/future_resolution.py
```diff
@@ -26,23 +27,28 @@ def _(fut: Future):
...
-    return [v for v in tup if isinstance(v, Future)]
+    # here to inspect whatever is inside the sequence
+
+    return [v for v in iterable if isinstance(v, Future)]
 
 
 @traverse_to_unwrap.register
 @singledispatch
-def _(tup: tuple):
+def _(iterable: Union[tuple, list, set]):
```
benclifford, line 43: at least with set, there is some subtlety that I think means this should be separately implemented for set:

an iterable of mapping .result() over a set behaves differently to creating a set of the unwrapped result and returning that: the set {f(1), f(2)} only has one element in some situations and this code should use `set()` rather than discarding the set semantics.

parsl/tests/test_python_apps/test_pluggable_future_resolution.py
```diff
@@ -29,14 +30,41 @@ def test_simple_pos_arg():
 
 
 @parsl.python_app
-def b_first(x: Sequence[int]):
+def b_first(x: Iterable[int]):
     return x[0] + 1
```
benclifford, line 34: you can't do [0] on an `Iterable` - mypy detects that and is giving an error here...
an Iterable style implementation here would be to call `next(iter(x))` I think.

parsl/tests/test_python_apps/test_pluggable_future_resolution.py
```diff
@@ -29,14 +30,41 @@ def test_simple_pos_arg():
...
+
+@parsl.python_app
+def make_path(s: str):
+    return Path(s)
+
+
+@parsl.python_app
+def append_paths(iterable: Iterable[Path], end_str: str = "end"):
+    type_ = type(iterable)
+    return type_([Path(s, end_str) for s in iterable])
```
benclifford, line 54: > parsl/tests/test_python_apps/test_pluggable_future_resolution.py:54: error: Too many arguments for "Iterable"  [call-arg]

there's no static guarantee that the constructor of an arbitrary type that happens to be an iterable takes one parameter that is a list (of things you want to somehow populate that iterable with)

more technically: this is coming from (I think) because we statically know that `iterable` is an Iterable, so _type = Iterable at least as far as finding a single value that always works, and then type_(....) is Iterable(....) but Iterable() takes no parameters.

parsl/dataflow/future_resolution.py
benclifford: actually I see what you're trying to do with invoking runtime-discovered constructors...

parsl/tests/test_python_apps/test_pluggable_future_resolution.py
Andrew-S-Rosen: Oops, good catch. I didn't mean to change this one.

parsl/tests/test_python_apps/test_pluggable_future_resolution.py
Andrew-S-Rosen: Makes sense. This was me trying to be too smart, I suppose...

Andrew-S-Rosen:
Thanks for the comments, @benclifford. I believe I have addressed them. That said, I am still unsure how to resolve the

> TypeError: Invalid annotation for 'iterable'. typing.Union[tuple, list, set] is not a class.

in the older versions of Python.

benclifford:
try writing three registers in a row with explicit types instead of using Union - you can write multiple register annotations, like this in `parsl/dataflow/memoization.py`:

```
@id_for_memo.register(str)
@id_for_memo.register(int)
@id_for_memo.register(float)
@id_for_memo.register(type(None))
def id_for_memo_pickle(obj: object, output_ref: bool = False) -> bytes:
```

Andrew-S-Rosen:
Thanks, @benclifford! That did the trick. Feel free to merge this into your PR if you wish. The next step will be `dict` handling.

PR #3119, opened 2024-03-04
Add more typing around scale-in decision state

benclifford:
Use a typed dataclass with human readable names, instead of a numerically index list for the decision state itself, and add type anontations on the method signature.

# Changed Behaviour

This is intended to make some of the decision code easier to read and typecheck, but should not change behaviour.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Nice.

khk-globus (approved):
Nice.

PR #3120, opened 2024-03-04
Add process context switch counts to resource monitoring

benclifford:
# Changed Behaviour

There is a schema change in the monitoring database which will break across Parsl versions before/after this PR.

## Type of change

- New feature


parsl/monitoring/remote.py
```diff
@@ -218,6 +220,15 @@ def accumulate_and_prepare() -> Dict[str, Any]:
...
 
         d["psutil_cpu_count"] = psutil.cpu_count()
+
+        # note that this will be the CPU number of the base process, not anything launched by it
+        d["psutil_cpu_num"] = pm.cpu_num()
+
+        pctxsw = pm.num_ctx_switches()
+
+        d["psutil_process_num_ctx_switches_voluntary"] = pctxsw.voluntary
+        d["psutil_process_num_ctx_switches_involuntary"] = pctxsw.involuntary
```
khk-globus, line 230: Inching closer and closer to just getting the current reading from `/proc/self/stat` and storing it for later.

khk-globus (approved):

parsl/monitoring/remote.py
benclifford: there's no "just" there: it needs parsing into the structured database by someone, either the end-user or Parsl takes on that parsing burden.

Issue #3105, opened 2024-02-23
Outreachy: getting familiar with the development environment


benclifford:
This issue is targeted at potential applications for Parsl's participation in Outreachy. The tasks in this issue can be performed by any number of applicants. Talk to @benclifford on Slack if you are interested.

The point of the tasks in this issue are to walk you through getting a development environment set up, making a trivial change to the code, seeing how our tooling tests things, deliberately introducing a problem, and then fixing that problem. By the end you should have a trivial 1-byte change to submit to parsl as a PR, to try out the PR process.

There are lots of steps because I have tried to make each step small.

If you get stuck or don't understand any of these steps - you can ask @benclifford on Slack or write publicly in the #parsl-hackers channel on  Slack.

1. Get a Python environment - parsl is supported on Python versions 3.8 - 3.12.
2. Sign up for a GitHub account
3. Go to the parsl repository at https://github.com/Parsl/parsl and near the top right, click "Forks / create a new fork". This will make a fork of parsl in your own github account.
5. Install parsl - follow the instructions in the "For developers" section of the Parsl README at https://github.com/Parsl/parsl - but when asked to run the command `git clone https://github.com/parsl/parsl`, instead use the address of your fork created in the previous step.
6. You can run some local tests by running `make local_thread_test`, and some more advanced tests with `make htex_local_alternate_test`
7. You can test our code linting by running `make flake8' which checks all of our Python code to see if it matches the Python style rules in [PEP-8](https://peps.python.org/pep-0008/)
8. We have configuration to let us break the style rules sometimes. This lives in the file called `.flake8`. You can open that file and make a change to move us towards proper PEP-8 compliance: find the line that says `max-line-length = 158` and make it 1 smaller. (PEP-8 says the max line length should be 80... but we can't change it all at once)
9. Run `make flake8` again - you should see that you can an error: flake8 should have found a line that is now too long.
10. Edit that line - you'll have to figure out how to make it shorter, perhaps by splitting it into two valid lines of Python
11. Run `make flake8` again and you should see that flake8 passes again
12. Now you can make a git commit: type `git commit -a -m "Reduce flake8 max-line-length`. Don't worry if you don't understand what git is doing: you can learn more about that during the internship.
13. Now push that commit up to github: type `git push`
14. Visit your repository on github in your web browser - you should see your commit there now.
15. Click "Pull request" to make a pull request to the main Parsl repository.
16. Fill out the template - you don't need to write too much
[...]

PR #3122, opened 2024-03-04
Reduce flake8 max-line-length

Harichandra-Prasath:
# Description

Reduced  flake-8 maximum length and changed some lines which violate the new configuation

# Changed Behaviour

# Fixes

Fixes #3105 

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@
 # W504: line break after binary operator
 #       (Raised by flake8 even when it is followed)
 ignore = E126, E402, E129, W504
-max-line-length = 158
+max-line-length = 157
 exclude = test_import_fail.py,
   parsl/executors/workqueue/parsl_coprocess.py
 # E741 disallows ambiguous single letter names which look like numbers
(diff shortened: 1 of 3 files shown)


benclifford (approved):

benclifford:
your PR failed in CI, with this error from `mypy`, the static type checking tool we used:

```
parsl/executors/workqueue/executor.py:65: error: List or tuple literal expected as the second argument to "namedtuple()"  [misc]
```

That's because you can't use a variable as a parameter to `namedtuple` when doing static typechecking (even if the variable is a constant on the line before...)

You should be able to recreate this error yourself in your development environment by running `make mypy` and see that it happens in your own environment.

Then some different ideas:

idea 1: You can split the string into multiple lines directly in the call to namedtuple - for example, like this:

```
def f(x: str):
 print(x)


f('a'
  'b')
```

idea 2, if we knew all the types of each piece of the named tuple, we could use a different syntax that I think is much nicer - but you would have to know all the types, so perhaps that is something to try in a later pull request: https://docs.python.org/3/library/typing.html#typing.NamedTuple

PR #3123, opened 2024-03-04
Reduce flake8 max-line-length

Unique-Usman:
# Description

Reduced flake-8 maximum length and changed some lines which violate the new configuation

# Changed Behaviour

# Fixes

Fixes #3105

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -413,8 +413,9 @@ class Manager:
                             raise WorkerLost(worker_id, platform.node())
                         except Exception:
                             logger.info("Putting exception for executor task {} in the pending result queue".format(task['task_id']))
-                            result_package = {'type': 'result', 'task_id': task['task_id'], 'exception':
-                                              serialize(RemoteExceptionWrapper(*sys.exc_info()))}
+                            result_package = {'type': 'result',
+                                              'task_id': task['task_id'],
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -413,7 +413,8 @@ def worker_watchdog(self, kill_event: threading.Event):
                             raise WorkerLost(worker_id, platform.node())
                         except Exception:
                             logger.info("Putting exception for executor task {} in the pending result queue".format(task['task_id']))
-                            result_package = {'type': 'result', 'task_id': task['task_id'], 'exception': serialize(RemoteExceptionWrapper(*sys.exc_info()))}
+                            result_package = {'type': 'result', 'task_id': task['task_id'], 'exception':
+                                              serialize(RemoteExceptionWrapper(*sys.exc_info()))}
```
benclifford, line 417: Hi. Thanks for your PR.

For style, I would maybe write this with on three different lines - so that each key and value is next to each other on its own line, like this:

type: result
task_id: ....
exeception:  serialize(...)

That might make this easier to read when someone looks at this code later.

Unique-Usman: Hello @benclifford thanks for the suggestion. I should have figured that out.

benclifford (approved):

PR #3125, opened 2024-03-04
Reduce flake8 max-line-length

error-4u:
# Description

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@
 # W504: line break after binary operator
 #       (Raised by flake8 even when it is followed)
 ignore = E126, E402, E129, W504
-max-line-length = 153
+max-line-length = 152
 exclude = test_import_fail.py,
   parsl/executors/workqueue/parsl_coprocess.py
 # E741 disallows ambiguous single letter names which look like numbers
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -621,7 +621,7 @@ def start_file_logger(filename: str, level: int = logging.DEBUG, format_string:
         None.
     """
     if format_string is None:
-        format_string = "%(asctime)s.%(msecs)03d %(name)s:%(lineno)d %(processName)s(%(process)d) %(threadName)s %(funcName)s [%(levelname)s]  %(message)s"
+        format_string = "%(asctime)s.%(msecs)03d %(name)s:%(lineno)d %(processName)s %(threadName)s %(funcName)s [%(levelname)s]  %(message)s"
```
benclifford, line 624: it's not so good to remove information from the logging messages - in the past we've been making these log lines *longer* not shorter, to get more information - especially to do with processes and threads because that can be hard to debug.

probably a better thing to do here would have been to split the format string over multiple lines - have a look at how PR #3130 addressed this line being too long, for example

benclifford (approved):

PR #3129, opened 2024-03-05
Reduce flake8 max-line-length

nupurkale78:
# Description

The issue involves reducing the max-line-length configuration in Flake8 to adhere to PEP-8 standards.

# Changed Behaviour

# Fixes

Fixes #3105 

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -1,3 +1,4 @@
+
 import logging
 import threading
 import queue
@@ -544,7 +545,10 @@ class DatabaseManager:
                 if reprocessable_last_resource_messages:
                     self._insert(table=STATUS, messages=reprocessable_last_resource_messages)
             except Exception:
-                logger.exception("Exception in db loop: this might have been a malformed message, or some other error. monitoring data may have been lost")
+                logger.exception(
+                    "Exception in db loop: this might have been a malformed message,"
+                    "or some other error. monitoring data may have been lost"
+                )
                 exception_happened = True
         if exception_happened:
             raise RuntimeError("An exception happened sometime during database processing and should have been logged in database_manager.log")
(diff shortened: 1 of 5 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -1,3 +1,4 @@
+
```
benclifford, line 1: This looks like you accidentally put a blank line in at the start of the file. For cleanliness, it's good to not change whitespace that isn't directly part of what you are trying to achieve; and our style is to not have blank lines at the top of files

nupurkale78: Sorry, I've fixed it.

benclifford (approved):

parsl/monitoring/db_manager.py
```diff
@@ -544,7 +544,10 @@ def start(self,
                 if reprocessable_last_resource_messages:
                     self._insert(table=STATUS, messages=reprocessable_last_resource_messages)
             except Exception:
-                logger.exception("Exception in db loop: this might have been a malformed message, or some other error. monitoring data may have been lost")
+                logger.exception(
+                    "Exception in db loop: this might have been a malformed message,"
```
benclifford, line 548: here's something I didn't notice on my last review! there's a common mistake with multiline strings (that I often make too) -- there is no space between the command at the end of this line and the word "or" on the next line!
(compare to how you split up format_string in interchange.py with spaces at the end of each string.

benclifford:
ok, looks like #3130 was ready slightly before yours - you can look and see that they both make basically the same change, so I have merged #3130. If you want to get a PR for this task #3105 you can make a new PR making the length even shorter, but you can also instead move on to looking at some of the tasks that we've tagged with the `outreachy` tag: https://github.com/Parsl/parsl/issues?q=is%3Aissue+is%3Aopen+label%3Aoutreachy

PR #3130, opened 2024-03-05
Reduce flake8 max-line-length

mercybassey:
# Description

Reduced flake8 maximum line length 

# Changed Behaviour

# Fixes

Fixes # 3105

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -544,11 +544,10 @@ class DatabaseManager:
                 if reprocessable_last_resource_messages:
                     self._insert(table=STATUS, messages=reprocessable_last_resource_messages)
             except Exception:
-                log_message = (
+                logger.exception(
                     "Exception in db loop: this might have been a malformed message, "
                     "or some other error. monitoring data may have been lost"
                 )
-                logger.exception(log_message)
                 exception_happened = True
         if exception_happened:
             raise RuntimeError("An exception happened sometime during database processing and should have been logged in database_manager.log")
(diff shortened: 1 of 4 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -544,7 +544,11 @@ def start(self,
                 if reprocessable_last_resource_messages:
                     self._insert(table=STATUS, messages=reprocessable_last_resource_messages)
             except Exception:
-                logger.exception("Exception in db loop: this might have been a malformed message, or some other error. monitoring data may have been lost")
+                log_message = (
```
benclifford, line 547: Hi, thanks for your PR. You could avoid using an intermediate variable here. You could use multi-line string syntax directly on the method call to logger.exception. Here's an example function call like that to show what I mean:

```
def f(x: str):
 print(x)

f('a'
  'b')
```

benclifford:
It looks like I'll probably merge PR #3129 because it will be ready slightly before yours. But this PR is fine otherwise. If you want, you can wait a few minutes and reduce the max line length even further, or you can move on to looking at some of the other outreachy issues - I'm going to tag some more now.

mercybassey:
> It looks like I'll probably merge PR #3129 because it will be ready slightly before yours. But this PR is fine otherwise. If you want, you can wait a few minutes and reduce the max line length even further, or you can move on to looking at some of the other outreachy issues - I'm going to tag some more now.

Alright, I'll wait a few minutes and reduce the max line length even further. However, I'll also move on to looking at some of the other outreachy issues too.

benclifford (approved):

benclifford:
turns out this PR got merged after all and #3129 didn't - just because the fixes and tests passed a few minutes before the #3129 changes (which look basically the same)

PR #3134, opened 2024-03-05
Implement Atexit Cleanup for Enhanced Resource Management

### LINKED ISSUE #3104: `atexit` cleanup will not work properly in Python 3.12 and beyond
**Describe the bug**
Python has documented that it is unsafe to do thread and fork operations in `atexit` handlers.

Each DFK installs an `atexit` handler so that it will attempt clean up its external resources (various processes on the same host, batch job submissions, for example) if a user workflow falls off the end of a script without shutting down that DFK.

Python 3.12 has moved to a more aggressive model of failing on thread and fork operations in exit handlers. See issues #3102 and #3103 and Python documentation at https://docs.python.org/3/library/atexit.html

That means that `atexit` cleanup will, long term, probably have to go away.

* Most immediately, it should probably be labelled as deprecated (in all Python versions) and documentation updated now to suggest to users that they should be calling dfk `cleanup()` themselves
* Eventually `atexit` cleanup should be replaced with either nothing, or a warning if the DFK has not been cleaned up, indicating to a user how this should be implemented in their code.
* I've been pushing a bit on resources going away without needing some fragile DFK-driven shutdown protocol - this issue suggests that that is even more important now, so that resources will be cleaned up with out an `atexit` driven shutdown protocol.

nupurkale78:
# Description

This update includes `self.cleanup()` commands at various points in the script to better manage resources. By making sure that resources are cleaned up when they are no longer needed, this change improves stability, performance, and prevents resources from being wasted.

# Changed Behaviour

This update ensures smoother operation and resource handling without altering the core functionality of existing workflows.

# Fixes

Fixes #3104 

## Type of change

- Bug fix

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -70,15 +70,14 @@ class DataFlowKernel:
     @typechecked
     def __init__(self, config: Config) -> None:
         """Initialize the DataFlowKernel.
-    # Consider adding 'self.cleanup()' here if this method allocates resources.
-
+        
         Parameters
         ----------
         config : Config
             A specification of all configuration options. For more details see the
             :class:~`parsl.config.Config` documentation.
         """
-
+        self.cleanup()
         # this will be used to check cleanup only happens once
         self.cleanup_called = False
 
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -1410,18 +1409,24 @@ class DataFlowKernelLoader:
     def clear(cls) -> None:
         """Clear the active DataFlowKernel so that a new one can be loaded."""
         cls._dfk = None
+<<<<<<< HEAD
+    self.cleanup()
 
+=======
+        self.cleanup()
+>>>>>>> c4f3087c (fix for 3104)
```
khk-globus, line 1417: This is an uncleaned up merge conflict.  If this is a new concept, one phrase for the search engines is `merge conflict`.

PR #3135, opened 2024-03-05
Make htex scale_in pick longest idle blocks

benclifford:
Prior to this, scale_in picked the shortest-idle (i.e. most recently used but now idle) blocks.

In the scaling code right now, there isn't any particular reason to order idle blocks in any particular order...

... but the next PR following on from this one will also consider unstarted blocks with infinite idle time, and in that situation, the policy will be that unstarted idle blocks should be scaled down in preference to currently running blocks. (other policies are possible and reasonable but not addressed by either this PR or the upcoming PR)

I think this was intended to work this way to begin with (although I'm unsure of the justification) but there is a sign error that is flipped by this PR.

@yadudoc or @ZhuozhaoLi do you have any insight into the purpose of this sorting?

# Changed Behaviour

Block scale in order is changed - I don't think that will bother people.

## Type of change

- Bug fix? new feature?


khk-globus (approved):
Subtle; a unit test might be in order.

PR #3141, opened 2024-03-05
Stop DFK cleanup in atexit, because this doesn't work in Python >3.12"

error-4u:
# Description

atexit handlers work differently in the latest version of Python (version 3.12) and it is not safe to automatically call cleanup any more in Python 3.12

# Changed Behaviour

Instead of cleaning up, the code will now emit a WARNING log message to help users understand the change.

# Fixes

Fixes # (https://github.com/Parsl/parsl/issues/3104)

## Type of change

- Bug fix

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@
 # W504: line break after binary operator
 #       (Raised by flake8 even when it is followed)
 ignore = E126, E402, E129, W504
-max-line-length = 153
+max-line-length = 152
 exclude = test_import_fail.py,
   parsl/executors/workqueue/parsl_coprocess.py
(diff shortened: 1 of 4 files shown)


benclifford:
right now, this PR has a big jumble of changes that are unrelated to issue #3104 - look at https://github.com/Parsl/parsl/pull/3141/files and see them all. i'll work through fixing that with you.

benclifford (changes requested):
discussing how to fix this in Slack

benclifford:
This PR was tangled up, but we untangled things into a new PR #3165 that is now merged. So I'll close this issue.

PR #3144, opened 2024-03-05
Reduce flake8 max-line-length

marameref:
# Description

I reduced flake8 max-length to 113

# Changed Behaviour

- flake8 

# Fixes

- Reduce flake8 max-line-length to 113

## Type of change

- Code maintenance/cleanup


marameref:
@benclifford pls review

.flake8
```diff
@@ -8,9 +8,10 @@
 # W504: line break after binary operator
 #       (Raised by flake8 even when it is followed)
 ignore = E126, E402, E129, W504
-max-line-length = 152
+max-line-length = 131
 exclude = test_import_fail.py,
-  parsl/executors/workqueue/parsl_coprocess.py
+  parsl/
```
benclifford, line 13: I think this is a mistake here: it looks like you've split the line containing `parsl/executors/workqueue/parsl_coprocess.py` into two separate lines, and now the new version says to ignore all errors in the whole of parsl!

benclifford (changes requested):
see comment on .flake8 accidental change

PR #3157, opened 2024-03-06
Dict memoization sorting before normalizing keys is unsafe #2140

### LINKED ISSUE #2140: Dict memoization sorting before normalizing keys is unsafe
This line struck me as odd:

https://github.com/Parsl/parsl/blob/1449ea7d019bfac7448a1260de91e613dda137e3/parsl/dataflow/memoization.py#L89

My first thought was that it might not give you a stable sort if objects are hashable (can be used in a dict), but define something meaningless for comparison (my thought was comparison of memory addresses).

The case I found when trying to exploit this to make something unstable is a type which is hashable but raises an error on comparison. Functions:
```python
>>> def foo(): return 1
...
>>> def bar(): return 2
...
>>> d = {foo: 1, bar: 2}
>>> sorted(d)
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'function' and 'function'
```

To show that this doesn't only happen in "silly" cases, I also tried out `enum.auto`, and found that it also is hashable but not comparable:
```python
>>> class Foo(enum.Enum):
...     x = enum.auto()
...     y = enum.auto()
...
>>> Foo.x < Foo.y
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'Foo' and 'Foo'
```

_In theory_ we might be able to find a type which does define comparison, but does it in an unreliable way which is tied to the state of the current process. I think given the presence of types which define `__hash__` but not `__lt__`, we should assume that such objects exist.

## Resolution

To solve this, I believe that all we need to do is do the key normalization step first, saving the normalization results in a temporary dict, and then operate on the normed keyspace.

Something to the effect of
```python
def id_for_memo_dict(d):
    keymap = {id_for_memo(k): k for k in d}
    normed_keys = sorted(keymap.values())

    normalized_list = []
    for k in normed_keys:
        normalized_list.append(k)
        normalized_list.append(id_for_memo(d[keymap[k]]))
    return serialize(normalized_list)
```

MundiaNderi:
# Description

Stability of sorting  and transforming keys  into a format suitable for sorting.

# Changed Behaviour

Prior sorting was done before normalising the keys. This implementation normalises the keys and then sorts them.  Added a test case that fails should this not be the case.

# Fixes

Fixes #2140 

## Type of change

- Bug fix


MundiaNderi:
@benclifford I've worked on the memoization issue but I still have some conflicts with the max-length-files . How do I proceed ?

benclifford:
You've got a lot of other random stuff in your PR that isn't related to issue #2140 - maybe you typed `git add .` or some command like that which added everything in your working directory?

For example, lots of test output in runinfo/ is there in your PR...

Have a look at this URL to see it all

https://github.com/Parsl/parsl/pull/3157/files

You can remove the files you don't want with `git rm`

It also looks like you have a few other changes with whitespace and new lines that aren't relevant to your PR.

MundiaNderi:
You're right. I used the git add command.  Let me remove the files I don't want

parsl/dataflow/memoization.py
```diff
@@ -94,10 +98,13 @@ def id_for_memo_dict(denormalized_dict: dict, output_ref: bool = False) -> bytes
     if type(denormalized_dict) is not dict:
         raise ValueError("id_for_memo_dict cannot work on subclasses of dict")
 
-    keys = sorted(denormalized_dict)
+    # keys = sorted(denormalized_dict)  Line that sirosen commented on
+    # Proposed solution was to normalize the keys and then sort them
+    keymap = {id_for_memo(k): k for k in denormalized_dict}
+    normed_keys = sorted(keymap.values())
```
benclifford, line 104: ok, I think this is right thing to be doing - but I don't think your test case that you added really tests that something was broken before, and is now fixed by this change.

I will try to add some comments on the test case if I can figure out a stronger test case.

MundiaNderi: Thank you , I've also realised that and was trying to think though it. Where I fell short was I corrected the implementation first, then created the test case later. I should have done vice versa.

benclifford:
there are still a lot of changes in this pull request that are not directly related to the dictionary sorting issue: look in the "Files Changed" tab on this pull request in the Github web interface, and you will see a lot of unrelated changes still.

MundiaNderi:
Okay @benclifford let me work on resolving this.

MundiaNderi:
@benclifford I've cleaned up the files less one file which got messed up during rebase. How would you advise me to proceed ?

benclifford:
> @benclifford I've cleaned up the files less one file which got messed up during rebase. How would you advise me to proceed ?

Which file? What do you think it should look like?

MundiaNderi:
This file: parsl/dataflow/dflow.py. git rm parsl/dataflow/dflow.py deleted the file locally instead of from the commit. Ideally it should not be present within this PR but rebasing and merging .

benclifford:
It looks like the change to dflow.py is a straightforward line break which you can fix by joining the lines again and making a new commit with that fix.

MundiaNderi:
Thank you @benclifford I've fixed the file.

PR #3159, opened 2024-03-06
Set up GitHub Actions Workflow for Testing Parsl with Flux

### LINKED ISSUE #2713: Flux + parsl testing

mercybassey:
# Description

This pull request introduces a new GitHub Actions workflow aimed at testing Parsl's integration with Flux. 

# Changed Behaviour

The primary focus of this PR is to enhance Parsl's testing infrastructure by integrating a new GitHub Actions workflow aimed at validating Parsl's compatibility and functionality with Flux.

# Fixes

Fixes #2713 

## Type of change

- Enhancement


benclifford:
ok, great.

here are some things that I think should happen next in this PR:

i) Install the "right" parsl version

These Github Actions are run to test specific new versions of parsl that only exist in pull requests. This step in your newly added Actions workflow checks out the correct version of Parsl (with whatever changes the submitter has made):

```
      - name: Checkout
        uses: actions/checkout@v3
```

but later on you install the latest official release of Parsl (as released on Pypi.org), not the version that was checked out from the pull request:

```
      - name: Install Parsl
        run: |
          pip3 install parsl
```

To install the version that is checked out by the checkout step (instead of using the version from Pypi), I think it will be enough to instead call `pip3 install .`  - in that command, `.` means "install whatever is in the current directory`, and we hope it will be parsl... if you look in .github/workflows/ci.yaml you should both a checkout action and then later on some `pip install .` commands, and basically I think you should do something like that.


ii) Run some tests to check you installed parsl right.

You can check that the basic parsl install worked, before running anything with parsl+flux, with `make local_thread_test` (which actually just runs: `pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/local_threads.py --random-order --durations 10`)  - that should run for about 10 seconds, and end like this:

```
=============================================== 148 passed, 222 skipped, 4 warnings in 5.94s ================================================
```

You might encounter some problems here with stuff not being installed right (missing dependencies, for example)

iii) test parsl+flux

There's one basic flux test file here `parsl/tests/test_flux.py` so I think you should be able to run:

```
pytest parsl/tests/test_flux.py
```

Extras to think about after the above is working, but don't attempt until the basics above work:

iv) These tests have a problem that I dislike: if they detect flux is not working, they will not fail (and so CI will not fail, and so developers will not know that flux is no longer being tested). Probably in a separate PR after this PR is done, we can work on changing that behaviour.

v) Our other executors, we try to run a big range of general Parsl tests to check that lots of different parsl usage patterns and options work with this executor - that would be a useful thing to add in. If you look in the makefile, you will see various executors tested by running `pytest` with `--config <some config>` configured for different kinds of executors, and it would be good to make that work for a flux configuration too.

mercybassey:
Hi @benclifford I've implemented the `pip3 install parsl .` and integrated the installation of dependencies from `test-requirements.txt` into the CI pipeline, so all necessary libraries for testing are installed which worked well.

Following that, I attempted to verify the Parsl installation by running a subset of the test suite as suggested (pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/local_threads.py --random-order --durations 10). However, I've encountered a persistent issue where the test fails due to an unrecognized argument `--random-order`.  After adjusting the test command to remove potentially problematic flags, it did not resolve the issue.

Given this context, I'll be needing your suggestions on potential adjustments or alternative approaches to verify the Parsl installation.

benclifford:
The implementation for --random-order is installed from a extra pytest plugin module: https://pypi.org/project/pytest-random-order/

that's usually installed as part of the testing requirements... in the main Parsl dev instructions, that happens as part of "make deps", or you could change your test install to `pip install . -r test-requirements.txt` which will install both `.` (the parsl code under test) and also all the dependencies in test-requirements.txt

mercybassey:
@benclifford During the CI run, I encountered a test failure in `parsl/tests/test_bash_apps/test_stdout.py`, which asserts a failure condition related to handling stdout and stderr paths for bash apps.

`=================================== FAILURES ===================================
_____________________________ test_bad_stderr_file _____________________________

    @pytest.mark.issue363
    def test_bad_stderr_file():
        """Testing bad stderr file"""
    
        err = "/bad/dir/t2.err"
    
        fn = echo_to_streams("Hello world", stderr=err)
    
        try:
            fn.result()
        except perror.BadStdStreamFile:
            pass
        else:
>           assert False, "Did not raise expected exception BadStdStreamFile"
E           AssertionError: Did not raise expected exception BadStdStreamFile
E           assert False

parsl/tests/test_bash_apps/test_stdout.py:70: AssertionError
----------------------------- Captured stdout call -----------------------------
Hello world`

Upon investigating, I identified that the failure was due to the test expecting an exception when attempting to write to a non-existent directory, but the exception was not raised as expected. 

Could you please provide some guidance on how you would like me to address this test failure? Should I attempt to modify the test or the underlying functionality to resolve the issue, or is there a different approach you'd prefer?

benclifford:
that test looks like this problem some people have had on WSL2: https://github.com/Parsl/parsl/issues/3160

in that test, the assumption is that `/` will not be writable, which is true on a traditional multi-user unix system. but apparently not true for that WSL users in #3160, and looks like its not true in the flux container that you're using.

I think seeing this test break in two different places is good motivation to fix the test - maybe you're interested in a side-quest to fix #3160 (as a separate PR) which would help both the WSL users and this flux testing.

mercybassey:
> that test looks like this problem some people have had on WSL2: #3160
> 
> in that test, the assumption is that `/` will not be writable, which is true on a traditional multi-user unix system. but apparently not true for that WSL users in #3160, and looks like its not true in the flux container that you're using.
> 
> I think seeing this test break in two different places is good motivation to fix the test - maybe you're interested in a side-quest to fix #3160 (as a separate PR) which would help both the WSL users and this flux testing.

Okay, ill take a look

mercybassey:
@benclifford, based on your suggestions on #3160  I created two test cases to address the issue. The first test verifies the writability of a directory before attempting a write operation. The second test, `test_write_to_non_writable_directory` is designed to validate that it handles and raises exceptions when writing to such directories.

Despite these efforts and confirming the expected behavior locally, the CI run does not yield the expected `BadStdStreamFile` exception for the `test_write_to_non_writable_directory` case.

benclifford:
ah, that's interesting and I think I see what's happening there -- I think your tests in the new environment are running as the `root` user, and that user is allowed to write everywhere, overriding file system permissions.

So I'll have to have a think about this test a lot more.

For now, to proceed with your development work testing flux, I think you should comment out the failing tests with a note about why, and make further progress with flux testing, and I will work on a better solution to the failing test. Then before you merge, we'll fix up this test in your branch with whatever I figure out.

mercybassey:
Hi @benclifford Following your guidance for testing Parsl with Flux, I initially ran `pytest parsl/tests/test_flux.py` in the CI setup. However, this attempt resulted in a failure due to a missing `--config` argument requirement. So I executed `pytest parsl/tests/test_flux.py --config parsl/tests/configs/local_threads.py`, which worked.

While this solution worked, I realized that using the `local_threads.py` configuration may not directly apply or fully test the integration with `FluxExecutor` which seems to align with your suggestion about testing a wide range of general Parsl tests with different executors to ensure broad compatibility and functionality.

benclifford:
see here in your test output:

```
parsl/tests/test_flux.py ssssssssss                                      [100%]
```

Those `s` symbols mean "skipped`. Have a look at the other test runs and you should (mostly) see that tests that pass are labelled with `.` (although some are skipped for various reasons)

Scroll down a few lines and you'll see this:

```
SKIPPED [1] parsl/tests/test_flux.py:34: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:47: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:58: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:71: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:92: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:108: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:120: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:135: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:152: intended for local config
SKIPPED [1] parsl/tests/test_flux.py:165: intended for local config
```

which gives the reason for each of those tests being skipped, and it turns out to be exactly your other observation:

> I realized that using the local_threads.py configuration may not directly apply

Basically our tests are divided into two broad types, all mingled together: some parsl tests can run with any configuration and we want to try those tests with lots of different configurations, specified by the `--config` parameter.

Some tests, however, don't work that way: they either bring along their own specific configuration (or sometimes don't even need a full Parsl configuration at all). Those tests get run with `--config local`. (Its why you'll probably see people have the most problems getting `--config local` tests running more than other tests, because they test a bunch of very specific configurations which often have quite specific dependencies).

In the case of the flux tests, they're implemented as tests that "know" they are only for testing flux, and bring along as much or as little configuration as they need - if you look through the test, you'll see some tests instantiate a `FluxExecutor` which is the top level component under test, but some other tests in there do some more specific unit testing of individual classes and don't even need a flux executor.

So `--config local` should get you further along here.

I suspect from looking at how the tests were run in the yaml that came from the Flux team that you're going to have to do something with that `flux start` command to get flux running before the tests will pass, but I don't have a lot of experience there -- so you might have to do some reading about Flux yourself to figure it out.

mercybassey:
Hi @benclifford the CI ran to completion. Initially, I added a step to start Flux with some parameters to ensure initialization, but encountered an error: 

`Run flux start --test-size=2
flux-broker: stdin is not a tty - can't run interactive shell
flux-start: 0 (pid 823) exited with rc=1
flux-start: 1 (pid 824) Killed
Error: Process completed with exit code 137.
` 
So, I configured Flux to start and run the tests in a single command. This approach worked, but the job was running indefinitely. I couldn't find a way to cancel the long-running job via the GitHub Actions UI manually. I then added a concurrency configuration to the CI workflow so that any in-progress jobs would be automatically canceled when new commits are pushed (which I will remove now that the CI has run to completion).

After revisiting the example YAML provided by @vsoch, I noticed the `-vvv` flag used in his setup. I included the `-vv` flag in my implementation for a less verbose output than @vsoch's, so I could see why the step was running without exiting. With this adjustment, the CI ran to completion.

benclifford:
>  This approach worked, but the job was running indefinitely.

In the main parsl CI, the approach to this is the yaml configures a 60 minute timelimit because it's fairly common for buggy code to hang there.

benclifford:
OK this is looking good and almost ready to merge, I think.

I'm extremely suspicious of any kind of test result thats look like "I ran it, and it hung; i ran it again with a change that shouldn't change anything (such as changing log output) and it did not hang" - that's often a result of some kind of race condition bug somewhere in the code, where the tests don't always fail, but that you should treat the occasional failure (or hang) as a real bug that users are going to encounter, just not every time.

Can you run this a few more times (with different `-v` log levels like you tried already) and see if there are any more hangs? (the point of these tests is to discover problems like this, not just to mindlessly pass...)

The tests run successfully in 2 minutes, so you might set a 5 minute timeout - (see the main parsl ci.yaml for how that is set)

benclifford:
OK, the most recent one hung - if I look at the test results with timestamps turned on, I see this:

```
2024-03-08T12:55:12.7492058Z parsl/tests/test_flux.py::test_future_callback_flux_exception PASSED     [ 10%]
2024-03-08T12:55:12.7510181Z parsl/tests/test_flux.py::test_future_cancel_no_underlying_future PASSED [ 20%]
2024-03-08T12:55:12.7526443Z parsl/tests/test_flux.py::test_future_cancel PASSED                      [ 30%]
2024-03-08T12:55:13.2387868Z parsl/tests/test_flux.py::test_cancel PASSED                             [ 40%]
2024-03-08T12:55:13.2401628Z parsl/tests/test_flux.py::test_future_callback_nofile PASSED             [ 50%]
2024-03-08T12:55:16.2338552Z parsl/tests/test_flux.py::test_multiply PASSED                           [ 60%]
2024-03-08T12:55:17.6339861Z parsl/tests/test_flux.py::test_affinity PASSED                           [ 70%]
2024-03-08T12:55:19.1458371Z parsl/tests/test_flux.py::test_except PASSED                             [ 80%]
2024-03-08T12:55:19.1471496Z parsl/tests/test_flux.py::test_future_callback_returncode PASSED         [ 90%]
2024-03-08T12:58:06.0629403Z parsl/tests/test_flux.py::test_future_running 
2024-03-08T12:58:06.0724215Z ##[error]The operation was canceled.
```

so that's good in the sense that this test finding problems, like it is meant to!

benclifford:
My gut feeling from this is there is some kind of hang at shutdown here. I think it would be good to merge this PR (with occasionally hanging tests) and then open a new issue about the demonstrated behaviour.

std.out
```diff
@@ -0,0 +1 @@
+hello there
```
benclifford, line 1: this file shouldn't be added in your PR, probably added by mistake

test.memo.stdout.x
```diff
@@ -0,0 +1 @@
+X
```
benclifford, line 1: this file shouldn't be added in your PR, probably added by mistake

parsl/tests/test_bash_apps/test_stdout.py
```diff
@@ -34,42 +34,74 @@ def echo_to_streams(msg, stderr=None, stdout=None):
...
-
-    try:
-        fn.result()
-    except perror.BadStdStreamFile:
-        pass
-    else:
-        assert False, "Did not raise expected exception BadStdStreamFile"
-
-    return
+# Skipping these tests temporarily due to root user permissions
```
benclifford, line 37: I'll have to do something about this - to merge this PR, this test will have to be re-enabled, so I will try to work on that to unblock things.

test.memo.stdout.x
mercybassey: This is a mistake. I'll remove it.

benclifford:
my gut feeling is this hang is happening at shutdown of the test suite, rather than in the individual test cases - for whatever reason, shutdown often ends up being unreliable in various pieces of code in our environment. If you're interested in digging into that, you poke inside this code a bit more (although debugging this sort of hang is often frustrating and unrewarding).

If you want to do that, if you have a docker environment handy, you can probably try to recreate this hang in your own environment - I think this is just a regular docker image being used here - and then try to trace through the code, and add more logging to try to understand where it is hanging...

mercybassey:
> my gut feeling is this hang is happening at shutdown of the test suite, rather than in the individual test cases - for whatever reason, shutdown often ends up being unreliable in various pieces of code in our environment. If you're interested in digging into that, you poke inside this code a bit more (although debugging this sort of hang is often frustrating and unrewarding).
> 
> If you want to do that, if you have a docker environment handy, you can probably try to recreate this hang in your own environment - I think this is just a regular docker image being used here - and then try to trace through the code, and add more logging to try to understand where it is hanging...

Okay, I'll take a look.

mercybassey:
@benclifford running the command `flux start pytest parsl/tests/test_flux.py --config local --random-order` with and without verbosity does not hang. See the following screenshots.

benclifford:
ok. that's frustrating. I've encountered problems before which I can only make happen inside the CI environment (sometimes I think because the CI runners are relatively underpowered virtual machines, which flushes out a different kind of race condition). You can make a new draft PR with a fork of this #3159 work and hack away inside that if you want to be able to try out changes to debug this more in the CI environment. Don't feel obliged to do that though - it's a pretty painful debugging experience.

mercybassey:
> ok. that's frustrating. I've encountered problems before which I can only make happen inside the CI environment (sometimes I think because the CI runners are relatively underpowered virtual machines, which flushes out a different kind of race condition). You can make a new draft PR with a fork of this #3159 work and hack away inside that if you want to be able to try out changes to debug this more in the CI environment. Don't feel obliged to do that though - it's a pretty painful debugging experience.

I'll give it a try

mercybassey:
Hi, @benclifford I think I have resolved the issue. I used an Ubuntu version `20.04`; the test passed and exited without verbosity. Here's the [draft](https://github.com/Parsl/parsl/pull/3218).

mercybassey:
> so what you're saying is, if a user wants to run flux, they have to install a different operating system? that's quite frustrating (by which, I mean, you cannot seriously tell a user that is the solution to their hang) but it's interesting that changing this has changed whether the hang happens or not.

I'm currently running Ubuntu 22.04 LTS on my local machine, where the tests execute smoothly (no hangs). However, it appears that any Ubuntu version beyond `20.04` encounters problems when running in the CI environment.

benclifford:
ok this is good information to help with debugging. I'll try to have a poke at this hang in the next few days too.

another thing you might look at is this that I mentioned right near the start:

> v) Our other executors, we try to run a big range of general Parsl tests to check that lots of different parsl usage patterns and options work with this executor - that would be a useful thing to add in. If you look in the makefile, you will see various executors tested by running pytest with --config <some config> configured for different kinds of executors, and it would be good to make that work for a flux configuration too.

This would run a completely different set of tests than `--config local` and hopefully give the flux executor a more intense test. It might also reveal some other problems that haven't been detected by other testing.

mercybassey:
Hi @benclifford Could you please clarify if there is a standard Flux configuration available for testing the Flux executor, or should I proceed with creating a custom configuration?

Makefile
```diff
@@ -120,6 +120,14 @@ coverage: ## show the coverage report
 	# coverage report
 	echo no-op coverage report
 
+
 .PHONY: clean
 clean: ## clean up the environment by deleting the .venv, dist, eggs, mypy caches, coverage info, etc
-	rm -rf .venv $(DEPS) dist *.egg-info .mypy_cache build .pytest_cache .coverage runinfo $(WORKQUEUE_INSTALL)
+	parsl+flux
```
benclifford, line 126: what is this parsl+flux line for?

mercybassey: Hi @benclifford this is a mistake, I'll proceed with removing it.

Makefile
```diff
@@ -120,6 +120,13 @@ coverage: ## show the coverage report
 	# coverage report
 	echo no-op coverage report
 
+
 .PHONY: clean
 clean: ## clean up the environment by deleting the .venv, dist, eggs, mypy caches, coverage info, etc
-	rm -rf .venv $(DEPS) dist *.egg-info .mypy_cache build .pytest_cache .coverage runinfo $(WORKQUEUE_INSTALL)
+	rm -rf .venv $(DEPS) dist *.egg-info .mypy_cache build .pytest_cache .coverage runinfo_* $(WORKQUEUE_INSTALL)
```
benclifford, line 126: This line is merged badly - a recent PR #3236 removed that _ on the end of runinfo

parsl/tests/test_bash_apps/test_stdout.py
benclifford: @mercybassey ok, I think the right thing to do for now here is instead of invoking `make local_thread_test` (which then invokes `pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/local_threads.py --random-order --durations 10`), the Github Actions yaml could invoke this command instead:

 `pytest parsl/tests/ -k "not cleannet and not issue363" --config parsl/tests/configs/local_threads.py --random-order --durations 10`

That would skip a bunch of tests related to handling of stderr and stdout - which is fine, as this line is only to verify that basic Parsl runs before running the real tests against flux. For example, taskvine and work queue tests also skip those tests marked `issue363` as well, in our main test run.

Then you don't need to modify this file test_stdout.py at all and can set it back to how it is in `master`

benclifford:
OK, this PR is almost ready to merge. There are some tidyup comments that you marked as resolved without actually resolving them. I'll re-open those. And I wrote about how I think that broken test case should be handled.

benclifford:
@mercybassey have a look through the files here https://github.com/Parsl/parsl/pull/3159/files when you think you've got all the extra files removed, and check they really are gone.

benclifford:
looks like recent tests are hitting that hang-at-exit problem again. I'll have a look at that right now.

mercybassey:
@benclifford the failing test passed this time.

benclifford:
> @benclifford the failing test passed this time.

yes annoying. but we can merge this PR, and just accept that these tests will sometimes fail this way until someone diagnoses and fixes whatever is broken.  so I think if you do the tidyups of messy files in this PR, it would then be ready to merge.

mercybassey:
> > @benclifford the failing test passed this time.
> 
> yes annoying. but we can merge this PR, and just accept that these tests will sometimes fail this way until someone diagnoses and fixes whatever is broken. so I think if you do the tidyups of messy files in this PR, it would then be ready to merge.

@benclifford I have removed all the messy files.

benclifford:
While trying to understand what's happening with hung runs, I realised that this PR does not record test artefacts in the same way that the main github actions workflow does. I put this commit onto my testing branch to capture a test artefact when a run finishes, with the hope that when I later see a hanging run it will give some more clues.

Artefacts like this have been really useful for debugging tests in general.

You can look in ci.yaml to see what I copied this from

https://github.com/Parsl/parsl/pull/3259/commits/a41b732cdc82eb4b85336d7da0f76d394847f86a

mercybassey:
Hi @benclifford. Are there any updates on this issue? I'd like to begin working on this [one](https://github.com/Parsl/parsl/issues/3189).

benclifford:
@mercybassey you can work on the os x model before this PR #3159 is merged, on another branch - I need to write up an issue for what I discovered with flux.

.github/workflows/parsl+flux.yaml
```diff
@@ -0,0 +1,47 @@
+name: Test Flux Scheduler
+on:
+  pull_request: []
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    permissions:
```
khk-globus, line 8: This is now 4 years out of date.  Are we interested in using a more recent distro for the test and CI framework?

khk-globus (approved):

.github/workflows/parsl+flux.yaml
benclifford: The flux container image is tagged for `jammy` which is the subsequent ubuntu LTS release (22.04).  The most recent ubuntu LTS is 24.04 which was released just over a month ago, and it looks like there isn't a flux container for that (on https://hub.docker.com/r/fluxrm/flux-sched/tags).

So I'll upgrade this to 22.04 and see what happens.

benclifford:
I pulled some of the test changes I made here into PR #3483 for merge first separately.

PR #3165, opened 2024-03-06
Stop DFK cleanup in atexit, because this doesn't work in Python >3.12

error-4u:
# Description

atexit handlers work differently in the latest version of Python (version 3.12) and it is not safe to automatically call cleanup any more in Python 3.12

# Changed Behaviour

Instead of cleaning up, the code will now emit a WARNING log message to help users understand the change.

# Fixes

Fixes # (https://github.com/Parsl/parsl/issues/3104)

## Type of change

- Bug fix


error-4u:
should i reduce it to 150 @benclifford

benclifford:
@error-4u you just need to make the line shorter than 151 characters to make the error go away - any length shorter than 151 will be ok. For example, I can see that you used `.` at the end of your log message to end the sentence, but in most log messages in Parsl, log messages do not end that way - so removing that `.` would both make your line shorter *and* make it more consistent with other Parsl log messages

error-4u:
i have to make a seperate branch for flake8 or can work in the atexit branch only @benclifford

benclifford:
use your atexit-fix branch, because the problem is with the code you added on that branch. for example:

git fetch original
git merge original/master
make flake8

and you should see the error.

then fix the error in `dflow.py` without changing anything else

run `make flake8` again and see that the error is fixed

then `git commit` and `git push origin HEAD`

benclifford (approved):

Issue #3161, opened 2024-03-06
SimpleLauncher should help the user understand why they are not getting multiple nodes


benclifford:
**Is your feature request related to a problem? Please describe.**

When Parsl runs workers inside a batch system, each block of workers can contain multiple compute nodes.

In order to run across all of the compute nodes inside that block, users can specify a `Launcher`, which is usually specific to the particular kind of HPC system that they are using: https://parsl.readthedocs.io/en/stable/reference.html#launchers

There is also the `SimpleLauncher` which doesn't try to run workers across multiple nodes, but only runs on the single node it is already running on.

This can be confusing to users when asking for multiple worker nodes, which are then unused if they do not specify the correct launcher but instead choose the SimpleLauncher.

**Describe the solution you'd like**

SimpleLauncher should inspect its `nodes_per_block` parameter and raise an exception if users are trying to use it with multiple nodes in a block, rather than 1 node per block.

Sometimes advanced users genuinely do want to run this way (for example in @yadudoc 's HTEX MPI prototype) and so the `__init__` method should get a new parameter "permit_multiple_nodes" which should allow users to specify multiple nodes per block without getting this exception. This new parameter should have a default value of `False` so that users who do not specify this option will get the new exception, and the old behaviour is only seen by advanced users.

In addition to implementing this, there should be three test cases:
* test that the launcher works with one node per block specified
* test that the launcher raises an exception with some other number of nodes per block, for example 64 nodes per block
* test that advanced users can disable the exception when using 64 nodes per block but setting permit_multiple_nodes=True

**Additional context**
this comes from discussion with @Andrew-S-Rosen

Andrew-S-Rosen:
CCing @WardLT since the exception (and proposed kwarg to circumvent it) is relevant to his workflows as well.

Andrew-S-Rosen:
Food for thought: would it not make more sense to log a warning?

benclifford:
Other similar code, the wrapped launcher, logs a warning in similar circumstances and it would be more internally consistent to do follow that style. Mostly this comes from my preference of "you aren't going to read the warnings, and this is almost definitely the wrong thing to be doing". But yeah, I'd be fine with logging a warning like the wrapped launcher does a few lines below.

benclifford:
That is, my view is:
The behaviour right now is you are asking the launcher to do X and it cannot do X, but instead it does Y, which is not X. It might be more desirable to raise an exception that X cannot be done, rather than doing something else that is not X.

[...]

PR #3167, opened 2024-03-06
Updated simpleLauncher Function 

christailu:
# Description

The SimpleLauncher class has been updated to raise an exception when users attempt to use it with multiple nodes per block without specifying the `permit_multiple_nodes` parameter. Additionally, the __init__ method of SimpleLauncher now includes a new parameter `permit_multiple_nodes` to allow advanced users to disable this exception and use multiple nodes per block if needed.

# Changed Behaviour

After this change, users attempting to use SimpleLauncher with multiple nodes per block without setting `permit_multiple_nodes=True` will receive an exception. This behavior ensures that users are aware of the limitations of SimpleLauncher and can make informed decisions when specifying launcher options.

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/docs/userguide/execution.rst
+++ b/docs/userguide/execution.rst
@@ -39,7 +39,7 @@ parameters include access keys, instance type, and spot bid price
 Parsl currently supports the following providers:
 
 1. `parsl.providers.LocalProvider`: The provider allows you to run locally on your laptop or workstation.
-2. `parsl.providers.CobaltProvider`: This provider allows you to schedule resources via the Cobalt scheduler.
+2. `parsl.providers.CobaltProvider: This provider allows you to schedule resources via the Cobalt scheduler. Please note that the CobaltProvider is deprecated and will be removed by 2024.04.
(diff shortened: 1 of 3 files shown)


docs/userguide/execution.rst
```diff
@@ -39,7 +39,7 @@ parameters include access keys, instance type, and spot bid price
 Parsl currently supports the following providers:
 
 1. `parsl.providers.LocalProvider`: The provider allows you to run locally on your laptop or workstation.
-2. `parsl.providers.CobaltProvider`: This provider allows you to schedule resources via the Cobalt scheduler.
+2. `parsl.providers.CobaltProvider: This provider allows you to schedule resources via the Cobalt scheduler. Please note that the CobaltProvider is deprecated and will be removed by 2024.04.
```
benclifford, line 42: this looks you like were also working on a different issue, but in the same git branch? This isn't part of issue #3161, and you should try to keep pull requests focused on just one thing

PR #3172, opened 2024-03-06
Reduce flake8 max-line-length to 150

marameref:
# Description

Reduced flake-8 maximum length (to 150) and refactored some lines which which were not compatible with the new configuation

# Changed Behaviour

This PR wont change any behaviour of the project

# Fixes

Fixes https://github.com/Parsl/parsl/issues/3105

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/docs/userguide/mpi_apps.rst
+++ b/docs/userguide/mpi_apps.rst
@@ -1,14 +1,6 @@
 MPI Apps
 ========
 
-.. note::
-
-    Parsl's support for MPI Apps described below is pending release.
-    Please use the ``mpi_experimental_3`` branch to use the functionality
-    described in this document. To install directly from github:
-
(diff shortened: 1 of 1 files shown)


benclifford (changes requested):
ok, this is generally good, but:

* there are a few files added by mistake. look here: https://github.com/Parsl/parsl/pull/3172/files and you will see a few files (like hello1.txt) that I think you did not intend to add. you can remove those with `git rm`.

* you need to merge in the latest changes from GitHub master - there will be a few conflicts there, because someone recently made a PR to reduce the max-line-length from 152 to 151 (in PR #3148). You will need to resolve the conflicts that happen when you merge that latest Parsl `master` into your branch. Be careful to keep their changes from #3148 and to not overwrite them with your own, in db_manager.py

PR #3177, opened 2024-03-06
 Upadated the SimpleLauncher Class

christailu:
# Description

This pull request enhances the behavior of the SimpleLauncher class in the Parsl library. It addresses the issue where users could inadvertently attempt to use the SimpleLauncher with multiple nodes per block, which could lead to confusion and unexpected behavior. This enhancement ensures that an exception is raised when users try to use the SimpleLauncher with more than one node per block, unless explicitly permitted.

# Changed Behaviour

After this PR, if users attempt to use the SimpleLauncher with multiple nodes per block without explicitly permitting it, an exception will be raised. This helps clarify the behavior of the SimpleLauncher and prevents potential confusion for users.

# Fixes

Fixes # (issue)

## Type of change

- Bug fix

### DIFF
--- a/parsl/launchers/launchers.py
+++ b/parsl/launchers/launchers.py
@@ -8,7 +8,7 @@ logger = logging.getLogger(__name__)
 class SimpleLauncher(Launcher):
     """ Does no wrapping. Just returns the command as-is
     """
-    def __init_(self, debug: bool = True) -> None
+    def __init_(self, debug: bool = True) -> None:
         super().__init__(debug=debug)
 
     def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> str:
--- a/parsl/launchers/launchers.py
+++ b/parsl/launchers/launchers.py
@@ -8,16 +8,19 @@ logger = logging.getLogger(__name__)
 class SimpleLauncher(Launcher):
     """ Does no wrapping. Just returns the command as-is
     """
-    def __init_(self, debug: bool = True) -> None:
+    def __init__(self, debug: bool = True) -> None:
         super().__init__(debug=debug)
 
-    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> str:
+    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int, permit_multiple_nodes: bool = False) -> str:
         """
         Args:
         - command (string): The command string to be launched
-        - task_block (string) : bash evaluated string.
-
+        - tasks_per_node (int): Number of tasks to launch per node
+        - nodes_per_block (int): Number of nodes per block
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Hi @christailu! Thanks for this. While I will not be the one to formally review this PR, I did want to point out that ultimately, it would likely be best to log a warning rather than raise an error, as noted in https://github.com/Parsl/parsl/issues/3161#issuecomment-1980399458.

christailu:
@benclifford , I tried to update the class, can you take a look at what should be improved? thanks!

parsl/launchers/launchers.py
```diff
@@ -8,16 +8,20 @@
 class SimpleLauncher(Launcher):
     """ Does no wrapping. Just returns the command as-is
     """
-    def __init_(self, debug: bool = True) -> None:
+    def __init__(self, debug: bool = True) -> None:
```
benclifford, line 11: that's an interesting bugfix  - i hadn't noticed this was broken!

parsl/launchers/launchers.py
```diff
@@ -8,16 +8,20 @@
...
         Args:
         - command (string): The command string to be launched
-        - task_block (string) : bash evaluated string.
-
+        - tasks_per_node (int): Number of tasks to launch per node
+        - nodes_per_block (int): Number of nodes per block
+        - permit_multiple_nodes (bool): Whether to allow multiple nodes per block
         """
+        if nodes_per_block > 1 and not permit_multiple_nodes:
+            logger.warning("SimpleLauncher only supports 1 node per blok. "
```
benclifford, line 23: spell block "block"

parsl/launchers/launchers.py
```diff
@@ -8,16 +8,20 @@
...
-    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> str:
+    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int, permit_multiple_nodes: bool = False) -> str:
         """
         Args:
         - command (string): The command string to be launched
-        - task_block (string) : bash evaluated string.
-
+        - tasks_per_node (int): Number of tasks to launch per node
+        - nodes_per_block (int): Number of nodes per block
+        - permit_multiple_nodes (bool): Whether to allow multiple nodes per block
```
benclifford, line 20: In the discussion of issue #3161 , @Andrew-S-Rosen and I concluded that we probably didn't need this parameter if issuing a warning (instead of raising an exception) so it would probably be good to remove this extra complexity.

benclifford (approved):

benclifford:
I updated your description @christailu and merged this PR.

Issue #2604, opened 2023-02-20
docs source links often point to missing/incorrect source file


benclifford:
**Describe the bug**

Example:
condorprovider docs point to non-existent source https://parsl.readthedocs.io/en/stable/stubs/parsl.providers.CondorProvider.html -> http://github.com/Parsl/parsl/blob/master/parsl/providers.py

benclifford:
docs/conf.py contains this function which resolves a link into a github link. It looks like it is making the assumption that an imported module parsl.provider.CondorProvider would have sourcecode at parsl.provider.~~CondorProvider~~, which is untrue in two ways:

CondorProvider is re-exported from parsl.provider and the source code is defined in a different import.

The source for parsl.provider lives in parsl/provider/__init__.py, not in parsl/provider.py

```
def linkcode_resolve(domain, info):
    if domain != 'py':
        return None
    if not info['module']:
        return None
    filename = info['module'].replace('.', '/')
    return "http://github.com/Parsl/parsl/blob/master/{}.py".format(filename)
```

christailu:
Hey @benclifford , I have tried to fixe the bug, can take a look please?#3179 Would like some feedbaack

PR #3179, opened 2024-03-06
Fixed the bug for linkcode_resolve function

christailu:
# Description

This change adjusts the linkcode_resolve function in the conf.py file to properly handle the mapping of module names to their corresponding GitHub URLs. Specifically, it addresses the issue where the assumption about the source code location for CondorProvider is incorrect.

# Changed Behaviour

After this change, when generating documentation, the links to the source code on GitHub for modules like parsl.provider.CondorProvider will correctly point to the appropriate location, ensuring accurate reference to the source code.

# Fixes

Fixes # (#2604 )

## Type of change

- Bug fix

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -52,8 +52,24 @@ def linkcode_resolve(domain, info):
         return None
     if not info['module']:
         return None
-    filename = info['module'].replace('.', '/')
-    return "http://github.com/Parsl/parsl/blob/master/{}.py".format(filename)
+    
+    module = info['module']
+    fullname = info['fullname']
+    obj = sys.modules.get(module)
+    if obj is None:
+        return None
+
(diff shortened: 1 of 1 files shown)


yadudoc:
Tagging @benclifford because I'm not super familiar with what exactly this bug is fixing.

benclifford (changes requested):
Have you tested this PR? You should be able to see, first with `master` Parsl, that the source code link described in the original issue is broken - when you click on the source code link you should get some kind of error page from GitHub; and then you should be able to see that when you build the documentation with your change, that the source code link now leads to the source code of the relevant class (CondorProvider is one class mentioned explicitly in the original issue, but there are other cases too)

benclifford:
@christailu do you still want to work on this PR?

PR #3186, opened 2024-03-07
Reduce flake8 max-line-length

dawit-andargachew:
# Description
Reduce flake8 max-line-length
When I set up the project, the max line length was 152 and I decreased it by one.

# Changed Behaviour
max line length is reduced by 1 and creates an error when we run this command
`make flake8`

# Fixes
Reduce flake8 max-line-length

Fixes # (issue)
[#3105](https://github.com/Parsl/parsl/issues/3105)

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@
 # W504: line break after binary operator
 #       (Raised by flake8 even when it is followed)
 ignore = E126, E402, E129, W504
-max-line-length = 151
+max-line-length = 150
 exclude = test_import_fail.py,
   parsl/executors/workqueue/parsl_coprocess.py
 # E741 disallows ambiguous single letter names which look like numbers


benclifford (changes requested):
when you've reduced the maximum line length by one, and now see errors in flake8, you shoudl go through and fix each of those errors, until flake8 is working again.

Then submit those fixes as part of this pull request.

dawit-andargachew:
> when you've reduced the maximum line length by one, and now see errors in flake8, you shoudl go through and fix each of those errors, until flake8 is working again.
> 
> Then submit those fixes as part of this pull request.

dawit-andargachew:
Thank you for the suggestion. 

I had the impression that the task simply required changing the value, which would have been a very trivial task if I had been aware that fixing those errors was the main task.

I understand the true nature of the task, I am more than willing to contribute to the project in any way I can.

PR #3203, opened 2024-03-07
Reduce flake8 max-line-length to 147

Liz-Adhiambo:
# Description

Reduce flake8 max-line to 147

# Changed Behaviour

No change

# Fixes

Fixes # (#3105)

## Type of change

- Code maintenance/cleanup


Liz-Adhiambo:
Hi @benclifford could you please review  the changes I've made? I'll appreciate your feedback and thoughts on them.

yadudoc (changes requested):
Hi @Liz-Adhiambo, your changes look all look correct to me. I appreciate the effort on tracking the changes on master and updating the PR, nice job!

Right now, it looks like there are a couple of merge conflicts, could you take fix those please? Once those are fixed, I think this PR will be ready to merge. 

Some notes on fixing merge conflicts on github: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github

Liz-Adhiambo:
@yadudoc,  I have fixed the conflicts. Thank you for the notes

yadudoc:
@Liz-Adhiambo Please note that the line limit is at 147 now.

Liz-Adhiambo:
@yadudoc I have reduced the max-line length to 146 now . I did experience sort of a blocker when I tried working on line 496 in the monitoring.py file, breaking the line always gave me a statement is unreachable error so I opted to shorten the word "message" to "msg". Are there any other solutions I could have tried? What do you think?

benclifford:
Changing the word "message" to "msg" seems like the wrong way to fix a "statement is unreachable" error - changing just a word shouldn't change whether a statement is reachable or not!

Have a closer read of the code, and of the error messages you got from mypy, and see if you can understand what was happening before and after your change.

Liz-Adhiambo:
@benclifford Okay  let me try that again

Liz-Adhiambo:
I tried a different approach, introducing a variable name for the error message. What do you think?

parsl/monitoring/monitoring.py
```diff
@@ -429,6 +429,7 @@ def start(self,
 
                         msg_0: AddressedMonitoringMessage
                         msg_0 = (msg, 0)
+                        unknown_type_mg = f"Discarding message from interchange with unknown type {msg[0].value}"
```
benclifford, line 432: It would be better style here to format the message only if it is needed - for example right before the error message is logged around line 452

benclifford (changes requested):

parsl/monitoring/monitoring.py
Liz-Adhiambo: I had a look at this and since msg_0 is initialized at line 475 the best possible place for me to initialize the error message was line 476. I looked at another possible solution and realized that breaking the line and having the '# type: ignore[unreachable]' comment on the first line of the broken sentence solves the unreachable issue I previously had. I don't know why I had not thought of this before

parsl/monitoring/monitoring.py
```diff
@@ -449,7 +449,8 @@ def start(self,
                             # but there is no verification that the message
                             # received from ic_channel.recv_pyobj() is actually
                             # of that type.
-                            self.logger.error(f"Discarding message from interchange with unknown type {msg[0].value}")  # type: ignore[unreachable]
+                            self.logger.error(f"Discarding message"  # type: ignore[unreachable]
```
benclifford, line 452: You don't need that `f` on the front because this string isn't being formatted - only the next one is. That's not a breaking problem, but its another style thing.

This code is also missing out on a space after the word "message" and before the word "from" because putting two strings next to each other doesn't add a space in automatically.

Liz-Adhiambo: Oh yes, I just updated my code to reflect this

benclifford (approved):

PR #3205, opened 2024-03-07
TaskVine: add config and links to CCL documentation

colinthomas-z80:
# Description

Adding a short bit about TaskVine to the config documentation.

## Type of change


colinthomas-z80:
The warnings in the log had escaped me the first time around...

docs/userguide/configuring.rst
```diff
@@ -493,12 +493,12 @@ This system uses Grid Engine which Parsl interfaces with using the `parsl.provid
 .. literalinclude:: ../../parsl/configs/cc_in2p3.py
 
 
-CCL (Notre Dame, with Work Queue)
----------------------------------
+CCL (Notre Dame, TaskVine - Work Queue)
```
benclifford, line 496: I think your changes look like they're a mix of "we are going to tell people to use task vine and not tell them about work queue any more" and "we're going to tell people about *both* work queue and task vine". I think you should pick one of those two approaches: either switch this documentation to be entirely task vine, or be clear that you are telling people about two different executors (that, mostly irrelevantly from a user perspective,  just happen to be written by the same humans).

It's probably lower maintenance to "task vine only" in this documentation?

benclifford (changes requested):

docs/userguide/configuring.rst
colinthomas-z80: I understand. We certainly want to encourage TaskVine usage, though I do not think it would be entirely helpful to remove reference to WorkQueue altogether. I think a couple sentences and link will suffice.

benclifford (approved):

PR #3210, opened 2024-03-07
Corrected type hint

WelileDlamini:
# Description

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup


benclifford (changes requested):
This PR changes a lot of things in small ways, without being very clear *why* you are making those changes:

for example, is there a bug you are trying to fix?
Does that bug have an issue number? If so write, the issue number in the description of this PR. If not, write a short description of the bug you are trying to fix.

If you aren't trying to fix a bug, but have some other motivation, then write that motivation in the description of the PR.

Many of these changes seem unrelated - if they are all related in some way, then it would be good to write what that overarching theme is, in the description of the PR.

If the different changes are unrelated, then it can be easier to understand (both now when reviewing, and later when someone is looking back at history to understand the changes) if you split these changes into multiple pull requests, each with their own description of what the PR is intended to achieve.

PR #3214, opened 2024-03-08
SimpleLauncher should help the user understand why they are not getting multiple nodes #3161

Ytemiloluwa:
# Description

Test cases for: 

- test that the launcher works with one node per block specified
- test that the launcher raises an exception with some other number of nodes per block, for example 64 nodes per block
- test that advanced users can disable the exception when using 64 nodes per block but setting permit_multiple_nodes=True

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup


benclifford:
I merged #3177 to do this, so I will close this issue now. Good work on putting tests in your PR!

PR #3216, opened 2024-03-08
Reduce flake8 max-line-length

nawarajshah:
# Description

Reduce flake8 max-line-length
Issue https://github.com/Parsl/parsl/issues/3105

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup


benclifford:
It looks like this PR actually deletes all of parsl - is that what you intended?

benclifford (changes requested):
probably this is not what you intended

nawarajshah:
> probably this is not what you intended

For get the put the issue

benclifford:
The Outreachy application period is over, so I'm closing your PR - thanks for your participation in Parsl+Outreachy so far!

PR #3229, opened 2024-03-11
Report usage using a richer configuration traversing protocol

benclifford:
This PR introduces code to traverse the configuration object (in a similar manner to the RepresentationMixin style of logging the supplied configuration object) with the intention of giving each object a chance to report its own usage information.

The protocol now reports configured objects either as a JSON string class name, or as a JSON object containing the class name and any additional information that class wishes to report for usage (via the `UsageInformation` abstract class)

This PR modifies the HighThroughputExecutor to use this API to report richer usage information: a specific usage tracking query is to ask about use of the enable_mpi_mode parameter, and so the HighThroughputExecutor will now report the boolean value of that parameter.

Beware that this reports on configuration, not use, of components: for example, configurations by default will include three staging providers, even though I believe it is extremely rare that either the FTP or HTTP staging providers are actually used to stage data. The `UsageInformation` API is intended to support reporting whether these staging providers actually stage anything, but this PR does not implement that in those staging provider components.

To support UsageInformation instances which report on usage during a run, the component tree is traversed both for the start message and the end message, and may result in different information in each message - for example, the DFK report occurs only end in the end message.

An example start message looks like this: (pretty-formatted)
```
{'correlator': 'f7595d08-7b94-49bc-b3d7-1ea7532b2f51',
 'parsl_v': '1.3.0-dev',
 'python_v': '3.12.2',
 'platform.system': 'Linux',
 'start': 1710150467,
 'components': ['parsl.config.Config',
                {'c': 'parsl.executors.high_throughput.executor.HighThroughputExecutor', 'mpi': False},
                'parsl.providers.local.local.LocalProvider',
                'parsl.channels.local.local.LocalChannel',
                'parsl.launchers.launchers.SingleNodeLauncher',
                'parsl.data_provider.ftp.FTPInTaskStaging',
                'parsl.data_provider.http.HTTPInTaskStaging',
                'parsl.data_provider.file_noop.NoOpFileStaging',
                'parsl.monitoring.monitoring.MonitoringHub']}
```

# Changed Behaviour

Different information will be reported via usage tracking - anyone processing that usage data will need to adapt their code.

## Type of change

- New feature


parsl/usage_tracking/api.py
```diff
@@ -0,0 +1,66 @@
+from parsl.utils import RepresentationMixin
+
+from abc import abstractmethod
+from functools import singledispatch
+from typing import Any, List, Sequence
+
+
+# Traverse the configuration heirarchy, returning a JSON component
+# for each one. Configuration components which implement
```
khk-globus, line 9: `hierarchy` (sp)

parsl/usage_tracking/api.py
```diff
@@ -0,0 +1,66 @@
...
+        # report the class name of this component
+        me = [qualified_name]
+
+    # unwrap typeguard-style unwrapping
+    init: Any = type(obj).__init__
+    if hasattr(init, '__wrapped__'):
+        init = init.__wrapped__
+
+    import inspect
+
```
khk-globus, line 43: Odd place for a lazy import?  Not wrong just ... not expected.

parsl/usage_tracking/api.py
```diff
@@ -0,0 +1,66 @@
...
+    argspec = inspect.getfullargspec(init)
+
+    for arg in argspec.args[1:]:  # skip first arg, self
+        arg_value = getattr(obj, arg)
+        d = get_parsl_usage(arg_value)
+        me += d
```
khk-globus, line 49: Stylistic alternative (but by no means a blocker):
```python
    me.extend(get_parsl_usage(getattr(obj, arg)) for arg in argspec.args[1:])
```

parsl/usage_tracking/api.py
```diff
@@ -0,0 +1,66 @@
...
+    return me
+
+
+@get_parsl_usage.register(list)
+@get_parsl_usage.register(tuple)
+def get_parsl_usage_sequence(obj: Sequence):
+    result = []
+    for v in obj:
+        result += get_parsl_usage(v)
+    return result
```
khk-globus, line 60: I'm not clear on the expected call frequency, but list comprehension is historically slightly (but noticeably) more performant in some situations:
```python
    return [get_parsl_usage(v) for v in obj]
```
For this reviewer's eye, that's easier to read/grok as well.

parsl/usage_tracking/api.py
```diff
@@ -0,0 +1,66 @@
...
+        arg_value = getattr(obj, arg)
+        d = get_parsl_usage(arg_value)
+        me += d
+
+    return me
+
+
+@get_parsl_usage.register(list)
+@get_parsl_usage.register(tuple)
+def get_parsl_usage_sequence(obj: Sequence):
```
khk-globus, line 56: Thoughts on return type information?

parsl/usage_tracking/usage.py
```diff
@@ -130,43 +133,50 @@ def check_tracking_enabled(self):
...
+        message = {'correlator': self.correlator_uuid,
+                   'end': int(time.time()),
+                   'components': [dfk_component] + get_parsl_usage(self.dfk._config)}
+        logger.debug(f"Usage tracking end message (unencoded): {message}")
+
+        return self.encode_message(message)
 
-        return json.dumps(message)
+    def encode_message(self, obj):
+        return PROTOCOL_VERSION + json.dumps(obj).encode('utf-8')
```
khk-globus, line 178: FWIW, `.encode()` [defaults to 'utf-8' in Python](https://docs.python.org/3/library/stdtypes.html#str.encode).  Not wrong, but redundant here.

khk-globus (approved):
Looks good; a couple of minor comments/suggestions but no blockers.

parsl/usage_tracking/api.py
benclifford: should be once at start once at end of each workflow, so efficiency isn't really a big deal here.

benclifford:
I converted this to draft status, because others are working on usage tracking now (which @kylechard and @yadudoc are driving I think?) - so this can be merged if they want it, otherwise we can close it.

parsl/usage_tracking/api.py
benclifford: i moved it - it's an import style i tend to use when i'm hacking stuff into things that I didn't get round to tidying

parsl/usage_tracking/api.py
benclifford: they're all List[Any] -- more tightly List[X] where X is anything that matches:

> # Traverse the configuration hierarchy, returning a JSON component
> # for each one.

rendered into code as anything which can go into `json.dump`

So, there's possibly a more nuanced type annotation to be made to express that `X` but it's not immediately in my head.

PR #3230, opened 2024-03-11
Change API usage to explict start/end style

benclifford:
Previously the message format was driven by a small "invoke me repeatedly, I'll change my behaviour based on how many times you've invoked me" state machine.

This PR removes that state machine and relies on the DFK knowing whether it is starting up or shutting down - that information is available statically inside the DFK code.

# Changed Behaviour

This should not change any behaviour

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/usage_tracking/usage.py
+++ b/parsl/usage_tracking/usage.py
@@ -109,7 +109,6 @@ class UsageTracker:
                                                 sys.version_info.micro)
         self.tracking_enabled = self.check_tracking_enabled()
         logger.debug("Tracking status: {}".format(self.tracking_enabled))
-        self.initialized = False  # Once first message is sent this will be True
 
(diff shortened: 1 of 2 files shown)


parsl/usage_tracking/usage.py
```diff
@@ -109,7 +109,6 @@ def __init__(self, dfk, port=50077,
                                                 sys.version_info.micro)
         self.tracking_enabled = self.check_tracking_enabled()
         logger.debug("Tracking status: {}".format(self.tracking_enabled))
-        self.initialized = False  # Once first message is sent this will be True
```
khk-globus, line 112: Always thought this was ugly.  Heh.

khk-globus (approved):

PR #3232, opened 2024-03-11
Scale in unregistered htex blocks

### LINKED ISSUE #2627: htex termination (and other scale-in-by-quantity) does not scale in blocks that haven't registered yet
**Describe the bug**
When scaling in by quantity of blocks (at parsl exit, or in the htex scaling strategy), the htex scale in code chooses which blocks to scale in, up to the specified quantity.

It does not choose blocks which have not got as far as a manager in that block registering - on my laptop, because execution is happening so fast, but I think in the HPC case, this will apply to blocks waiting in the queue.

These blocks are then abandoned by parsl - left running in my laptop, and in the HPC case, I expect occupying space in the queue, to eventually run.

The code in question is here:
 https://github.com/Parsl/parsl/blob/33398b15cbb20ddc5c273d22015d6c18351337c4/parsl/executors/high_throughput/executor.py#L666

Crossref related #2120 which talks about picking available blocks in an unusual order.

**Expected behavior**
Blocks in queues should be scaled in when scaling by quantity.

Especially, they might be considered ahead of blocks which are already running (although that order is probably a policy decision, not a technical one, taking into account things like cost of occupying idle nodes, immediacy of worker availability on workload increase, "fresh" walltime vs partially used walltime...)

**Environment**
my laptop, parsl master 33398b15cbb20ddc5c273d22015d6c18351337c4

benclifford:
This PR will make the htex scale-in code consider blocks which have been launched through the provider mechanism but have not registered to the interchange. This is especially relevant at shutdown time, when previous blocks still waiting in the queue (and so, not registered) would be abandoned by parsl shutdown.

[first commit on this PR is a test that should break; when that happens, I'll push the rest; this is to get a demonstrated failure of test in CI]

# Changed Behaviour

More blocks scaling in. Tidier shutdown.

# Fixes

Fixes #2627 

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -17,7 +17,7 @@ import parsl.launchers
 from parsl.serialize import pack_res_spec_apply_message, deserialize
 from parsl.serialize.errors import SerializationError, DeserializationError
 from parsl.app.errors import RemoteExceptionWrapper
-from parsl.jobs.states import JobStatus, JobState
+from parsl.jobs.states import JobStatus, JobState, TERMINAL_STATES
 from parsl.executors.high_throughput import zmq_pipes
 from parsl.executors.high_throughput import interchange
 from parsl.executors.errors import (
@@ -730,6 +730,26 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
 
         managers = self.connected_managers()
         block_info: Dict[str, BlockInfo] = {}
+
+        # First populate the list of blocks with blocks known to the provider
+        # mechanism. Then update that list with information from the
+        # interchange which will include idle times and task counts. Blocks
+        # with no interchange registration (for example because they are still
+        # in a batch queue) will be regarded as idle for a long time (so will
+        # be preferred when killing most idle blocks) in preference to any
+        # blocks which have already registered.
+
+        provider_blocks = self.status()
+        for (b_id, job_status) in provider_blocks.items():
+
+            # skip blocks that have already finished...
+            if job_status.state in TERMINAL_STATES:
+                continue
+
+            if b_id not in block_info:
+                block_info[b_id] = BlockInfo(tasks=0, idle=float('inf'))
+
+        managers = self.connected_managers()
         for manager in managers:
             if not manager['active']:
                 continue
--- a/parsl/tests/test_scaling/test_scale_down_htex_unregistered.py
+++ b/parsl/tests/test_scaling/test_scale_down_htex_unregistered.py
@@ -1,0 +1,70 @@
+import logging
+import time
+
+import pytest
+
+import parsl
+
+from parsl import File, python_app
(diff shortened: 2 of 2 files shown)


benclifford:
Since I wrote this code a few weeks ago, there's been a change to the high throughput executor status method to introduce a "MISSING" state, and I've also gone over the general status code with @yadudoc a bit.

Most importantly, I'd like to check that this code isn't actually calling the provider status calls (eg. invoking `sbatch`) unnecessarily: the current code is quite unclear as to when asking for status returns some in-memory cache of blocks (and block status) vs calling out to the LRM.

There might also be opportunities for aligning this code better with the code that handles `MISSING` status.

yadudoc:
This draft PR fixes the issue described here: #3235.

benclifford:
@yadudoc - I don't think it does? this PR doesn't fix any cache-based status update delays? I thought we tested a different (no-PR-number) patch?

parsl/executors/high_throughput/executor.py
```diff
@@ -730,6 +730,26 @@ class BlockInfo:
...
+        for (b_id, job_status) in provider_blocks.items():
+
+            # skip blocks that have already finished...
+            if job_status.state in TERMINAL_STATES:
+                continue
+
+            if b_id not in block_info:
+                block_info[b_id] = BlockInfo(tasks=0, idle=float('inf'))
+
+        managers = self.connected_managers()
```
khk-globus, line 752: Makes the earlier assignment redundant, yeah?

parsl/executors/high_throughput/executor.py
```diff
@@ -730,6 +730,26 @@ class BlockInfo:
...
+
+        provider_blocks = self.status()
+        for (b_id, job_status) in provider_blocks.items():
+
+            # skip blocks that have already finished...
+            if job_status.state in TERMINAL_STATES:
+                continue
+
+            if b_id not in block_info:
+                block_info[b_id] = BlockInfo(tasks=0, idle=float('inf'))
```
khk-globus, line 750: `provider_blocks` is a dictionary, yeah?  If so, then `b_id` is guaranteed to be unique through the loop.  Coupled with the fact that `block_info` is empty when the loop starts, the if-conditional is a waste.  Could merge it all into the instantiation:
```python
block_info = {
  block_id: BlockInfo(tasks=0, idle=float('inf')
  for block_id, job_status in provider_blocks.items()
  if job_status.state not in TERMINAL_STATES
}
```
Later on, `block_info` keys are set to zero on first use as well, so could make that structure a `defaultdict`:
```python
block_info: Dict[str, BlockInfo = defaultdict(lambda: BlockInfo(0, float('inf')))
for b_id, job_status in provider_blocks.items():
    # skip blocks that have already finished...
    if job_status.state in TERMINAL_STATES:
        continue

    # the defaultdict logic is now the authority on "empty BlockInfo" objects
    block_info[b_id]
```
And could remove the if-branch in the following for-loop as well.

PR #3245, opened 2024-03-12
Make JobStatusPoller parameters mandatory and remove defaults

benclifford:
Making the parameters mandatory and without defaults helps in debuggability and in code understanding.

Defaults come from closer to the user, and so the defaults removed here are never used.

To support Parsl users using JobStatusPoller as part of a DFK-free environment (for example, Globus Compute), this PR leaves the dfk parameter to have a meaningful default.

# Changed Behaviour

Runtime behaviour should not be changed, as this only removes unused defaults and requires parameters that were already supplied.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/job_status_poller.py
+++ b/parsl/jobs/job_status_poller.py
@@ -106,7 +106,7 @@ class PollItem:
 
 
 class JobStatusPoller(Timer):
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Good call; thanks for this.

Issue #3235, opened 2024-03-11
HTEX is slow to report broken config


yadudoc:
**Describe the bug**

When I run a basic config on my laptop that uses the `SlurmProvider` incorrectly, Parsl detects the failure to find `sbatch` but tries about ~10 times and fails, before finally the periodic pollers kick in and detect that the executor has reached a bad state.

Here's my test script:
```
from parsl.providers import SlurmProvider
from parsl.executors import HighThroughputExecutor
from parsl import Config
from parsl import python_app
import parsl


config = Config(
    executors=[
        HighThroughputExecutor(
            label="Fake_broken_provider",
            provider=SlurmProvider(
                "debug",
                init_blocks=0,
            ),
        )
    ],
)
parsl.load(config)

@python_app
def double(x):
    return x * 2

future = double(5)
future.result()
```

Here's the output:
```
Traceback (most recent call last):
  File "/private/tmp/test.py", line 28, in <module>
    future.result()
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/site-packages/parsl/dataflow/dflow.py", line 302, in handle_exec_update
    res = self._unwrap_remote_exception_wrapper(future)
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/site-packages/parsl/dataflow/dflow.py", line 570, in _unwrap_remote_exception_wrapper
    result = future.result()
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/Users/yadu/opt/anaconda3/envs/parsl_py3.9/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
parsl.executors.errors.BadStateException: Executor Fake_broken_provider failed due to: Error 1:
	Failed to start block 0: Cannot launch job parsl.Fake_broken_provider.block-0.1710183543.7008908: Could not read job ID from submit command standard output; recode=127, stdout=, stderr=/bin/sh: sbatch: command not found

Error 2:
[...]

PR #3247, opened 2024-03-13
Move provider status caching closer to providers

benclifford:
parsl.provider.* status() calls can be expensive, in the sense that they do things like execute batch system commands which system administrators dislike being run automatically and periodically, and even in some situations ban that behavior.

So, Parsl tries to not do those calls too often - providers can define a status_polling_interval (which returns an interval in seconds) and the strategy code will call (through a chain of calls) the provider status method at most every status_polling_interval.

Prior to this PR, this was implemented in the job status poller code, which makes the following chain of calls at most every status polling interval

poller/strategy iterator =>
  poll_item.poll (runs "often" - eg every 5 seconds driven by job status poller?)
    => determine if the rest of this stack should run, or otherwise use data cached as poll item._status attribute
      => executor.status()
        => provider.status()
          => (in the case of cluster provider, provider._status and LRM command line callout)

There are a couple of issues related to this call stack:

* #3235 - slow update of jobs that failed to submit - when a job fails to submit, at present it is recorded in a separate table, the _simulated_status dictionary, in BlockProviderExecutor. Entries added to that are not exposed to the scaling code until executor.status() is called, which may be many strategy iterations later. In that time, the scaling strategy make make the wrong decisions, based on incomplete information. That is the core of issue #3235

* #2627 - htex scale in needs to be aware of job status. Scale in code is not always driven in by the poller/strategy code; for example it executes at shutdown driven by the DataFlowKernel. That scale in code then is not in a position to make use of the cached data inside poll_item and it is more natural to call executor.status() to determine status - see PR #3232. That results in calls to provider.status() without any status_polling_interval rate limiting/caching.

This current PR moves that rate limit/cache deeper into the stack to the BlockProviderExecutor.

The job status poller code now polls the executor on every iteration, relying on the executor to perform rate limiting and caching.

That has the effect of exposing simulated statuses to the scaling strategy code on each iteration, so that the delays in issue #3235 will not be experienced.

This change also means that PR #3232 can safely call executor.status() without exceeding the rate limit.

This PR changes when provider status gets called, because it is now a cache that refreshes whenever status is called past the expiry time, rather than more explicitly driven in a loop. That might change scaling behaviour, but I think not in any significant way.

This PR changes the decision of when an executor is to be polled for status: previously non-positive values of status_polling_interval were used as magic values to inhibit provider status polling. Now, any executor with a provider will be polled.

This PR makes the call stack above provider.status a little more data driven and less effectful (the effect being load on the LRM), which might make refactoring this stack a bit easier. However, it will also make that a bit more computation heavy - I think not much because usually there are not many blocks in existence.

This PR adds a test that repeatedly calling BlockProviderExecutor.status does not result in an excessive number of calls to the provider status method but still updates eventually.

# Changed Behaviour

LRM interactions might be substantially different, and scaling behaviour may be different while still fitting within the broad descriptions of what scaling is meant to do.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 import logging
+import math
 import threading
+import time
 from itertools import compress
 from abc import abstractmethod, abstractproperty
 from concurrent.futures import Future
@@ -65,6 +67,8 @@ class BlockProviderExecutor(ParslExecutor):
         self._executor_bad_state = threading.Event()
         self._executor_exception: Optional[Exception] = None
 
+        self._provider_status_cache: Tuple[float, List[JobStatus]] = (-math.inf, [])
+
         self._block_id_counter = AtomicIDCounter()
 
         self._tasks = {}  # type: Dict[object, Future]
@@ -87,21 +91,6 @@ class BlockProviderExecutor(ParslExecutor):
 
         return d
 
-    @property
-    def status_polling_interval(self):
-        """Returns the interval, in seconds, at which the status method should be called. The
-        assumption here is that, once initialized, an executor's polling interval is fixed.
-        In practice, at least given the current situation, the executor uses a single task provider
-        and this method is a delegate to the corresponding method in the provider.
-
-        :return: the number of seconds to wait between calls to status() or zero if no polling
-                 should be done
-        """
-        if self._provider is None:
-            return 0
-        else:
-            return self._provider.status_polling_interval
-
     def _fail_job_async(self, block_id: Any, message: str):
         """Marks a job that has failed to start but would not otherwise be included in status()
         as failed and report it in status()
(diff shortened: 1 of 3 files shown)


benclifford:
saw this test failure in Python 3.8 github actions:

```
    
        def check_one_block():
            return len(dfk.executors['htex_local'].connected_managers()) == 1
    
>       try_assert(
            check_one_block,
            fail_msg="Expected 1 managers during a single long task",
        )
E       AssertionError: Expected 1 managers during a single long task
E         (Still failing after timeout [5000ms], with attempts every 20ms)
parsl/tests/test_scaling/test_scale_down_htex_auto_scale.py:102: AssertionError
```

I'm unclear if it's related to this PR - definitely this PR is peturbing behaviour vaguely in this area.

benclifford:
it's recreatable on my laptop, definitely related to this PR... ^

benclifford:
Deeper in, it's this exception happening as data structures get out of sync:

```
  File "/home/benc/parsl/src/parsl/parsl/executors/status_handling.py", line 124, in status
    status = self._make_status_dict(block_ids, self._provider_status_cache[1])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benc/parsl/src/parsl/parsl/executors/status_handling.py", line 87, in _make_status_dict
    raise IndexError("block id list and status string list differ in size")
IndexError: block id list and status string list differ in size
```

but most of Parsl is capable of recovering from that error (eventually it converges to a non-exception case, when the cache is refreshed)

benclifford:
Based on further testing, I will withdraw this PR - the various scaling block structures are tangled a bit more than I hoped, and this PR breaks things that isn't well tested (although the test suite did find the breakage). This rearrangement needs to happen in some more subtle way, probably as part of a bigger structure tidyup around the various data structures which have partial status info

PR #3121, opened 2024-03-04
Test that monitoring is resistant to helper process termination

benclifford:
Monitoring shutdown has previously been quite hangy, and in those situations those tests are likely to hang rather than raise some assertion error.

## Type of change

- Code maintenance/cleanup

PR #3253, opened 2024-03-17
Return None rather than raise an exception for Gantt chart of 0 tasks

benclifford:
Prior to this PR, visualizing a workflow with no tasks in it would cause an HTTP 500 server error, caused by an exception raised in the gantt generation code.

However, a workflow with no tasks is a legitimate workflow.

This PR makes task_gantt_plot return None, rather than attempting a plot, which then places no Gantt chart in workflow visualization, rather than breaking.

This bug was exposed through PR #3121 which sometimes (as a legitimate part of that test) results in a monitoring DB with a workflow with no tasks (but not always - that test is testing shutdown behaviour, not database contents).

## Type of change
- Bug fix


khk-globus (approved):
:facepalm:

PR #3254, opened 2024-03-17
Use parsl.utils log helpers for monitoring, rather than own variant

benclifford:
This PR removes start_file_logger which was similar, but not the same as parsl.utils.set_file_logger. It makes set_file_logger return the chosen root logger, which was the main difference between start_file_logger and set_file_logger.

# Changed Behaviour

After this PR, monitoring logs will now use the verbose log format used by parsl.log, instead of their own more terse format.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Ah, nice-found DRY-ification.

PR #3255, opened 2024-03-17
Tidyup interchange text on monitoring port descriptions

benclifford:
This PR is clearer about connecting specifically to the MonitoringHub, and removes a sentence that describes how parameters are passed around which is not monitoring specific.

## Type of change


khk-globus (approved):
Might change "at which" to "to which", but small potatoes.

PR #3260, opened 2024-03-18
Support context manager protocol in DFK

### LINKED ISSUE #3258: `with` style context manager for Parsl DFK

Users should be able to use parsl DFKs using the `with` statement, to cause the DFK to be cleaned up when it is finished with.

Previously, if a user did not explicitly clean up a DFK, an `atexit` handler would do this when the process exits. See issue #3104 for why that doesn't work in recent Python, PR #3165 for the PR that removes that handling and issue #3251 for an example of a user encountering this as a practical problem.

I would like some syntax like this:

```
with parsl.load(config):
    f = my_app()
    f.result()
# implicit contextmanager cleanup of the loaded DFK happens here by calling dfk.cleanup()
```

In addition to introducing this context manager, documentation should be changed to encourage users to use this syntax instead of explicitly calling cleanup() in most scripted cases (although not always, for example, I think syntactically this doesn't work with jupyter notebooks as I think a `with` block cannot span multiple notebook cells)

It might be possible to rework exit handling in an atexit handler to still be able to do shutdown, but I don't feel like that will be easy - because a lot of stuff can happen at DFK shutdown.

**Additional context**
This is coming from changed (more strict) behaviour around threads and processes in Python 3.12.

Harichandra-Prasath:
# Description

Recently, PR #3165 removed at-exit cleanup of DFKs, so after that PR, a user has to explicitly invoke the DFK cleanup method. More justification for this is in issue #3104 

This PR adds context manager support to the DFK, so that it can be used in a `with` block around an entire workflow.

Earlier description:
Adding the support of "with" style context manager for implicit cleaning of DFK when it exits the context manager. The cleanup will be called from `__exit__` method of the context manager.

The need for this feature is previously, atexit has a registered method that will call the cleanup and it was working fine, but for  python 1.12 and  beyond, Python has documented that it is unsafe to do thread and fork operations in atexit handlers, so atexit  cleanup behaved differently than what it is expected. After the PR #3165, atexit registered method dont call dfk cleanup rather log a warning that dfk is not cleaned up. It makes the user has to explicitly call the dfk.cleanup() to method to release the resources. 

It pushed into the situation to not rely on atexit handlers for cleaning up the dfk rather need a new way to implicitly cleanup the dfk. So loading the dfk under a context manager allows to call the cleanup when it exits the context manager.

# Changed Behaviour

From this PR, user can load the DFK within context manager like "with". A simple example can be 

with parsl.load(config):
        # user workflow

User need not to call the cleanup explicitly after this PR. This allows to cleanup the dfk relatively easier.

# Fixes

Fixes #3258 

## Type of change

- New feature


benclifford:
You should write at least one test for this new feature: a test that shows an example of how a user would run a Parsl app when using this new structure.

You should also work on the documentation - for example, look in the user guide for places where it would be right to use this new style, and make changes there. If you aren't comfortable with the documentation, maybe ask in the Outreachy slack channel to see if anyone there can help: there are lots of people interested in documentation.

Harichandra-Prasath:
fine sir

Pazil24:
@Harichandra-Prasath You might want to do a git rebase to keep the branch up to date.
Also your branch should be in line with the changes you're making or made.

docs/userguide/configuring.rst
```diff
@@ -75,9 +77,11 @@ Consider a configuration function if the application will shut down and re-launc
     def make_config() -> Config:
         return Config(...)
 
-    parsl.load(make_config())
+    with parsl.load(make_config()):
+        pass
```
benclifford, line 81: it would be better to write something like ' # Your workflow here' instead of 'pass'

parsl/tests/test_python_apps/test_basic_within_context_manager.py
```diff
@@ -0,0 +1,39 @@
...
+from parsl.errors import NoDataFlowKernelError
+
+
+@parsl.python_app
+def square(x):
+    return x * x
+
+
+@parsl.bash_app
+def foo(x, y, z=30, stdout='foo.stdout', label=None):
```
benclifford, line 13: this test function doesn't need to be so complex: make a simple function so that people reading this code do not try to understand all of the parameters. for example use a function foo(x) return "echo {x+1}"

benclifford (approved):

PR #3265, opened 2024-03-19
Use module-scope logger in MonitoringHub

benclifford:
It already does! except indirected via self.logger. This PR removes that indirection.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Oof.

PR #3266, opened 2024-03-19
Rename monitoring port variables to reflect use

benclifford:
The naming of the interchange port suggests that is for the interchange to send monitoring messages, but it is a general ZMQ listening port used by at least the DFK too.

There is another general listening port, listening on UDP, used by the UDPRadio on remote workers.

This PR renames most variables referencing the above connections to have zmq_ and udp_ prefixes, instead of a mix of hub_ and ic_ prefixes, to clarify which variables and parameters apply to which of those two ports.

This clarifies in most of the codebase, for example, that the former hub_port and hub_port_range variables do not control the same thing in different ways, but that one is for ZMQ and ones is for UDP.

This PR avoids changing user-facing parameters, so some ambiguosly named parameters still exist: those should be dealt with by a deprecation process over a longer time period in a different PR.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -108,12 +108,12 @@ class DataFlowKernel:
 
         # hub address and port for interchange to connect
         self.hub_address = None  # type: Optional[str]
-        self.hub_interchange_port = None  # type: Optional[int]
+        self.hub_zmq_port = None  # type: Optional[int]
         if self.monitoring:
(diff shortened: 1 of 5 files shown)


khk-globus (approved):
One of many, many, many low-hanging fruit items that have been on my mind.  Thanks for getting to such mundane details as this that Actually Do Matter.  :pray:

PR #3268, opened 2024-03-20
Fix with context manager

error-4u:
# Description

The goal is to address this issue by enabling the use of DFKs within context managers, which would allow for easier cleanup of resources associated with the DFK when exiting the context manager. By implicitly handling the cleanup 

# Changed Behaviour

Users can seamlessly utilize the DFK within context managers such as with. There's no need for users to manually trigger the cleanup process.

# Fixes

Fixes # (https://github.com/Parsl/parsl/issues/3258)

## Type of change

- New Feature

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -206,6 +206,14 @@ class DataFlowKernel:
 
         atexit.register(self.atexit_cleanup)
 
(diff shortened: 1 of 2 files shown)


benclifford:
This PR was implemented in PR #3260 so I am closing this PR.

PR #3269, opened 2024-03-20
WQExecutor + TVExecutor: Automatic cleanup by adding their own atexit handlers.

tphung3:
# Description

Previously these executors don't exit on their own but wait for the DFK to call their shutdown methods from the DFK's atexit handler. Now the call is removed so these executors add their own atexit handlers to clean themselves up, otherwise a Parsl program using these executors will hang at the end of the program.

These handlers never spawn new threads or processes so they should be safe according to Python3.12's atexit standard.

# Changed Behaviour

TaskVineExecutor and WorkQueueExecutor.

## Type of change
- Bug fix
- New feature


tphung3:
It should be good for review now @benclifford

benclifford (approved):

PR #3271, opened 2024-03-21
Clearing the active DFK on cleaning

### LINKED ISSUE #3263: Clearing loaded DFK on cleaning the DFK in dfk.cleanup()
Currently after a DFK is cleaned up by dfk.cleanup() ,  it cleans the DFK but not clearing the reference of the DFK. So, we have to call parsl.clear() to remove the loaded dfk to load another new config.  And if we try to load a new without parsl.clear(), it raises a error indicating there is already a loaded config. 

Try 

parsl.load()
parsl.dfk().cleanup()
parsl.load()

it would be better if dfk.cleanup() can clear the reference as well,  as after dfk.cleanup() we cant load new config as well as pushing new tasks.

Harichandra-Prasath:
# Description

Currently now, on cleaning the Active DFK ( dfk.cleanup ), It releases all the resources and cleans the DFK, but it doesnt remove the reference of the cleaned dfk from the KernelLoader, It makes the user to run .clear() one more time to remove the reference of the DFK, and there are no things can be done with cleaned up dfk , even cant load a new config, So its better to clear the reference as well at cleanup, so that user can directly load another config to work on

# Changed Behaviour
After this pr, at cleanup of dfk , reference of the cleaned dfk will be removed

# Fixes

Fixes #3263 

## Type of change

- New feature

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -1254,6 +1254,8 @@ class DataFlowKernel:
             logger.info("Terminated monitoring")
 
         logger.info("DFK cleanup complete")
+        logger.info("Removing the reference of cleaned DFK")
+        DataFlowKernelLoader.clear()
 
     def checkpoint(self, tasks: Optional[Sequence[TaskRecord]] = None) -> str:
         """Checkpoint the dfk incrementally to a checkpoint file.


benclifford:
I ended up doing this slightly differently, in PR #3444, and in that PR I also cleaned up a lot of tests: even though they were not failing in this PR, there would now be a lot of unnecessary/redundant clear calls happening.

PR #3274, opened 2024-03-21
Annotate scale_in and scale_out with types in JobStatusPoller

benclifford:
These types align with the scale_in and scale_out types defined on BlockProviderExecutor.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Easy peazy.

PR #3275, opened 2024-03-21
Remove unused code path, driven by new type annotation

benclifford:
Block IDs are strings, and _fail_job_async is never used without an allocated block ID.

This was revealed by mypy after tightening the type annotation in this PR.

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
This looks correct.

PR #3281, opened 2024-03-22
WQ + TV Executors: Making their shutdown methods idempotent

tphung3:
# Description

Previously, WQ and TV Executors can be shut down twice, one by the registration of `shutdown` to atexit, one by the user explicitly calling `parsl.dfk().cleanup()`. This PR makes the effect of these executors shutdown happen only once. Note that previously the shutdown methods don't create bugs as they only call process joins and a python process can be joined many times, but future bugs might come.

# Changed Behaviour

WQ and TV Executors

## Type of change

- Code maintenance/cleanup


tphung3:
I'm not sure why python 3.8 takes a long time here

tphung3:
@benclifford This PR is RTM

PR #3282, opened 2024-03-22
Test that init_blocks blocks are recorded in the monitoring database

benclifford:
This is a test for an uncoming rearrangement of init_blocks handling.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py
+++ b/parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py
@@ -1,0 +1,86 @@
+import logging
+import os
+import parsl
+import pytest
+import time
+
+from parsl.providers import LocalProvider
+from parsl.channels import LocalChannel
+from parsl.launchers import SimpleLauncher
+
+from parsl.config import Config
+from parsl.executors import HighThroughputExecutor
+from parsl.monitoring import MonitoringHub
+
+
+def fresh_config(run_dir, strategy):
+    return Config(
+        run_dir=os.fspath(run_dir),
+        executors=[
+            HighThroughputExecutor(
+                label="htex_local",
+                cores_per_worker=1,
+                encrypted=True,
+                provider=LocalProvider(
+                    channel=LocalChannel(),
+                    init_blocks=1,
+                    # min and max are set to 0 to ensure that we don't get
+                    # a block from ongoing strategy scaling, only from
+                    # init_blocks
+                    min_blocks=0,
+                    max_blocks=0,
+                    launcher=SimpleLauncher(),
+                ),
+            )
+        ],
+        strategy=strategy,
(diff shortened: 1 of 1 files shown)


parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # at this point, we should find one row in the monitoring database.
+
+    engine = sqlalchemy.create_engine("sqlite:///runinfo/monitoring.db")
+    with engine.begin() as connection:
+
+        # we should see a single block:
+        result = connection.execute(text("SELECT COUNT(DISTINCT block_id) FROM block"))
+        (c, ) = result.first()
+        assert c == 1
```
khk-globus, line 76: This is a slightly subtle "what's going on here?" assertion.  For these, I often write something to communicate the assertion's purpose.  In this case, this feels to me like a "test setup verification" in that it's verifying that something larger isn't in play (e.g., the sqlite DB being shared between tests and not otherwise cleaned up).  Consider:
```python
    assert c == 1, "Verify test setup"
```

parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # there should be a pending status
+        result = connection.execute(text("SELECT COUNT(*) FROM block WHERE block_id = 0 AND status = 'PENDING'"))
+        (c, ) = result.first()
+        assert c == 1
+
+        # there should be a cancelled status
+        result = connection.execute(text("SELECT COUNT(*) FROM block WHERE block_id = 0 AND status = 'CANCELLED'"))
+        (c, ) = result.first()
+        assert c == 1
```
khk-globus, line 86: And then here, the only substantive difference is in the SQL statements ... but those won't come through PyTest's output in the failing case.  So I might suggest something like:
```python
assert c == 1, "Unexpected number of PENDING statuses for test block_id."
...
assert c == 1, "Unexpected number of CANCELLED statuses for test block_id."
```
Perhaps verbose, but the key in my mind is to reduce the "question marks" in the next dev's brain when looking at these in a failed PyTest run.

(Even better if you've can make the text more qualitative to clue the next dev in than my rudimentary suggestions.)

khk-globus (approved):

khk-globus (approved):

parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py
benclifford: I already had inline comments with this sort of text, so I moved them from being `# comments` into the corresponding assert statements.

Issue #3278, opened 2024-03-22
Centralising block and job control and status


benclifford:
Currently there are several separate code paths that deal with launching jobs and storing job/block status.

In the "init blocks" path, which is not handled by scaling strategy: blocks are launched and status is recording in monitoring db by a call from the DB into the executor, without any interaction with the parsl.jobs.* status poller / strategy code.

In the JobStatusPoller/scaling strategy path:

The JobStatusPoller periodically refreshes from the executor layer, but also between those refreshes adds its own mutations for block scale in and out (which i think do not take into account any init_blocks blocks?)

Inside the high throughput executor, two hooks take data from the provider layer (which is sourced by asking the actual local resource manager like slurm) and mutate it in two different ways to add information on failed jobs.

Some providers also keep track of jobs that they have launched in further data structures.

This very distributed way of managing blocks is making it quite hard to reason about, debug and enhance the scaling code.

I'd like to push on getting a more centralised place for block and job scale in, out and status, rather than having so many different routes and structures.

For example:

init_blocks should be handled as part of the strategy code: the history here I think is that you could use `none` strategy (in the codebase sense) in which case your scaling strategy (in a what happens sense) is: launch init_blocks and do nothing else. This is implemented by the `none` strategy doing nothing, and some other part of Parsl handles the implementation of this "what happens" strategy without expressing it as a "codebase" strategy. 

The nuances of sourcing job and block status from places other than an LRM call have historically been treated as "this is a kludge" rather than "genuine first order" information sources. This is hard to reason about, and the data structures at different layers could be simplified, documented, merged... taking into consideration that these other information sources are *not* hacks but genuine information that deserves proper treatment.

PR #3283, opened 2024-03-22
Handle init_blocks in scaling strategy, rather than special-casing it

benclifford:
This is part of issue #3278 tidying up job and block management.

# Changed Behaviour

Now init_blocks scale out happens on the first strategy poll, not at executor start - that will often delay init_blocks scaling by one strategy poll period compared to before this PR.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/strategy.py
+++ b/parsl/jobs/strategy.py
@@ -129,8 +129,8 @@ class Strategy:
         self.executors = {}
         self.max_idletime = max_idletime
 
-        self.strategies = {None: self._strategy_noop,
-                           'none': self._strategy_noop,
+        self.strategies = {None: self._strategy_init_only,
+                           'none': self._strategy_init_only,
                            'simple': self._strategy_simple,
                            'htex_auto_scale': self._strategy_htex_auto_scale}
 
@@ -146,10 +146,17 @@ class Strategy:
         for executor in executors:
             self.executors[executor.label] = {'idle_since': None}
 
-    def _strategy_noop(self, status: List[jsp.PollItem]) -> None:
-        """Do nothing.
+    def _strategy_init_only(self, status_list: List[jsp.PollItem]) -> None:
+        """Scale up to init_blocks at the start, then nothing more.
         """
-        logger.debug("strategy_noop: doing nothing")
+        for exec_status in status_list:
+            if exec_status.first:
+                executor = exec_status.executor
+                logger.debug(f"strategy_init_only: scaling out {executor.provider.init_blocks} initial blocks for {executor.label}")
+                exec_status.scale_out(executor.provider.init_blocks)
+                exec_status.first = False
+            else:
+                logger.debug("strategy_init_only: doing nothing")
 
     def _strategy_simple(self, status_list: List[jsp.PollItem]) -> None:
         self._general_strategy(status_list, strategy_type='simple')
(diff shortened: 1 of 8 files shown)


benclifford:
failing a precondition in test_drain for exactly the "Changed Behaviour" reason - now that test will be reliant on strategy poll period so it can probably be made to use a smaller strategy period

parsl/jobs/strategy.py
```diff
@@ -146,10 +146,17 @@ def add_executors(self, executors: Sequence[ParslExecutor]) -> None:
...
         """
-        logger.debug("strategy_noop: doing nothing")
+        for exec_status in status_list:
+            if exec_status.first:
+                executor = exec_status.executor
+                logger.debug(f"strategy_init_only: scaling out {executor.provider.init_blocks} initial blocks for {executor.label}")
+                exec_status.scale_out(executor.provider.init_blocks)
+                exec_status.first = False
+            else:
+                logger.debug("strategy_init_only: doing nothing")
```
khk-globus, line 159: Is this prefix necessary?  I thought we included the method name in the log line already, per the log formatter.

parsl/jobs/strategy.py
```diff
@@ -146,10 +146,17 @@ def add_executors(self, executors: Sequence[ParslExecutor]) -> None:
...
-        """Do nothing.
+    def _strategy_init_only(self, status_list: List[jsp.PollItem]) -> None:
+        """Scale up to init_blocks at the start, then nothing more.
         """
-        logger.debug("strategy_noop: doing nothing")
+        for exec_status in status_list:
+            if exec_status.first:
+                executor = exec_status.executor
+                logger.debug(f"strategy_init_only: scaling out {executor.provider.init_blocks} initial blocks for {executor.label}")
+                exec_status.scale_out(executor.provider.init_blocks)
```
khk-globus, line 156: For new SLOCs, might consider passing the args to `.debug()` rather that doing the formatting up front.  No sense in doing the work if debug-level logs aren't emitted:
```python
logger.debug("scaling out %s initial blocks for %s", executor.provider.init_blocks, executor.label)
```

khk-globus (approved):
Looks good, with a couple of minor comments.  A good addition to this would be a test verifying use of the `.first` branch.

parsl/jobs/strategy.py
benclifford: It's a vestige from the line it used to be, which included that (renamed) prefix.

The method name is indeed in the log formatter

parsl/jobs/strategy.py
benclifford: this isn't a new line - it's a move and a keyword rename!

benclifford:
> A good addition to this would be a test verifying use of the .first branch

From an end functionality perspective, that's checking that init_blocks actually gets used to start blocks, rather than a later part of the strategy code: with strategy `none` that's pretty straightforward (and it's what happens by chance in parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py already).

With the other two strategies that's less observable from the outside, I think: `init_blocks` vs `min_blocks` doesn't necessarily even make sense because of that difficulty in observing the difference...

parsl/jobs/strategy.py
khk-globus: :boom:

PR #3289, opened 2024-03-24
Add support for dicts in future-resolution-plugin (#3111)

Andrew-S-Rosen:
# Description

Adds support for shallow resolution of Futures in `dict` types for #3111.

Thoughts, @benclifford? This takes care of key futures. It also takes care of value futures, but in this spirit of "shallow" future resolution, it only does so if it `isinstance(value, Future)` and won't (yet) handle more complex cases like a dictionary where the value is a `list` containing a future.

### DIFF
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -1,6 +1,6 @@
+from collections.abc import Mapping
 from concurrent.futures import Future
 from functools import singledispatch
-from collections.abc import Mapping
 
 
 @singledispatch
--- a/parsl/dataflow/future_resolution.py
+++ b/parsl/dataflow/future_resolution.py
@@ -55,3 +55,28 @@ def _(iterable):
 
     type_ = type(iterable)
     return type_(map(unwrap, iterable))
+
+
+@traverse_to_gather.register(dict)
+def _(dictionary):
+    futures = []
+    for key, value in dictionary.items():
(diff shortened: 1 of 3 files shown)


parsl/dataflow/future_resolution.py
```diff
@@ -55,3 +56,28 @@ def unwrap(v):
 
     type_ = type(iterable)
     return type_(map(unwrap, iterable))
+
+
+@traverse_to_gather.register(Mapping)
+def _(dictionary):
+    futures = []
+    for key, value in dictionary.items():
+        if isinstance(key, Future):
```
benclifford, line 65: you should be able to call traverse_to_gather here without an if statement and it will recursively find the futures for you. at least thats the intention.

parsl/dataflow/future_resolution.py
```diff
@@ -55,3 +56,28 @@ def unwrap(v):
 
     type_ = type(iterable)
     return type_(map(unwrap, iterable))
+
+
+@traverse_to_gather.register(Mapping)
```
benclifford, line 61: not a fan of the conflation of mapping vs dictionary here... possibly my own personal taste, but this sort of stuff should be as type-specific as possible, mostly because traverse_to_unwrap is making a dictionary, not whatever type T it happened to be called with, and that's some surprise to the user if they're expecting to see a T in their args (as that's what they're passing in...)

PR #3299, opened 2024-03-27
Use .pytest rundir hook for disconnection test

benclifford:
By leaving the test config's rundir as default, parsl/tests/conftest.py will reassign that attribute to a dynamically generated directory under .pytest/parsltest-current

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Ugh; so many of these.  And then I get lost on entire framework fixups.  :facepalm:  Thanks for this one.

PR #3300, opened 2024-03-27
Use a defaultdict for BlockInfo in htex scale-in

benclifford:
This is to accomodate upcoming expansion of the block_info dictionary to include data from an additional source (the list of blocks known to the scaling code)

This change arises from a review of PR #3232, an earlier prototype of that expansion work.

# Changed Behaviour

Behaviourally, this shouldn't change anything: it changes invalid block accesses from key errors into blocks that are infinitely idle and completely unloaded. However there are no indexes into block_info except in code that is adding block information, so this situation should not arise.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Yep.  Nice.

PR #3302, opened 2024-03-27
Scale in blocks at shutdown using Job Status Poller

benclifford:
This is part of a series of PRs to move all scaling behaviour to the job status poller (see PR #3283 for a corresponding init_blocks move)

This will now scale in blocks using the job status poller scale in code, which means the (moved) shutdown scale in loop does not need to send its own BLOCK_INFO messages.

The scale_in call is changed from the executor version to the status poller version. That latter version is mostly a passthrough to the executor version, with the above-mentioned monitoring code change. So the behaviour of the scale in call + monitoring should be the same.

The parameter to the scale_in call is a count of blocks. This is intended to mean "all of them" - by counting however many blocks are known and using that count. This code will work for any sufficiently large number: it needs to be at least as large as the number of blocks that should be scaled in.

Prior to this PR, the number of known blocks was the number of jobs in the provider's `resources` dictionary - a structure that should otherwise be private to the particular provider.

Post this PR, the number of known blocks is the number of blocks known to the JobStatusPoller module: this is a possibly different number, although it should be at least the number of blocks which have been submitted (either by getting a new status in PollItem.scale_out or by being present in the most recent provider cache refresh in PollItem.poll).

Because it should be at least the number of submitted blocks, this number should be large enough to represent "all of them" (with the above meaning).

# Changed Behaviour

none - see above reasoning

## Type of change

- Code maintenance/cleanup


benclifford:
See PR #3304 for a test for this

khk-globus (approved):
Heh; couldn't keep the test and changed code together in the same PR?  ;-)

benclifford:
> Heh; couldn't keep the test and changed code together in the same PR? ;-)

It was originally that way on my laptop - but then I decided I also wanted to have the test in history *before* this PR #3302 to be clear that the test should pass both before and after the change introduced here.

khk-globus:
Yeesh!  Bring a well-reasoned thought process to the discussion why don't you!!!

PR #3311, opened 2024-03-29
Move scale in at exit code into close method of job status poller

benclifford:
# Changed Behaviour

Changed behaviour: this will now happen slightly before it used to, but the differences are only changed place in the call stack, and log messages will appear in a different order now - so the actual scaling-in behaviour should be unchanged.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Ugh; that `bad_state_is_set` leave a sour taste in my mouth, currently.  Ick.

PR #3312, opened 2024-03-29
Force kwargs for monitor wrapper for reduced maintenance fragility

benclifford:
This function has a lot of parameters, and long sequences of positional parameters are hard to align. See #2973 for similar work elsewhere.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Heh; works for me.

PR #3147, opened 2024-03-05
add diaspora radio and is able to send resource info

ClaudiaCumberbatch:
# Description

Add a diaspora radio to send resource information. Also add a test_radio directory to test this functionality.

# Changed Behaviour

The resource information will not be inserted into database, instead, it will be sent to diaspora. If the db insertion is still required, just change the `radio_mode` in `executors/high_throughput/executors.py` .

## Type of change

- New feature

PR #3315, opened 2024-03-31
[not for merge] start a monitoring radio plugin API

benclifford:
# Description

This is the beginning of a monitoring radio plugin API, driven by some work with desc (where the choice of monitoring radio has been something to pay attention to performance-wise), by @ClaudiaCumberbatch who wants to plug in another radio, and the chronolog project where a hack sesssion resulted in sending some monitoring messages into chronolog. There are also possible uses in Globus Compute when wanting to observe task behaviour in future without having any supporting Parsl-style monitoring infrastructure.

# Changed Behaviour

This API is a partial prototype and might not work quite right, but there is enough there to get feedback from potential users.

Prior to this PR monitoring configuration is sent as pair: a radio name (hard-coded in executor definitions since PR #2274, and not easily user selectable) along with information that can be used by some radios (for example, the monitoring router UDP listener endpoint).

After this PR, those fields are all replaced by a pickled configuration object, some instance of `RadioConfig`, which will be sent to the worker and used to create an appropriate radio. A user can now specify a RadioConfig instance (most likely a subclass of RadioConfig such as UDPRadioConfig) which contains whichever configuration values are relevant for that particular radio mode - for example, HTEXRadioConfig needs no further configuration, but UDPRadioConfig contains an IP address and UDP port.

This allows plugin-style out-of-codebase radios to be implemented without modifying the parsl source tree - that's a feature Parsl has been driving towards recently in many places to accomodate differing levels of code quality and support ability: most recently @ClaudiaCumberbatch's diaspora PR #3147, which this PR should accomodate without any diaspora-specific additions to the main Parsl codebase.

## Type of change

- New feature

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -13,7 +13,6 @@ Core
     parsl.app.app.join_app
     parsl.dataflow.futures.AppFuture
     parsl.dataflow.dflow.DataFlowKernelLoader
-    parsl.monitoring.MonitoringHub
     parsl.dataflow.dependency_resolvers.DependencyResolver
     parsl.dataflow.dependency_resolvers.DEEP_DEPENDENCY_RESOLVER
     parsl.dataflow.dependency_resolvers.SHALLOW_DEPENDENCY_RESOLVER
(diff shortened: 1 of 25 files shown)


benclifford:
all work for this patch stack has been submitted as other PRs. closing.

PR #3317, opened 2024-04-08
Move 'first' field into strategy state

benclifford:
This is part of work to move JobStatusPoller facade state into other classes, as part of job handling rearrangements in PR #3293

This should not change behaviour: each executor has a single PolledExecutorFacade and a single strategy.ExecutorState, and this PR moves the 'first' field from one to the other.

parsl/tests/test_monitoring/test_htex_init_blocks_vs_monitoring.py tests that init_blocks handling still fires properly - that's what is switched by this 'first' field.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/strategy.py
+++ b/parsl/jobs/strategy.py
@@ -26,6 +26,10 @@ class ExecutorState(TypedDict):
     If the executor is not idle, then None.
     """
 
+    first: bool
+    """Is this the first poll for this executor?
+    """
+
 
 class Strategy:
     """Scaling strategy.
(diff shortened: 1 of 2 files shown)


parsl/jobs/strategy.py
```diff
@@ -26,6 +26,10 @@ class ExecutorState(TypedDict):
     If the executor is not idle, then None.
     """
 
+    first: bool
+    """Is this the first poll for this executor?
+    """
+
```
khk-globus, line 32: Stating what the boolean does without being explicit _can_ (though obviously not always) lead to ambiguous understanding.  Suggest being explicit, like:
```python
    first: bool  # True if executor has not yet been polled
```

khk-globus (approved):

PR #3319, opened 2024-04-08
Add ZipFileStaging to archive outputs into a zip file

benclifford:
# Description

This output file staging mechanism provides a zip: URL scheme, naming
a file inside a zip file; output files from apps can be staged out into
this location. Input staging is not supported.
    
The use-case for this is as part of a sequence of PRs separating
PR #3306, leading to stdout staging into zipfiles.

# Changed Behaviour

This is new functionality only activated by use of the `zip:` URL scheme, which previously would have raised an error.

## Type of change

- New feature

### DIFF
--- a/parsl/data_provider/zip.py
+++ b/parsl/data_provider/zip.py
@@ -1,0 +1,84 @@
+import filelock
+import logging
+import os
+import parsl
+import zipfile
+
+from typing import Tuple
+
+from parsl.data_provider.staging import Staging
+from parsl.data_provider.files import File
+
+
+logger = logging.getLogger(__name__)
+
+
+class ZipFileStaging(Staging):
+    """A stage-out provider for zip files.
+
+    This provider will stage out files by writing them into the specified zip
+    file.
+
+    The filename of both the zip file and the file contained in that zip are
+    specified using a zip: URL, like this:
+
+    zip:/tmp/foo/this.zip/inside/here.txt
+
+    This URL names a zip file ``/tmp/foo/this.zip`` containing a file
+    ``inside/here.txt``.
+
+    The provider will use the Python filelock package to lock the zip file so
+    that it does not conflict with other instances of itself. This lock will
+    not protect against other modifications to the zip file.
+    """
+
+    def can_stage_out(self, file: File) -> bool:
+        logger.debug("archive provider checking File {}".format(repr(file)))
+        return file.scheme == 'zip'
+
+    def stage_out(self, dm, executor, file, parent_fut):
+        assert file.scheme == 'zip'
+
+        zip_path, inside_path = zip_path_split(file.path)
+
+        working_dir = dm.dfk.executors[executor].working_dir
+
+        if working_dir:
+            file.local_path = os.path.join(working_dir, inside_path)
+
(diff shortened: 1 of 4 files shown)


parsl/tests/test_staging/test_zip_out.py
```diff
@@ -0,0 +1,99 @@
...
+
+    return "; ".join(cmds)
+
+
+@pytest.mark.local
+def test_zip_out(tmpd_cwd):
+    # basic test of zip file stage-out
+    zip_path = str(tmpd_cwd) + "container.zip"
+    file_base = "data.txt"
+    of = File("zip:" + zip_path + "/" + file_base)
```
khk-globus, line 43: From the nitpicky dept, `tmpd_cwd` does the `/` for you.  Consider:
```python
zip_path = tmpd_cwd / "container.zip"
file_base = "data.txt"
of = File(f"zip:{zip_path / file_base}")
```

Note, though, that there is a slight difference between these two:
```python
str(tmpd_cwd) + "container.zip"  # --> /some/tmp/dircontainer.zip
str(tmpd_cwd / "container.zip")  # --> /some/tmp/dir/container.zip
```

So, mostly a bike-shed, and some other instances of same in this file, but a suggestion.

parsl/data_provider/zip.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def _zip_stage_out(zip_file, inside_path, working_dir, parent_fut=None, inputs=[], _parsl_staging_inhibit=True):
+    file = inputs[0]
+
+    os.makedirs(os.path.dirname(zip_file), exist_ok=True)
+
+    with filelock.FileLock(zip_file + ".lock"):
+        with zipfile.ZipFile(zip_file, mode='a', compression=zipfile.ZIP_DEFLATED) as z:
+            z.write(file, arcname=inside_path)
```
khk-globus, line 67: Is there any value in attempting to test the lock?  It's a new requirement in `requirements.txt` and as discussed out-of-band, a real consideration, so perhaps worth an AT of some kind?

khk-globus, line 67: This is appending; should there be any provision for _updating_ an entry with a newer variant?  (Open question, not sure if that would be wise, but asking.)

khk-globus (approved):

parsl/data_provider/zip.py
benclifford: What you end up with this way is a zip file with multiple files with the same name, which the format can accomodate, at least. Removing/replacing an older entry is, I think, awkward with the Python zipfile library. In the basic Parsl DAG workflow model, it would be unusual to write the same file multiple times, so this doesn't bother me too much - but in a more developed version of this code, replacing would probably be the right thing to do.

khk-globus: Ohh, that's right.  Thanks for that reminder.  Of course the format handles it.  And with that reminder, I'm now thinking that this:

> but in a more developed version of this code, replacing would probably be the right thing to do.

is only relevant if there would be serious performance or efficiency gains.  Just not worth rewriting the entire file without some significant motivator.

PR #3325, opened 2024-04-09
Unshare "now" timestamp across executor polls

benclifford:
This PR removes the notion of a shared current time among all polls.

Instead each facade poll looks up the current time individually.

Although executor polls are all launched from the same loop, there's no need for them to act as if they are being polled at the same moment in time, which in the case of multiple executors, will be some time in the past.

This means there is one less variable shared between polls of different executors, which is part of work pushing executor polling deeper into each executor away from the per-DFK JobStatusPoller.

# Changed Behaviour

This will change the cadence of polling: `now` may be measured a bit later, especially in the presence of multiple executors, causing meaning a poll that might have previously skipped to now run, and as a knock on, all subsequent skips/updates will be jiggled around based on that change. This should not have any serious effect on observable behaviour.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/job_status_poller.py
+++ b/parsl/jobs/job_status_poller.py
@@ -37,7 +37,8 @@ class PolledExecutorFacade:
     def _should_poll(self, now: float) -> bool:
         return now >= self._last_poll_time + self._executor.status_polling_interval
 
-    def poll(self, now: float) -> None:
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Indeed; an obvious cleanup.  Thanks.

PR #3326, opened 2024-04-09
Remove unused 'noci' pytest marker.

benclifford:
Not-running-tests-in-CI is achieved by more subtle means, such as topic and issue specific test markers.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
:boom:

PR #3338, opened 2024-04-10
Replace DFK and JobStatusPoller monitoring zmq channels with Queue radio

benclifford:
Before this PR, the DataFlowKernel and each JobStatusPoller PolledExecutorFacade each opened a ZMQ connection to the monitoring router. These connections are not threadsafe, but (especially in the DFK case) no reasoning or checking stops the DFK's connection being used in multiple threads.

Before this PR, the MonitoringHub presented a 'send' method and stored the DFK's ZMQ connection in self._dfk_channel, and each PolledExecutorFacade contained a copy of the ZMQ channel code to open its own channel, configured using parameters from the DFK passed in at construction.

This PR:

* moves the above uses to use the MonitoringRadio interface (which was originally designed for remote workers to send monitoring information, but seems ok here too)

* has MonitoringHub construct an appropriate MonitoringRadio instance for use on the submit side, exposed as self.radio;

* replaces the implementation of send with a new MultiprocessingQueueRadio which is thread safe but only works in the same multiprocessing environment as the launched monitoring database manager process (which is true on the submit side, but for example means this radio cannot be used on most remote workers)

This work aligns with the prototype in #3315 (which pushes on monitoring radio configuration for remote workers) and pushes in the direction (without getting there) of allowing other submit-side hooks.

This work removes some monitoring specific code from the JobStatusPoller, replacing it with a dependency injection style.
This is part of work to move JobStatusPoller facade state into other classes, as part of job handling rearrangements in PR #3293

This PR will change how monitoring messages are delivered from the submitting process, and the most obvious thing I can think of that will change is how this will behave under load: heavily loaded messaging causing full buffers and other heavy-load symptoms will now behave as multiprocessing Queues do, rather than as ZMQ connections do. I have not attempted to characterise either of those modes.

# Description

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/radios.py
+++ b/parsl/monitoring/radios.py
@@ -6,6 +6,7 @@ import logging
 
 from abc import ABCMeta, abstractmethod
 
+from multiprocessing.queues import Queue
 from typing import Optional
 
 from parsl.serialize import serialize
@@ -173,3 +174,17 @@ class UDPRadio(MonitoringRadio):
             logging.error("Could not send message within timeout limit")
             return
         return
+
+
+class MultiprocessingQueueRadio(MonitoringRadio):
+    """A monitoring radio intended which connects over a multiprocessing Queue.
+    This radio is intended to be used on the submit side, where components
+    in the submit process, or processes launched by multiprocessing, will have
+    access to a Queue shared with the monitoring database code (bypassing the
+    monitoring router).
+    """
+    def __init__(self, queue: Queue) -> None:
+        self.queue = queue
+
+    def send(self, message: object) -> None:
+        self.queue.put((message, 0))
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -177,10 +177,11 @@ class DataFlowKernel:
 
         # this must be set before executors are added since add_executors calls
         # job_status_poller.add_executors.
+        radio = self.monitoring.radio if self.monitoring else None
         self.job_status_poller = JobStatusPoller(strategy=self.config.strategy,
                                                  strategy_period=self.config.strategy_period,
                                                  max_idletime=self.config.max_idletime,
-                                                 dfk=self)
+                                                 monitoring=radio)
 
         self.executors: Dict[str, ParslExecutor] = {}
 
--- a/parsl/jobs/job_status_poller.py
+++ b/parsl/jobs/job_status_poller.py
@@ -1,7 +1,6 @@
 import logging
 import parsl
 import time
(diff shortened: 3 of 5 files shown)


parsl/monitoring/radios.py
```diff
@@ -173,3 +174,17 @@ def send(self, message: object) -> None:
...
+    This radio is intended to be used on the submit side, where components
+    in the submit process, or processes launched by multiprocessing, will have
+    access to a Queue shared with the monitoring database code (bypassing the
+    monitoring router).
+    """
+    def __init__(self, queue: Queue) -> None:
+        self.queue = queue
+
+    def send(self, message: object) -> None:
+        self.queue.put((message, 0))
```
khk-globus, line 190: `message` is a generic, nominally pickleable object.  Okay ... and we're wrapping into a tuple for the other end.  What is the magic number `0`?

Also, where is the end of this queue?  As in, where is it drained?

benclifford: It's the IP address this UDP packet was received from! (in the original UDP implementation of monitoring radios)

Really that should get removed from the MonitoringRadio API but not in this PR...

The other end of the queue is in the bowels of the monitoring database manager... it's super confusing to follow that code because there are many different queues all draining into each other...  it's this block_msgs queue in MonitoringHub:

``` self.radio = MultiprocessingQueueRadio(self.block_msgs)```

which is consumed in the dbm_proc process started inside the MonitoringHub, which is processed in this thread in `parsl/monitoring/db_manager.py`

```
        self._block_queue_pull_thread = threading.Thread(target=self._migrate_logs_to_internal,
                                                         args=(
                                                             block_queue, 'block', self._kill_event,),
                                                         name="Monitoring-migrate-block",
                                                         daemon=True,
                                                         )
```

The queue already exists and is passed to the router code to be populated from a ZMQ receiver and a UDP receiver too - so this radio is providing another way to get messages into that queue that doesn't use ZMQ or UDP.

In general that code has parallel queues with different names (resources, blocks, priority) because they all used to have different processing behaviour in various places. Over time, I've worked towards making them all behave the same, with the intention of having a single set of queues, but that work is incomplete. So the message needs to go into one of these queues and block_msgs is a fairly arbitrary choice.

khk-globus: Heh.  That's ... both :+1: and :-1:.  Thanks.

khk-globus (approved):

PR #3339, opened 2024-04-10
Unabstract executor.create_monitoring_info

benclifford:
This method makes sense for `BlockProviderExecutors`, where it is used by the `JobStatusPoller`. It does not make sense for Parsl Executors in general and is not otherwise used. So remove from the `ParslExecutor` base class, which removes the dependency of `parsl.executors.base` on `parsl.jobs.*`, a cleaner separation of concerns.

This method is implemented in `HighThroughputExecutor`, but the information generated only uses `BlockProviderExecutor` information and would make sense for other BlockProviderExecutors (Work Queue, Task Vine and the erstwhile IPP and Extreme Scale executors). Using this implementation has been prototyped in the desc branch of parsl since 2021. So move that implementation out of `HighThroughputExecutor` and into `BlockProviderExecutor`.

This should not change any behaviour when using `HighThroughputExecutor`. When using Work Queue or Task Vine, this change will bring block monitoring support to those executors. As noted above, this has been used successfully with Work Queue since 2021 in the desc branch. This PR modifies the basic monitoring test to test against those executors in addition to the `HighThroughputExecutor`.

This PR modifies the type annotations of `create_monitoring_info` to type check properly - the previous htex implementation was not type-checked. The return type is modified to a sequence, because `List` is invariant on its parameter type, which is incompatible with adding `Dict[str, Any]` to it.

# Changed Behaviour

see main description

## Type of change

- New feature

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
+import datetime
 import logging
 import threading
 from itertools import compress
 from abc import abstractmethod, abstractproperty
 from concurrent.futures import Future
-from typing import List, Any, Dict, Optional, Tuple, Union, Callable
+from typing import List, Any, Dict, Optional, Sequence, Tuple, Union, Callable
 
 from parsl.executors.base import ParslExecutor
 from parsl.executors.errors import BadStateException, ScalingFailed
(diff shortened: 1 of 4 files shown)


parsl/executors/status_handling.py
```diff
@@ -234,3 +235,18 @@ def _get_block_and_job_ids(self) -> Tuple[List[str], List[Any]]:
...
+        for bid, s in status.items():
+            d: Dict[str, Any] = {}
+            d['run_id'] = self.run_id
+            d['status'] = s.status_name
+            d['timestamp'] = datetime.datetime.now()
+            d['executor_label'] = self.label
+            d['job_id'] = self.blocks_to_job_id.get(bid, None)
+            d['block_id'] = bid
+            msg.append(d)
+        return msg
```
khk-globus, line 252: I recognize this is largely copying, but could this be reduced to a single list comprehension?  Consider:
```python
# are all items effectively collected at the same time?  Or is the loop
# iteration time pertinent, perhaps for concurrent data structure modification?
now = datetime.datetime.now()
return [
    {
        'run_id': self.run_id,
        ...
        'timestamp': now,
        ...
    }
    for bid, s in status.items()
]
```

khk-globus (commented):
Fundamental change looks good, with an inline suggestion.  I'll wait to approve until you've hammered out the testing issue.

benclifford:
tests are failing - see #3339 for environment fix

parsl/executors/status_handling.py
benclifford: sure. eventually I'd also like these monitoring messages to be typed more strongly: they have strongly defined formats that are not declared in the type system but sort of are implied by the SQL schema declared in SQLAlchemy deep in the monitoring DB manager...

khk-globus (approved):

PR #3340, opened 2024-04-10
Fix typing on general strategy

benclifford:
This PR adds a type annotation to the function signature.

That has the knock-on effect of enabling type checking within the method, revealing two mypy errors.

One is unreachable code: because the strategy will only be called with a BlockProviderExecutor executor, it does not need to check if it has been called with a BlockProviderExecutor.

The other type error comes from mypy not realising that a path is unreachable: the idle time of an executor may be None, but at the point of the type error, an `if` statement has ensured that it is never none. This PR adds an assert statement to assert that assertion to the type checker (and, incidentally, to the logs via wrap_with_logs).

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/strategy.py
+++ b/parsl/jobs/strategy.py
@@ -9,7 +9,6 @@ import parsl.jobs.job_status_poller as jsp
 
 from parsl.executors import HighThroughputExecutor
 from parsl.executors.base import ParslExecutor
-from parsl.executors.status_handling import BlockProviderExecutor
 from parsl.jobs.states import JobState
 from parsl.process_loggers import wrap_with_logs
 
@@ -183,15 +182,12 @@ class Strategy:
         self._general_strategy(executor_facades, strategy_type='htex')
 
     @wrap_with_logs
-    def _general_strategy(self, executor_facades, *, strategy_type):
+    def _general_strategy(self, executor_facades: List[jsp.PolledExecutorFacade], *, strategy_type: str) -> None:
         logger.debug(f"general strategy starting with strategy_type {strategy_type} for {len(executor_facades)} executors")
 
         for ef in executor_facades:
             executor = ef.executor
             label = executor.label
-            if not isinstance(executor, BlockProviderExecutor):
-                logger.debug(f"Not strategizing for executor {label} because scaling not enabled")
-                continue
             logger.debug(f"Strategizing for executor {label}")
 
             if self.executors[label]['first']:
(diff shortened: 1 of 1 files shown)


parsl/jobs/strategy.py
```diff
@@ -253,8 +249,9 @@ def _general_strategy(self, executor_facades, *, strategy_type):
                     if not self.executors[executor.label]['idle_since']:
                         logger.debug(f"Starting idle timer for executor. If idle time exceeds {self.max_idletime}s, blocks will be scaled in")
                         self.executors[executor.label]['idle_since'] = time.time()
-
                     idle_since = self.executors[executor.label]['idle_since']
+                    assert idle_since is not None, "The `if` statement above this assert should have forced idle time to be not-None"
+
```
khk-globus, line 254: Would it work to force it with `: float = `?

benclifford: it's extracted from `self.executors[executor.label]['idle_since']` which is optional so you need to tell the type checker to disbelieve the type annotation: annotating idle_since as `float` would move the type error up to line 252, like this:

```
parsl/jobs/strategy.py:252: error: Incompatible types in assignment (expression has type "float | None", variable has type "float")  [assignment]
```

benclifford: you need to have some assert in here about "believe me... the types say you can have got an idle_since from this shared data structure that everyone can be poking at but I am promising you, it will be not None, even though I cannot prove some other thread did not modify the shared structure"

benclifford: (cannot prove *using the language of Python types...*)

khk-globus: Yeah; can't keep the mypy idiosyncrasies clear sometimes.  In my mind, that's what `: float =` is doing.  "No, really, this is the case."  Okay.  It is what it is.

khk-globus (approved):

parsl/jobs/strategy.py
benclifford: :float is saying this is a place for a float so in two directions:

i) you can rely on the value to be a float (assuming you believe the annotation...)
ii) you should be sure you believe that what is put in that place is a float

The second of those is failing here: mypy has evidence that we can be assigning a None here, not a float (from the type signature of self.executors[executor.label]['idle_since'].

But because idle_since is a local variable (which self.executors[executor.label]['idle_since'] is not), we can believe that some other force isn't narrowing it down, and eliminate the None case.

If the `if` statement above had been working with locals rather than globals, I think it's possible mypy would have inferred that case-narrowing itself: there is more reasoning available about locals because they're not so prone to being poked at by code outside the current block.

PR #3344, opened 2024-04-11
Rework monitoring recording of invalid stdout/err streams

benclifford:
Before this PR, an invalid stdout/err stream was logged at WARNING level and the filename as recorded in the monitoring database was set to the exception message.

This PR elevates that log to an ERROR, because this is an error, including the exception message to give a debugging hint.

In this error case, a blank stdout or stderr will be recorded in the monitoring database. Generally the task will fail later due to other parts of the code being unable to interpret the specification.

# Changed Behaviour

less confusing stuff in the monitoring database. more noise in logs.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_bash_apps/test_stdout.py
+++ b/parsl/tests/test_bash_apps/test_stdout.py
@@ -1,3 +1,4 @@
+import logging
 import os
 
 import pytest
@@ -72,7 +73,7 @@ def test_bad_stderr_file():
 
 
 @pytest.mark.executor_supports_std_stream_tuples
-def test_stdout_truncate(tmpd_cwd):
+def test_stdout_truncate(tmpd_cwd, caplog):
     """Testing truncation of prior content of stdout"""
 
     out = (str(tmpd_cwd / 't1.out'), 'w')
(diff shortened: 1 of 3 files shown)


parsl/tests/test_bash_apps/test_stdout.py
```diff
@@ -87,8 +88,11 @@ def test_stdout_truncate(tmpd_cwd):
     assert len1 == 1
     assert len1 == len2
 
+    for record in caplog.records:
+        assert record.levelno < logging.ERROR
```
khk-globus, line 92: These hunks are a good addition to these ATs; what about testing the exception paths as well?  Possibly a new test each for stderr and stdout.  Without thinking harder about it than ~30s, the first approach that comes to mind is a kernel like:
```python
assert "Could not parse stdout" in ...
# and
assert "Could not parse stderr" in ...
```
Or `mock.patch` the `get_std_fname_mode` and ensure that the mock was called.  Something like:
```python
exc_text = randomstring()  # See note below
mock_get_mode = mock.patch('...', side_effect=MemoryError(exc_text)  # an arbitrary exception
...
assert mock_get_mode.called
assert exc_text in ...
```

I don't think Parsl has implemented a `randomstring()` fixture, but [Compute has one we could borrow](https://github.com/globus/globus-compute/blob/2.18.0/compute_sdk/tests/unit/conftest.py#L11-L16).

Not mandatory, obviously, but sharing this here as a suggestion after talking out-of-band.

benclifford:
This failed a run of test_monitoring/test_basic.py, with a missing resource message saying that the task started executing on an htex worker. That's in monitoring but shouldn't be directly affected by this change. The resource monitoring messages come from the worker, via htex results channel and then ZMQ into the monitoring router and onwards via multiprocessing.Queue into the db_manager. But there isn't enough logging in the recorded artefacts to trace what happened to that message in the abovementioned pipeline and I think it's likely that it's one of the many mysterious behaviours of monitoring.

PR #3347, opened 2024-04-11
Add more cases for std stream logging

benclifford:
Previously, stdout/err logs no output if the streams were not set (i.e. defaulting to None) and logged with the entire stream target if set.

This PR adds more cases here with case specific logging: if stdout/err is not redirected, that is now logged. If tuple mode is used, the tuple is unpacked and logged as separate fields. If an unknown format is specified an error is logged (but no exception is raised - it is not the job of this code to validate the stdout/err specifications)

This PR is in preparation for a new case to be added in an upcoming PR which requires more serious formatting that was not accomodated by the previous implementation.

# Changed Behaviour

log messages - including a new error log

## Type of change

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -1391,10 +1391,20 @@ class DataFlowKernel:
 
     @staticmethod
     def _log_std_streams(task_record: TaskRecord) -> None:
-        if task_record['app_fu'].stdout is not None:
-            logger.info("Standard output for task {} available at {}".format(task_record['id'], task_record['app_fu'].stdout))
-        if task_record['app_fu'].stderr is not None:
-            logger.info("Standard error for task {} available at {}".format(task_record['id'], task_record['app_fu'].stderr))
+        tid = task_record['id']
+
+        def log_std_stream(name: str, target) -> None:
+            if target is None:
+                logger.info(f"{name} for task {tid} will not be redirected.")
+            elif isinstance(target, str):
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -1391,10 +1391,20 @@ def load_checkpoints(self, checkpointDirs: Optional[Sequence[str]]) -> Dict[str,
...
+
+        def log_std_stream(name: str, target) -> None:
+            if target is None:
+                logger.info(f"{name} for task {tid} will not be redirected.")
+            elif isinstance(target, str):
+                logger.info(f"{name} for task {tid} will be redirected to {target}")
+            elif isinstance(target, tuple) and len(target) == 2:
+                logger.info(f"{name} for task {tid} will be redirected to {target[0]} with mode {target[1]}")
+            else:
+                logger.error(f"{name} for task {tid} has unknown specification: {target!r}")
```
khk-globus, line 1404: If this is an error, should it also raise an exception?  Otherwise, it feels more like a warning, no?

benclifford: It's an error. It should be validated elsewhere and this code path should be unreachable (and so do something more like raise an assertion that this path is "impossible".

"should" in the sense of thats how things could be but aren't. If you get this message, there's a problem and your workflow won't run as you expect. But this log statement is not the place to be validating that user input: prior to this, this code would log at info level whatever garbage you had fed in, and it will continue to  log that garbage here in this case.

khk-globus (approved):

PR #3348, opened 2024-04-11
Treat stdout and stderr as stageable Files

benclifford:
Prior to this PR, stdout and stderr were treated as string pathnames interpreted in an executor specific context: for example, htex treats them as paths on the worker filesystem; Task Vine treats these paths as names of files that should be staged to the submitting system.

This is/was different to the treatment of files placed in the outputs kwarg of an app using File objects. Those file objects can be staged out arbitrarily using Parsl file staging.

This PR allows File objects to be specified for bash_app stdout and stderr kwargs, and stages those files out in the same way as outputs-kwarg files are staged.

This PR preserves the previous str-based behaviour without staging. This is inconsistent with the data staging mechanism, which only accepts Files, but preserves backwards compatibility, which is important here because stdout and stderr kwargs are used by users a lot.

When this new staging mechanism is used, DataFutures for stdout/err Files are made available to the user via app_future.stdout and app_future.stderr. This properties existed previously, and when using strings for stdout/err, retain their previous behaviour of returning those strings.

This mode is incompatible with tuple-based stdout/err specifications, which allow opening stdout/err with specific file modes, and rely on the user to manage those files outwith Parsl: that concept is not compatible with the file staging model of "write a file once, don't ever modify it".

Fixes issue #363

# Changed Behaviour

Things using the existing stdout/err behaviour should behave the same - this PR should turn some previously invalid stdout/err specifications into valid ones.

## Type of change

- New feature

### DIFF
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -5,6 +5,7 @@ import logging
 
 from parsl.app.errors import wrap_error
 from parsl.app.app import AppBase
+from parsl.data_provider.files import File
 from parsl.dataflow.dflow import DataFlowKernelLoader
 
 logger = logging.getLogger(__name__)
@@ -54,7 +55,12 @@ def remote_side_bash_executor(func, *args, **kwargs):
         if stdfspec is None:
             return None
 
-        fname, mode = get_std_fname_mode(fdname, stdfspec)
+        if isinstance(stdfspec, File):
+            fname = str(stdfspec)
+            mode = "w"
+        else:
+            fname, mode = get_std_fname_mode(fdname, stdfspec)
+
         try:
             if os.path.dirname(fname):
                 os.makedirs(os.path.dirname(fname), exist_ok=True)
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -56,7 +56,9 @@ def remote_side_bash_executor(func, *args, **kwargs):
             return None
 
         if isinstance(stdfspec, File):
-            fname = str(stdfspec)
+            # a File is an os.PathLike and so we can use it directly for
+            # the subsequent file operations
+            fname = stdfspec
             mode = "w"
         else:
             fname, mode = get_std_fname_mode(fdname, stdfspec)
@@ -66,7 +68,7 @@ def remote_side_bash_executor(func, *args, **kwargs):
                 os.makedirs(os.path.dirname(fname), exist_ok=True)
             fd = open(fname, mode)
         except Exception as e:
-            raise pe.BadStdStreamFile(fname) from e
+            raise pe.BadStdStreamFile(str(fname)) from e
         return fd
 
     std_out = open_std_fd('stdout')
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -831,11 +831,11 @@ class DataFlowKernel:
             func, outputs[idx], o = stageout_one_file(file, func)
             app_fut._outputs.append(o)
 
-        file = kwargs.get('stdout', None)
+        file = kwargs.get('stdout')
         if isinstance(file, File):
             func, kwargs['stdout'], app_fut._stdout_future = stageout_one_file(file, func)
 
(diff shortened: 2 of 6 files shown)


parsl/dataflow/dflow.py
```diff
@@ -262,6 +262,12 @@ def _create_task_log_info(self, task_record):
                 stderr_name = ""
         else:
             stderr_name = ""
+            # TODO:
+            # When stdout_spec is a File, the above fails, but we can't then use str, because str for a File
+            # gives the localpath, which doesn't exist on the submit side for a non-file: File object.
+            # Reporting the stderr path here probably should be done differnetly for File objects, perhaps
+            # reporting the URL without using get_std_fname_mode which only makes sense in the case of
+            # shared filesystem stdout/stderrs?
```
khk-globus, line 270: Typically, when I see a TODO in a code base, it's a clue to me that we need a ticket.  That is, recognizing that, when I'm planning my next foray between tasks, I don't (usually) take my cues of "what to do next"  by searching for TODO comments, the best place for task visibility is a well-formed ticket.  Especially if planning with a team.

Parsl is not developed via a more formal sprintly arrangement, so this is perhaps "just fine," but I observe it nonetheless.

parsl/dataflow/dflow.py
```diff
@@ -786,33 +796,56 @@ def _add_output_deps(self, executor: str, args: Sequence[Any], kwargs: Dict[str,
...
+
+        for idx, file in enumerate(outputs):
+            func, outputs[idx], o = stageout_one_file(file, func)
+            app_fut._outputs.append(o)
+
+        file = kwargs.get('stdout', None)
+        if isinstance(file, File):
+            func, kwargs['stdout'], app_fut._stdout_future = stageout_one_file(file, func)
+
+        file = kwargs.get('stderr', None)
```
khk-globus, line 845: FWIW, `.get()`'s `default` argument, ahm, _defaults_ to `None`.  So could write these as simply:
```python
file = kwargs.get('stdout')
...
file = kwargs.get('stderr')
```
Same difference.

parsl/dataflow/futures.py
```diff
@@ -70,13 +70,32 @@ def __init__(self, task_record: TaskRecord) -> None:
...
+        self._stderr_future: Optional[DataFuture] = None
+
     @property
-    def stdout(self) -> Optional[str]:
-        return self.task_record['kwargs'].get('stdout')
+    def stdout(self) -> Union[None, str, DataFuture]:
+        """Return app stdout. If stdout was specified as a string, then this
+        property will return that string. If stdout was specified as a File,
+        then this property will return a DataFuture representing that file
+        stageout."""
```
khk-globus, line 81: Thank you for updating this.

khk-globus (approved):

parsl/dataflow/dflow.py
benclifford: oh i thought i had fixed this before I pushed it up -- that's unintentional...

khk-globus (changes requested):
Actually, I misunderstood the intention of that TODO.  Let's get that addressed before merge time!

parsl/dataflow/dflow.py
benclifford: should be addressed now, with testing building on the tests introduced in #3363

parsl/dataflow/dflow.py
benclifford: I have changed these to remove `None`

benclifford:
I think i've addressed all comments so submitting this again for review

parsl/app/bash.py
```diff
@@ -54,7 +55,12 @@ def open_std_fd(fdname):
         if stdfspec is None:
             return None
 
-        fname, mode = get_std_fname_mode(fdname, stdfspec)
+        if isinstance(stdfspec, File):
+            fname = str(stdfspec)
```
benclifford, line 59: wait, this should be fspath, not str

khk-globus (approved):

PR #3352, opened 2024-04-11
Move PolledExecutorFacade functionality into BlockProviderExecutor

benclifford:
This should be code movement only and should not change functionality.

This consolidates scale in/out and status functionality in BlockProviderExecutor, which hopefully makes future work rearranging the three sources of block status information easier.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Thanks to your precursor set of PRs, this is super, super clear.  Thanks!

PR #3353, opened 2024-04-11
Make HTEX scale-down be aware of unstarted blocks

benclifford:
# Description

Prior to this PR, the scale_in code for the `HighThroughputExecutor` will not scale in any block that has not had at least one manager register with the interchange, because it retrieves the list of blocks from the interchange. This is documented in issue #3232 

This PR makes the htex scale in code also pay attention to blocks in the status_facade list - which includes blocks that have been submitted, and blocks which have been reported by the provider mechanism.

# Changed Behaviour

Scaling in will now scale in more blocks.

# Fixes

Fixes #3232 

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -724,8 +724,6 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin):
 
         for block_id, job_status in self._status.items():
             if job_status.state not in TERMINAL_STATES:
-                # TODO: is there a nicer way to make block_info come into existence?
-                # can i write just the expression block_info[block_id] on its own?
                 block_info[block_id] = new_block_info()
 
         managers = self.connected_managers()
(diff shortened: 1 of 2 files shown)


benclifford:
there's one todo here aimed at @khk-globus - what's the nicest way to make a defaultdict entry come into existence without specifying any changes to the default?

parsl/executors/high_throughput/executor.py
```diff
@@ -713,8 +713,20 @@ class BlockInfo:
...
+        # which have connected to the interchange.
+
+        def new_block_info():
+            return BlockInfo(tasks=0, idle=float('inf'))
+
+        block_info: Dict[str, BlockInfo] = defaultdict(new_block_info)
+
+        for block_id, job_status in self._status.items():
+            if job_status.state not in TERMINAL_STATES:
+                block_info[block_id] = new_block_info()
```
khk-globus, line 727: Per the PR comment to me, I think this looks fine.  The only suggestion I might make is to take advantage of the `dataclass`'s default handling via:
```python
@dataclass
class BlockInfo:
    tasks: int = 0
    idle: float = float('inf')
```
And then there's no need for the `new_block_info` function.  Just:
```python
block_info = defaultdict(BlockInfo)
```
Beyond that, per our side-band conversation, I don't think now's the time to tackle the code smell you've discovered.  This whole method could be cleaner yes, but suggest "not now."

khk-globus (approved):
Ugh; I hate that `15000`ms (test `try_assert) is where we're at currently.  But it is.

PR #3355, opened 2024-04-14
Add taskvine temp/url handling to parsl/taskvine executor 

colinthomas-z80:
# Description

Using the stub staging provider in the taskvine executor will give control of staging https files to taskvine. Declaring a file with the protocol `taskvinetemp://` will also utilize the temporary/intermediate data features of taskvine.

## Type of change
- New feature


benclifford:
@colinthomas-z80 this is failing on one of the `File` tests

colinthomas-z80:
@benclifford This is working now on my end. Interested to hear any thoughts you have. This is only regarding temp files fyi, the task grouping we discussed will be another endeavor. 

It seems now that we must declare the stub staging provider we will have to explicitly state all other provider types. Maybe the stub staging provider should be hard-coded in the manager? I'm not sure if that would make a difference though

parsl/executors/taskvine/executor.py
```diff
@@ -363,6 +366,7 @@ def submit(self, func, resource_specification, *args, **kwargs):
 
         # Also consider any *arg that looks like a file as an input:
         input_files += [self._register_file(f) for f in args if isinstance(f, File)]
+        logger.debug(f"registered input files {input_files}")
```
benclifford, line 369: for debug logs in Parsl, @khk-globus and I have been pushing on a style that defers string interpolation until the log system has decided if it really wants to render that message - often that's not the case and rendering can be expensive. that format looks like this:

https://github.com/Parsl/parsl/blob/64e163ceaf4b43746909f30c9659738f29dd84e1/parsl/channels/local/local.py#L58

rather than using Python-level string formatting (either f-strings or `"".format(`

parsl/executors/taskvine/manager.py
```diff
@@ -128,6 +128,15 @@ def _prepare_environment_regular(m, manager_config, t, task, poncho_env_to_file,
         t.add_environment(poncho_env_file)
 
 
+def _handle_file_declaration_protocol(m, filename, cache):
+    if "taskvinetemp://" in filename:
+        return m.declare_temp()
+    elif "https://" in filename or "http://" in filename:
+        return m.declare_url(filename, cache=cache, peer_transfer=True)
+    else:
+        return m.declare_file(filename, cache=cache, peer_transfer=True)
```
benclifford, line 137: its a bit ugly that this isn't somehow using the parsed URL form of Parsl files to check the namespace, rather than doing ad-hoc basic URL parsing in this `if` statement... this is switching on the URI scheme only, I think, with the intention that it works like, for example, `StubStaging.can_stage_*` I think?

parsl/executors/taskvine/manager.py
```diff
@@ -364,18 +373,26 @@ def _taskvine_submit_wait(ready_task_queue=None,
...
                         else:
-                            task_in_file = m.declare_file(spec.parsl_name, cache=spec.cache, peer_transfer=True)
+                            task_in_file = _handle_file_declaration_protocol(m, spec.parsl_name, spec.cache)
                             parsl_file_name_to_vine_file[spec.parsl_name] = task_in_file
-                        t.add_input(task_in_file, spec.parsl_name)
+                        logger.debug("Adding input file {}, {} to TaskVine".format(spec.parsl_name, task.executor_id))
+                        if spec.netloc == '':
+                            t.add_input(task_in_file, spec.parsl_name)
+                        else:
+                            t.add_input(task_in_file, spec.netloc)
```
benclifford, line 382: I'm a bit unclear what the use of spec.netloc is here... is it the netloc from the user specified URL? does that mean that if you're saying http://foo/bar.txt, the task-vine side filename will be foo?  (rather than bar.txt or something else?)

parsl/executors/taskvine/stub_staging_provider.py
```diff
@@ -0,0 +1,37 @@
...
+from parsl.data_provider.files import File
+from parsl.data_provider.staging import Staging
+from parsl.utils import RepresentationMixin
+
+logger = logging.getLogger(__name__)
+
+known_url_schemes = ["http", "https", "taskvinetemp"]
+
+
+class StubStaging(Staging, RepresentationMixin):
```
benclifford, line 15: this name (which i probably invented) looks kinda lame when thinking about how a user would configure things: they'd maybe want to say they're using "TaskVineStaging" or "TaskVineAwareStaging" or something like that?

parsl/executors/taskvine/utils.py
```diff
@@ -69,12 +69,14 @@ class ParslFileToVine:
     """
     def __init__(self,
                  parsl_name: str,   # name of file
+                 netloc: str,       # name of file if url
```
benclifford, line 72: so why can't you use parsl_name for the name of file like the documentation one line before says? is it parsl_name is the parsl-supplied URL when the file is a URL?

parsl/tests/configs/taskvine_ex.py
```diff
@@ -2,9 +2,12 @@
...
+from parsl.data_provider.zip import ZipFileStaging
 from parsl.executors.taskvine import TaskVineExecutor, TaskVineManagerConfig
+from parsl.executors.taskvine.stub_staging_provider import StubStaging
 
 
 def fresh_config():
     return Config(executors=[TaskVineExecutor(manager_config=TaskVineManagerConfig(port=9000),
-                                              worker_launch_method='factory')])
+                                              worker_launch_method='factory',
+                                              storage_access=[FTPInTaskStaging(), StubStaging(), NoOpFileStaging(), ZipFileStaging()])])
```
benclifford, line 13: this drops off the default HTTP provider, so it's testing that HTTP works via this new mechanism not the other Parsl mechanisms. that's good.

Maybe you can drop off NoOpFileStaging too? (NoOpFileStaging really means "use the shared filesystem" staging, and I think taskvine in general is trying to manage those files better than the shared filesystem can?)

or at least, what breaks if you take away NoOpFileStaging here?

benclifford: i think it would be good to add a taskvine specific local test case that validates that taskvinetemp: URLs work (at least that the file contents get moved around correctly) when used with the TaskVineExecutor - seeing as that's the core functionality of this PR

parsl/executors/taskvine/stub_staging_provider.py
```diff
@@ -0,0 +1,37 @@
...
+    def can_stage_in(self, file):
+        logger.debug("Task vine staging provider checking passthrough for {}".format(repr(file)))
+        return file.scheme in known_url_schemes
+
+    def can_stage_out(self, file):
+        logger.debug("Task vine staging provider checking passthrough for {}".format(repr(file)))
+        return file.scheme in known_url_schemes
+
+    def stage_in(self, dm, executor: str, file: File, parent_fut: Optional[Future]) -> Optional[DataFuture]:
+        if file.netloc == '':
```
benclifford, line 26: i don't really understand the logic of these lines - theres a lot of moving fields around in weird ways in those two lines. for example using `netloc` for a filename is unusual. what's wrong with using `local_path` there?

parsl/executors/taskvine/executor.py
```diff
@@ -215,9 +218,9 @@ def __create_data_and_logging_dirs(self):
 
         # Create directories for data and results
         log_dir = os.path.join(run_dir, self.label)
-        self._function_data_dir = os.path.join(run_dir, self.label, "function_data")
+        self._function_data_dir = os.path.join("/tmp/function_data/", self.label, str(uuid.uuid4()))
```
benclifford, line 221: is this a change to say "taskvine is going to manage movements of all these files off this local temporary directory"? if so, is that something that works independently of this PR #3355? (eg. if you did it as a different PR, is that what would happen without any other changes?)

parsl/executors/taskvine/manager.py
colinthomas-z80: This change, and other netloc vs. filename related items are a result of some tricky naming items going on here.

My observation was that the current url parsing will fill in parsl_file.netloc with the "remote name", so bar.txt in your example. This is what we wish to refer to it as within the application

colinthomas-z80:
I brushed up on the urlparse documentation and found that netloc, is indeed supposed to give the "host" part of a url string. Why in most cases it seems to get filled out with the terminal filename I am not certain, but I will start using local_path anyway.

parsl/tests/configs/taskvine_ex.py
colinthomas-z80: We definitely want to keep the capability of using the shared filesystem. Removing NoOp staging fails some basic tests staging out local files. I don't see any reason at the moment to re implement local file handling through the new staging provider.

parsl/executors/taskvine/manager.py
colinthomas-z80: It is a bit redundant compared to what the staging provider does, but there is not a clear pass-through to declare our vine_file objects through the staging provider. I made a change here which brings the protocol/scheme through to this spot where we can switch on that, a little nicer than url string comparisons.

PR #3359, opened 2024-04-17
Demo of hacking in custom registration behaviour into interchange registrations

benclifford:
# Description

This is a demo for @cms21 of hacking in custom registration behaviour when a manager registers to an interchange.

(Longer term, I'd like this sort of thing to be possible through configurable monitoring radios, something like PR #3315, allowing people to plug in whatever behaviour they want, throughout parsl, when "interesting things happen")

# Changed Behaviour

you should see a file benc.demo appear in the working directory of the interchange, with a new line appended every time a manager registers

## Type of change

- demo

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -451,6 +451,8 @@ class Interchange:
                 logger.info("Registration info for manager {!r}: {}".format(manager_id, msg))
                 self._send_monitoring_info(hub_channel, m)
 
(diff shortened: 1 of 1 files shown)


benclifford:
no need for this to remain open

cms21:
Thanks for this @benclifford.  Could you keep this branch around for a little bit?

benclifford:
@cms21 yeah no problem - I just closed it to get it off the open PRs list

PR #3362, opened 2024-04-18
Remove nested exception from BadStdStreamFile exception

benclifford:
Prior to this PR, this exception class had a mandatory nested exception attibute.

This PR removes that exception, because Python already has some nesting mechanisms for exceptions which would serve well enough here: https://docs.python.org/3/tutorial/errors.html#exception-chaining

This change allows exceptions to be optionally chained like any other Python exception; or not chained at all, simplifying one use of BadStdStreamFile where a contrived TypeError was constructed to fill the slot.

# Changed Behaviour

Any user code catching BadStdStreamFile exceptions and extracting the chained exception will have to be adapted to live without that chained exception.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/app/bash.py
+++ b/parsl/app/bash.py
@@ -60,7 +60,7 @@ def remote_side_bash_executor(func, *args, **kwargs):
                 os.makedirs(os.path.dirname(fname), exist_ok=True)
             fd = open(fname, mode)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
Oh jeez; this is a "Duh!"

PR #3363, opened 2024-04-18
Fix broken stdstream handling for bytes and arbitrary os.PathLike objects; test harder.

benclifford:
# Description

Prior to this PR, an arbitrary os.PathLike was rendered to monitoring using that PathLike's str method. What should be appearing in the monitoring database should be a string representation of the path. This is an issue in (rare in practice?) cases where `__str__` and `__fspath__` return different things. This PR changes rendering to use `fspath` rather than `str`.

os.PathLike objects can also render as `bytes`, not `str`. This PR decodes those bytes using `.decode()`.

This PR adds tests for various valid data types going from app stdout/err parameters to the corresponding column in the monitoring database, including test cases that drove the above fixes.

# Changed Behaviour

This will change how some stdout/stderr paths are rendered to the monitoring database, hopefully for the better.

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_monitoring/test_stdouterr.py
+++ b/parsl/tests/test_monitoring/test_stdouterr.py
@@ -4,7 +4,10 @@
 import os
 import parsl
 import pytest
+import re
 import time
+
+from typing import Union
 
 from parsl.tests.configs.htex_local_alternate import fresh_config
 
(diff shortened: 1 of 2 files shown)


khk-globus (approved):
Nicely done with the test!

PR #3380, opened 2024-04-20
Decrease strategy poll period in stdout/err vs monitoring test

benclifford:
This removes 16 x 5s = 80 seconds of realtime during test execution.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Did I miscount?  I trust the 80 seconds removed per (I'm guessing) empirical evidence, but aren't there 10 x 2 parameterized tests, not 8 x 2?

Doesn't matter one iota; thanks for this.

PR #3385, opened 2024-04-22
Shutdown multiprocessing queues at monitoring shutdown

benclifford:
This frees up some threads and file descriptors.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
"Ouch!" that we've missed this for so long.  Good find.

PR #3387, opened 2024-04-23
Reduce test sleep()

khk-globus:
# Description

The checkpoint is the threshold of interest, so rather than waiting for an inordinate amount of time, dynamically ascertain what the checkpoint wait time is.  Add 1 second to that value for good measure.

Test time reduced from ~20s locally to ~4s.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/configs/local_threads_checkpoint_periodic.py
+++ b/parsl/tests/configs/local_threads_checkpoint_periodic.py
@@ -1,13 +1,11 @@
 from parsl.config import Config
 from parsl.executors.threads import ThreadPoolExecutor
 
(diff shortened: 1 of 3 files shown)


benclifford (commented):
now that there's nothing special happening in local_setup/local_teardown, that pair could be replaced by local_config returning the desired config. which will bring in some tmpd_cwd handling that is (deliberately) avoided by local_setup.

previously the special handling was capturing `dfk` as a global, but this PR changes that to use the *other* global that contains the test dfk reference.

benclifford (approved):

PR #3388, opened 2024-04-23
Make parsl.AUTO_LOGNAME generation pluggable

benclifford:
A generation function can now be specified at configuration time, with the previous behaviour invoked as default.

# Changed Behaviour

None: If this parameter is not specified, parsl.AUTO_LOGNAME should generate the same names as before this PR.

## Type of change

- New feature

### DIFF
--- a/parsl/config.py
+++ b/parsl/config.py
@@ -50,6 +50,9 @@ class Config(RepresentationMixin):
         of 1.
     run_dir : str, optional
         Path to run directory. Default is 'runinfo'.
+    std_autopath : function, optional
+        Sets the function used to generate stdout/stderr specifications when parsl.AUTO_LOGPATH is used. If no function
+        is specified, generates paths that look like: ``rundir/NNN/task_logs/X/task_{id}_{name}{label}.{out/err}``
     strategy : str, optional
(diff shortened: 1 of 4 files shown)


parsl/config.py
```diff
@@ -50,6 +50,9 @@ class Config(RepresentationMixin):
         of 1.
     run_dir : str, optional
         Path to run directory. Default is 'runinfo'.
+    std_autopath : function, optional
+        Sets the function used to generate stdout/stderr specifications when parsl.AUTO_LOGPATH is used. If no function
+        is specified, generates paths that look like: ``rundir/NNN/task_logs/X/task_{id}_{name}{label}.{out/err}``
```
khk-globus, line 55: `NNN` is clear to me; what is `X` in the path name?

benclifford: it's ` str(int(task_id / 10000)).zfill(4)` - the 10000s part of the task ID and upwards. I omitted a human description because it's awkwardly verbose for what I intended to be a shortish docstring... I guess something based around "dividing into directories with 10000 tasks each" would work?

khk-globus (approved):

PR #3393, opened 2024-04-23
Remove some local-dev test detritus

khk-globus:
## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_bash_apps/test_memoize_ignore_args.py
+++ b/parsl/tests/test_bash_apps/test_memoize_ignore_args.py
@@ -1,7 +1,5 @@
 import os
 
-import pytest
-
 import parsl
 from parsl.app.app import bash_app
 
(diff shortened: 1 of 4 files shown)


parsl/tests/test_staging/test_file.py
```diff
@@ -22,11 +22,11 @@ def test_files():
...
+def test_open(tmpd_cwd):
+    fpath = tmpd_cwd / 'test-open.txt'
+    fpath.write_text('Hello')
 
-    pfile = File('test-open.txt')
+    pfile = File(fpath)
 
-    with open(str(pfile), 'r') as opfile:
-        assert (opfile.readlines()[0] == 'Hello')
+    with open(pfile) as opfile:
```
benclifford, line 31: this is changing what's being tested: before, it tested that whatever a File returned as its `str` could be opened. Now it's testing that whatever comes back as its `fspath` can be opened.

That `str` implementation is a bit horrible (it raises an exception for File objects which don't have a local representation) but I think it's still needed as part of usability: for example, people writing bash_apps would traditionally write `f"cat {myfile}"` when forming their bash command lines rather than `f"cat {os.fspath(myfile)}"

So maybe this should be testing *both* behaviours?

parsl/tests/test_bash_apps/test_memoize_ignore_args_regr.py
```diff
@@ -30,21 +28,17 @@ def no_checkpoint_stdout_app(stdout=None):
...
 
-    path_x = "test.memo.stdout.x"
-    if os.path.exists(path_x):
-        os.remove(path_x)
+    path_x = tmpd_cwd / "test.memo.stdout.x"
 
     # this should run and create a file named after path_x
-    no_checkpoint_stdout_app(stdout=path_x).result()
-    assert os.path.exists(path_x)
+    no_checkpoint_stdout_app(stdout=str(path_x)).result()
```
benclifford, line 37: Recent work *should* have this still work with path_x being an os.PathLike, which I think is the case here - for example, see the monitoring tests in parsl/tests/test_monitoring/test_stdouterr.py (although those tests only check that the path goes through to monitoring correctly - not that the path really is used by the executing app)

benclifford: (i.e. this `str` should be unnecessary)
(and maybe should be `os.fspath`)
(at which point you might get a `bytes`)
(and have to deal with that)

parsl/tests/test_bash_apps/test_memoize_ignore_args.py
```diff
@@ -22,24 +21,18 @@ def no_checkpoint_stdout_app_ignore_args(stdout=None):
...
 
-def test_memo_stdout():
+def test_memo_stdout(tmpd_cwd):
+    path_x = tmpd_cwd / "test.memo.stdout.x"
 
     # this should run and create a file named after path_x
-    path_x = "test.memo.stdout.x"
-    if os.path.exists(path_x):
-        os.remove(path_x)
+    no_checkpoint_stdout_app_ignore_args(stdout=str(path_x)).result()
```
benclifford, line 28: see other comment about hopefully-redundant `str`

benclifford:
@khk-globus these seem good. I made some comments you might want to address before merge if you have time/capacity, but I will label approved so you can also merge without.

PR #3395, opened 2024-04-24
Implement stage-in for zip staging provider

benclifford:
## Type of change

- New feature

### DIFF
--- a/parsl/data_provider/zip.py
+++ b/parsl/data_provider/zip.py
@@ -42,6 +42,12 @@ class ZipFileStaging(Staging):
     """
 
     def can_stage_out(self, file: File) -> bool:
+        return self.is_zip_url(file)
+
+    def can_stage_in(self, file: File) -> bool:
+        return self.is_zip_url(file)
+
+    def is_zip_url(self, file: File) -> bool:
         logger.debug("archive provider checking File {}".format(repr(file)))
 
         # First check if this is the scheme we care about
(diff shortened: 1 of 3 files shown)


parsl/data_provider/zip.py
```diff
@@ -76,6 +82,20 @@ def stage_out(self, dm, executor, file, parent_fut):
...
+        if working_dir:
+            file.local_path = os.path.join(working_dir, inside_path)
+
+        stage_in_app = _zip_stage_in_app(dm)
+        app_fut = stage_in_app(zip_path, inside_path, working_dir, outputs=[file], _parsl_staging_inhibit=True, parent_fut=parent_fut)
+        return app_fut._outputs[0]
+
 
 def _zip_stage_out(zip_file, inside_path, working_dir, parent_fut=None, inputs=[], _parsl_staging_inhibit=True):
     file = inputs[0]
```
khk-globus, line 101: For both `_zip_stage_in` and `_zip_stage_out`, are these interfaces set by another driver?  I'm observing that in _these_ internal methods, the following arguments don't currently appear to be utilized:
- `working_dir`
- `parent_fut`
- `_parsl_staging_inhibit`

parsl/data_provider/zip.py
```diff
@@ -93,6 +113,18 @@ def _zip_stage_out_app(dm):
     return parsl.python_app(executors=['_parsl_internal'], data_flow_kernel=dm.dfk)(_zip_stage_out)
 
 
+def _zip_stage_in(zip_file, inside_path, working_dir, *, parent_fut, outputs, _parsl_staging_inhibit=True):
+    with filelock.FileLock(zip_file + ".lock"):
+        with zipfile.ZipFile(zip_file, mode='r') as z:
+            content = z.read(inside_path)
+        with open(outputs[0], "wb") as of:
+            of.write(content)
```
khk-globus, line 121: I wonder if there's a way to think in lists.  This method:
1. takes out a lock
2. opens the (zip) file
3. reads the content from the zip file
4. closes the (zip) file
5. opens a new file
6. writes the content
7. closes the new file

There's clearly no working around steps 3, 5, 6, and 7, but steps 1, 2, and 4 could be aggregated for multiple reads, if, say, `inside_path` were of type `List` and there were an appropriate number of outputs.

That could further be better for memory by reading and writing in chunks.  Consider:
```python
with filelock.FileLock(...):
    with zipfile.ZipFile(...):
        for inp, outp in zip(inside_paths, outputs):
            with open(outp, "wb") as of:
                shutil.copyfileobj(inp, outp)
```

I won't write a similar comment for `_zip_stage_out`, but I'm thinking in similar terms.  "It would be nice if this could be implemented with list-think."

PR #3396, opened 2024-04-24
Add first unit test for File

khk-globus:
We may have other UTs littered among the other tests, but this file now sets the stage (via the new `unit/` directory) for where to place them.  A not-complete metric for UTs: do they they run _fast_?  If not, or if the test includes `.sleep()`, it's probably not a UT.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/unit/test_file.py
+++ b/parsl/tests/unit/test_file.py
@@ -1,0 +1,99 @@
+import os
+from unittest import mock
+
+import pytest
+
+from parsl import File
+
+_MOCK_BASE = "parsl.data_provider.files."
+
+
+@pytest.mark.local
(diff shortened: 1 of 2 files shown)


parsl/tests/unit/test_file.py
```diff
@@ -0,0 +1,99 @@
...
+@pytest.mark.parametrize("scheme", ("https", "ftps", "", "file", "asdfasdf"))
+def test_file_repr(scheme):
+    netloc = "some.netloc"
+    filename = "some_file_name"
+    path = f"/some/path/{filename}"
+    if scheme and scheme != "file":
+        url = f"{scheme}://{netloc}{path}"
+    else:
+        scheme = "file"
+        url = path
```
benclifford, line 85: you *should* be able to use a file: scheme in a full URL too

benclifford (approved):

PR #3397, opened 2024-04-25
[not for merge] tidying threads and file descriptors at shutdown

benclifford:
this is a patch stack in my dev environment where i have been pushing on cleanup of threads and file descriptors at shutdown, with the eventual goal of having pytest fixtures assert that only the main thread is running and the same number of file descriptors are open at the start and the end. I've been merging work from this already in small pieces and will continue to merge small tidyups from it.

There is at least one large part which needs some thought and discussion, which is how to shut down the htex queue processing thread - the engine of almost everything in htex that doesn't happen immediately on user task submission.


benclifford:
the thread check for a moment worked but then was broken by #3607 which starts a second thread at test initialization time and does not clean it up - there is a fragile workaround (checking that the number of threads is the same before and after a test) that would fail in an obvious test rearrangement of importing GC inside tests, so it isn't required to be installed for running tests. See for example this which does not require sqlalchemy to be installed for non-monitoring tests: https://github.com/Parsl/parsl/blob/6ac930eabab520c3de8f2d88f7940ae5a17e11ca/parsl/tests/test_monitoring/test_app_names.py#L54

benclifford:
The radical MPI test is also failing: parsl/tests/test_radical/test_mpi_funcs.py

```
            post_ac = threading.active_count()
>           assert pre_ac == post_ac, "test left threads running: " + repr(threading.enumerate())
E           AssertionError: test left threads running: [<_MainThread(MainThread, started 139679236183872)>, <Thread(Thread-20 (_listener), started daemon 139676128106176)>]
```

Issue #34, opened 2017-10-04
Usage Tracking and logging for Parsl project reporting


yadudoc:
For reporting purposes we need to capture the following information

 - [ ] Hash of Username, hostname (ip) 
 - [ ] Hash of Parsl script (if possible)
 - [ ] Count of apps launched by DFK
 - [ ] Site IDs used
 - [ ] Total run time

A UDP packet with this info could be sent out from the Parsl script and this could be captured and logged on a server hosted on AWS.
Here are the components we'll need
 - [ ] Log info on the client side and send a UDP packet
 - [ ] Receive and log info to stable storage on our side.
 - [ ] Config option to opt-out of tracking
 - [ ] Documentation to clearly indicate the info collected and how to opt-out

kylechard:
At this stage we should focus on the bare minimum (and least invasive). 

Some of these stats can be sent at the beginning while others are sent at the end. Sending at the end is slightly risky as there might be failures that would kill the script.  

Begininig
- [ ] hash of username+IP (to avoid the millions of 'ubuntu' users)
- [ ] UDP packet will include sending IP
- [ ] start time

End
- [ ] number of apps launched by DFK
- [ ] number of app failures
- [ ] number of sites used
- [ ] total compute hours
- [ ] end time

In the future we could add the following
- [ ] site ID (when we have a site registry)
- [ ] hash of the script

annawoodard:
We might also consider a username-based filter for devs, which would automatically set a flag to lump testing stats separately from user stats. Presumably for project reporting we only want user stats, and it could be a pain to disentangle them post hoc. Alternatively we could have a 'testing mode' configuration flag, but that makes it more likely we forget to set it sometimes and inflate the user stats.

kylechard:
Agreed. We actually discussed this same topic this afternoon. The thought was to have a testing mode for our usage and for our automated tests.

annawoodard:
Sounds good! Were you guys thinking a testing mode configuration flag, or an automated filter? Everyone makes mistakes, so if we let it be possible for us to forget, I think we will forget from time to time.

Filter pros:
1) Eliminates per-workflow possibility dev forgets to set the flag

Filter cons:
1) Introduces per-dev possibility we forget to add dev to the filter
2) Possible collisions (user gets filtered as dev)-- probably unlikely for now
3) Less flexible: there may be situations where devs should count as users
4) More work for new people to contribute

[...]

PR #3400, opened 2024-04-25
Allow users to select their preferred level of usage tracking

NishchayKarle:
# Description

This PR introduces a choice of 3 levels for users to select based on their preferred level of usage reporting. It introduces updates on top of #3229.

This PR modifies ```UsageTracker``` to check for level of usage tracking set, and create start and end messages based on it.

### Tracking Levels
1. Level 1:  python version, parsl version, operating system details. 
2. Level 2: configuration details + Level 1
3. Level 3: total apps run, total failed apps, execution time + Level 2

Usage tracking if currently enabled will be defaulted to level 1. 

### Usage Data sent at launch (Levels 1 and 2)
• Capture Parsl version, Python version, and environment details at startup.
• Configuration Reporting: Log details about providers, launchers, execu-
tors, channels, and storage access methods used.

### Usage Data sent on closure (Level 3 only)
• Number of apps ran
• Number of failed apps 
• Total time elapsed

An example start message for level 1 looks like this: (pretty-formatted)
```
{
    "correlator": "5ef1538b-c03c-4f7c-a832-12a69fa270c6",
    "parsl_v": "1.3.0-dev",
    "python_v": "3.12.2",
    "platform.system": "Darwin",
    "tracking_level": 1,
}
```

Similarly, start message for level 2
```
{
    "correlator": "d2c5507a-52b5-4482-a8e0-b0636d79b03e",
    "parsl_v": "1.3.0-dev",
    "python_v": "3.12.2",
    "platform.system": "Darwin",
    "tracking_level": 2,
    "components": [
        {"c": "parsl.config.Config", "executors_len": 1},
        ["parsl.executors.threads.ThreadPoolExecutor"],
    ],
}
```

An example end message for level 3 looks like this: 
```
{
    "correlator": "f49c2956-faee-4b1f-bd0c-e4227d358af6",
    "end": 1714038716,
    "execution_time": 15,
    "components": [
        {"c": "parsl.dataflow.dflow.DataFlowKernel", "app_count": 3, "app_fails": 0},
        {"c": "parsl.config.Config", "executors_len": 1},
        ["parsl.executors.threads.ThreadPoolExecutor"],
    ],
}
```

# Changed Behaviour

Different information will be reported via usage tracking - anyone processing that usage data will need to adapt their code.

## Type of change

- New feature


benclifford (changes requested):
This uses lots of code from PR #3229 which I initially put up as an example, but it looks like it was OK for your work, and that @khk-globus has subsequently reviewed - so I think it would be better to get #3229 merged and then your work on top of that will be clearer.

benclifford:
I'll work on getting #3229 merged and then you can update against that.

NishchayKarle:
Okay, thanks. I did make the changes suggested from the  #3229 review.

parsl/usage_tracking/usage.py
```diff
@@ -98,90 +99,121 @@ def __init__(self, dfk, port=50077,
...
-        self.tracking_enabled = self.check_tracking_enabled()
-        logger.debug("Tracking status: {}".format(self.tracking_enabled))
+        self.python_version = (
+            f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
+        )
+        self.tracking_level = self.check_tracking_level()
+        self.start_time = None
+        logger.debug(f"Tracking level: {self.tracking_level}")
+
+    def check_tracking_level(self):
```
benclifford, line 115: its good to add type annotations on new functions

parsl/usage_tracking/usage.py
```diff
@@ -98,90 +99,121 @@ def __init__(self, dfk, port=50077,
...
+
+        Returns: int
+            - 0 : Tracking is disabled
+            - 1 : Tracking is enabled with level 1
+                  Share info about Parsl version, Python version, platform
+            - 2 : Tracking is enabled with level 2
+                  Share info about config + level 1
+            - 3 : Tracking is enabled with level 3
+                  Share info about app count, app fails, execution time + level 2
+        """
```
benclifford, line 126: can this code ever return a different value than the above?

parsl/usage_tracking/usage.py
```diff
@@ -98,90 +99,121 @@ def __init__(self, dfk, port=50077,
...
 
-        Tracking will be enabled unless either of these is true:
+        if self.config.usage_tracking:
+            level = 1 if self.config.usage_tracking is True else int(self.config.usage_tracking)
 
-            1. dfk.config.usage_tracking is set to False
-            2. Environment variable PARSL_TRACKING is set to false (case insensitive)
+        envvar = str(os.environ.get("PARSL_TRACKING", False)).lower()
+        if envvar in {"true", "1", "2", "3"}:
+            level = 1 if envvar == "true" else int(envvar)
```
benclifford, line 135: what happens if I set the envvar to something else? (such as Yes or On or Disable)?

parsl/usage_tracking/usage.py
```diff
@@ -98,90 +99,121 @@ def __init__(self, dfk, port=50077,
...
+                  Share info about app count, app fails, execution time + level 2
+        """
 
-    def check_tracking_enabled(self):
-        """Check if tracking is enabled.
+        level = 0
 
-        Tracking will be enabled unless either of these is true:
+        if self.config.usage_tracking:
+            level = 1 if self.config.usage_tracking is True else int(self.config.usage_tracking)
```
benclifford, line 131: I think this code is expecting that self.config.usage_tracking might be a general int. `Config` should be constraining self.config.usage_tracking to always be a `bool`.

benclifford:
@NishchayKarle I merged PR #3229. There are a few uncontroversial looking merge conflicts you'll have to resolve, and then your PR #3400 should look much more like only what you've been working on.

parsl/usage_tracking/usage.py
NishchayKarle: Yes. Config does constrain usage_tracking to be a bool. I can change it to be a an int, and set it within `Config` as
```python
class Config(...):
  def __init__(self, ..., usage_tracking: int = 0, ... ):
    ...
    self.usage_tracking = int(usage_tracking) # we change currently set True or False to be Level 1 and 0
    ...
```

Or we can add another Parameter ```usage_tracking_level: int``` in `Config`

What do you think?

benclifford: better to have one parameter that contains the whole config.

you could maybe validate its range within `Config.__init__` too, to raise an exception as early as possible.

benclifford: this is also an indication that another programmer working on the code might make similar mistakes in future, so it would be a good thing to add some kind of test case in parsl/tests/ - for example, you could add a test that checks that Config objects can be instantiated with some appropriate usage tracking values (True, False, 0, 3), and fail with some inappropriate ones ("yes", 4, -1)

benclifford:
@NishchayKarle the diff (https://github.com/Parsl/parsl/pull/3400/files) for this PR contains a lot of noise that is not part of the subject of this PR of allowing users to change usage tracking - for example, changing quote symbols between `"` and `'`, capitalisation changes in symbol names, other fields added to the monitoring info, code reformatting, override of encoding defaults.

For all of those changes, you should open separate PRs, with justification about why they're needed. For example, if there are code style changes, make a code style change PR - you should see a lot of PRs in parsl history that only rearrange code and deliberately don't change functionality, so that problems later can be tracked down to bad code reformatting rather than functionality changes.

parsl/usage_tracking/usage.py
benclifford: this is still an open question

NishchayKarle: Anything other than 'True', 'true', '1', '2' or '3' will result in usage_tracking level 0

parsl/usage_tracking/usage.py
NishchayKarle: No.
config will raise an Exception if usage_tracking values are other than 0 (or False), 1 (or True), 2, 3
envvar can only be 'true', '1', '2' or '3'. Anything else is treated as level 0.

parsl/usage_tracking/usage.py
```diff
@@ -110,28 +110,46 @@ def __init__(self, dfk, port=50077,
...
+        Checks the following in order:
+            1. PARSL_TRACKING environment variable
+                - Possible values:
+                    ["true", "false", "True", "False", "0", "1", "2", "3"]
+                - Other values are treated as Level 0 (disabled)
+
+            2. usage_tracking in Config
+                - Possible values: [True, False, 0, 1, 2, 3]
+
+            Config.usage_tracking overrides PARSL_TRACKING if both are set.
```
benclifford, line 129: I think this is not entirely true: I set PARSL_TRACKING=3, and Config(usage_tracking=0) and I see two UDP packets being sent.

NishchayKarle: Since I chose to override env variable if usage_tracking is set in config, we cannot differentiate setting```usge_tracking = 0```in the config and usage tracking being disabled by default. So in this specific case I had the env variable override the config. 

I can fix the unnecessary confusing by having PARSL_TRACKING env variable override what's set in the config if usage_tracking is set in both places. What do you think?

benclifford: another policy with that extra "not set" info available would be "parsl will choose the lowest level that is specified by any of the parameters, and 0 if none of them are set"

benclifford: That symmetric policy appeals to me because it's symmetric... There's no arbitrary priority of one setting over the other. And it uses the lowest value that the user has chosen, so there is no "I set a low level but high level info got reported" behaviour.


I think you could compute it something like take the min of (+infinity, usage_tracking if set, PARSL_TRACKING if set) and it you end up with +infinity, that means default to 0

NishchayKarle: Made changes to set usage_tracking parameter default to `None` and choose the lowest level that is specified by any of the parameters, and 0 if none of them are set

parsl/usage_tracking/usage.py
```diff
@@ -110,28 +110,53 @@ def __init__(self, dfk, port=50077,
...
-            2. Environment variable PARSL_TRACKING is set to false (case insensitive)
+        envvar_level = inf
+        envvar = str(os.environ.get("PARSL_TRACKING", 0)).lower()
+        if envvar in {"true", "false"}:
+            envvar_level = 1 if envvar == "true" else 0
 
-        """
-        track = True
+        elif envvar in {"0", "1", "2", "3"}:
+            envvar_level = int(envvar)
```
benclifford, line 150: I'm a bit uncomfortable with the style of this `if` statement:

i) try to have an `else` statement, unless there is some reason for not having one: That `if` won't fire always (that's why it is there) and if you write an explicit `else`, you can be clearer about what you intend to happen in that case. For example, in this case, I think the `else` behaviour is that we end up with envvar_level set to inf.

So you can set envvar_level to that in an else statement instead of assigning it first and then adjusting it later.

(that's a functional programming style that I've picked up from eg Haskell and Rust)

ii) In line 147, avoid doing an if than then immediately contains another if: 
the true and false cases are different, so make them different top level clauses in this if statement

So you could end up with a sequence of if/elif/elif/.../else where envvar_level is explicitly assigned in each of those clauses.

then, iii) we can look at that `else envvar_level = inf`, and ask if that is the right behaviour for every other value of PARSL_TRACKING that a user might specify?

NishchayKarle: I've refactored the code to address suggestions from i and ii. 

For point iii, I've assigned `envvar_level = inf` when the `PARSL_TRACKING` value does not match
any conditions specified in the if statements, as we anticipate only the values specified by these conditions to be valid.

benclifford: I absolutely anticipate users setting PARSL_USAGE_TRACKING to other values, for example, `trrue` as a typo, or 99999 to get the maximum tracking by using the biggest number they can imagine is plausible.

It would be good to lead users to a solution to why this code is ignoring their specified tracking level, rather than silently failing: at the very least a `logger.warn`, but I think because this code ultimately runs in DFK startup (`DataFlowKernel.__init__` -> `UsageTracker.__init__` -> `UsageTracker.check_tracking_level`, it would be good to raise a configuration exception in that case.

parsl/usage_tracking/usage.py
```diff
@@ -110,28 +110,60 @@ def __init__(self, dfk, port=50077,
...
+            - 2 : Tracking is enabled with level 2
+                  Share info about config + level 1
+            - 3 : Tracking is enabled with level 3
+                  Share info about app count, app fails, execution time + level 2
+        """
+        inf = sys.maxsize
 
-    def check_tracking_enabled(self):
-        """Check if tracking is enabled.
+        envvar = str(os.environ.get("PARSL_TRACKING", 0)).lower()
```
benclifford, line 144: does this default to 0, and then force tracking off if PARSL_TRACKING is not set?

NishchayKarle: Changed this to default to None and setting `envvar_level = inf` in that case

parsl/usage_tracking/usage.py
NishchayKarle: Added Exceptions for invalid values for PARSL_TRACKING set and also for invalid values set in the config

parsl/usage_tracking/usage.py
```diff
@@ -110,28 +111,70 @@ def __init__(self, dfk, port=50077,
...
+        else:
+            raise ConfigurationError(
+                f"PARSL_TRACKING values must be true, false, 0, 1, 2, or 3 and not {os.environ.get('PARSL_TRACKING')}"
+            )
+
+        if self.config.usage_tracking is None:
+            config_level = inf
+
+        elif 0 <= int(self.config.usage_tracking) <= 3:
+            config_level = int(self.config.usage_tracking)
```
khk-globus, line 167: Thanks to `typeguard` this is a useless conversion.  `self.config.usage_tracking`, if not None, is guaranteed to be an integer.  Further, if the `usage_tracking` is always set (as discussed in another comment), the previous branch (`is None`) is also unnecessary.

parsl/config.py
```diff
@@ -132,6 +135,7 @@ def __init__(self,
         self.strategy_period = strategy_period
         self.max_idletime = max_idletime
         self.usage_tracking = usage_tracking
+        self._validate_usage_tracking()
```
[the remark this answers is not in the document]
NishchayKarle: Ben had the suggestion in an earlier comment of having usage_tracking parameter default to `None` meaning unspecified in the same way that `PARSL_TRACKING` env variable can also represent unspecified by being not set. 

I can change `None` to be a constant greater than `3` to represent the unspecified state. Making it `-1`, I think would be confusing when selecting the minimum level if usage tracking is set both in the config and the environment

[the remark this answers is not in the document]
benclifford: @khk-globus `None` represents "no preference specified here" - it doesn't have any concrete integer tracking level equivalent, but instead indicates to the defaulting code later on that when choosing between: the `Config` specified tracking level, the `$PARSL_TRACKING` specified tracking level and the default tracking level, this config has no preference in that choice.

The integer representation comes along in code that is also aware of `$PARSL_TRACKING` in parsl/usage_tracking/usage.py to evaluate that "what's every bit of preference combined?".

So the value in `Config` is "user tracking preference" (?) that can be "no preference" or "one of these levels", the same possibilities as expressed in the `PARSL_TRACKING` tracking preference environment variable  - vs the actual tracking level later, which must be just one of the concrete levels synthesised from all preferences.

It's probably a bit over-engineered to even have these two different ways of specifying tracking level at all, given the last 6 years neglect of the usage project, and the ridiculous discrete maths we're ending up using to model and resolve this...

... but that's what posterity has delivered us (see issue #34 and related work in 2018).

Later on, usage_tracking makes use of an approximation of positive infinity (in place of None) to evaluate the minimum value of the ordered set 0<1<2<3<None... potentially this code could make that substitution right here, but it feels like a weird user experience when logging out configs: the requested tracking level is not infinity. the infinity is there because it's the 0-element of the `min` monoid, not because it's a desired tracking level.

benclifford:
Now that PR #3466 is merged, I think you (@NishchayKarle ) can revisit the work in this PR with a much simpler configuration interface.

benclifford (approved):

PR #3402, opened 2024-04-26
Rework description of --config local tests

benclifford:
The previous description of putting `--config local` tests into /sites/ and /integration/ does not reflect the reality of current Parsl development. cc @khk-globus who has put some thought into test organisation, but for now remove that advice.

## Type of change

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -50,18 +50,16 @@ There are two broad groups of tests: those which must run with a
 specific configuration, and those which should work with any
 configuration.
 
-Tests which run with a specific configuration live under the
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Makes sense; I don't actually know when we do run those site-specific tests.  Is that something you do?

benclifford:
parsl/tests/sites/ is "just another directory" of tests, including tests that are nothing to do with tests.

No one runs any site-specific tests ever - that *cough* principle is one of the drivers for moving to weekly auto-releases with explicitly accepting there is no human testing. See the work leading up to PR #2612 which made the first autorelease after no release at all for more than a year.

PR #3403, opened 2024-04-29
Remove return bool from Channel.close()

benclifford:
Theres was no defined meaning to the return bool either in the code or in the human text:

- no code calls close to give an example of interpretation

- it's not success/fail because local channel returns False to indicate that it didn't need to do anything, not that there was a failure.

Other .close() style methods return None and raise an exception if there is a problem. This PR pushes Channel.close() in this direction.

A separate PR will actually invoke this .close() method.

# Changed Behaviour

This is an apparently unused return value so should not affect anything.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/channels/base.py
+++ b/parsl/channels/base.py
@@ -89,15 +89,8 @@ class Channel(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def close(self) -> bool:
-        ''' Closes the channel. Clean out any auth credentials.
-
-        Args:
-            None
-
-        Returns:
-            Bool
-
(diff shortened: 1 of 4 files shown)


khk-globus (approved):
I'm not sure I buy the argument in this case that just because it's not utilized that it should be removed.  On the other hand, "meh."  We can re-add it at a future date, concomitant with its actual use-case, if needs be.

benclifford:
@khk-globus the argument isn't that it isn't used - it's that it has no defined meaning if you try to use it for ... something?

PR #3408, opened 2024-04-30
fix variable name

patzm:
# Description

Fixed a typo / variable name bug in the docs

# Changed Behaviour

nothing

# Fixes

no issue opened yet

## Type of change


WardLT:
Good catch, but could you change `exec` -> `pool` instead of `pool`->`exec`.

In retrospect, `exec` is a reserved word in Python so we shouldn't be using it in the example.

Thanks!

patzm:
sure, done

WardLT (approved):

benclifford:
merged, thanks!

PR #3410, opened 2024-05-02
Close monitoring processes at shutdown, freeing some fds

benclifford:
Before this PR, pytest parsl/tests/test_monitoring/ leaves 1093 fds open (which is too much for the test suite to complete in some default configurations of max fds 1024)

After this PR, that test run leaves 895 fds open.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Wow; amazing what repetition will do.  Three `.close()` calls to reap ~200 fds.  Heh.  Keep on, good sir!  I'm liking this line of work a lot.

PR #3411, opened 2024-05-02
Remove vestigial monitoring parameter

benclifford:
The use of this parameter was removed in PR #3346

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Och; I even took a mental note about when looking at the other PR.  It obviously overflowed my mental cache before I put it down.  Thanks for this.

Issue #3414, opened 2024-05-04
WQ executor over-scales resources because it assumes one WQ worker only ever executes one task


svandenhaute:
**Describe the bug**
In Parsl+WQ, for each block that is provided, a WQ worker is spawned using `work_queue_worker` which checks (or is given a list of) the available resources. Based on that, it executes tasks. The number of tasks it can execute at any given point is variable based on the resource annotations of the specific app invocations that have been submitted to the dfk at that time.

At the moment, Parsl kind of assumes that each worker (i.e. each block from the provider) can only ever execute one task at a time. If 10 new tasks roll in which are resource-annotated with `cores: 1`, there is only a total of 10 cores required to execute these tasks, so let's say 1 block from a provider with 64 cores is required as scale out. Unfortunately, at the moment, Parsl+WQ seems to assume that these 10 tasks should get executed on separate workers, which therefore requires 10 blocks from the provider. Especially for smaller tasks, this induces massive over-scaling.

**To Reproduce**
Run any WQ workflow in which the resources required per task are significantly smaller than the total resources in a single node that are available to `work_queue_worker`

**Expected behavior**
In the example above, scale out should only request one extra block (0 blocks < 10 cores < 1 block) instead of 10 blocks.

**Environment**
Any

**Distributed Environment**
`BlockProvider`

benclifford:
cc @dthain - I'm not sure what the work queue story is in general on this.

In the parsl block (aka batch job) scaling code, this behaviour is coming from the assumption that we know ahead of time a count of "task slots" available on each node and hence in each batch job. That works for htex which allocates based on a fixed task size.

With WQ-style different-resources-per-task model, that number doesn't exist. Instead, there's a hard coded `1` here: https://github.com/Parsl/parsl/blob/ef1ff6c2925323ec4a2058c857084f0335ab14fd/parsl/executors/workqueue/executor.py#L670 (read `workers_per_node` as "task slots per node" for scaling purposes)

I'm personally not particularly beholden to the Parsl block scaling code being used for *everything* and I know that Task Vine has some more intricate task factory configuration...

benclifford:
If you're in a position to estimate a "task slots per worker" value based on your own understanding of your workload, it is hopefully straightforward to add that as a configuration parameter to the work queue executor and return that value instead of `1` mentioned above.

In more complicated land, let's see if @dthain has anything already or opinions on how this might go in the future.

benclifford:
[...]

Issue #3323, opened 2024-04-09
Outreachy project: Worker selection algorithms in the High Throughput Executor Interchange


benclifford:
Parsl will hopefully supervise one intern with [Outreachy](https://outreachy.org) in the middle of 2024 on a project to add extra worker selection algorithms to the High Throughput Executor. This issue should track activity on this project.

The short project description from the outreachy website is:

===

Project Description
More Efficient Worker Selection in Parsl's High Throughput Executor

Parsl's High Throughput Executor distributes tasks given to it by the core of Parsl to be executed on worker nodes. Right now, it picks an available worker randomly. This has worked well over the last 5 years, but there are some situations where a different algorithm could be used to pick a worker, and we think that we would get better performance if so.

We have four different ideas for algorithms. We dive further into these algorithms in the “Intern Tasks” section.

The goal of this internship is to implement some of those algorithms, to understand why they might (or might not) be helpful, and to measure how well they do or do not work.

 

Internship Tasks
Getting Started

Get Parsl running and a development environment set up
Running Parsl on a bigger computer system, which we'll get you access to
Making a Pull Request – we'll walk you through making one to gain confidence
Learn about Parsl's task distribution system (part of Parsl that you'll be working on the most) and Parsl's monitoring system (measure how efficiently things are working). This involves reading the source code, experimenting to understand the flow, and having hack sessions with mentors.

The First Big Change: After understanding how to implement different selection algorithms, where to make them, and why, you will:

Make those changes!
Test them out with your mentor
Write tests to check your changes work (so that when other people change the code, they don't break your stuff)
Write documentation so users understand when and why they would want to use what you have implemented
Make a pull request and experience the review process
Get your changes merged and ready for release!
There are four algorithms we would like to try (descriptions of each will be provided):

1: Scaling down worker nodes more effectively
2: Avoiding nodes that will run out of execution time
3: Multi-block MPI
4: Spreading the load evenly
You don't have to do them all – even one would be a great addition to our codebase and open the way for other people to continue. And don't worry if you don't understand everything written here right now – you will learn more about what's going on during the internship.

Intern Benefits
We hope you'll get some good experience writing code that will be used by real users. We'd like you to get experience with how to take a new feature all the way to being released for users to use, so as well as writing code; you'll get some experience writing documentation, making sure things are tested, and explaining how your code works to other developers. In addition, you will make valuable connections and gain professional references.

===

benclifford:
[...]

Issue #3404, opened 2024-04-29
interchange hangs on command server thread exit


benclifford:
**Describe the bug**
Under high file descriptor load, I have seen this (entire) log file, along with a test hang. The interchange process was still running, but it looks like the command thread was broken.

```
2024-04-29 06:23:33.162 interchange:139 HTEX-Interchange(77941) MainThread __init__ [INFO] Attempting connection to client at 127.0.0.1 on ports: 55953,55652,55843
2024-04-29 06:23:33.163 interchange:151 HTEX-Interchange(77941) MainThread __init__ [INFO] Connected to client
2024-04-29 06:23:33.163 interchange:182 HTEX-Interchange(77941) MainThread __init__ [INFO] Bound to ports 54690,54379 for incoming worker connections
2024-04-29 06:23:33.164 interchange:198 HTEX-Interchange(77941) MainThread __init__ [INFO] Platform info: {'parsl_v': '1.3.0-dev', 'python_v': '3.12.2', 'os': 'Linux', 'hostname': 'parsl-dev-3-12-9270', 'dir': '/home/benc/parsl/src/parsl'}
2024-04-29 06:23:33.164 interchange:354 HTEX-Interchange(77941) MainThread start [INFO] Incoming ports bound
2024-04-29 06:23:33.165 interchange:248 HTEX-Interchange(77941) MainThread _create_monitoring_channel [INFO] Connecting to MonitoringHub
2024-04-29 06:23:33.165 interchange:253 HTEX-Interchange(77941) MainThread _create_monitoring_channel [INFO] Connected to MonitoringHub
2024-04-29 06:23:33.165 interchange:229 HTEX-Interchange(77941) Interchange-Task-Puller task_puller [INFO] Starting
2024-04-29 06:23:33.165 interchange:248 HTEX-Interchange(77941) Interchange-Command _create_monitoring_channel [INFO] Connecting to MonitoringHub
2024-04-29 06:23:33.165 interchange:31 HTEX-Interchange(77941) Interchange-Command wrapped [ERROR] Exceptional ending for _command_server on thread Interchange-Command
Traceback (most recent call last):
  File "/home/benc/parsl/src/parsl/parsl/process_loggers.py", line 27, in wrapped
    r = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/interchange.py", line 275, in _command_server
[...]

PR #3415, opened 2024-05-05
Make WorkQueue scaling aware of core counts

benclifford:
@svandenhaute was getting frustrated by having different sized tasks and the current parsl-vs-wq scaling not being able to deal with that - see issue #3414 - as of existing parsl `master` the best we could come up with was estimating an average task size; but that does not work well when tasks are far from average (eg a period with above size tasks, followed by a period with under-sized tasks).

this PR demonstrates changing scaling (WQ only) to pay attention to number of cores requested by a task in its parsl_resource_specification and use that in the scaling calculation

In @svandenhaute actual workload, number of cores is used as a proxy for "amount of node needed" - both CPU cores and number of GPUs, and this PR is probably better developed in that direction.

This probably has some relevance to an idea I talked with @yadudoc in his HTEX MPI prototype where we considered how to support multiple blocks, in #3323. That issue #3323 talks about how the interchange might allocate MPI tasks to different blocks; this current PR might lead in the direction of the scaling code being able to understand how many blocks are needed for the currently submitted workload - so, adding https://github.com/Parsl/parsl/labels/MPI


svandenhaute:
Parsl/WQ seems to hang at the end of workflow execution with this particular version of the code. Any ideas on why this could be the case? Happens for multiple providers and multiple types of apps, and systematically disappears when I switch to 2024.04.29.

benclifford:
since 2024.04.29, there was #3407 which was part of ongoing messing around in Parsl in general with how to do shutdowns - WQ hangs at shutdown if you didn't shut Parsl down.

Since #3165 parsl has not done this automatically at exit, because Python 3.12 doesn't allow that any more, and users (i.e. you) have more responsibility on shutting things down more explicitly - eg by putting your workflow into a `with` block (eg like the README change in #3404) or less preferable by calling `parsl.dfk().cleanup()` at the end of things.

I'd like Work Queue and Task Vine to *not* hang when the process falls off the end without a parsl shutdown, but that's not where those executors are right now.

svandenhaute:
ah right, thanks!

khk-globus (approved):

PR #3418, opened 2024-05-07
Replace SingleNodeLauncher with SimpleLauncher for MPI Mode

WardLT:
# Description

Corrects a problem in the MPI mode implementation of HTEx.
SingleNodeLauncher creates too many workers when running in MPI mode because it
launches multiple copies of the executor onto the head node.

We want SimpleLauncher, which (instead) launches a single copy of the executor.

# Changed Behaviour

Anything that uses the MPI mode will now create the desired number of workers.

# Fixes

No issue, but see discussion in Slack.

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_mpi_apps/test_mpi_mode_disabled.py
+++ b/parsl/tests/test_mpi_apps/test_mpi_mode_disabled.py
@@ -1,4 +1,3 @@
-import logging
 from typing import Dict
 import pytest
 import parsl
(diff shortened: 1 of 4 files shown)


parsl/tests/test_mpi_apps/test_mpi_mode_disabled.py
```diff
@@ -1,4 +1,3 @@
-import logging
```
benclifford, line 1: I have been trying to keep tidyups in separate PRs from functionality changes, with a goal of the changes being on-topic - this stuff is super confusing to debug when it turns out eg. something broke because of a changed import rather than a launcher change.

benclifford: I made this into PR #3436

benclifford (approved):

benclifford:
i changed "executor" to "worker pool" in the merge message - another note for #3426

PR #3420, opened 2024-05-08
Ignore mypy errors and skip tests that don't work on MacOS

yadudoc:
# Description

This PR makes two changes:
1. `os.sched_(get/set)affinity` is not supported on MacOS. This PR ignores mypy errors from code blocks that touch on affinity that are not used in MacOS.
2. Skip tests that touch on affinity that are not expected to work on MacOS.

## Type of change

- Bug fix
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -677,7 +677,8 @@ def worker(
     # If desired, set process affinity
     if cpu_affinity != "none":
         # Count the number of cores per worker
-        avail_cores = sorted(os.sched_getaffinity(0))  # Get the available threads
+        # OSX does not implement os.sched_getaffinity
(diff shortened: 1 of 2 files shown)


benclifford:
this is breaking in CI

benclifford:
@yadudoc i just pushed a possible fix for the most recent mypy error on Linux - I haven't tested if this works on OS X.

yadudoc:
@benclifford This works on OSX. Let's get this merged?

benclifford (approved):

benclifford:
I'm a bit uncomfortable with this test style of "we know this test fails so we ignore that failure" and I've been pushing a bit to make developers explicitly acknowledge that some tests are not being disabled... there is no "skip" in CI, just pass or fail, and this style forces this test to "pass" in the sense of not failing, on platforms without sched_getaffinity.

```
@pytest.mark.skipif('sched_getaffinity' not in dir(os), reason='System does not support sched_setaffinity')
```

PR #3423, opened 2024-05-08
Add an MPIExecutor

WardLT:
# Description

Add MPIExecutor -- a wrapper class over HTEx which fixes or removes options irrelevant when `enable_mpi_mode=True`.

To do list:
- [x] Gather consensus on what to call "max_workers"? I've started with `max_tasks`, but might change that to `max_tasks_per_block`.
- [x] Write a test case ensuring the docstring and init function are not out of date (mechanism for ensuring changes to HTEx don't break MPIEx)
- [x] Write a test case showing MPIEx works as intended
- [x] Update Parsl documentation with new example

Also, whoops, I included a commit from #3418. LMK if I should bother rebasing

# Changed Behaviour

No behavior was changed, only added

# Fixes

N/A

## Type of change

- New feature
- Code maintenance/cleanup

### DIFF
--- a/docs/userguide/mpi_apps.rst
+++ b/docs/userguide/mpi_apps.rst
@@ -15,7 +15,7 @@ The broad strokes of a complete solution involves the following components:
 2. Specify an MPI Launcher from one of the supported launchers ("aprun", "srun", "mpiexec") for the
    :class:`~parsl.executors.high_throughput.executor.HighThroughputExecutor` with: ``mpi_launcher="srun"``
 3. Specify the provider that matches your cluster, (eg. user ``SlurmProvider`` for Slurm clusters)
-4. Set the non-mpi launcher to :class:`~parsl.launchers.SingleNodeLauncher`
+4. Set the non-mpi launcher to :class:`~parsl.launchers.SimpleLauncher`
 5. Specify resources required by the application via ``resource_specification`` as shown below:
 
 
@@ -54,7 +54,7 @@ Configuring the Provider
 
 Parsl must be configured to deploy workers on exactly one node per block. This part is
 simple. Instead of defining a launcher which will place an executor on each node in the
-block, simply use the :class:`~parsl.launchers.SingleNodeLauncher`.
+block, simply use the :class:`~parsl.launchers.SimpleLauncher`.
 The MPI Launcher that the application will use is to be specified via ``HighThroughputExecutor(mpi_launcher="LAUNCHER")``
 
 It is also necessary to specify the desired number of blocks for the executor.
--- a/docs/userguide/mpi_apps.rst
+++ b/docs/userguide/mpi_apps.rst
@@ -1,32 +1,104 @@
-MPI Apps
-========
+MPI and Multi-node Apps
+=======================
+
+The :class:`~parsl.executors.MPIExecutor` support running MPI applications or other computations which can
+run on multiple compute nodes.
+
+Background
+----------
 
 MPI applications run multiple copies of a program that complete a single task by
 coordinating using messages passed within or across nodes.
+
 Starting MPI application requires invoking a "launcher" code (e.g., ``mpiexec``) from one node
 with options that define how the copies of a program should be distributed to others.
-Parsl simplifies this by composing the "launcher" command from the resources specified at the time
-each app is invoked.
 
-The broad strokes of a complete solution involves the following components:
+The launcher includes a series of options that control how copies of the program are distributed
+across the nodes (e.g., how many copies per node) and
+how each copy is configured (e.g., which CPU cores it can use),
+among other options.
 
-1. Configuring the :class:`~parsl.executors.high_throughput.executor.HighThroughputExecutor` with:
-    ``enable_mpi_mode=True``
-2. Specify an MPI Launcher from one of the supported launchers ("aprun", "srun", "mpiexec") for the
-   :class:`~parsl.executors.high_throughput.executor.HighThroughputExecutor` with: ``mpi_launcher="srun"``
-3. Specify the provider that matches your cluster, (eg. user ``SlurmProvider`` for Slurm clusters)
-4. Set the non-mpi launcher to :class:`~parsl.launchers.SimpleLauncher`
-5. Specify resources required by the application via ``resource_specification`` as shown below:
+The options for launchers vary slightly between MPI implementations and compute clusters.
 
+Configuring ``MPIExecutor``
+---------------------------
+
+The :class:`~parsl.executors.MPIExecutor` is a wrapper over
+:class:`~parsl.executors.high_throughput.executor.HighThroughputExecutor`
+which eliminates options that are irrelevant for MPI applications.
+
+Define an
+
+1. Set ``max_workers_per_block`` to maximum number of tasks to run per block of compute nodes.
+   This value is typically the number of nodes per task divided by the number of nodes per task.
+2. Set ``mpi_launcher`` to launcher used for your application.
+3. Specify the provider that matches your cluster and use the :class:`~parsl.launchers.SimpleLauncher`,
+   which will ensure that no Parsl processes are placed on the compute nodes.
+
+An example for ALCF's Polaris supercomputer:
 
 .. code-block:: python
 
-    # Define HighThroughputExecutor(enable_mpi_mode=True, mpi_launcher="mpiexec", ...)
+    config = Config(
+        executors=[
+            MPIExecutor(
+                address=address_by_interface('bond0'),
+                max_workers_per_block=4,  # Up to 4 tasks
+                provider=PBSProProvider(
+                    account="parsl",
+                    worker_init=f"""module load miniconda; source activate /lus/eagle/projects/parsl/env""",
+                    walltime="1:00:00",
+                    queue="debug",
+                    scheduler_options="#PBS -l filesystems=home:eagle:grand",
+                    launcher=SimpleLauncher(),
+                    select_options="ngpus=4",
+                    nodes_per_block=4,
+                    max_blocks=1,
+                    cpus_per_node=64,
+                ),
+            ),
+        ]
+    )
+
+
+Writing an MPI App
+------------------
+
+:class:`~parsl.executors.high_throughput.MPIExecutor` can execute both Python or Bash Apps which invoke an MPI application.
+
+Create the app by first defining a function which includes ``parsl_resource_specification`` keyword argument.
+
+Then, replace the call to the MPI launcher with ``$PARSL_MPI_PREFIX``.
+``$PARSL_MPI_PREFIX`` references an environmental variable which will be replaced with
+the correct MPI launcher configured for the resource list provided when calling the function
+and with options that map the task to nodes which Parsl knows to be available.
+
+The function can be a Bash app
+
+.. code-block:: python
 
     @bash_app
     def lammps_mpi_application(infile: File, parsl_resource_specification: Dict):
         # PARSL_MPI_PREFIX will resolve to `mpiexec -n 4 -ppn 2 -hosts NODE001,NODE002`
         return f"$PARSL_MPI_PREFIX lmp_mpi -in {infile.filepath}"
+
+
+or a Python app:
+
+
+.. code-block:: python
+
+    @bash_app
+    def lammps_mpi_application(infile: File, parsl_resource_specification: Dict):
+        from subprocess import run
+        with open('stdout.lmp', 'w') as fp, open('stderr.lmp', 'w') as fe:
+            proc = run(['$PARSL_MPI_PREFIX', '-i', 'in.lmp'], stdout=fp, stderr=fe)
+            return proc.returncode
+
+
+Run either App by calling with its arguments and a resource specification which defines how to execute it
+
+.. code-block:: python
 
     # Resources in terms of nodes and how ranks are to be distributed are set on a per app
     # basis via the resource_spec dictionary.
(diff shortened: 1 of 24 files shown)


Andrew-S-Rosen:
>  Gather consensus on what to call "max_workers"? I've started with max_tasks, but might change that to max_tasks_per_block.

I'm going to think out loud here. In the "old" way, I usually set `max_workers_per_node=nodes_per_job * nodes_per_block`. So, let's consider the following scenario: 5 Slurm allocations (blocks), each reserving 4 nodes, and I want each task (App) to run via MPI over 2 nodes in the same block. This means I'd usually set `max_workers_per_node` to 2*4=8.

Now let's consider what happens with the same scenario where we call this parameter `max_tasks_per_block=8`. This doesn't seem to check out because I have 4 nodes on a block, but only 2 (not 8) tasks are running on that block. 

Is there a mistake somewhere in my logic?

WardLT:
My understanding is that the `max_workers_per_node` should be `nodes_per_block // nodes_per_app` and not `nodes_per_block * nodes_per_app`.

The weird nomenclature is that, normally, each node of each block gets one executor and each executor gets up to `max_workers_per_node`.

Only one executor gets launched per block in MPI mode (regardless of node count), which means the total number of workers available to launch equals `max_workers_per_node`.

Does that clarify? If so, maybe I rename `max_workers_per_node` to `max_workers_per_block` to avoid introducing another term?

Andrew-S-Rosen:
Thanks @WardLT! On one hand, that makes sense. On the other hand, it's not the behavior I am observing.

With my example (`max_workers_per_node` = nodes_per_app * nodes_per_block), I see each task seemingly running on two 128-core nodes as noted in the OUTCAR:

```
 vasp.6.4.1 05Apr23 (build May 18 2023 13:19:41) complex                 
 executed on    LinuxNV_CrayMPICH date 2024.05.08  20:35:01
 running  256 mpi-ranks, with    1 threads/rank, on    2 nodes
 distrk:  each k-point on  256 cores,    1 groups
 distr:  one band on NCORE=   1 cores,  256 groups
```

However, in your case (`max_workers_per_node` = nodes_per_app // nodes_per_block), I can't even get Parsl to launch tasks. It is complaining in the logs about "more slots than tasks".

That said, I haven't done a resource check on the compute node to be 100% certain about the first case. If you are confident, I will defer to your expertise, but I bring this up just in case I happen to be correct on this.

Andrew-S-Rosen:
Thanks @WardLT! That makes perfect sense. I tested things out, and it looks like I was misled by what was being printed in the VASP OUTCAR. This is even more justification for the need to have a dedicated `MPIExecutor` 😉 

As for the original question about the name, I think adding a `_per` qualifier is worthwhile for clarity's sake. Using `max_tasks_per_block` seems wise then. As a sidenote, I feel like we should be using the phrase "app" instead of "task" since task is such an overloaded phrase in the land of Slurm and friends, but I don't think that's critical.

parsl/executors/high_throughput/mpi_executor.py
```diff
@@ -0,0 +1,105 @@
...
+                 worker_ports: Optional[Tuple[int, int]] = None,
+                 worker_port_range: Optional[Tuple[int, int]] = (54000, 55000),
+                 interchange_port_range: Optional[Tuple[int, int]] = (55000, 56000),
+                 storage_access: Optional[List[Staging]] = None,
+                 working_dir: Optional[str] = None,
+                 worker_debug: bool = False,
+                 max_tasks: int = 1,
+                 prefetch_capacity: int = 0,
+                 heartbeat_threshold: int = 120,
+                 heartbeat_period: int = 30,
```
Andrew-S-Rosen, line 40: Do we need to enumerate all of base class' (`HighThroughputExecutor`) keyword arguments if we aren't modifying the defaults? My concern is if we update the base class, then we might forget to update `MPIExecutor`. It seems like a safer approach might be to only explicitly have new and updated keyword arguments and then pass any remaining `**kwargs` to the base class. Thoughts?

WardLT: Using **kwargs was my first thought too, but I wanted to try enumerating them to give better IDE support.

I made a unit test that will fail of the signature of the base class changes. Does that seem sufficient?

Andrew-S-Rosen: Sounds reasonable to me!

(I really wish IDEs supported better inspection of classes...)

benclifford:
no brainpower for this for the next day or so but will give some more in-depth opinions after that

parsl/executors/high_throughput/mpi_executor.py
```diff
@@ -0,0 +1,105 @@
...
+                 drain_period: Optional[int] = None,
+                 poll_period: int = 10,
+                 address_probe_timeout: Optional[int] = None,
+                 worker_logdir_root: Optional[str] = None,
+                 mpi_launcher: str = "mpiexec",
+                 block_error_handler: Union[bool, Callable[[BlockProviderExecutor, Dict[str, JobStatus]], None]] = True,
+                 encrypted: bool = False):
+        super().__init__(
+            # Hard-coded settings
+            cores_per_worker=1e-9,  # Ensures there will always be enough workers
```
benclifford, line 50: This is not *always* - there's a number of workers where this will no longer work - because this is not 1/∞, just an approximation of that.

parsl/executors/high_throughput/mpi_executor.py
```diff
@@ -0,0 +1,105 @@
...
+            address_probe_timeout=address_probe_timeout,
+            worker_logdir_root=worker_logdir_root,
+            mpi_launcher=mpi_launcher,
+            block_error_handler=block_error_handler,
+            encrypted=encrypted
+        )
+
+
+# Update the docstring on import
+def _update_from_htex_docstring():
```
benclifford, line 79: :exploding_head:

benclifford:
for the above max_workers_per_* discussion, more generally I'd like to push the internals of htex to talk about  "per_pool" when it's relevant, although that doesn't necessarily need to be reflected in the user facing parameters that this PR is dealing with. but as the above discussion shows "per node" and "per block" are internally not going to be the right words to use.

parsl/executors/high_throughput/mpi_executor.py
WardLT: Do you hate this? I felt a little foolish writing it, but don't think any Python documentation tools have a "get parameter definitions from the superclass" function

parsl/executors/high_throughput/mpi_executor.py
tomdemeyere: In any case, isn't the edge case 1/max_tasks here? Since we know that there will never be more than max_tasks workers? It might be cleaner that way?

benclifford: (the original assumption for this parameter is "of course a task needs cores on the manager node to execute" but that's not how htex-mpi does things)

WardLT:
I added a first draft of the MPI Executor draft. @Andrew-S-Rosen and @tomdemeyere , you are a good target audience. How does it seem to you? https://parsl--3423.org.readthedocs.build/en/3423/

tomdemeyere:
Nice! I am excited to try it with the modified MPI mode in https://github.com/Parsl/parsl/pull/3418 but even without it, this seems much simpler and accessible.

I see everything that I would need to launch MPI apps, would this work with max_block > 1 ?

benclifford:
@tomdemeyere there's not enough awareness of tasks having different resource requirements (number of ranks for each task) in the interchange, which dispatched tasks to (in the htex-mpi case) different blocks. So you might end up with some tasks queued in one block due to insufficient resources while another block has capacity.

That awareness was one of the targets of this project which was ultimately not funded: #3323

docs/userguide/mpi_apps.rst
```diff
@@ -1,335 +1,132 @@
...
-                future = cosmic_tagger(workdir="/home/yadunand/CosmicTagger",
-                                       datatype="float32",
-                                       batchsize=str(batchsize),
-                                       parsl_resource_specification=parsl_res_spec)
-
-
-                print(f"Stdout : {future.stdout}")
-                print(f"Stderr : {future.stderr}")
-                futures[(num_nodes, batchsize)] = future
+    # Resources in terms of nodes and how ranks are to be distributed are set on a per app
```
Andrew-S-Rosen, line 103: Do we _always_ need a `parsel_ressource_specification`? I'm just thinking back to the approach used in your older videos where we never used such a thing with the `HighThroughputExecutor`. If it's omitted, I assume the implicit assumption is that all tasks have the same resource requirements but it will still function?

docs/userguide/mpi_apps.rst
```diff
@@ -1,335 +1,132 @@
...
+    # basis via the resource_spec dictionary.
+    resource_spec = {
+        "num_nodes" = 2,
+        "ranks_per_node" = 2,
+        "num_ranks" = 4,
```
Andrew-S-Rosen, line 108: Presumably these should be `:` instead of `=`.

docs/userguide/mpi_apps.rst
```diff
@@ -1,335 +1,132 @@
...
+    }
+    future = lammps_mpi_application(File('in.file'), parsl_resource_specification=resource_spec)
 
+Advanced: More Environment Variables
```
Andrew-S-Rosen, line 112: Very selfish question:

I am usually running VASP (via ASE) on the GPU nodes of Perlmutter. I assume one is still using a `SimpleLauncher()` here since it's an "internal" call to `srun`?

For context, a Slurm script usually looks something like the following:

```python
#!/bin/bash
#SBATCH -N 1
#SBATCH -C gpu
#SBATCH -G 4
#SBATCH -q debug
#SBATCH -J vasp_job
#SBATCH -t 00:30:00
#SBATCH -A matgen_g

module load vasp/6.4.1-gpu
export OMP_NUM_THREADS=8
export OMP_PLACES=threads
export OMP_PROC_BIND=spread
--gpu-bind=single:1
export ASE_VASP_COMMAND="srun -n 4 -c 32 --cpu_bind=cores -G 4 --gpu-bind=none vasp_std"

python run_my_jobs.py
```

where `run_my_jobs.py` is some arbitrary program that in it launches a bunch of `@python_app`s representing one VASP task per App.

Andrew-S-Rosen (commented):
@WardLT: Awesome!! This is looking great. Some minor corrections below :)

docs/userguide/mpi_apps.rst
benclifford: There has to be some default made in that case. In the workqueue case, for example, ommitting a core count (or memory specification) for a task will allocate one entire worker node (whatever the relevant worker node happens to have on it).

That is one obvious default. The other is that you get 1 rank.

WardLT: Yes, you do always need one and - at the moment (see #3427) - Parsl will crash without it being provided as a keyword argument.

docs/userguide/mpi_apps.rst
WardLT: I think this application is possible by replacing the parts of the VASP command definition which deal with the total ranks to `\$PARSL_MPI_PREFIX`. Escaping the `$` will (hopefully!) result in the environment variable not being resolved with `ASE_VASP_COMMAND` is defined but instead when the contents `ASE_VASP_COMMAND` is used to start a subprocess somewhere deep in the bowels of ASE.

```bash
export ASE_VASP_COMMAND="\$PARSL_MPI_PREFIX -c 32 --cpu_bind=cores -G 4 --gpu-bind=none vasp_std"
```

That said, I'm not sure if `-G` is the "total" number of GPUs or the number per node. You might to switch to `--gpus-per-node` instead of `-G`.

WardLT:
Alright. I've addressed Andrew's comments and fixed a bunch of other grammar issues. 

I think this is ready for review.

parsl/tests/test_mpi_apps/test_mpiex.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+        ]
+    )
+    parsl.load(config)
+
+
+def local_teardown():
+    parsl.dfk().cleanup()
+    parsl.clear()
+
```
benclifford, line 36: You can use `local_config()` for the above, instead of `local_setup` and `local_teardown`: return `config` from `local_config()` without calling `parsl.load` yourself, and conftest.py will handle the loading, unloading, and will also handle some extra stuff like directing your run files into `.pytest` - pretty much `local_setup` and `local_teardown` are only needed for tests that *aren't* loading following the: load DFK... test some stuff... shutdown DFK pattern.

parsl/tests/test_mpi_apps/test_mpiex.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.mark.local
+def test_get_env():
+    future = get_env_vars(parsl_resource_specification={
+        "num_nodes": 2,
+        "ranks_per_node": 2,
+    })
+    env_vars = future.result()
+    assert env_vars['PARSL_NUM_RANKS'] == '4'
```
benclifford, line 70: the mathematician in me would use two different prime numbers here, not both 2, to get some difference in the values of the parameters (eg so that you're additionally testing that your aren't being allocated num_nodes * num_nodes ranks). then CS people can be forever confused why you're using not using a power of two...

docs/userguide/mpi_apps.rst
```diff
@@ -1,173 +1,72 @@
...
-                        nodes_per_block=user_opts["nodes_per_block"],
-                        init_blocks=1,
-                        min_blocks=0,
-                        max_blocks=1, # Can increase more to have more parallel jobs
-                        walltime=user_opts["walltime"]
-                    ),
+        executors=[
+            MPIExecutor(
+                address=address_by_interface('bond0'),
+                max_workers_per_block=2,  # Assuming 2 nodes per task
```
benclifford, line 45: here's that use of 2 twice for different meanings I mentioend elsewhere!

docs/userguide/mpi_apps.rst
Andrew-S-Rosen: Makes sense. This is not your problem to deal with so to speak, but this just makes it even more awkward that `parsl_resource_specification` is not a keyword argument of the decorator itself. Having to redefine all of your functions to take this (hidden) keyword argument just to parallelize things is a bit hacky and counter-intuitive.

parsl/executors/high_throughput/mpi_executor.py
benclifford: I really dislike this and have been trying to think of nicer ways to do this.

One way I came up with is to stop trying to regard MPIExecutor as "real HTEX but different", which leads to this _update_from_htex_docstring, and instead view htex as existing in two equal first order modes: mpi and not-mpi.

Then theres a shared docstring, which you could define (in HighThroughputExecutor) as  _shared_docstring= """
...
"""

and then the HTEX docstring is  `__doc__ = """whatever non-mpi htex intro""" + _shared_docstring
and the MPI docstring is "__doc__ = """whatever mpi htex intro""" + _shared_docstring

That removes any parsing from this function, by making the docstrings already separated in source into three chapters, to be combined using `+`.

WardLT: Good idea! I implemented something like that, after seeing how to set docstrings programmatically https://stackoverflow.com/questions/70863543/can-a-python-docstring-be-calculated-f-string-or-expression

parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -208,6 +208,8 @@ def get_result(self, block: bool, timeout: float):
         """Return result and relinquish provisioned nodes"""
         result_pkl = self.pending_result_q.get(block, timeout=timeout)
         result_dict = pickle.loads(result_pkl)
+        # TODO (wardlt): If the task did not request nodes, it won't be in `self._map_tasks_to_nodes`. Causes Parsl to hang
+        #  See Issue #3427
```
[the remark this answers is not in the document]
WardLT: Yea, I should have done that. Was leery to touch code, but it's already broken 🤷🏼

parsl/tests/test_mpi_apps/test_mpiex.py
WardLT: Done. I also added a little to the CONTRIBUTING.rst document describing this feature

benclifford (approved):

parsl/executors/high_throughput/mpi_executor.py
```diff
@@ -0,0 +1,85 @@
...
+class MPIExecutor(HighThroughputExecutor):
+    __doc__ = f"""A version of :class:`~parsl.HighThroughputExecutor` tuned for executing multi-node (e.g., MPI) tasks.
+
+    The Provider _must_ use the :class:`~parsl.launchers.SimpleLauncher`,
+    which places a single pool of workers on the first node of a block.
+    Each worker can then make system calls which use an MPI launcher (e.g., ``mpirun``, ``srun``)
+    to spawn multi-node tasks.
+
+    Specify the maximum number of multi-node tasks to run at once using ``max_workers_per_block``.
+    The maximum number should be smaller than the ``nodes_per_block`` in the Provider.
```
Andrew-S-Rosen, line 23: @WardLT: Just looking at this again. Should it be "The value should be less than _or equal to_ the ``nodes_per_block``?

WardLT: Yes, it should be less than or equal to

PR #3428, opened 2024-05-13
Use concurrency in the README

WardLT:
# Description

It bothers me that there is no parallelism in the README.

# Changed Behaviour

N/A

# Fixes

N/A

## Type of change

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,17 +22,19 @@ Parsl lets you chain functions together and will launch each function as inputs 
         return x + 1
 
     @python_app
-    def g(x):
-        return x * 2
+    def g(x, y):
+        return x + y
 
     # Start Parsl on a single computer
     with parsl.load():
-        # These functions now return Futures, and can be chained
+        # These functions now return Futures
         future = f(1)
         assert future.result() == 2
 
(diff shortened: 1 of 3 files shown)


benclifford:
This failed because of a possibly unrelated possibly one off failure - it would be good if you could grab that whole failure into a separate issue to track down later before rerunning these tests.

benclifford:
The background I come from this maybe would be phrased more as concurrent rather than parallel (which in my head comes more with the implication of closely synchronized MPI style code)

WardLT:
The failure has been happening since Friday afternoon in all my branches. I think it's stemming from a bug in RADICAL Pilot

WardLT:
I do like "concurrent" better, and that's what Python calls their `concurrent.futures` module anyway.

benclifford:
> The failure has been happening since Friday afternoon in all my branches. I think it's stemming from a bug in RADICAL Pilot

Probably even more reason to open an issue, then

benclifford:
see https://github.com/Parsl/parsl/issues/3429 for broken CI issue

benclifford (approved):

PR #3432, opened 2024-05-15
Factor radio selection in monitoring worker-side code

benclifford:
# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/remote.py
+++ b/parsl/monitoring/remote.py
@@ -96,6 +96,22 @@ def monitor_wrapper(*,
     return (wrapped, args, new_kwargs)
 
 
+def get_radio(radio_mode: str, monitoring_hub_url: str, task_id: int, run_dir: str) -> MonitoringRadio:
+    radio: MonitoringRadio
+    if radio_mode == "udp":
+        radio = UDPRadio(monitoring_hub_url,
+                         source_id=task_id)
+    elif radio_mode == "htex":
+        radio = HTEXRadio(monitoring_hub_url,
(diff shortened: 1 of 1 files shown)


parsl/monitoring/remote.py
```diff
@@ -96,6 +96,22 @@ def wrapped(*args: List[Any], **kwargs: Dict[str, Any]) -> Any:
...
+    radio: MonitoringRadio
+    if radio_mode == "udp":
+        radio = UDPRadio(monitoring_hub_url,
+                         source_id=task_id)
+    elif radio_mode == "htex":
+        radio = HTEXRadio(monitoring_hub_url,
+                          source_id=task_id)
+    elif radio_mode == "filesystem":
+        radio = FilesystemRadio(monitoring_url=monitoring_hub_url,
+                                source_id=task_id, run_dir=run_dir)
```
khk-globus, line 109: This is not new code, so just an observation that the non-congruence of the first two args vs the `filesystem` path feels weird.

Otherwise, nice DRY.

khk-globus (approved):

parsl/monitoring/remote.py
benclifford: yeah, that's part of the motivation behind #3315  - to move to a more object-oriented configuration which both identifies the desired radio and carries along whatever the necessary config is for that radio in a single object... so that his above call becomes something like asking the relevant config object "give me a suitably configured radio, whatever that means"

PR #3433, opened 2024-05-15
Remove double init of some HTEX/monitoring attributes

benclifford:
These have been initialised in the ParslExecutor base class since #3361

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
:+1:

PR #3440, opened 2024-05-17
Remove $ when writing job name

### LINKED ISSUE #3437: Parsl creates hidden log files, doesn't clean them up
I didn't know Parsl was creating so many log files because they were hidden. 
I'm OK with files appearing in my run directory, but not accumulating them endlessly without my notice.

To be specific, I don't know what is responsible for `.workers.block-0.17158*` files but I'd like them to be visible.

No hidden files.

Adding a "clean up Parsl's logging files" CLI tool

**Additional context**
Is anyone else concerned?

WardLT:
# Description

Fixes #3437, a problem which was created by me in the first place.

# Changed Behaviour

Does not remove the `parsl.` prefix from the hostfile written when using localprovider.

# Fixes

Fixes #3437 

## Type of change

- Bug fix

### DIFF
--- a/parsl/providers/local/local.py
+++ b/parsl/providers/local/local.py
@@ -206,7 +206,7 @@ class LocalProvider(ExecutionProvider, RepresentationMixin):
         script_path = "{0}/{1}.sh".format(self.script_dir, job_name)
         script_path = os.path.abspath(script_path)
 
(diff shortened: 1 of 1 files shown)


benclifford:
this is a problem with all the providers, right?

WardLT:
No, the other providers use the `${jobname}` syntax correctly. They use it to describe a string Template, which means the `$` is removed when substituting the variable name.

LocalProvider uses f-strings to replace `jobname`, which means the `$` stays.

benclifford (approved):

Issue #3376, opened 2024-04-20
htex command channel use is not thread safe


benclifford:
**Describe the bug**

The interchange command channel is presented on the submit side as an object `executor.command_client` that is invoked from any thread. However, it uses a ZMQ channel which is not thread safe. When two threads hit this too close together in time, this causes weird behavior in ZMQ: at least sometimes a segfault style crash, and I suspect possibly also some hangs.

The command channel accesses should be made less thread safe. I don't have a favoured solution. Some ideas that I am not super fond of but I guess would be ok:
i) make a new ZMQ connection (in the opposite TCP-direction to the current command channel) for every thread that makes a command client invocation. 
ii) use an internal cross-thread RPC mechanism and gateway those RPCs in the thread that owns the ZMQ connection.

In `desc` branch parsl, I've put a lot of locks around ZMQ command channel accesses, but I don't believe this is sufficient to guarantee thread-safety given ZMQ's discussion of eg. memory barriers

**To Reproduce**
it's a race condition

**Expected behavior**
this thread unsafety should not happen

benclifford:
Another observation I made, that I think is relevant to this issue:


```
The interchange command client contains this:
                except zmq.ZMQError:
                    logger.exception("Potential ZMQ REQ-REP deadlock caught")
                    logger.info("Trying to reestablish context")
                    self.zmq_context.recreate()
                    self.create_socket_and_bind()
The only reason I can figure out for this is that the command client is used in a non-thread-safe manner and that multiple commands can be invoked from multiple threads without receiving responses - which is non-thread-safe in two ways: 1) you can't send two REQs on the same ZMQ socket in a row, without waiting for the first REP; and 2) you can't use the same ZMQ socket across multiple threads even if you sequence your operations properly.

#3376 talks about non-threadsafe use of the command client, and I guess I hope that if #3376 is fixed, the above can go away.
```

https://github.com/Parsl/parsl/issues/3376 talks about non-threadsafe use of the command client, and

benclifford:
Further to that last comment, I'm unclear in which situations that re-establish code will work. I think it makes the following assumptions:
i) the previous TCP-level listening socket has been closed due to this ZMQError - because otherwise, the rebind in this `except` block will not be able to rebind
[...]

Issue #3373, opened 2024-04-20
Interchange: launch as a fork/exec not Python multiprocessing


benclifford:
**Is your feature request related to a problem? Please describe.**
The interchange is launched using Python `multiprocessing` but mostly wants to be its own freestanding process: for example, it awkwardly works around inheriting the main process logger by avoiding the "parsl" logging topic; and it communicates with the main process mostly via ZMQ (with only a small bit of multiprocessing-based communication near the start).

Multiprocessing-fork is a Bad Thing (see issue #2343) so launching the interchange this way should go away.

Launching as a fresh process would also allow more normal logging (for example, logging everything to the interchange log file rather than carefully avoiding what was inherited).
 
The only use of multiprocessing queues, which is to communicate the ZMQ port numbers to be used by workers, could be transmitted in a different way: for example add a `WORKER_PORTS` command to the command channel.

**Describe the solution you'd like**
Use normal fork/exec so this behaves like a freestanding process

benclifford:
> Launching as a fresh process would also allow more normal logging (for example, logging everything to the interchange log file rather than carefully avoiding what was inherited).

regarding this point, I think I have encountered a situation (although it's quite awkward to trace) where some changes I would like to contribute to Parsl make use of regular Parsl code that logs to a parsl.* logging channel, and then hangs - I think (although I am not completely convinced) that this is because of a fork-but-no-exec lock held over the fork for some part of the regular parsl logging setup.

Implementing this issue would remove that possibility.

benclifford:
in the immediately preceeding comment, the hang looks like this:

```
(gdb) bt
#0  __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x6473d2908b30)
    at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (futex_word=futex_word@entry=0x6473d2908b30, expected=expected@entry=0, clockid=clockid@entry=0, 
    abstime=abstime@entry=0x0, private=<optimized out>, cancel=cancel@entry=true) at ./nptl/futex-internal.c:87
#2  0x00007b717460befb in __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x6473d2908b30, expected=expected@entry=0, 
    clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=<optimized out>) at ./nptl/futex-internal.c:139
[...]

PR #3441, opened 2024-05-17
Add a timeout option to htex command client

benclifford:
This is in preparation in an upcoming PR for replacing the current multiprocessing.Queue based report of interchange ports (which has a 120s timeout) with a command client based retrieval of that information (so now the command client needs to implement a 120s timeout to closely replicate that behaviour). That work is itself part of using fork/exec to launch the interchange, rather than multiprocessing.fork (issue #3373)

When the command client timeouts after sending a command, then it sets itself to permanently bad: this is because the state of the channel is now unknown. eg. Should the next action be to receive a response from a previously timed out command that was eventually executed? Should the channel be recreated assuming a previously sent command was never sent?

Tagging issue #3376 (command client is not thread safe) because I feel like reworking this timeout behaviour and reworking that thread safety might be a single piece of deeper work.

# Changed Behaviour

Timeout functionality is unused at the moment, so this shouldn't change things.

## Type of change

- New feature


benclifford:
(needs a bit more testing with the no-timeout case, wups)

benclifford:
tagging @rjmello for review because he most recently touched anything to do with zmq

parsl/tests/test_htex/test_command_client_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+
+    req = ic_channel.recv_pyobj()
+    assert req == "SLOW_COMMAND", "Should have received command on interchange side"
+    assert not cc.ok, "CommandClient should have set itself to bad"
+
+    with pytest.raises(CommandClientBadError):
+        cc.run("ANOTHER_COMMAND")
+
+    cc.close()
+    ctx.term()
```
rjmello, line 66: `cc.close()` should take care of `ctx.term()`.

parsl/tests/test_htex/test_command_client_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+    ctx = curvezmq.ClientContext(None)
+
+    # RFC6335 ephemeral port range
+    cc = CommandClient(ctx, "127.0.0.1", (49152, 65535))
+
+    # cc will now wait for a connection, but we won't do anything to make the
+    # other side of the connection exist, so any command given to cc should
+    # timeout.
+
+    with pytest.raises(CommandClientTimeoutError):
```
rjmello, line 30: We may want to include a test for the `InternalConsistencyError` path as well.

rjmello (changes requested):

parsl/tests/test_htex/test_command_client_timeout.py
benclifford: It's meant to be "impossible" in the sense of an assertion failure

parsl/tests/test_htex/test_command_client_timeout.py
benclifford: ok, I see it does that now. Although that's a bit weirdly asymmetric if you ever imagine sharing one context amongst multiple channels...

rjmello: Good point; maybe leave your code as-is then follow up with a PR to remove the automatic `ctx.term()` call. I can take care of the latter.

parsl/tests/test_htex/test_command_client_timeout.py
rjmello: Makes sense; your call.

parsl/tests/test_htex/test_command_client_timeout.py
benclifford: I noted this potential term change in issue #3457

rjmello (approved):
🎸

PR #3442, opened 2024-05-20
Make HTEX reject empty resource specification in MPI mode

### LINKED ISSUE #3427: Parsl hangs when using MPI Mode HTEx if resource specs are not found
**Describe the bug**
I found, accidentally, that Parsl will hang if a task is sent to HTEx running in "MPI Mode" if you fail to provide a `parsl_resource_dict` as an explicitly-named keyword argument.

It appears that the the resource requirements are not sent along with the task if you provide the resource specification as a keyword argument. 
The lack of task records means that the task is not registered in the `tasks_to_nodes` map in `MPIScheduler` and therefore not found when the last completes.

```
2024-05-13 12:07:17.349 parsl:409 8518 Result-Pusher [ERROR]  Got an exception: 1
  File "/home/lward/Work/ExaLearn/parsl/parsl/executors/high_throughput/process_worker_pool.py", line 403, in push_results
    r = self.task_scheduler.get_result(block=True, timeout=push_poll_period)
  File "/home/lward/Work/ExaLearn/parsl/parsl/executors/high_throughput/mpi_resource_management.py", line 213, in get_result
    nodes_to_reallocate = self._map_tasks_to_nodes[task_id]
KeyError: 1
```

**To Reproduce**

Remove the `parsl_resource_requirements` from the input arguments to an MPI app in any of the tests.

**Expected behavior**

Parsl at least fails with an error message, at least. It might be good for Parsl to substitute some default values (1 rank per node, 1 node) with a warning so that tasks still run.

**Environment**
 - OS: Ubuntu
 - Python version: 
 - Parsl version: `master` as of today.

**Distributed Environment**
- Where are you running the Parsl script from ? Desktop
- Where do you need the workers to run ? Same as parsl

R1j1t:
# Description

Added a check in `HighThroughputExecutor.submit` in cases where `parsl_resource_specification` is empty when `enable_mpi_mode=True`

Possible changes:
- [ ] ~update user doc~ ([user doc](https://parsl.readthedocs.io/en/stable/userguide/mpi_apps.html#writing-an-mpi-app) does mention the need to configure `parsl_resource_specification `
- [x] Update exception `InvalidResourceSpecification` as it currently expects a set
- [ ] fix test case?
- [ ] ~Refactor `test_mpi_mode_enabled.py`? Many function arguments are never used.~

# Changed Behaviour

Parsl fails with an error message, instead of hanging on the user

# Fixes

Fixes #3427 

## Type of change

- Bug fix


R1j1t:
Also, happy to hear feedback on PR!

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -18,13 +18,19 @@ def __str__(self):
         return f"Invalid resource specification options supplied: {self.invalid_keys}"
 
 
-def validate_resource_spec(resource_spec: Dict[str, str]):
+def validate_resource_spec(resource_spec: Dict[str, str], is_mpi_enabled: bool = False):
```
benclifford, line 21: I prefer to not specify defaults like `= False` here unless there is a reason (for example, for better usability if the function is exposed to end users). Here I think it's more important to be clear on every use of validate_resource_spec whether it is validating for mpi mode or non-mpi mode, and removing this default will ensure that happens.

parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -18,13 +18,19 @@ def __str__(self):
...
 
     Raises: InvalidResourceSpecification if the resource_spec
         is invalid (e.g, contains invalid keys)
     """
     user_keys = set(resource_spec.keys())
+
+    # empty resource_spec when mpi_mode is set causes parsl to hang
+    # ref issue #3427
+    if is_mpi_enabled and len(user_keys) == 0:
+        raise InvalidResourceSpecification({'mpi_mode requires parsl_resource_specification to be configured'})
```
benclifford, line 32: InvalidResourceSpecification takes one parameter, which is a set of invalid resource keys. The parameter on this line is not a set of invalid resource keys!

benclifford (changes requested):
see comments on specific lines

benclifford:
ok, reviewing this again, I feel like you should take a step back and review what `validate_resource_spec` is meant to be doing - rather than what it actually does.

For both MPI mode and not-MPI mode, there are some keys that are required, some that are optional and some that are forbidden - different for each mode.

The current validation code in `master` doesn't distinguish between those two modes - and in this PR you've added in that distinction, which is probably a good thing.

And the current validation code doesn't look for which keys/combinations are required at all. Your current PR doesn't look to see which keys are required either, it just requires that *something* is specified.

So maybe it would be good, perhaps with help from @WardLT and @yadudoc , to figure out and write down for both modes which keys are required, optional and forbidden - based on understanding what the code does with those keys. And then rewrite `validate_resource_spec` to validate those rules.

R1j1t:
Thanks @benclifford for the comment and it makes perfect sense to check for required parameters be passed. Let me dig in the code for the required parameters as well. I will confirm my finding with @WardLT and @yadudoc!

benclifford:
another question i have is in mpi mode, there are three interacting parameters and this validate function does more than validate - it computes new values. but what happens in a user specifies all three values? (ranks per node, num nodes and num ranks?) - for example if a user specifies 2 ranks per node, 500 ranks, 3 nodes? I feel like there is something bad here that should be validated. I've also raised that question with @rjmello offline who is doing some other validation work.

WardLT (approved):
Thanks, @R1j1t ! I'm inclined to approve this PR as it takes a good step in the direction of making resource spec validation errors clearer for users.

Building on @benclifford , I think we can do better by deviating farther from the initial implementation. For example, we could use a data class to handle the validation logic rather than duplicate logic dataclasses already handles well. We should also move the checks from buried within HTEx to part of the app module to encourage use of the same spec across different Executors.

Those ideas and the ones Ben already has feel like things we should spread over multiple PRs. Do you agree?

PS. I'm about to leave on vacation. Reviews are going to be slower over the next few weeks.

.gitignore
```diff
@@ -120,3 +122,7 @@ ENV/
 
 # emacs buffers
 \#*
+
+#deps
+/.deps
+
```
benclifford, line 128: don't change .gitignore with changes that are not relevant to the topic of the PR: none of these changes seem relevant to the point of the PR.

R1j1t: I have removed any changes in this file. The reason I added them was because of the following files getting created on every test execution
image

Let me open another PR for this!

R1j1t:
> For both MPI mode and not-MPI mode, there are some keys that are required, some that are optional and some that are forbidden - different for each mode.


I am not sure if I am missing something, but based on the below code `num_ranks` and `ranks_per_node` should be present post `validate_resource_spec`.  https://github.com/Parsl/parsl/blob/146a6afd3ef3606ff5801527d1eee5487b080f47/parsl/executors/high_throughput/mpi_prefix_composer.py#L72-L75 
cc: @yadudoc

benclifford (approved):

.gitignore
benclifford: ok - you can also fix the MPI test perhaps. It looks like: parsl/tests/test_mpi_apps/test_mpi_mode_enabled.py uses `local_setup` and `local_teardown` when it could use `local_config` instead to supply the test configuration. If the test used `local_config`, then the test system should make those files appear under `.pytest/parsltest-current` like it does for many other tests.

R1j1t:
Thanks @benclifford, @WardLT for the help! Looking forward to contributing to parsl :)

PR #3443, opened 2024-05-20
Move CoC file to .github folder to transform into a community health file

sophie-bui:
# Description

Moving this CoC file to the .github folder transforms it into a community health file, which allows us to use it in all of our repos without needing to create new CoC files in separate repos. This also enables us to maintain this file in a central location.

[More details on Community health files
](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file#supported-file-types)

# Changed Behaviour

The only thing we need to do is remember that it lives in the .github folder and to make updates there.

# Fixes

N/A

## Type of change

- Code maintenance/cleanup


benclifford:
I read through to this link https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file and my understanding from that is that this needs to go in a new repository called `.github`, not the (same-named) `.github` folder in the parsl/parsl repository, to be treated as a default for all repositories?

benclifford:
(see step 3 of https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file#creating-a-repository-for-default-files )

sophie-bui:
Ah, okay. I thought it could be in the .github folder for our main repository. But after rereading, I believe you are correct. I was trying to find a way to feature the CoC in our playbook without having multiple copies of it. Do you think this is a good solution for that?

benclifford:
I think the CoC is project scoped, not per repo (for example: there probably shouldn't be a different Code of Conduct for the website vs the python code...) and so I think moving the code of conduct into a .github repo to get project-wide defaults is the right thing to do.

benclifford:
@sophie-bui alternatively, pick some other place, like still in the root of the parsl/parsl repo - and link there from the playbook. i think you did it like that for the contributing guide, right?

sophie-bui:
Thanks for your advice, Ben! I'll think about this a bit before implementing anything. I'll go ahead and close this PR.

PR #3444, opened 2024-05-21
Do parsl.clear() at DFK cleanup when relevant

benclifford:
It was UI wart that parsl.clear() is necessary for a user to do, arising from a long-ago distinction between creating a DFK (of which potentially a user might make many) and asking some other piece of Parsl to manage a single default DFK (parsl.load). It has been an ongoing usability problem for users to understand why they need to do both parsl.dfk().cleanup() and parsl.clear().

This PR makes a lot of changes throughout the test suite to remove parsl.clear statements that are unnecessary after this PR - tested in development by a removable assert.

# Changed Behaviour

parsl.clear() will no longer be necessary - although it will be harmless to continue to do so.

When invoking an app after the default DFK has been shut down, the error message will now be that a config has not been loaded (`must first load config`) rather than `Cannot submit to a DFK that has been cleaned up` - because the cleaned-up default DFK will have been unset, rather than left as default.

## Type of change

- New feature


parsl/tests/test_htex/test_cpu_affinity_explicit.py
```diff
@@ -37,16 +37,9 @@ def test_cpu_affinity_explicit():
     config.executors[0].max_workers_per_node = 1
 
     logger.debug(f"config: {config}")
-    # TODO: is there a `with` style for this, to properly deal with exceptions?
-
-    parsl.load(config)
-    try:
 
+    with parsl.load(config):
```
khk-globus, line 41: :boom:  Heh.  "We'll get there.  One day."

parsl/dataflow/dflow.py
```diff
@@ -1271,6 +1271,11 @@ def cleanup(self) -> None:
         atexit.unregister(self.atexit_cleanup)
         logger.info("Unregistered atexit hook")
 
+        if DataFlowKernelLoader._dfk == self:
+            logger.info("Unregistering default DFK")
+            parsl.clear()
+            logger.info("Unregistered default DFK")
+
```
khk-globus, line 1278: If `==` check makes me wonder if there's a `!=` path to be considered?  Is it _possible_?  Would some kind of (at least a) log be warranted in the else-case?

khk-globus (approved):

parsl/dataflow/dflow.py
benclifford: that path can exist, in the somewhat rare situation that you make a DFK by instantiating `DataFlowKernel` yourself, rather than using `parsl.load` -- in that case, the DataFlowKernelLoader (which manages this default DFK that comes from calling `parsl.load`) isn't aware of that DFK, and so it doesn't need to be told that it no longer needs to know about this DFK.

If you're doing this, you aren't going to get any of the defaulting behaviour (such as app invocations being sent automatically to that DFK) and you especially aren't going to hit the problem of not being able to get another DFK without clearing this one.

it's an artefact of history that users can make DFKs that way and I don't think I've really seen anyone do it since singleton-style loading was introduced in 2018 in da5935213490416d33af2ee77adb624db9f818e0

benclifford: (might be worth a debug message saying this isn't happening, but I don't think anything more)

benclifford: this test should be `is` not `==` though I think -- it's checking if the DFK objects are the same objects. (although right now, DFK `==` will behave the same as `is` because it is not overridden)

PR #3445, opened 2024-05-21
Extend dependency error change in #2989 to remaining two dependency types

benclifford:
# Description

PR #2989 made dependency errors look nicer, but only did it for positional 
parameters. This PR extends that to kwarg and inputs=... dependencies.
    
PR #2989 describes these changes in more depth.
    
This PR adds a type annotation onto DFK._unwrap_futures that drives this
change - preventing _unwrap_futures from returning None task ids for failed
dependencies and instead requiring the task IDs to be strings.

## Type of change

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -870,7 +870,8 @@ class DataFlowKernel:
 
         return depends
 
-    def _unwrap_futures(self, args, kwargs):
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Subtle ... I'm going to call it a fix.  Thanks DRY.

PR #3455, opened 2024-05-22
Clarify docstring for `MPIExecutor`

Andrew-S-Rosen:
As noted in https://github.com/Parsl/parsl/pull/3423#discussion_r1610611500, the docstring for `MPIExecutor` said `max_workers_per_block` should be less than `nodes_per_block` when in reality it should be less than or equal to.

CC @WardLT.


WardLT (approved):
Thanks!

benclifford (approved):

PR #3463, opened 2024-05-28
Launch interchange as a fresh process

benclifford:
This PR removes a use of multiprocessing fork-without-exec.

At heart, this is how the interchange has wanted to be launched for some time (because of earlier remote interchange work).

Launching multiprocessing fork caused a bunch of problems related to inheriteing state from from the parent submitting process that go away with this (jumbled logging topics, race conditions around at least logging-while-forking, inherited signal handlers).

The configuration dictionary, previously passed in memory over a fork, is now sent in pickled form over `stdin`. Using pickle here rather than (eg.) JSON keeps the path open for sending richer configuration objects, beyond what can be encoded in JSON. This isn't something needed right now, but at least configurable monitoring radios (the immediate driving force behind this PR) are modelled around passing arbitrary configuration objects around to configure things - and so it seems likely that if interchange monitoring configuration is exposed to the user, richer objects would be passed here. See PR #3315 for monitoring radio prototype.

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -673,17 +673,6 @@ def start_file_logger(filename: str, level: int = logging.DEBUG, format_string: 
     logger.addHandler(handler)
 
 
-# @wrap_with_logs(target="interchange")
-# def starter(*args: Any, **kwargs: Any) -> None:
-#    """Start the interchange process
-
-#    The executor is expected to call this function. The args, kwargs match that of the Interchange.__init__
-#    """
-#    setproctitle("parsl: HTEX interchange")
-#    # logger = multiprocessing.get_logger()
-#    ic = Interchange(*args, **kwargs)
-#    ic.start()
-
 if __name__ == "__main__":
     setproctitle("parsl: HTEX interchange")
 
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -743,9 +743,6 @@ if __name__ == "__main__":
     # TODO: all these ad-hoc parsers are pretty horrible - using a command line
     # is a bit of a horrible way to do this...
 
-    # TODO: initialize logging here and log any exceptions raised during
-    # these two lines into the log file:
-
     ic = Interchange(client_address=args.client_address,
                      interchange_address=parseNone(args.interchange_address),
                      client_ports=parseInt3(args.client_ports),
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-import argparse
+import base64
 import datetime
 import json
 import logging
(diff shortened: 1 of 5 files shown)


benclifford:
commit e7d18aa4725310534e0ec4aa1ce5601d01ec2ecc swaps this PR from using `argparse` to passing a base64 pickled config dict on the command line. that removes a lot of code around "we have to define a schema here", at the expense of a much more opaque interchange command line. (although prior to this PR, there was no visible interchange command line at all, in that sense: the same configuration dictionary was passed by remaining in memory across a fork)

I did it so that people can compare the argparse approach (before this commit) and the pickle approach (after this commit)

benclifford:
(@rjmello @yadudoc @khk-globus are the people that probably care about my last comment - https://github.com/Parsl/parsl/pull/3463#issuecomment-2160918089

)

benclifford:
@khk-globus suggested using stdin rather than a command line parameter, and I have made that change.

We also briefly discussed what protocol to use over stdin (primarily: pickle vs JSON). This PR sticks with pickle and I added a note about that into the PR description:

> Using pickle here rather than (eg.) JSON keeps the path open for sending richer configuration objects, beyond what can be encoded in JSON. This isn't something needed right now, but at least configurable monitoring radios (the immediate driving force behind this PR) are modelled around passing arbitrary configuration objects around to configure things - and so it seems likely that if interchange monitoring configuration is exposed to the user, richer objects would be passed here. See PR https://github.com/Parsl/parsl/pull/3315 for monitoring radio prototype.

khk-globus:
> @khk-globus suggested using stdin rather than a command line parameter, and I have made that change.

Hmm; I appear to have [commented in a weird place](https://github.com/Parsl/parsl/commit/e7d18aa4725310534e0ec4aa1ce5601d01ec2ecc#commitcomment-143018894); pulling up from the commit-comment so the PR has an easily discoverable record of it:

> Merely looking at the code removed, this is a beautiful addition.  The key insight being that _we_ are the consumers of this interface, so we don't have to support debuggability-for-humans, per se.  That's good.
>
> As discussed in Slack; I wonder if using pipe-to-stdin would be a better approach than faffing about with the CLI, "security concerns" there abouts, and potential (if nominal) size limits.  A secondary question is whether pickle (over base64) is the correct serialization, as opposed to a "simple" `repr`/`ast.literal_eval` cycle, or even JSON.

parsl/executors/high_throughput/interchange.py
```diff
@@ -672,13 +672,10 @@ def start_file_logger(filename: str, level: int = logging.DEBUG, format_string:
...
-    The executor is expected to call this function. The args, kwargs match that of the Interchange.__init__
-    """
+if __name__ == "__main__":
     setproctitle("parsl: HTEX interchange")
-    # logger = multiprocessing.get_logger()
-    ic = Interchange(*args, **kwargs)
+
+    config = pickle.load(sys.stdin.buffer)
+
+    ic = Interchange(**config)
```
khk-globus, line 680: Is there any reason to keep stdin open after this?  We might consider clearing that fd as soon as we're reasonably done with it.

khk-globus (approved):
Love it.  Nice work.

parsl/executors/high_throughput/interchange.py
benclifford: Feels kinda weird closing stdin at all, although should be legit and lets us open another half-a-worker-pool worth of connections (with fds being the limit right now on maximum worker pool connections, and each one requiring two fds), I guess.

Less weird to me is closing at the other end of the pipe in executor.py

benclifford: related, see #3501 which closes the executor end of this pipe

PR #3465, opened 2024-06-03
Remove unused-in-production HTEX interchange default values

benclifford:
This PR makes all parameters to the Interchange class into mandatory keyword-only arguments.

The removed defaults were not used in production use, because they were all specified explicitly in parsl/executors/high_throughput/executor.py too.

The single exception to this was client_address, which was defaulted in the interchange and never specified by the exeuctor. This PR moves that default into executor.py too, to work like all the other defaults.

See similar changes to the process worker pool, PR #2973, for more detailed justification.

test_zmq_binding.py is the only test which instantiates Interchange objects directly (rather than testing the executor as a whole) and this PR modifies that test to explicitly specify all interchange parameters rather than relying on the otherwise-unused defaults.

# Changed Behaviour

This should not change user-facing behaviour

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -527,7 +527,8 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         get the worker task and result ports that the interchange has bound to.
         """
         self.interchange_proc = ForkProcess(target=interchange.starter,
-                                            kwargs={"client_ports": (self.outgoing_q.port,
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
Makes sense.

yadudoc (approved):
I've looked through this a few times over and these changes look right to me. I know there's a plan in #3463 to switch to using subprocess to launch instead of ForkProcess but this would be good to get in now.

PR #3466, opened 2024-06-03
Remove environment variable configuration for usage tracking

### LINKED ISSUE #3456: [proposal] remove PARSL_TRACKING environment variable

Currently Parsl takes usage tracking configuration from both `Config` (like other configuration) and from the `$PARSL_TRACKING` environment variable.

The interactions between these two parameters, as well as two conflicting defaults (one in `Config` and one in `usage.py`), are quite subtle and hard to understand intuitively and in the code.

PR #3400 makes that configuration more complicated: instead of a `bool`, the usage tracking level becomes an `int` (of restricted range). In reviewing that PR, along with @khk-globus, it's become pretty apparent that fusing two configurations and defaults together is even more complex in this situation.

Other configuration options in Parsl do not have this "specify multiple configuration values, fuse them together" behaviour: configuration comes from the `Config` object (and objects contained within that class).

I would like to simplify this piece of Parsl by removing `$PARSL_TRACKING`: this removes a bunch of complexity both for user understanding of what usage level they have actually configured, and for programmers trying to understand these code paths.

A more generalised fuseable configuration mechanism that can fuse *any* configuration options, not only usage tracking level.

**Additional context**
This is part of a funded deliverable for usage tracking coordinated by @kylechard and @yadudoc, which @NishchayKarle is working on right now.

NishchayKarle:
# Description

This PR removes the use of `$PARSL_TRACKING` env variable to setup usage tracking configuration. Parsl will now only take usage tracking configuration from the `config`

# Changed Behavior
Anyone using only the env variable `PARSL_TRACKING` to enable usage tracking will need to set `usage_tracking = True` within the `config` to enable usage tracking.

# Fixes

Fixes #3456 

## Type of change

- Code maintenance/cleanup
- Update documentation

### DIFF
--- a/parsl/usage_tracking/usage.py
+++ b/parsl/usage_tracking/usage.py
@@ -59,7 +59,6 @@ def udp_messenger(domain_name: str, UDP_PORT: int, sock_timeout: int, message: b
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)  # UDP
         sock.settimeout(sock_timeout)
         sock.sendto(message, (UDP_IP, UDP_PORT))
-        print(message)
         sock.close()
 
     except socket.timeout:
--- a/parsl/usage_tracking/usage.py
+++ b/parsl/usage_tracking/usage.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import os
 import platform
 import socket
 import sys
@@ -60,6 +59,7 @@ def udp_messenger(domain_name: str, UDP_PORT: int, sock_timeout: int, message: b
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)  # UDP
         sock.settimeout(sock_timeout)
         sock.sendto(message, (UDP_IP, UDP_PORT))
+        print(message)
         sock.close()
 
     except socket.timeout:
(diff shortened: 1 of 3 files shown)


benclifford (changes requested):
There are a couple other places where that variable is mentioned. I used `git grep` to find them:

```
$ git grep PARSL_TRACKING
docs/quickstart.rst:You can opt-in by setting ``PARSL_TRACKING=true`` in your environment or by 
docs/userguide/usage_tracking.rst:If you wish to opt in to usage reporting, set ``PARSL_TRACKING=true`` in your environment or set ``usage_tracking=True`` in the configuration object (`parsl.config.Config`).
```

This PR should fix those references too

parsl/usage_tracking/usage.py
```diff
@@ -117,21 +116,16 @@ def __init__(self, dfk, port=50077,
...
-        Tracking will be enabled unless either of these is true:
+        Tracking will be enabled unless the following is true:
 
             1. dfk.config.usage_tracking is set to False
-            2. Environment variable PARSL_TRACKING is set to false (case insensitive)
 
         """
         track = True
 
         if not self.config.usage_tracking:
```
benclifford, line 126: this logic can be tidied up a lot! I think this function returns exactly the value of self.config.usage_tracking now? so the `if` statement and local variable are not needed at all?

NishchayKarle: changed this to return `self.config.usage_tracking`

benclifford (approved):

PR #3468, opened 2024-06-04
Add exit choices for the DFK context manager

benclifford:
Most important is a ``wait`` exit mode which will make the context manager wait for tasks to complete at exit, unless an exception is raised. This addresses a common usability problem where users forget to wait for their tasks and are then confused by their workflow exiting immediately, but does not defer exceptions - this block-on-exception behaviour was the main usability problem when attempting this before in PR #610.

# Changed Behaviour

The default mode should preserve existing behaviour - so none.

## Type of change

- New feature

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -217,9 +217,24 @@ class DataFlowKernel:
     def __enter__(self):
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
-        logger.debug("Exiting the context manager, calling cleanup for DFK")
-        self.cleanup()
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        mode = self.config.exit_mode
+        logger.debug("Exiting context manager, with exit mode '%s'", mode)
+        if mode == "cleanup":
+            logger.info("Calling cleanup for DFK")
+            self.cleanup()
+        elif mode == "skip":
+            logger.info("Skipping all cleanup handling")
+        elif mode == "wait":
+            if exc_type is None:
+                logger.info("Waiting for all tasks to complete")
+                self.wait_for_current_tasks()
+                self.cleanup()
+            else:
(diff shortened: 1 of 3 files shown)


benclifford:
@WardLT you might be interested in this - I think the same choices could be applied to the ParslPoolExecutor when used as a context manager.

parsl/dataflow/dflow.py
```diff
@@ -217,9 +217,24 @@ def __init__(self, config: Config) -> None:
...
+        elif mode == "skip":
+            logger.info("Skipping all cleanup handling")
+        elif mode == "wait":
+            if exc_type is None:
+                logger.info("Waiting for all tasks to complete")
+                self.wait_for_current_tasks()
+                self.cleanup()
+            else:
+                logger.info("There was an exception - cleaning up without waiting for task completion")
+                self.cleanup()
```
khk-globus, line 235: If `cleanup` is the default in almost all cases, could make the code follow "moral" suit.  Consider:
```python
# with typeguard, is this, in fact, necessary?
if mode not in ("cleanup", "wait", "skip"):
    raise InternalConsistencyError(f"Exit case for {mode} should be unreachable, validated by typeguard on Config()")

if mode == "skip":
    logger.info("Skipping all cleanup handling")
    return

if mode == "wait":
     if exc_type is None:
        logger.info("Waiting ...")
        self.wait_for_current_tasks()
     else:
        logger.info("was an exc ...")

logger.info("Calling cleanup for DFK")
self.cleanup()
```

khk-globus (approved):

parsl/dataflow/dflow.py
benclifford: could do - I think it comes down to seeing this case statement as choosing three distinct behaviours vs a "normal flow" that gets ornamented by different things on the way. I've ended up with more of a preference for being explicit about all three flows as explicit separate options, but I think it would be fine either way.

PR #3473, opened 2024-06-05
Made paramiko and globus optional dependencies to fix #2218

### LINKED ISSUE #2218: make globus and paramiko optional dependencies

arhag23:
# Description

Removed paramiko and globus_sdk from requirements.txt and added them as optional modules in setup.py. Added OptionalModuleMissing errors for the affected files if usage is attempted without the correct module being installed. Removed test that required globus_sdk to test external exceptions.

# Changed Behaviour

If users have code that depends on these now optional modules they might have to ensure that the proper extra modules are installed as well.

# Fixes

Fixes #2218

## Type of change

- Bug fix

### DIFF
--- a/parsl/channels/oauth_ssh/oauth_ssh.py
+++ b/parsl/channels/oauth_ssh/oauth_ssh.py
@@ -1,10 +1,14 @@
 import logging
 import socket
 
-import paramiko
-
 from parsl.channels.ssh.ssh import SSHChannel
 from parsl.errors import OptionalModuleMissing
+
+try:
+    import paramiko
+    _ssh_enabled = True
+except (ImportError, NameError, FileNotFoundError):
+    _ssh_enabled = False
 
 try:
     from oauth_ssh.oauth_ssh_token import find_access_token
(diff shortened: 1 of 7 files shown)


benclifford:
Hi. These changes mostly look ok. Can you check off which of the following you have tested still works:

- [ ] globus transfer
- [ ] ssh non-interactive
- [ ] ssh interactive login
- [ ] ssh oauth

then we can figure out what else needs testing.

benclifford:
if you aren't in a position to test globus changes, you might submit the ssh/paramiko-only changes as a separate PR. That would fit well with issue #3515 which is making SSH related stuff much more optional in preparation for removal.

PR #3476, opened 2024-06-07
Added options to set annotations and a service account in the Kuberne…

shishichen:
…tes worker pod configuration

# Description

Added options for pod annotations and a service account to be used by the pod, which are passed through to the Kubernetes client. These options are needed when using the provider on Google Kubernetes Engine for pods to mount Google Cloud Storage persistent volumes.

# Changed Behaviour

Enables additional options to be configured on the Kubernetes worker pods.

# Fixes

Fixes # n/a - minor change

## Type of change

- New feature

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -83,6 +83,10 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
     persistent_volumes: list[(str, str)]
         List of tuples describing persistent volumes to be mounted in the pod.
         The tuples consist of (PVC Name, Mount Directory).
+    service_account_name: str
+        Name of the service account to run the pod as.
+    annotations: Dict[str, str]
+        Annotations to set on the pod.
     """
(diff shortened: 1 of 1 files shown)


benclifford:
cool I'll try this out on my kubernetes cluster. @trey-stafford @mfisher87 might be in a better place to review this PR than me though.

benclifford (approved):
I tested out that pods start up with an alternate service account and with some basic annotations.

shishichen:
Thanks for the review!

mfisher87:
@benclifford thanks for looping us in! Unfortunately, our time commitment is temporarily coming to an end this week and don't have time for an in-depth review amongst the other tie-up tasks. We'll be looking at this work again in October! The Python looks great, but my context is super limited :)

One comment not related to this PR: I see some docstrings that have type hints in them, but there are also real Python type annotations on those methods & classes. Does your autodoc engine read the Python type annotations? In which case the docstring hints could be removed to simplify maintenance.

PR #3481, opened 2024-06-07
Provide a BlockProviderExecutor default scale-in method

### LINKED ISSUE #3471: Work Queue using Parsl scaling sometimes tries to scale in an old block and then breaks
**Describe the bug**

Sometimes work queue scale-in will generate a log message like this on repeated strategy runs, and will not scale in blocks it should be scaling in.

```

1717584031.221508 2024-06-05 03:40:31 MainProcess-256259 JobStatusPoller-Timer-Thread-140411461217168-140411449100032 parsl.process_loggers:30 wrapped ERROR: Exceptional ending for _general_strategy on thread JobStatusPoller-Timer-Thread-140411461217168
  File "/global/u1/b/bxc/tmp/tmp-202406-slurm/parsl/parsl/process_loggers.py", line 26, in wrapped
    r = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/global/u1/b/bxc/tmp/tmp-202406-slurm/parsl/parsl/jobs/strategy.py", line 256, in _general_strategy
    executor.scale_in_facade(active_blocks - min_blocks)
  File "/global/u1/b/bxc/tmp/tmp-202406-slurm/parsl/parsl/executors/status_handling.py", line 305, in scale_in_facade
    del self._status[block_id]
        ~~~~~~~~~~~~^^^^^^^^^^
KeyError: '0'
```

@svandenhaute reported this on his cluster and I have recreated it on perlmutter.

It looks like this comes from `WorkQueueExecutor.scale_in` choosing block IDs from `self.blocks_to_job_id` which lists all block IDs that have ever existed, not active+pending status blocks that should be used here.

# History

PR #3308 renamed that structure which makes this bug a bit more obvious to diagnose. Pior to that PR, `self.blocks_to_job_id` was called `self.blocks`.

This code was copied from `HighThroughputExecutor`, but `HighThroughputExecutor` now uses block IDs from self._status and from the interchange, rather than from self.blocks_to_job_id so should not have this same problem.

`TaskVineExecutor` also copies this code and likely also shows this problem, and this issue should not be closed until both TaskVine and WorkQueue executors are fixed.

**To Reproduce**

This program on perlmutter will scale up and down in the right ways to demonstrate this bug:

```

import parsl
from parsl.launchers import SrunLauncher
from parsl.executors import WorkQueueExecutor
from parsl.addresses import address_by_interface

config = Config(
    max_idletime=30,
    executors=[
        WorkQueueExecutor(
            provider=SlurmProvider(
                'debug',  # Partition / QOS
                init_blocks=0,
                min_blocks=0,
                max_blocks=3,
                nodes_per_block=1,
                # string to prepend to #SBATCH blocks in the submit
                scheduler_options='#SBATCH -C cpu',
                # Command to be run before starting a worker
                worker_init='module load python; source activate tmp-202406-slurm',
                # We request all hyperthreads on a node.
                launcher=SrunLauncher(overrides='-c 128'),
                walltime='00:05:00',
                # Slurm scheduler can be slow at times,
                # increase the command timeouts
                cmd_timeout=120,
            ),
        )
    ]
)

with parsl.load(config):


  @parsl.python_app
  def foo():
    import time
    time.sleep(1)


  print("first job")
  foo().result()
 
  print("now idling long enough to scale in")
  import time
  time.sleep(70)

  print("second job")
  foo().result()
 
  print("now idling long enough to scale in")
  import time
  time.sleep(70)


  print("now running another task")
  foo().result()
  print("done")
```

**Expected behavior**
scaling should wor

**Environment**

perlmutter, configured as above, on a slightly hacked up fork of bf98e5049c34eaf1400663404e29672b85d3ac5e

benclifford:
Prior to this PR, there was obvious-but-buggy (see issue #3471) scaling in code copy-paste shared by Work Queue and Task Vine (and prior to PR #1563 by High Throughput Executor)

This code moves that shared behavior into BlockProviderExecutor, and attempts to bugfix it there.

This PR is draft because I would like to write some tests around the scaling behaviour but don't have my head round what I want to test.

# Changed Behaviour

Different blocks will be chosen for scale in - hopefully a better choice.

# Fixes

#3471 

## Type of change

- Bug fix
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -213,28 +213,34 @@ class BlockProviderExecutor(ParslExecutor):
         """
         logger.debug("Number of blocks requested for scale in: %s", count)
 
-        # TODO: prefer pending blocks? because that's the only judgement we have about
-        # which blocks are empty?
-
         # TODO: prefer pending blocks with largest block ID? because they're likely the
-        # ones with least valuable queue position.
+        # ones with least valuable queue position - this assumes the block ID is an integer...
+        # so ugh... that's not type-safe info we have at the moment...
 
-        candidate_blocks = [block_id for block_id, job_status in self._status.items() if job_status.state not in TERMINAL_STATES]
+        candidate_blocks = [(block_id, job_status) for block_id, job_status in self._status.items() if job_status.state not in TERMINAL_STATES]
+
+        # this will prefer pending blocks (TODO: test that)
+        # because this is the only judgement we have here for preferring
+        # empty blocks (blocks with no tasks on them) - and we should prefer
+        # those for scaling in.
+
+        candidate_blocks.sort(key=lambda t: t[1].state)
+
         logger.debug("Candidate blocks for scale in: %s", candidate_blocks)
 
-        # Obtain list of blocks to kill
-        to_kill = candidate_blocks[:count]
+        # Obtain list of block IDs to kill
+        to_kill = [block_id for (block_id, _) in candidate_blocks[:count]]
 
         logger.debug("Blocks chosen to scale in: %s", to_kill)
 
-        kill_ids = [self.blocks_to_job_id[block] for block in to_kill]
+        kill_job_ids = [self.blocks_to_job_id[block] for block in to_kill]
 
         # Cancel the blocks provisioned
         if self.provider:
-            logger.info(f"Scaling in jobs: {kill_ids}")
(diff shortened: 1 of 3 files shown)


benclifford:
addressing this in PR #3526 and a subsequent PR, as part of work on draft PR #3295

PR #3485, opened 2024-06-12
Remove unused inputs and outputs fields for AppBase

arhag23:
# Description

Testing to see if `AppBase` needs `outputs` and `inputs` fields or the lines changed are necessary.

# Changed Behaviour

Removed the last 2 lines of `AppBase.__init__` which set `AppBase.inputs` and `AppBase.outputs` to default values if they exist.

## Type of change

- Code maintenance/cleanup


benclifford (approved):
I also looked through the current code, and also git history, and I think these are two fields that were introduced back at the very start of parsl in 2017 when the task and decorator model was still being unexplored - as far as I can tell, they've been unused for a long time, but not causing anyone problems.

PR #3486, opened 2024-06-12
Fixed Kubernetes worker container launch command to remove trailing semicolon.

shishichen:
# Description

The launch command is formatted with a newline at the end so when the trailing semicolon is added here, it causes the semicolon to be run as a separate bash command by the worker. A single semicolon is a syntax error in bash and produces an error in container logs when the container tries to run it. Removing it is safe since it's at the end of the command and nothing is concatenated after it.

# Changed Behaviour

Remove an error in the Kubernetes worker container logs.

# Fixes

Fixes # n/a - minor issue

## Type of change

- Bug fix


benclifford:
I haven't tried it out, but that ; is maybe unnecessary and could be removed? At first glance it looks like noise to me.

shishichen:
Yeah, that works too - updated.

benclifford (approved):

PR #3493, opened 2024-06-20
Add debug logging around local channel process launch

benclifford:
This is driven by a case where it is unclear when the Parsl scaling code is launching processes to monitor batch queue status.

# Changed Behaviour

This will add 4 debug lines per block started, block cancelled, and per poll (limited by provider.status_polling_interval() and config.strategy_period)

## Type of change

- New feature

### DIFF
--- a/parsl/channels/local/local.py
+++ b/parsl/channels/local/local.py
@@ -68,13 +68,13 @@ class LocalChannel(Channel, RepresentationMixin):
             logger.debug("Created process with pid %s. Performing communicate", proc.pid)
             (stdout, stderr) = proc.communicate(timeout=walltime)
             retcode = proc.returncode
-            logger.debug("Process returned %s", proc.returncode)
+            logger.debug("Process %s returned %s", proc.pid, proc.returncode)
 
         except Exception:
-            logger.exception(f"Execution of command '{cmd}' failed")
+            logger.exception(f"Execution of command failed:\n{cmd}")
             raise
         else:
-            logger.debug("Execution of command completed normally")
+            logger.debug("Execution of command in process %s completed normally", proc.pid)
 
         return (retcode, stdout.decode("utf-8"), stderr.decode("utf-8"))
 
--- a/parsl/channels/local/local.py
+++ b/parsl/channels/local/local.py
@@ -55,6 +55,7 @@ class LocalChannel(Channel, RepresentationMixin):
         current_env.update(envs)
 
         try:
+            logger.debug("Creating process with command '%s'", cmd)
             proc = subprocess.Popen(
                 cmd,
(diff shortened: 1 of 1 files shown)


benclifford:
cc @cms21 - this is extra logging I mentioned in Globus Compute's `#help` slack channel for debugging excessive `qstat` calls

parsl/channels/local/local.py
```diff
@@ -64,12 +65,16 @@ def execute_wait(self, cmd, walltime=None, envs={}):
...
             )
+            logger.debug("Created process with pid %s. Performing communicate", proc.pid)
             (stdout, stderr) = proc.communicate(timeout=walltime)
             retcode = proc.returncode
+            logger.debug("Process returned %s", proc.returncode)
 
-        except Exception as e:
-            logger.warning("Execution of command '{}' failed due to \n{}".format(cmd, e))
+        except Exception:
+            logger.exception(f"Execution of command '{cmd}' failed")
```
khk-globus, line 74: For clarity in the logs, you might move the command to the end of the string, perhaps delineated by a newline.  Consider:
```python
logger.exception(f"Execution of command failed:\n  {cmd}")
```

parsl/channels/local/local.py
```diff
@@ -64,12 +65,16 @@ def execute_wait(self, cmd, walltime=None, envs={}):
...
             raise
+        else:
+            logger.debug("Execution of command completed normally")
```
khk-globus, line 77: In a multiple-concurrent-commands scenario, is this enough information?  Would adding the command that completed normally to the log line be of use?  Something like:
```python
logger.debug("Execution of command completed normally: (%s) %s", retcode, cmd)
```
I don't have enough context here to know if that would just be noise, but thinking in terms of a retrospective and "how do I correlate which subprocess finished at what time?" and so forth.

khk-globus (approved):

parsl/channels/local/local.py
benclifford: PR #2328 was intended to address pretty much exactly this kind of ambiguity by adding a bunch more context to every log line to give that kind of context. Adding the process ID of the Popened process to all of these log lines would maybe make things more explicit.

benclifford: pid added to some log lines in 9745d88

parsl/channels/local/local.py
benclifford: rearranged like that in 9745d88

cms21:
Thanks @benclifford, when will this be available in Globus Compute? As soon as it's released in the public Parsl version?

rjmello:
We expect to include this in the July 10th Globus Compute release.

PR #3494, opened 2024-06-24
enable usage tracking in example configs

NishchayKarle:
# Description

Enable usage tracking with level 1 in the configs

# Changed Behaviour

Example configs will have `usage_tracking` set to the lowest level and share the following user data: `python-version`, `parsl-version` and `platform`

## Type of change

- New feature


benclifford (approved):

danielskatz:
thanks @NishchayKarle for doing this!

Issue #3495, opened 2024-06-26
HTEX interchange ManagerLost/VersionMismatch reporting is broken by PR #3463


benclifford:
PR #3463 rearranges how classes are defined in the interchange, as an effect of rearranging how the interchange is launched. Prior to PR #3463, the interchange code was imported as a module, `parsl.executors.high_throughput.interchange` and all top level definitions in that module existed as, for example, `parsl.executors.high_throughput.interchange.ManagerLost`.

When sending a `ManagerLost` exception back to the interchange, this would be serialized (by pickle) including a reference to module+name `parsl.executors.high_throughput.interchange.ManagerLost`. There is an implicit requirement that the interchange is running from the same environment as the submit side so that module+name will be resolvable/`import`able on the submit side when unpickling the exception.

PR #3463 changes how the interchange is launched - interchange.py now runs as the `__main__` module of the interchange process, and top level definitions there exist as (for example) `__main__,ManagerLost`. References to `__main__.ManagerLost` cannot in general be deserialized on the submit side, because `__main__` refers to the user workflow process.

When sending a `ManagerLost` back to the submit side, this appears on the submit side as a deserialization error, rather than a properly deserialized `ManagerLost`:

```
parsl.serialize.errors.DeserializationError: Failed to deserialize objects. Reason: Received exception, but handling also threw an exception: Can't get attribute 'ManagerLost' on <module '__main__' from '/home/benc/parsl/src/parsl/wl.py'>
```

Discovered by @khk-globus as part of Globus Compute work.

**To recreate**

Start a long running `python_app` in a parsl process. killall -STOP process_worker_pool.py to suspend the worker pool, leading to a heartbeat timeout and `ManagerLost` (don't terminate the worker pool because Parsl will sometimes notice that the process has died on the submit side). Observe the broken exception.

**Proposed fix**
Move `ManagerLost` and `VersionMismatch` into `parsl.executors.high_throughput.errors` (alongside `WorkerLost`).

Guiding principle: objects which will be serialized should be importable, not defined locally. (dill has some complications to make that not entirely true, to do with code movement, but that isn't relevant here, I think)

PR #3496, opened 2024-06-26
Move ManagerLost and VersionMismatch to errors.py

khk-globus:
Per the analysis in #3495, defining the `ManagerLost` and `VersionMismatch` errors in the `interchange.py` became a problem in #3463, where the interchange now runs as `__main__`.  This makes it difficult for Dill to get the serde correct.

The organizational fix is simply to move these classes to an importable location, which follows the expectation that classes are available in both local and remote locations, which defining in `__main__` can't easily guarantee.

Fixes: #3495

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_serialization/test_3495_deserialize_managerlost.py
+++ b/parsl/tests/test_serialization/test_3495_deserialize_managerlost.py
@@ -1,0 +1,47 @@
+import os
+import signal
+
+import pytest
+
+import parsl
+from parsl import Config, HighThroughputExecutor
+
+
+@parsl.python_app
+def get_manager_pgid():
+    import os
+    return os.getpgid(os.getpid())
+
+
+@parsl.python_app
+def lose_manager():
+    import os
+    import signal
+
+    manager_pid = os.getppid()
+    os.kill(manager_pid, signal.SIGSTOP)
+
+
+@pytest.mark.local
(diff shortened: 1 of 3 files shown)


benclifford (approved):

parsl/tests/test_serialization/test_3495_deserialize_managerlost.py
```diff
@@ -0,0 +1,47 @@
...
+        heartbeat_period=1,
+        heartbeat_threshold=1,
+    )
+    c = Config(executors=[hte], strategy='simple', strategy_period=0.1)
+
+    with parsl.load(c):
+        manager_pgid = get_manager_pgid().result()
+        try:
+            lose_manager().result()
+        except Exception as e:
```
benclifford, line 42: i guess here I would have put in a test that what comes back really is a ManagerLost

khk-globus: Yeah; I see that.  That was actually the first iteration of my test, but I lost the thread yesterday when sifting through the "why won't it work?!" details.  I'll look at a follow-up PR later today.

khk-globus: Addressed in #3497.

PR #3501, opened 2024-07-01
Close stdin used to send config to Interchange

benclifford:
This was introduced in PR #3463 and at the time I incorrectly assumed that interchange exit would close both ends of the pipe. That is untrue.

For example: pytest parsl/tests/test_htex/ --config local ends with 341 fds open before this PR,
and 327 file descriptors open after this PR.

# Changed Behaviour

nothing should be user visible, but less file descriptor usage

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -551,6 +551,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         logger.debug("Popened interchange process. Writing config object")
         stdin.write(config_pickle)
         stdin.flush()
+        stdin.close()
         logger.debug("Sent config object. Requesting worker ports")
         try:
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -551,6 +551,7 @@ def _start_local_interchange_process(self) -> None:
         logger.debug("Popened interchange process. Writing config object")
         stdin.write(config_pickle)
         stdin.flush()
+        stdin.close()
         logger.debug("Sent config object. Requesting worker ports")
```
khk-globus, line 555: Suggest adding some version of a UT for this.  It's not an important detail, per se, but it's also a detail that I can see being easily forgotten/missed in a refactor.  Especially from newer contributors.

khk-globus (approved):

parsl/executors/high_throughput/executor.py
benclifford: #3397 is a longer term plan to catch all leftover fds and threads

PR #3502, opened 2024-07-02
Codify assumptions about provider.channel and provider.channels as asserts

benclifford:
The code will misbehave if a provider lacks channel(s) or if it has both.

Historically: a provider must have some way to run its commands, and that is what having a channel means. But the AdHocProvider wants to be given many channels, and the type system cannot express: "has .channel XOR has .channels"

## Type of change

- Code maintenance/cleanup


benclifford:
(fails at the moment because these assumptions are wrong...)

benclifford:
the assumptions are more subtle... so abandoning this PR...

Issue #3484, opened 2024-06-12
flux-in-parsl-CI testing is very hangy


benclifford:
**Describe the bug**

PR #3159 introduces per-PR flux testing. This hangs often - it's not immediately clear why.

This doesn't stop PRs being merged, because that flux test is not a mandatory test (because of lack of confidence in it being able to pass often) - but it means that we will be ignoring flux test failures, which makes the test pointless.

See notes on PR #3259 for some earlier investigations

**To Reproduce**
look at some recent CI builds

benclifford:
I've dug into this a bit more on my laptop (where I can use the flux testing container to see this bug happening).

With this `PYTHONFAULTHANDLER=1 python3 -m pytest parsl/tests/test_flux.py --config local  --random-order` I often get hangs.

Sending a SIGSEGV into that process causes the enabled `faulthandler` to dump a stack trace, and that basically shows that ZMQ is hanging in cleanup during a garbage collection.

All stack traces basically look like this (with the garbage collection happening at an arbitrary point in the test and FluxExecutor code - so `__del__` appears in the stack trace invoked from an arbitrary piece of code because thats where the garbage collection happened to execute.

```
Current thread 0x000076c1e68f01c0 (most recent call first):
  Garbage-collecting
  File "/usr/local/lib/python3.10/dist-packages/zmq/sugar/context.py", line 266 in term
  File "/usr/local/lib/python3.10/dist-packages/zmq/sugar/context.py", line 324 in destroy
  File "/usr/local/lib/python3.10/dist-packages/zmq/sugar/context.py", line 142 in __del__
  File "/usr/lib/python3.10/threading.py", line 247 in __init__
  File "/usr/lib/python3.10/threading.py", line 546 in __init__
  File "/parsl/parsl/executors/flux/executor.py", line 207 in __init__
  File "/parsl/parsl/tests/test_flux.py", line 37 in test_multiply
```

ZMQ `term` will (as documented) wait until all sockets are closed, but at least I don't see that there is anything that will deterministically close the ZMQ socket used in `FluxExecutor` - it might happen to be garbage collected and closed that way, I think, sometimes, but I haven't dug into which socket is being waited for, or into the garbage collection sequence.

My experience both with `__del__` (used in the Python ZMQ implementation) and with PyZMQ in general elsewhere in Parsl (in the High Throughput Executor and monitoring) is that letting this stuff happen at garbage collection time works out badly and that it would be better to close things explicitly because of random hangs like this.

[...]

PR #3503, opened 2024-07-02
Close Channels at DFK cleanup

benclifford:
Following on from #3403, which tightens the type signature of unused Channel.close(), this PR makes the DFK actually call Channel.close() when it shuts down an executor.

At startup, the DFK chooses whether to manage things with channels by checking for the presence of a `script_dir` attribute on the relevant provider (!) and this PR sticks with that same informal protocol.

On the LocalChannel, this has no effect. On the SSH Channels, this shuts down the SSH connection and removes an otherwise-abandoned thread.

# Changed Behaviour

Channels will be closed at shutdown.

## Type of change

- Bug fix


benclifford:
this is consistently(?) hanging in the flux tests... (in a similar way to the very-often-hanging flux tests, but seemly more so). I'll try to debug.

benclifford:
(crossreference #3484 for main issue to track flux CI hangs)

parsl/dataflow/dflow.py
```diff
@@ -1277,6 +1277,19 @@ def cleanup(self) -> None:
...
+            if hasattr(executor, 'provider'):
+                if hasattr(executor.provider, 'script_dir'):
+                    logger.info(f"Closing channel(s) for {executor.label}")
+
+                    if hasattr(executor.provider, 'channels'):
+                        for channel in executor.provider.channels:
+                            channel.close()
+                    else:
+                        assert hasattr(executor.provider, 'channel'), "If provider has no .channels, it must have .channel"
+                        executor.provider.channel.close()
```
khk-globus, line 1289: For debugging / retrospective, would `log.debug()` for each closed channel be of any use?  (And if so, do we have recognizable output from `channel.__repr__` or `.__str__` to make that feasible?)

khk-globus (approved):

parsl/dataflow/dflow.py
benclifford: yes to both: logging around closing is pretty useful here, even at INFO level, due to Parsl's history of hanging at shutdown; Channels are rendered as a recursive part of the start log-the-`Config`-object so they "should" already have an appropriate repr().

PR #3505, opened 2024-07-02
Join on stopped thread at end of LocalProvider+ssh test

benclifford:
This stops that thread being left still running over the end of the test - which is needed in work in PR #3397 to ensure that no threads are left running at the end of a test.

If that shutdown hangs, this test will now hang rather than leave the thread behind.

Along with PR #3504, this eliminates any threads left behind after parsl/tests/test_providers/test_local_provider.py

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Good call.

PR #3509, opened 2024-07-03
Allow multiple workers to share a CUDA device, intended for use with MPS mode

cms21:
Allow multiple workers to share a CUDA device, intended for use with MPS mode

# Description

This change allows in the case of CUDA devices the ability to set the same value of CUDA_VISIBLE_DEVICES for multiple Parsl workers on a node when using the high throughput executor.  This allows the user to make use of the MPS mode for CUDA devices to partition a GPU to run multiple processes per GPU.

To use MPS mode with this functionality several settings must be set by the user in their config.  

- `available_accelerators` should be set to the total number of GPU processes to be run on the node.  For example, for a node with 4 Nvidia GPUS, if you wish to run 4 processes per GPU, `available_accelerators` should be set to `16`.
- `worker_init` should include commands to start the MPS service and set any associated environment variables.  For example on the ALCF machine Polaris, it is recommended the user [make use of a bash script that starts the MPS service on a node called `enable_mps_polaris.sh`](https://docs.alcf.anl.gov/running-jobs/example-job-scripts/#using-mps-in-multi-node-jobs).  `worker_init` should then contain `worker_init='export NNODES='wc -l < $PBS_NODEFILE'; mpiexec -n ${NNODES} --ppn 1 /path/to/mps/script/enable_mps_polaris.sh'`

# Changed Behaviour

This PR should not change any existing behavior.

## Type of change

- New feature


benclifford:
@cms21 there are some linting errors, which if you have the development dependencies installed you should be able to recreate locally with `make flake8 mypy isort`

```
flake8 parsl/
parsl/executors/high_throughput/process_worker_pool.py:738:5: E722 do not use bare 'except'
parsl/executors/high_throughput/process_worker_pool.py:740:1: W293 blank line contains whitespace
parsl/executors/high_throughput/process_worker_pool.py:744:70: E228 missing whitespace around modulo operator
parsl/executors/high_throughput/process_worker_pool.py:744:82: E261 at least two spaces before inline comment
parsl/executors/high_throughput/process_worker_pool.py:745:9: E722 do not use bare 'except'
make: *** [Makefile:44: flake8] Error 1
```

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -731,9 +731,19 @@ def worker(
         os.sched_setaffinity(0, my_cores)  # type: ignore[attr-defined, unused-ignore]
         logger.info("Set worker CPU affinity to {}".format(my_cores))
 
+    # If CUDA devices, find total number of devices to allow for MPS
+    # See: https://developer.nvidia.com/system-management-interface
+    try:
+        mod_factor = int(os.popen("nvidia-smi -L | wc -l").read())
```
benclifford, line 737: probably should be calling .close() on the popened object, after its been read()

benclifford:
So this PR is splitting the accelerator notion into two different things, a bit like happens with regular cores: theres a notion of a "slot" which a worker can be allocated to, of which there may now be arbitrarily many; and there are the GPUs, and now there may be multiple slots per GPU, rather than one GPU = one slot.   (compare with htex worker slots, which might have a single core, or have multiple cores assigned to them -- this is the other way round (more cores than slots) but the basic idea is that the count in each set can now be different)

I have tried to think of some ways in which this can break in other situations:

if available_accelerators is specified as a list of identifiers, rather than as a count, and it turns out those identifiers can be parsed as an arbitrary int, and the user happens to have `nvidia-smi` installed, then the chosen accelerator ID will be interfered with by the `%` and this might result in an accelerator being chosen that is not in the original available_accelerator list. (say I'm manually specifying a set of integer accelerator IDs that I want a workflow to use that is not the full config)

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -731,9 +731,19 @@ def worker(
...
+    except:
+        mod_factor = None
+        
     # If desired, pin to accelerator
     if accelerator is not None:
-        os.environ["CUDA_VISIBLE_DEVICES"] = accelerator
+        try:
+            os.environ["CUDA_VISIBLE_DEVICES"] = str(int(accelerator)%mod_factor) # multiple workers will share a GPU
```
WardLT, line 744: This will order accelerators as [0, 1, 0, 1], which doesn't line up with the memory locality on Polaris (adjacent blocks of cores map to the same GPU). Could we do `int(accelerator) // mod_factor` instead so that it goes [0, 0, 1, 1]? 

Also (pendanty warning). what about naming `mod_factor` to "num_cuda_devices" instead for better clarity?

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -731,9 +731,19 @@ def worker(
...
+        except:
```
WardLT, line 745: The only time this should fail is if `mod_factor` is None. So, what about using `if mod_factor is not None ... else` instead of `try ... catch`?

WardLT (changes requested):
Thanks, @cms21 ! I think this is a good feature to add, but have some suggestions on how to make the implementation better reflect memory locality on Polaris.

Could you also update the documentation in `configuring.rst` too? Users should know that Parsl will auto-detect the number of devices for NVIDIA accelerators and can pin multiple workers per accelerator if you list more accelerators than devices.

benclifford:
@WardLT 

> Users should know that Parsl will auto-detect the number of devices for NVIDIA accelerators and can pin multiple workers per accelerator if you list more accelerators than devices.

what was the behaviour before this PR if you used non-existing accelerator numbers before this PR? did people get an error message? so is there/what is the behaviour change with this PR for that bad usage? (vs intentionally oversubscribing like the point of this PR)

@cms21 related to that, what happens to a user who deliberately puts multiple tasks on an accelerator but they haven't properly done the equivalent of `enable_mps_polaris` in their environment?

cms21:
> Thanks, @cms21 ! I think this is a good feature to add, but have some suggestions on how to make the implementation better reflect memory locality on Polaris.
> 
> Could you also update the documentation in `configuring.rst` too? Users should know that Parsl will auto-detect the number of devices for NVIDIA accelerators and can pin multiple workers per accelerator if you list more accelerators than devices.

I will add some documentation to the Parsl docs.  At some point, I will also update the Parsl docs on the ALCF website.

parsl/executors/high_throughput/process_worker_pool.py
cms21: I agree this should be the correct order.  To accomplish it, I think what's needed is something like:
```
procs_per_cuda_device = pool_size//mod_factor
os.environ["CUDA_VISIBLE_DEVICES"] = int(accelerator)//procs_per_cuda_device
```

cms21:
> 
> @cms21 related to that, what happens to a user who deliberately puts multiple tasks on an accelerator but they haven't properly done the equivalent of `enable_mps_polaris` in their environment?

I believe what happens is that multiple processes are run on the GPU, but they will not be concurrent because of context switching.  MPS should ensure concurrency of up to 48 processes per GPU.

One note, there is another way to run multiple processes per Nvidia GPU called MIG that does a physical partition of the GPU.  That is much more complicated and is only supported on a few Polaris queues I believe.  So I'm not attempting to address that with this PR.

parsl/executors/high_throughput/process_worker_pool.py
cms21: The one other time it would fail is if `accelerator` can't be cast as an `int`.  I'm focused on Polaris and Perlmutter for this development.  I'm not sure if there is a situation where `CUDA_VISIBLE_DEVICES` would be set to a non-integer, at least, I'm not aware of one.  So I think any situation that accelerator can't be cast as an `int`, `mod_factor` should also be `None`, so I think you are right, but if anyone knows better, please correct me.

cms21: So if someone wanted to set a worker to use 2 GPUS, accelerator should be `"0,1"` or `"2,3"`.

cms21: In that case, we want this to fail and fall back to using whatever string `accelerator` was set to.

cms21:
> if available_accelerators is specified as a list of identifiers, rather than as a count, and it turns out those identifiers can be parsed as an arbitrary int, and the user happens to have `nvidia-smi` installed, then the chosen accelerator ID will be interfered with by the `%` and this might result in an accelerator being chosen that is not in the original available_accelerator list. (say I'm manually specifying a set of integer accelerator IDs that I want a workflow to use that is not the full config)

I guess the last condition I'd add to this scenario is that `CUDA_VISIBLE_DEVICES` is being used by the code somehow.  This change does not affect any other GPU affinity environment variables.

parsl/executors/high_throughput/process_worker_pool.py
WardLT: Good catch!

parsl/executors/high_throughput/process_worker_pool.py
WardLT: Nice! I forgot that it could be a string.

parsl/executors/high_throughput/process_worker_pool.py
benclifford: this is polaris-specific, right, so we might expect in future users to want a different layout, in the same way that there are different worker<->core pins, right?

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -731,9 +732,30 @@ def worker(
...
+                procs_per_cuda_device = pool_size // num_cuda_devices
+                partitioned_accelerator = str(int(accelerator) // procs_per_cuda_device)  # multiple workers will share a GPU
+                os.environ["CUDA_VISIBLE_DEVICES"] = partitioned_accelerator
+                logger.info(f'Pinned worker to partitioned cuda device: {partitioned_accelerator}')
+            else:
+                os.environ["CUDA_VISIBLE_DEVICES"] = accelerator
+        except (TypeError, ValueError, ZeroDivisionError):
+            os.environ["CUDA_VISIBLE_DEVICES"] = accelerator
+        except Exception as e:
+            print(f"Unrecognized exception in setting cuda affinity: {e}\n")
```
benclifford, line 757: Perhaps do this to get it to go into the relevant worker_* log file rather than stdout.

```suggestion
            logger.exception("Unrecognized exception in setting cuda affinity")
```

... although I would also be happy with this not being caught, propagating out the top and killing the worker on the principle of "when there is weird stuff, let's not pretend it didn't happen"...

benclifford: after more thought, i am more convinced that this second catch of all exceptions should not be here

see https://gist.github.com/benclifford/1ad76311d082657bb8baf23de0063d18#exceptions for some notes i made

cms21: Okay, so do you want me to remove the second catch all together?

benclifford: yes

parsl/executors/high_throughput/process_worker_pool.py
cms21: The previous behavior put the GPU ids in ascending order on every machine, this change aligns with that behavior, so I'm not sure I see how this is Polaris specific.  Users still have the ability to tailor the CPU affinity to the GPU affinity however they choose.  I could imagine applications where you want certain ranks on certain GPUS for mpi communication, but that doesn't apply here?

However, I should modify the documentation to explain the gpus will be assigned in this fashion.

benclifford (approved):

benclifford:
I'm pretty uncomfortable with running that nvidia specific cuda command on every single worker ever and I've been trying to figure out what to do differently. I think this is something that only needs doing `if accelerator is not None` (that is, if the user has specified some accelerators )

cms21:
> I'm pretty uncomfortable with running that nvidia specific cuda command on every single worker ever and I've been trying to figure out what to do differently. I think this is something that only needs doing `if accelerator is not None` (that is, if the user has specified some accelerators )

Okay, I see you accepted this pr, do you want me to open another one to address this?

benclifford:
Yes, please

PR #3512, opened 2024-07-03
Fix debug code sample in FAQ

giordano:
# Description

`logging` is undefined in this code sample.  I think this has been broken at least since #392.

# Changed Behaviour

Make the example actually work.

# Fixes

N/A

## Type of change


docs/faq.rst
```diff
@@ -19,7 +19,7 @@ Alternatively, you can configure the file logger to write to an output file.
    parsl.set_stream_logger()
 
    # Write log to file, specify level of detail for logs
-   parsl.set_file_logger(FILENAME, level=logging.DEBUG)
+   parsl.set_file_logger(FILENAME, level=parsl.logging.DEBUG)
```
benclifford, line 22: I think the right thing to do here is `import logging`, rather than refer to `parsl.logging`

`parsl.logging` only exists because `parsl/__init__.py` imports logging athttps://github.com/Parsl/parsl/blob/4695c00ad9411967ee9c9ba7a7b2d771bb425603/parsl/__init__.py#L17  which causes the entire Python `logging` module to also appear under `parsl.logging`.

That's probably shouldn't be regarded as part of the API of Parsl - it is imported for some code used later on inside `parsl/__init__.py`.

benclifford (changes requested):
see comment for https://github.com/Parsl/parsl/pull/3512/files#r1667687807

docs/faq.rst
giordano: Uh, I didn't realise it was Python's `logging`, I thought it was an internal Parsl module called like that.

benclifford (approved):

PR #3514, opened 2024-07-05
Support custom interchange launch command

rjmello:
# Description

The command used to launch the HTEX interchange process can now be customized.

## Type of change

- New feature


benclifford:
@rjmello you need to edit MPIExecutor and tell it about the new htex parameter

rjmello:
> @rjmello you need to edit MPIExecutor and tell it about the new htex parameter

Good call; all set.

benclifford (approved):

PR #3516, opened 2024-07-09
Adding warning about provider options in MPI context

yadudoc:
# Description

Parsl providers assume a pilot job model, where a single manager runs per node. This model conflicts with the model that `MPIExecutor` uses where the scheduler starts a single manager on the entire job which then launches MPI apps. This PR adds a warning to documentation cautioning use of `cores_per_node` in the SlurmProvider and similar options.  

This PR comes Christopher Harrop from NOAA reported on the help channel. 

## Type of change

### DIFF
--- a/docs/userguide/mpi_apps.rst
+++ b/docs/userguide/mpi_apps.rst
@@ -60,6 +60,13 @@ An example for ALCF's Polaris supercomputer that will run 3 MPI tasks of 2 nodes
     )
 
 
+.. warning::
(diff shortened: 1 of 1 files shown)


benclifford:
"interact poorly" is kinda hands-oft phrasing I'd expect from a British person... Does this mean danger do not set these options?

yadudoc:
Hey @benclifford I've updated the text to be less "British" :)

benclifford (approved):

PR #3520, opened 2024-07-11
Free up the *Radio namespace for future config structures

benclifford:
Ongoing monitoring radio work (see PR #3315) introduces per-radio configuration classes using *Radio names.

This PR frees up the *Radio namespace for that use, by renaming non-user-exposed internal classes out of the way.

# Changed Behaviour

none

## Type of change

- New feature


khk-globus (approved):
Heh; in effect `<>s/Radio/RadioSender/`.

PR #3521, opened 2024-07-11
Move monitoring router parameters into object attributes

benclifford:
This is to support rearrangement of the structure of the router code into multiple threads and methods, without having to manually wire all of the multiprocessing objects between the new methods and threads.

These objects are part of the context of the router object, rather than parameters to individual methods which might change, and they are all multiprocessing objects which are thread-safe.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/router.py
+++ b/parsl/monitoring/router.py
@@ -32,7 +32,12 @@ class MonitoringRouter:
                  logdir: str = ".",
                  run_id: str,
                  logging_level: int = logging.INFO,
-                 atexit_timeout: int = 3    # in seconds
+                 atexit_timeout: int = 3,   # in seconds
+                 priority_msgs: "queue.Queue[AddressedMonitoringMessage]",
+                 node_msgs: "queue.Queue[AddressedMonitoringMessage]",
+                 block_msgs: "queue.Queue[AddressedMonitoringMessage]",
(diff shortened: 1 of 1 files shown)


parsl/monitoring/router.py
```diff
@@ -200,7 +215,7 @@ def router_starter(comm_q: "queue.Queue[Union[Tuple[int, int], str]]",
 
         router.logger.info("Starting MonitoringRouter in router_starter")
         try:
-            router.start(priority_msgs, node_msgs, block_msgs, resource_msgs, exit_event)
+            router.start()
```
khk-globus, line 218: Since making this semantic change, I might suggest similarly altering `.start()`, which begins to imply an asynchronous thought-process / operation, to `.run()`.  No functional difference, but at least follows the semantics of a few other libraries (e.g., `Threading.Thread`)

khk-globus (approved):
Just moving the attachment point, but it's all the difference.  Much cleaner.

parsl/monitoring/router.py
benclifford: ... but `multiprocessing` which I guess is what influenced this original uses `.start` and has a `.run` that means something completely different... :exploding_head:

PR #3526, opened 2024-07-16
Refactor naive scale in behaviour for Work Queue and Task Vine

benclifford:
The intended behaviour of this scale in code, which is only for scaling in all blocks (for example at the end of a workflow) makes sense as a default for all BlockProviderExecutors. This PR makes that refactor.

This code is buggy (before and after) - see issue #3471. This PR does not attempt to fix that, but moves code into a better place for bugfixing, and a subsequent PR will fix it.

# Changed Behaviour

This PR should not change any behaviour for any of the executors in the Parsl source tree. It might change behaviour for any out-of-tree executors.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -199,15 +199,32 @@ class BlockProviderExecutor(ParslExecutor):
                 self._simulated_status[block_id] = JobStatus(JobState.FAILED, "Failed to start block {}: {}".format(block_id, ex))
         return block_ids
 
-    @abstractmethod
     def scale_in(self, blocks: int) -> List[str]:
         """Scale in method.
 
         Cause the executor to reduce the number of blocks by count.
 
+        The default implementation will kill blocks without regard to their
+        status or whether they are executing tasks. Executors with more
+        nuanced scaling strategies might overload this method to work with
+        that strategy - see the HighThroughputExecutor for an example of that.
+
         :return: A list of block ids corresponding to the blocks that were removed.
         """
-        pass
+        # Obtain list of blocks to kill
+        to_kill = list(self.blocks_to_job_id.keys())[:blocks]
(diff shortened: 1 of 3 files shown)


benclifford:
@tphung3 @colinthomas-z80 you might be interested in looking at this

tphung3:
This makes sense to me. Thanks for letting me know!

parsl/executors/status_handling.py
```diff
@@ -199,15 +199,32 @@ def scale_out(self, blocks: int = 1) -> List[str]:
...
         """
-        pass
+        # Obtain list of blocks to kill
+        to_kill = list(self.blocks_to_job_id.keys())[:blocks]
+        kill_ids = [self.blocks_to_job_id[block] for block in to_kill]
+
+        # Cancel the blocks provisioned
+        if self.provider:
+            logger.info(f"Scaling in jobs: {kill_ids}")
+            r = self.provider.cancel(kill_ids)
```
khk-globus, line 221: - in the error case, no sense in defining `kill_ids`
- if it reads clearer to you (does to me) `kill_ids` could be inlined:

    ```python
    if self.provider:
        kill_ids = [v for i, v in enumerate(self.blocks_to_job_id.values()) if i < blocks]
        logger.info(f"Scaling in jobs: {kill_ids}")
        ...
    ```
- On the other hand, I see that this mostly consolidating already-existing code, and the PR description mentions fixing it in a follow-up.  Your call.

khk-globus (approved):

parsl/executors/status_handling.py
benclifford: yeah this PR i really wanted to just do copy/paste refactoring.

Looking at this again (not for this PR):

I also think the else path should properly be treated as an error:

either raise an exception, or be an `assert` instead of an `if` with the rule that scale_in should only be called from internal code, and that code shouldn't be running if there is no provider.

PR #3349, opened 2024-04-11
Move send_monitoring_info method into BlockProviderExecutor

benclifford:
This should not change behaviour: it is a code move.

## Type of change

- Code maintenance/cleanup

PR #3527, opened 2024-07-16
Update block monitoring log message

benclifford:
* This monitoring message is not coming from the job status poller - this moved in PR #3349.

* This monitoring message is not being sent to the hub, but rather to the monitoring router on the far end of the monitoring radio.

* Debug messages should be formatted with deferred logger formatting.

## Type of change

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -243,7 +243,7 @@ class BlockProviderExecutor(ParslExecutor):
         # Send monitoring info for HTEX when monitoring enabled
         if self.monitoring_radio:
             msg = self.create_monitoring_info(status)
-            logger.debug("Sending message {} to hub from job status poller".format(msg))
+            logger.debug("Sending block monitoring message: %s", repr(msg))
             self.monitoring_radio.send((MessageType.BLOCK_INFO, msg))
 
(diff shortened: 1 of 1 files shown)


parsl/executors/status_handling.py
```diff
@@ -243,7 +243,7 @@ def send_monitoring_info(self, status: Dict) -> None:
         # Send monitoring info for HTEX when monitoring enabled
         if self.monitoring_radio:
             msg = self.create_monitoring_info(status)
-            logger.debug("Sending message {} to hub from job status poller".format(msg))
+            logger.debug("Sending block monitoring message: %s", repr(msg))
```
khk-globus, line 246: Might as well let the formatting take care of the representation, no?  Consider:
```diff
- logger.debug("Sending block monitoring message: %s", repr(msg))
+ logger.debug("Sending block monitoring message: %r", msg)
```
Can't always do this (given TOCTOU and if the argument is subsequently changed), but I think here should be safe, no?

khk-globus (approved):

parsl/executors/status_handling.py
benclifford: so updated in 7f5b6e7

PR #3529, opened 2024-07-16
Mark scale_out method as internal to BlockProviderExecutor

benclifford:
See PEP-8 https://peps.python.org/pep-0008/#descriptive-naming-styles

# Changed Behaviour

This should not change any behaviour.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -180,7 +180,7 @@ class BlockProviderExecutor(ParslExecutor):
         # Filters first iterable by bool values in second
         return list(compress(to_kill, killed))
 
-    def scale_out(self, blocks: int = 1) -> List[str]:
+    def _scale_out(self, blocks: int = 1) -> List[str]:
         """Scales out the number of blocks by "blocks"
         """
         if not self.provider:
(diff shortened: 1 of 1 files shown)


parsl/executors/status_handling.py
```diff
@@ -180,7 +180,7 @@ def _filter_scale_in_ids(self, to_kill, killed):
         # Filters first iterable by bool values in second
         return list(compress(to_kill, killed))
 
-    def scale_out(self, blocks: int = 1) -> List[str]:
+    def _scale_out(self, blocks: int = 1) -> List[str]:
```
rjmello, line 183: Do we not expect any valid scenarios in which a dependent project calls `.scale_out()` for initial blocks from outside the subclass?

benclifford: there's a historical jumble between should you call what is now `scale_out_facade` (which calls `scale_out` and does other things) or should you call `scale_out` avoiding the other stuff in scale_out_facade that maybe you want or maybe you don't want.

Historically these were methods in different classes: until PR #3352, scale_out_facade was actually part of the status handling code in `parsl/jobs/job_status_poller.py`.

This work is make there be a single entry point for scaling out - `scale_out_facade` - without option to avoid the random subset of behaviour that happens to live in scale_out_facade and not scale_out. Avoiding that random subset of behaviour is probably a bug.

The overarching PR for all of this work, which i am pulling out commit by commit as PRs, contains this follow on PR which removes `scale_out` entirely and you can only use `scale_out_facade` then:

https://github.com/Parsl/parsl/pull/3295/commits/20890ac66c9cc03783ff68a72122b43c8c53d766

rjmello: I'm sold; thanks for the explanation.

rjmello (approved):

PR #3531, opened 2024-07-16
Only query cuda devices if available_accelerators is set

cms21:
# Description

Only query cuda devices if `available_accelerators` is set

## Type of change

- Code maintenance/cleanup


benclifford (approved):

benclifford:
looks like there's a superficially unrelated CI failure - I'll poke at that

PR #3536, opened 2024-07-20
Remove unused dfk.memo_lookup_table attribute

benclifford:
This attribute is initialised sometimes, but not always, and is never read from. There's a similarly named attribute in Memoizer, so I think this was mistakenly introduced in commit
307b419dbcc847aeaf021f04c45b5149aa81d190.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Hmm; definitely more to go on this front as I look around for memoizing in general, but piecemeal "along the way" sweepings are probably the best approach at this time.

PR #3537, opened 2024-07-22
Update checkpoint docs to follow #1945 and #2667

benclifford:
## Type of change

### DIFF
--- a/docs/userguide/checkpoints.rst
+++ b/docs/userguide/checkpoints.rst
@@ -57,7 +57,7 @@ Changes inside the app, or by functions called by an app will not invalidate
 cached values.
 
 There are lots of other ways functions might be compared for equivalence,
-and ``parsl.dataflow.memoization.id_for_memo`` provides a hook to plug in
+and `parsl.dataflow.memoization.id_for_memo` provides a hook to plug in
 alternate application-specific implementations.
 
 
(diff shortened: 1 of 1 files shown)


PfeifferMicha:
Thanks for the quick fix and the link to parsl.dataflow.memoization.id_for_memo, that's very useful... I hadn't realized before that the id_for_memo could also be overwritten for functions!

docs/userguide/checkpoints.rst
```diff
@@ -49,15 +49,17 @@ during development. Using app caching will ensure that only modified apps are re
...
 
-However, Parsl does not traverse the call graph of the app function,
-so changes inside functions called by an app will not invalidate
+Changes inside the app, or by functions called by an app will not invalidate
 cached values.
 
+There are lots of other ways functions might be compared for equivalence,
+and ``parsl.dataflow.memoization.id_for_memo`` provides a hook to plug in
+alternate application-specific implementations.
+
```
khk-globus, line 62: Any reason not to make this a clickable link?  (That is, a single `` ` `` backtick.)

khk-globus (approved):

docs/userguide/checkpoints.rst
benclifford: that was my intention but I copied it from some other location which is also broken (which I will also fix). Long live intuitive markup languages.

Issue #3540, opened 2024-07-25
parsl/tests/test_serialization/test_3495_deserialize_managerlost.py suffers from issue #3016


benclifford:
This test does not protect itself against enemy managers - see PR #3019 for more details. This is resulting in hangs that I see maybe once per week in CI. This issue is a reminder about that.

Things to think about:
* what othre test is leaving managers behind and why isn't that being detected/stopped?
* turn on encryption for this test so that it is solved in the same way as #3019 - see PR #3030
* turn on encryption by default for all uses of htex, with insecure mode being an explicit user choice.

cc for interest @rjmello who made PR #3030 - the encryption PR, and @khk-globus who has an especial interest in test reliability

PR #3543, opened 2024-07-25
Accept multi-token interchange launch commands 

rjmello:
# Description

We now accept a string or a list of strings for `interchange_launch_cmd` to support multi-token commands.

We also no longer encode the interchange launch command before passing it to `subprocess.Popen()` because strings are sufficient.

## Type of change

- Bug fix

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -548,7 +548,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
 
         config_pickle = pickle.dumps(interchange_config)
 
-        self.interchange_proc = subprocess.Popen(self.interchange_launch_cmd.encode("utf-8"), stdin=subprocess.PIPE)
+        self.interchange_proc = subprocess.Popen(self.interchange_launch_cmd, stdin=subprocess.PIPE)
         stdin = self.interchange_proc.stdin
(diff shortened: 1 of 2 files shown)


benclifford:
this is repeatedly failing in CI with failures that look race-condition like to me - for example, like #3540

Maybe it's bad luck, maybe this change happens to stir up those race conditions a bit more.

benclifford:
I don't have any strong opinions either way -  security wise from a pure Parsl perspective the config is trusted, although I appreciate that in Globus Compute multi-user land that might not be so (and there you have to make your own arguments).

I do feel that the Parsl test suite should be resilient to this kind of change though - it's shouldn't be breaking just because of some irrelevant process related administrivia like this.

rjmello:
I agree, but I think we should address that longstanding issue in a subsequent PR.

benclifford:
could make this only take a sequence of strings, and remove the before-this-PR accept-one-string behaviour? I don't think backwards compatibility matters too much here and it would remove the complexity of having two options option that isn't very interesting and is probably confusing for a user trying to specify "my-interchange --myopt" as a string, which doesn't work but does typecheck.

rjmello:
If we aren't worried about backwards compatibility, then I definitely agree. Just pushed an update.

benclifford:
i don't think it's a concern -- this only appears in #3514 released earlier this month and I haven't heard of or imagined anyone wanting to use it apart from the specific use case you are trying to address with Globus Compute.

PR #3547, opened 2024-07-26
Add Modular Manager Selector Interface

matthewc2003:
# Description

Added a Manager Selector interface which allows users to choose an algorithm to sort the interesting managers. This will allow for flexible testing and implementation of manager selection strategies to optimize efficiency of the interchange. 

# Changed Behaviour

When using defining a config for the HTEX, users can now optionally specify a ManagerSelector to change the strategy. Default algorithm is the original random shuffle.

# Fixes

Further development of project outlined by [issue 3323](https://github.com/Parsl/parsl/issues/3323)

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -265,7 +265,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
                  worker_logdir_root: Optional[str] = None,
                  enable_mpi_mode: bool = False,
                  mpi_launcher: str = "mpiexec",
-                 manager_selector: Optional[ManagerSelectorBase] = ManagerSelectorRandom(),
+                 manager_selector: ManagerSelectorBase = ManagerSelectorRandom(),
                  block_error_handler: Union[bool, Callable[[BlockProviderExecutor, Dict[str, JobStatus]], None]] = True,
                  encrypted: bool = False):
 
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -20,6 +20,10 @@ from parsl.data_provider.staging import Staging
 from parsl.executors.errors import BadMessage, ScalingFailed
 from parsl.executors.high_throughput import zmq_pipes
 from parsl.executors.high_throughput.errors import CommandClientTimeoutError
+from parsl.executors.high_throughput.manager_selector import (
+    ManagerSelectorBase,
+    ManagerSelectorRandom,
+)
 from parsl.executors.high_throughput.mpi_prefix_composer import (
     VALID_LAUNCHERS,
     validate_resource_spec,
@@ -261,6 +265,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
                  worker_logdir_root: Optional[str] = None,
                  enable_mpi_mode: bool = False,
                  mpi_launcher: str = "mpiexec",
+                 manager_selector: Optional[ManagerSelectorBase] = ManagerSelectorRandom(),
                  block_error_handler: Union[bool, Callable[[BlockProviderExecutor, Dict[str, JobStatus]], None]] = True,
                  encrypted: bool = False):
 
@@ -276,6 +281,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         self.prefetch_capacity = prefetch_capacity
         self.address = address
         self.address_probe_timeout = address_probe_timeout
+        self.manager_selector = manager_selector
         if self.address:
             self.all_addresses = address
         else:
@@ -544,6 +550,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
                               "poll_period": self.poll_period,
                               "logging_level": logging.DEBUG if self.worker_debug else logging.INFO,
                               "cert_dir": self.cert_dir,
+                              "manager_selector": self.manager_selector,
                               }
 
         config_pickle = pickle.dumps(interchange_config)
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -21,8 +21,8 @@ from parsl.executors.errors import BadMessage, ScalingFailed
 from parsl.executors.high_throughput import zmq_pipes
 from parsl.executors.high_throughput.errors import CommandClientTimeoutError
 from parsl.executors.high_throughput.manager_selector import (
-    ManagerSelectorBase,
-    ManagerSelectorRandom,
+    ManagerSelector,
+    RandomManagerSelector,
 )
 from parsl.executors.high_throughput.mpi_prefix_composer import (
     VALID_LAUNCHERS,
@@ -265,7 +265,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
                  worker_logdir_root: Optional[str] = None,
                  enable_mpi_mode: bool = False,
                  mpi_launcher: str = "mpiexec",
-                 manager_selector: ManagerSelectorBase = ManagerSelectorRandom(),
+                 manager_selector: ManagerSelector = RandomManagerSelector(),
                  block_error_handler: Union[bool, Callable[[BlockProviderExecutor, Dict[str, JobStatus]], None]] = True,
                  encrypted: bool = False):
 
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -18,7 +18,7 @@ from parsl import curvezmq
 from parsl.app.errors import RemoteExceptionWrapper
 from parsl.executors.high_throughput.errors import ManagerLost, VersionMismatch
 from parsl.executors.high_throughput.manager_record import ManagerRecord
-from parsl.executors.high_throughput.manager_selector import ManagerSelectorBase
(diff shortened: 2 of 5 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -160,6 +161,9 @@ def __init__(self,
 
         self.heartbeat_threshold = heartbeat_threshold
 
+        assert manager_selector is not None, "Undefined manager_selector"
```
benclifford, line 164: this probably shouldn't be allowed to be None here -- maybe you were getting a mypy error?

parsl/executors/high_throughput/interchange.py
```diff
@@ -53,6 +53,7 @@ def __init__(self,
                  logging_level: int,
                  poll_period: int,
                  cert_dir: Optional[str],
+                 manager_selector: Optional[ManagerSelectorBase] = None,
```
benclifford, line 56: have a look at PR #3465 and reconsider this line in that context

parsl/executors/high_throughput/manager_selector.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ManagerSelectorBase:
+
+    def sort_managers(self, ready_managers: Dict[bytes, ManagerRecord], manager_list: List[bytes]) -> List[bytes]:
+        raise NotImplementedError
+
+
+class ManagerSelectorRandom(ManagerSelectorBase):
+
+    def __init__(self):
```
benclifford, line 16: maybe you can omit this __init__ entirely?

parsl/executors/high_throughput/interchange.py
```diff
@@ -485,8 +489,8 @@ def process_tasks_to_send(self, interesting_managers: Set[bytes]) -> None:
             interesting=len(interesting_managers)))
 
         if interesting_managers and not self.pending_task_queue.empty():
-            shuffled_managers = list(interesting_managers)
-            random.shuffle(shuffled_managers)
+            unsorted_managers = list(interesting_managers)
```
benclifford, line 492: you could maybe avoid doing (effectively) a copy here and then another copy inside the manager by defining the interface to be that sort_managers will never modify its input - eg make the unsorted_managers parameter be labelled as `Iterable[bytes]` so type checking will push back on implementations that try to do more than just iterate over it, and change the copy of the list inside the random implementation to make a list from that iterable. (which would actually be a set in practice here)

This probably won't affect performance much in many cases, but with a thousand managers in the list it might be relevant.

parsl/executors/high_throughput/interchange.py
matthewc2003: Cutting this out in conjunction with removing the default value from the interchange init.

parsl/executors/high_throughput/interchange.py
matthewc2003: Will remove this default value.

parsl/executors/high_throughput/manager_selector.py
```diff
@@ -0,0 +1,22 @@
+import copy
+import random
+from typing import Dict, List
+
+from parsl.executors.high_throughput.manager_record import ManagerRecord
+
+
+class ManagerSelectorBase:
```
benclifford, line 8: have a look at how abstract base class is used elsewhere to force users to implement methods / prohibit instantiating classes that aren't subclassed as much as they need to.

this could force some errors when people are developing new managers to occur when the config object is created (at the point they define their configuration, early on in their run) rather than this raising a `NotImplementedError` inside the interchange code where it is harder to understand.

parsl.executors.base would give you an example of `ABCMeta` and `@abstractmethod`

parsl/executors/high_throughput/manager_selector.py
matthewc2003: Will do

parsl/executors/high_throughput/interchange.py
matthewc2003: Yeah, so I am getting an mypy error without this assertion, how should I address this?

parsl/tests/test_mpi_apps/test_mpiex.py
```diff
@@ -43,16 +43,17 @@ def test_init():
     """Ensure all relevant kwargs are copied over from HTEx"""
 
     new_kwargs = {'max_workers_per_block'}
-    excluded_kwargs = {'available_accelerators', 'enable_mpi_mode', 'cores_per_worker', 'max_workers_per_node',
-                       'mem_per_worker', 'cpu_affinity', 'max_workers'}
+    excluded_mpix_kwargs = {'available_accelerators', 'enable_mpi_mode', 'cores_per_worker', 'max_workers_per_node',
+                            'mem_per_worker', 'cpu_affinity', 'max_workers'}
+    excluded_htex_kwargs = {'manager_selector'}
```
benclifford, line 48: what are you claiming about behaviour with this?

before this PR, excluded_kwargs was used as part of assertions like:
"the mpi executor args must look something like:  the htex args minus the excluded_kwargs plus the new_args"

(so excluded_args is the arguments htex should have that mpix does not have, and new_kwargs is the arguments mpix should have that htex does not have)

so whats being implemented by this test you could visualise as a big venn diagram of who has which kwargs:

one circle is the htex args, one circle is the mpix args, with most args being in the central overlap, and new_kwargs and excluded_kwargs being the two side bits.

with that as the basic test model, what are you expressing by adding in excluded_htex_kwargs?

parsl/executors/high_throughput/executor.py
```diff
@@ -20,6 +20,10 @@
 from parsl.executors.errors import BadMessage, ScalingFailed
 from parsl.executors.high_throughput import zmq_pipes
 from parsl.executors.high_throughput.errors import CommandClientTimeoutError
+from parsl.executors.high_throughput.manager_selector import (
+    ManagerSelectorBase,
+    ManagerSelectorRandom,
```
benclifford, line 25: this naming could be more in line with the rest of the parsl abstractions:

The base names the think (`ExecutionProvider`, `Launcher`, `Channel`) - so that when you write code/docs you say "this argument is a `Launcher` and this other argument is a `Channel` or `provider: ExecutionProvider" - using `ManagerSelector` without `Base` would fit with that.

The other abstractions in Parsl put the adjective-like bit at the start of name for the concrete user facing instances so maybe this should be `RandomManagerSelector` to be consistent with that - for example:
```
parsl/providers/__init__.py:__all__ = ['LocalProvider',
parsl/providers/__init__.py:           'CobaltProvider',
parsl/providers/__init__.py:           'CondorProvider',
parsl/providers/__init__.py:           'GridEngineProvider',
parsl/providers/__init__.py:           'SlurmProvider',
parsl/providers/__init__.py:           'TorqueProvider',
parsl/providers/__init__.py:           'LSFProvider',
parsl/providers/__init__.py:           'AdHocProvider',
parsl/providers/__init__.py:           'PBSProProvider',
parsl/providers/__init__.py:           'AWSProvider',
parsl/providers/__init__.py:           'GoogleCloudProvider',
parsl/providers/__init__.py:           'KubernetesProvider',
parsl/providers/__init__.py:           'AzureProvider']
```

```
class SingleNodeLauncher(Launcher):
class GnuParallelLauncher(Launcher):
class MpiExecLauncher(Launcher):
class MpiRunLauncher(Launcher):
class SrunLauncher(Launcher):
class SrunMPILauncher(Launcher):
    Use this launcher instead of SrunLauncher if each block will execute multiple MPI applications
class AprunLauncher(Launcher):
class JsrunLauncher(Launcher):
```

parsl/tests/test_mpi_apps/test_mpiex.py
matthewc2003: That's on me. It wasn't clear to me whether excluded_args were the extra args for HTEX or mpi executor. Reverting changes and adding "manager_selector" to excluded_args.

parsl/executors/high_throughput/executor.py
matthewc2003: New commit changes the name of the ABC and its subclasses to the suggested naming scheme.

benclifford:
ok this is looking good.

Bonus points: what happens when a manager selector raises an exception?  That basically doesn't happen now in practice because shuffling a list has few opportunities for breakage - but as the plugins get more complicated, they're going to be more likely to raise an exception.

For example, make a test manager selector that *always* raises an exception rather than shuffling the list. What happens when I run a task?

Usually given how parsl is historically implemented that sort of stuff results in two things:

i) a hang forever (bad) - possibly without even a suitable log message anywhere.

ii) the task failing all the way back to the user (good) - and potentially *all* tasks failing all the way back to the user forever (i.e. a proper safe shutdown) which is also ok behaviour.

As a similar example, here's a test (parsl/tests/test_python_apps/test_memoize_bad_id_for_memo.py)  for the memoizer/checkpointing plugin API which makes a data type `FailingMemoizable` that when we try to checkpoint it, always raises an exception, and a corresponding test that when we try to use it as a parameter to an app, we check that the failure gets reported all the way back as a task result in `test_python_failing_memoizer`

Have a look in (for example) parsl/tests/test_mpi_apps/test_mpiex.py to see some tests which define a Parsl config just for that one test file: you need to define `local_config` to return the config needed for this test file, and to mark your test(s) with @parsl.mark.local to indicate they should be run with that local configuration. That would allow you to implement a test with a deliberately always-broken selector (and incidentally would also demonstrate how to define and use new manager selectors in the config API)

parsl/executors/high_throughput/interchange.py
benclifford: this was fixed by removing optionality...

benclifford (approved):

PR #3549, opened 2024-07-27
Add type annotation and error log to _filter_scale_in_ids

benclifford:
Especially this log message is intended to help user understanding when Parsl is not scaling in as they expected - before this PR, any blocks marked as not-scaled-in were not reported to the user (perhaps on the assumption that it might work next time round?)

# Changed Behaviour

Slightly more logging in the case of providers not killing blocks.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/status_handling.py
+++ b/parsl/executors/status_handling.py
@@ -167,10 +167,14 @@ class BlockProviderExecutor(ParslExecutor):
     def provider(self):
         return self._provider
 
-    def _filter_scale_in_ids(self, to_kill, killed):
+    def _filter_scale_in_ids(self, to_kill: Sequence[Any], killed: Sequence[bool]) -> Sequence[Any]:
         """ Filter out job id's that were not killed
         """
         assert len(to_kill) == len(killed)
+
+        if False in killed:
+            logger.warning(f"Some jobs were not killed successfully: to_kill list is {to_kill}, killed list is {killed}")
+
         # Filters first iterable by bool values in second
         return list(compress(to_kill, killed))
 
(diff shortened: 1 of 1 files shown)


parsl/executors/status_handling.py
```diff
@@ -167,10 +167,14 @@ def tasks(self) -> Dict[object, Future]:
...
         return self._provider
 
-    def _filter_scale_in_ids(self, to_kill, killed):
+    def _filter_scale_in_ids(self, to_kill: Sequence[Any], killed: Sequence[bool]) -> Sequence[Any]:
         """ Filter out job id's that were not killed
         """
         assert len(to_kill) == len(killed)
+
+        if False in killed:
+            logger.warning(f"Some jobs were not killed successfully: to_kill list is {to_kill}, killed list is {killed}")
```
khk-globus, line 176: Are `to_kill` and `killed` `zip()`pable?  If so, consider spelling out the ones that succeeded and failed.  Something like:
```python
if False in killed:
    s = (j for j, s in zip(to_kill, killed) if s)  # "success"
    f = (j for j, s in zip(to_kill, killed) if not s)  # "fail"
    logger.warning(
      "Failed to stop all jobs."
      "\n  Successfully stopped: {', '.join(s)}"
      "\n        Failed to stop: {', '.join(f)}"
    )
```
Format as you will, of course, the but the point is to save the user some log-trawling thinking but spelling it out just a smidgen more.

khk-globus (approved):
Minor formatting suggestion, but good general thought-process.

parsl/executors/status_handling.py
benclifford: yeah that is a better way to present the info. I ended up not trying to do join style formatting because it looks weird when the success list is empty, and rather than going down a rat-hole of dealing with that case, I let Python's default list rendering do it.

PR #3551, opened 2024-07-27
Remove unused codepath from executor scale-out

benclifford:
block_ids is populatd by _scale_out which always returns a list, according to its type signature. So the `None` codepath should not ever be reached.

mypy agrees: after adding an else clause (to act on the `None` case), mypy then marks the added clause as unreachable.

## Type of change

- Code maintenance/cleanup


khk-globus (commented):
Is it possible that `self._scale_out(n)` returns an _empty_ list?  And are `.send_monitoring_info()` and `.update()` robust to that?

benclifford:
Without digging too hard, i don't think any of the code requires that list to contain any entries - (in different phrasing, there shouldn't be anything that gives the 1st entry of the list some privileged status).

khk-globus (approved):

PR #3556, opened 2024-07-31
Factor interchange monitoring code into a ZMQRadioSender

benclifford:
From an interchange perspective: this is a refactoring intended to clarify that the interchange isn't doing anything special wrt. monitoring messages and that it can send monitoring messages in the same way that remote workers can.

From a monitoring perspective: this pulls ZMQ sender code out of the interchange and puts it in a place that is more natural for ongoing development. For example, a potential future use with Work Queue and Task Vine is that workers would also benefit from using ZMQ to send monitoring messages.

In some potential use cases, it might be desirable to configure the radio used by the interchange instead of the hard-coded ZMQRadio. On-going work in draft PR #3315 addresses configuration of different types of radio and that work should be relevant here too.

# Changed Behaviour

This should not change behaviour

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -20,6 +20,7 @@ from parsl.executors.high_throughput.errors import ManagerLost, VersionMismatch
 from parsl.executors.high_throughput.manager_record import ManagerRecord
 from parsl.executors.high_throughput.manager_selector import ManagerSelector
 from parsl.monitoring.message_type import MessageType
+from parsl.monitoring.radios import MonitoringRadioSender, ZMQRadioSender
 from parsl.process_loggers import wrap_with_logs
 from parsl.serialize import serialize as serialize_object
 from parsl.utils import setproctitle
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -333,9 +324,14 @@ def start(self) -> None:
...
         signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
-        logger.info("Incoming ports bound")
+        logger.info("Starting main interchange method")
 
-        hub_channel = self._create_monitoring_channel()
+        if self.hub_address is not None and self.hub_zmq_port is not None:
+            logger.debug("Creating monitoring radio")
+            monitoring_radio = ZMQRadioSender(self.hub_address, self.hub_zmq_port)
+            logger.debug("Created monitoring radio")
```
khk-globus, line 332: I'm sure we've `.debug()`ed the `hub_address` and `hub_zmq_port` elsewhere ... would it make sense to `.debug()` them here as well?  I thinking in terms of an edge case where they, for whatever reason, _don't_ line up with the other log lines.

khk-globus (approved):

parsl/executors/high_throughput/interchange.py
benclifford: I'd also like to log this inside the ZMQRadioSender itself, but multiprocessing-driven awkward log configuration means that parsl.* logs don't go into the interchange log (or anywhere since #3463 - prior to that they appeared inside the main parsl.log even though they came from the interchange process!)

PR #3558, opened 2024-07-31
Split monitoring router into two radio-specific receiver threads

benclifford:
This is part of two ongoing strands of development:

* Preparation for arbitrary monitoring radio receivers, where UDP is not special compared to other methods. This code separates out the UDP specific code making it easier to move around later (see development in PR #3315)

* Avoiding poll-one-thing, poll-another-thing tight polling loops in favour of multiple blocking loops. The two router receiver sections used to run in a single tight non-blocking loop, each component waiting loop_freq (= 10 ms) for something to happen. After this PR, the separated loops are more amenable to longer blocking times - they only need to discover when exit event is set which probably can be more on the order of 1 second.

# Changed Behaviour

no functional changes - time of delivery of monitoring messages will be very slightly different, which in the race-condition prone parsl environment might be relevant

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/router.py
+++ b/parsl/monitoring/router.py
@@ -5,6 +5,7 @@ import os
 import pickle
 import queue
 import socket
+import threading
 import time
 from multiprocessing.synchronize import Event
 from typing import Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


benclifford:
failed in CI with a shutdown-of-router related error - i'll investigate more

khk-globus (approved):

PR #3561, opened 2024-08-01
Make monitoring hub start into its own Exception

benclifford:
This is in line with the principle that Parsl exceptions should all be subclasses of ParslError.

# Changed Behaviour

Anyone catching this exception will have to catch it differently

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Good call.  Does this need any documentation added?  ... probably not; this is "just" monitoring.

PR #3565, opened 2024-08-02
Fix pytest caplog heisenbug introduced in PR #3559

benclifford:
Prior to this PR, this usage of caplog is dependent on the level of the root logger, which is not set by this test and so ends up being dependent on which tests have run before: sometimes the INFO logs output by htex.shutdown are not captured by caplog.

This PR explicitly tells caplog to get at least INFO level logs, to capture the expected messages.

# Changed Behaviour

eliminates a bug in test code

## Type of change

- Bug fix


khk-globus (approved):
Heh; I just solved this same problem in another project.  :partying_face:

Issue #3515, opened 2024-07-07
Remove channels


benclifford:
# Overview

Parsl has a facility called Channels intended to support use cases such as running a workflow on your laptop but with task execution on a supercomputer. This facility never evolved beyond prototype stage, and effort from University of Chicago and University of Illinois was rapidly distracted into the funcX project which aims (with substantially more code and developer time) to provide roughly the same facility - now named Globus Compute and developed by professional programmers rather than an academic research team.

The presence of channels as quasi-abandonware inside the modern Parsl codebase is a painful drain on both user and developer time: Users continue to be fooled into believing that this abandoned prototype facility, rather than the results of the funcX project, should be used for remote execution use cases. Developer time continues to be taken up dealing with the intricate lacing of channel handling through the core codebase.

Channels should be removed from Parsl, with the default behaviour of `LocalChannel` becoming the only behaviour.

# Proposed timeline

I propose this timeline as a default, unless a consensus forms for some other timeline. In the absence of any other consensus forming in the comments of this issue, I'll make this timeline happen.

now .. 7th August 2024: people paying attention to issues get to comment on this issue, including offering entirely different alternative paths forward
7th August 2024: channel code is marked as going away in the codebase, in some way that is hard to avoid for users (for example, renaming all the user-facing channel classes). Users should be encouraged to visit this issue and comment, and to seek alternatives such as Globus Compute.
7th November 2024: channel code is removed from Parsl

# Why can't this become an "abandoned component sitting in a directory"

The channel facility is not componentised - it contains prototype quality code inside the core of Parsl

# What about the AdHocProvider which attempts to use multiple channels to run on a cluster which has no resource manager?

This should go away too - without channels it is of little use. Users should be encouraged to find other ways to run a cluster without any resource manager. It should not be parsl's job to manage this aspect of a user's cluster.

# What if someone wants to take on Channel tech sponsorship?

This could happen, but I think it's unlikely.

The tech sponsor would need to be responsible for tidying up the architecture and implementing those fixes, with noise such as "why are channels initialised or not based on the presence of `script_dir` attribute on a provider?", test implementations and fixups such as testing parsl without shared file systems in a channel-like environment, and substantially more fundamental questions such as "how is htex supposed to work when channels do not provide a channel for htex network connections?".

[...]

PR #3569, opened 2024-08-07
Aggressively deprecate Channels and AdHocProvider

benclifford:
This is described in issue #3515.

Issue #3515 has not received any comments arguing in favour of retaining channels and the AdHocprovider, and several in support of removing them,
 and so this PR takes a heavy handed approach that is well on the way to
the end goal of #3515 of deleting channels and the AdHocProvider entirely:

Channels except LocalChannel are renamed, so that any users of other channels will have to make a change to their code and actively observe the word "Deprecated" in the name.

The AdHocProvider is renamed in the same way.

Most documentation (but not docstrings) about channels and the ad-hoc provider is removed or replaced with a link to issue

Tests which much be manually run, and so in effect are never run and shouldn't be expected to work now - parsl/tests/manual_tests and parsl/tests/integration/ - are deleted rather than fixed to follow the above naming change.

The tests for SSH channels and the AdHocProvider that run in CI are modified to continue passing.

Exposure of the deprecated components via top level parsl.providers and parsl.channels re-export is removed. To use this components, the deprecated modules must be imported directly.

# Changed Behaviour

Users who want channels will have to acknowledge by modifying their code that they understand that these components are very strongly deprecated.

## Type of change

- Feature removal


docs/historical/changelog.rst
khk-globus, line 1: Good change!  I don't know that I would have been thinking in terms of changelog links.  Nice.

docs/userguide/execution.rst
```diff
@@ -47,8 +47,7 @@ Parsl currently supports the following providers:
 7. `parsl.providers.AWSProvider`: This provider allows you to provision and manage cloud nodes from Amazon Web Services.
 8. `parsl.providers.GoogleCloudProvider`: This provider allows you to provision and manage cloud nodes from Google Cloud.
 9. `parsl.providers.KubernetesProvider`: This provider allows you to provision and manage containers on a Kubernetes cluster.
-10. `parsl.providers.AdHocProvider`: This provider allows you manage execution over a collection of nodes to form an ad-hoc cluster.
-11. `parsl.providers.LSFProvider`: This provider allows you to schedule resources via IBM's LSF scheduler.
```
khk-globus, line 51: I had thought that both Markdown and reStructured text followed the utility of just enumerating with the number 1:
```reST
1. list item 1
1. list item 2
1. list item 3
...
1. list item N
```
Both monotonically increasing and single numbers work, but for just instances like this, it's a shame the original author of this section didn't just use `1.`

:shrug:

docs/userguide/plugins.rst
```diff
@@ -34,10 +34,9 @@ add on any wrappers that are needed to launch the command (eg srun inside
 slurm). Providers and launchers are usually paired together for a particular
 system type.
 
-A `Channel` allows the commands used to interact with an `ExecutionProvider` to be
-executed on a remote system. The default channel executes commands on the
-local system, but a few variants of an `parsl.channels.SSHChannel` are provided.
-
+Parsl also has a deprecated ``Channel`` abstraction.  See
+`issue 3515 <https://github.com/Parsl/parsl/issues/3515>`_
+for further discussion.
```
khk-globus, line 39: I'm wondering if it would be smarter to go all the way and just remove the references in the documentation entirely.  That is, nothing will have changed functionally, so if folks still want to use it, they can reference an older document in the interim while complaining to us that we're removing a crucial element.  Meanwhile, under the expectation that ~no one _is_ using it, this saves work in the future&nbsp;&mdash;&nbsp;do the work once and completely.  If removing it from the title of this section, this paragraph becomes a bit of a non-sequitur.

parsl/channels/__init__.py
```diff
@@ -1,7 +1,4 @@
 from parsl.channels.base import Channel
 from parsl.channels.local.local import LocalChannel
-from parsl.channels.oauth_ssh.oauth_ssh import OAuthSSHChannel
-from parsl.channels.ssh.ssh import SSHChannel
-from parsl.channels.ssh_il.ssh_il import SSHInteractiveLoginChannel
 
-__all__ = ['Channel', 'SSHChannel', 'LocalChannel', 'SSHInteractiveLoginChannel', 'OAuthSSHChannel']
+__all__ = ['Channel', 'LocalChannel']
```
khk-globus, line 4: Heh; a LocalChannel is a "fake" channel, eh?  :innocent:

benclifford: `LocalChannel` gets used both as a default (which can be edited by this PR) and as an explicit fake-default by a bunch of people (who copied it from examples) which this PR cannot edit.

That explicit fake-default path will have to be changed by users eventually, but it's more likely to affect users who don't care about channels - so I was avoiding that in this PR and leaving that inevitable pain till later...

The main point of this PR is to flush out the people who *do* care about (non-local) channels.

docs/userguide/plugins.rst
benclifford: mostly I was erring on the side of putting in links to the issue in lots of places because of previous people popping up and saying "but i like channels".

it's possible that actually those people don't exist any more - but in that case, we could be deleting channels entirely right now.

docs/historical/changelog.rst
benclifford: the doc build fails if not, so it's forced by the test suite

PR #3572, opened 2024-08-08
Correct and check types on monitoring router and database processes

benclifford:
# Description

Prior to this PR, the startup code for the monitoring router and database processes had type annotations on queues; but these types were not checked, and were incorrect - they were labelled process-local `Queue` instead of multiprocessing queues. This did not cause much trouble execution- and mypy-wise, as the interfaces of those two classes are similar enough, but it is confusing to read in a part of the codebase that is already confusing (that confusion is probably what lead to the incorrect annotations in the first place...)

They were not checked because the informal policy of "internal stuff is checked with mypy, external interfaces are checked with typeguard" works badly here:

The startup methods are launched using multiprocessing.Process, and function invocations are not type-checked by mypy across a multiprocessing Process constructor.

# Changed Behaviour

This PR introduces typeguard decorators onto the router and database start methods so that this internal checking happens at runtime. This consequently reveals that the type annotations of these methods are incorrect, and so this PR makes those consequential changes.

Further, generic types (`Queue[MessageType]`) are not supported on multiprocessing.Queues before Python 3.12 - so those generic indices are removed from the type annotations. That is unfortunate and weakens in-process static verification - but they could be re-introduced after Parsl drops Python 3.11 support (around 2027 in the present informal support policy)

## Type of change

- Bug fix
- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import multiprocessing.queues as mpq
 import os
 import queue
 import threading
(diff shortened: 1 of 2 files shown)


benclifford:
waiting on #3573 for a bit of syntactic tidyup

parsl/monitoring/db_manager.py
```diff
@@ -719,11 +722,12 @@ def close(self) -> None:
...
-                priority_msgs: "queue.Queue[TaggedMonitoringMessage]",
-                node_msgs: "queue.Queue[MonitoringMessage]",
-                block_msgs: "queue.Queue[MonitoringMessage]",
-                resource_msgs: "queue.Queue[MonitoringMessage]",
+@typeguard.typechecked
+def dbm_starter(exception_q: mpq.Queue,
+                priority_msgs: mpq.Queue,
+                node_msgs: mpq.Queue,
+                block_msgs: mpq.Queue,
+                resource_msgs: mpq.Queue,
```
khk-globus, line 730: Per the Python 3.12 gate for generic types, I ~just learned from this PR~ gather that `mpq.Queue[TaggedMonitoringMessage]` won't work; is the same also true for `[MonitoringMessage]` and the tuple of strings?  Seems a partial effort would be of some utility here, even if full coverage is 3+ years away.

Issue #3519, opened 2024-07-10
Raise an error when a non-MPI-mode HTEX is passed a resource specification


chris-janidlo:
**Is your feature request related to a problem? Please describe.**
Globus Compute would like to warn users when they try to do MPI stuff on non-MPI Compute endpoints. Currently the only place that would raise an error is when workers try to access `PARSL_MPI_PREFIX`, and it would be nice to catch that even earlier.

**Describe the solution you'd like**
Add a check to `validate_resource_spec` which raises an error if `is_mpi_enabled` is false and any resource spec is passed in.

**Describe alternatives you've considered**
Compute could check this at the engine level, before submitting to the HTEX

**Additional context**
https://github.com/Parsl/parsl/issues/3498 is asking for almost the opposite, but it sounds like @benclifford is on the side of more restrictiveness here

benclifford:
This is implemented in #3582

PR #3575, opened 2024-08-08
Add priority queue functionality to HTEX

matthewc2003:
# Description

Adds second queue for pending task queue in the interchange. Used to prioritize certain user-labeled tasks when distributing tasks to managers.  

# Changed Behaviour

Users can now use the parsl_resource_specification keyword in python apps to specify a 'priority'. Lower value is higher priority, tasks with higher priority will be sent out first. 

Users can also specify a 'queue_threshold' in the HTEX config to decide the cutoff value for tasks to go to the priority queue or the general FIFO queue.

# Fixes

Part of #3323 work

## Type of change

- New feature


parsl/executors/high_throughput/executor.py
```diff
@@ -659,8 +662,20 @@ def submit(self, func, resource_specification, *args, **kwargs):
         Returns:
               Future
         """
-
-        validate_resource_spec(resource_specification, self.enable_mpi_mode)
+        if self.enable_mpi_mode:
+            validate_resource_spec(resource_specification, self.enable_mpi_mode)
+        else:
+            if resource_specification and isinstance(resource_specification, dict):
```
benclifford, line 668: this all belongs in validate_resource_spec()

try implementing https://github.com/Parsl/parsl/issues/3519 as a separate PR first to get the structure in place in that function for "mpi vs non-mpi" resource validation (with, I guess, no resource spec allowed for non-mpi mode?) and you can get a parsl issue fixed as a nice side effect

parsl/executors/high_throughput/interchange.py
```diff
@@ -131,7 +133,8 @@ def __init__(self,
         self.hub_address = hub_address
         self.hub_zmq_port = hub_zmq_port
 
-        self.pending_task_queue: queue.Queue[Any] = queue.Queue(maxsize=10 ** 6)
+        self.priority_pending_task_queue: SortedList[Any] = SortedList(key=lambda msg: -msg['resource_spec']['priority'])
```
benclifford, line 136: my gut says to avoid two queues and have a single queue. tasks without any priority effectively receive an infinitely-weak priority in your implementation, I think, and you can implement that same behaviour with a single priority queue?

parsl/executors/high_throughput/interchange.py
```diff
@@ -253,7 +276,7 @@ def _command_server(self) -> NoReturn:
                 command_req = self.command_channel.recv_pyobj()
                 logger.debug("Received command request: {}".format(command_req))
                 if command_req == "OUTSTANDING_C":
-                    outstanding = self.pending_task_queue.qsize()
+                    outstanding = self.priority_pending_task_queue.__len__() + self.general_pending_task_queue.qsize()
```
benclifford, line 279: use `len(priority_pending_task_queue)` rather than invoking __len__

PR #3585, opened 2024-08-15
Fixed SLURM Provider to actually set ntasks-per-node (Currently set t…

JamesOBrien2:
…o cpus_per_task

# Description

Edited Slurm.py to set tasks-per-node to {cpus_per_task}.

# Changed Behaviour

SLURM Provider

# Fixes

Fixes SLURM provider not setting the ntasks-per-node correctly. Ideally however this would be set through a variable in the provider itself.

## Type of change

- Bug fix

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -154,9 +154,14 @@ class MonitoringHub(RepresentationMixin):
         self.router_exit_event = Event()
 
         self.router_proc = ForkProcess(target=router_starter,
-                                       args=(comm_q, self.exception_q, self.priority_msgs, self.node_msgs,
-                                             self.block_msgs, self.resource_msgs, self.router_exit_event),
-                                       kwargs={"hub_address": self.hub_address,
+                                       kwargs={"comm_q": comm_q,
+                                               "exception_q": self.exception_q,
+                                               "priority_msgs": self.priority_msgs,
+                                               "node_msgs": self.node_msgs,
+                                               "block_msgs": self.block_msgs,
+                                               "resource_msgs": self.resource_msgs,
+                                               "exit_event": self.router_exit_event,
+                                               "hub_address": self.hub_address,
                                                "udp_port": self.hub_port,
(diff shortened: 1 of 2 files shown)


benclifford:
@JamesOBrien2 this looks like it's a PR to fix a specific problem that you are having at a specific site, i guess? (which wouldn't be too surprising given how varied slurm installations behave)

There's a task_per_node variable right there, and maybe that's the value that should be used for the tasks per node setting? and maybe this doesn't need setting only when cores_per_node is specified?

But I'm not super clear on the problem you're trying to solve - i.e. what the wrong behaviour at your site is. Can you elaborate a bit?

JamesOBrien2:
Sorry you're right, the change provided is very one-dimensional / site-specific. The issue we are having is that tasks_per_node appears to be hard-coded to equal 1. As we are using SLURM it is forcing the --ntasks-per-node to equal 1. I believe this is originating from the submit() function, and when it is called in the Executor (base.py / status_handling.py). A variable tasks_per_node defaulted to 1 could fix this, however this may be a misuse as I am not entirely familiar with the Parsl  executors. If you would instead recommend an alternative use or fix please let me know! Hope this made more sense.

benclifford:
that tasks_per_node parameter in the submit API is really not used any more, and is from earlier.

the usual model is something like:

one batch job runs across 1 or more nodes.

each node is dedicated entirely to that batch job.

the batch job runs one copy of the Parsl `process_worker_pool.py` worker script (this is what the provider layer means by tasks-per-node = 1)

the process_worker_pool.py spawns multiple workers (by default one per detected core, but it's configurable) which can run one task each - so "many tasks per node" execution behaviour comes from *this* layer, not from asking slurm to run multiple slurm-level tasks.

That's not the only way to do things:
Is this model the way you want your workers laid out?
  if yes:  is this the way parsl is behaving for you? - if not, try to describe the misbehaviour you are seeing, in terms of the above model.

 if no:  try to describe how you want things to be different compared to the above model.

PR #2117, opened 2021-08-30
Regularise monitoring RESOURCE_INFO messages

benclifford:
This PR makes RESOURCE_INFO messages be in the same format as other
monitoring messages:  ( ( tag, message ), address)

This is part of ongoing rationalisation of monitoring messaging that
will allow any kind of message to be sent over any monitoring channel,
from any location.

## Type of change

- Code maintentance/cleanup

PR #3587, opened 2024-08-16
Remove monitoring queue tag switch monitoring db pre-router

benclifford:
The main goals of this PR is to make _migrate_logs_to_internal much more clearly a message forwarder, rather than a message interpreter.

This follows on from PR #2168 which introduces _dispatch_to_internal to dispatches messages based on their tag rather than on the queue the message was received on, and is part of an ongoing series to simplify the queue and routing structure inside the monitoring router and database code.

Further PRs in preparation (in draft PR #3315) contain further simplifications building on this PR.

After this PR:

* the database manager will respond to a STOP message on any incoming queue, vs previously only on the priority queue. This is a consequence of treating the queues all the same now.

* the database manager will not perform such strong validation of message structure based on message tag at this point. That's part of expecting the code to forward messages, not inspect them, with later inspecting code being the place to care abou structure. This only affects behaviour when invalid messages are sent.

Related PRs and context:

#3567 changes the monitoring router to be more of a router and to not inspect and modify certain in-transit messages.

There is a long slow project to regularise queues: PR #2117 makes resource info messages look like other message so they can be dispatched alongside other message types.

The priority queue was initially (as I understand it) introduced to attempt to address a race condition of message order arrival vs SQL database key constraints. The priority queue is an attempt to force certain messages to be processed before others (not in the priority queue). However a subsequent commit in 2019, 0a4b68555ce1946e46b96a13f9003e0733252ec6, introduces a more robust approach because this priority queue approach does not work and so is not needed.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -316,7 +316,7 @@ class DatabaseManager:
         self._kill_event = threading.Event()
         self._priority_queue_pull_thread = threading.Thread(target=self._migrate_logs_to_internal,
                                                             args=(
-                                                                priority_queue, 'priority', self._kill_event,),
+                                                                priority_queue, self._kill_event,),
                                                             name="Monitoring-migrate-priority",
                                                             daemon=True,
                                                             )
@@ -324,7 +324,7 @@ class DatabaseManager:
 
         self._node_queue_pull_thread = threading.Thread(target=self._migrate_logs_to_internal,
                                                         args=(
-                                                            node_queue, 'node', self._kill_event,),
+                                                            node_queue, self._kill_event,),
                                                         name="Monitoring-migrate-node",
                                                         daemon=True,
                                                         )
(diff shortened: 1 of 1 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -577,43 +577,26 @@ def start(self,
             raise RuntimeError("An exception happened sometime during database processing and should have been logged in database_manager.log")
 
     @wrap_with_logs(target="database_manager")
-    def _migrate_logs_to_internal(self, logs_queue: queue.Queue, queue_tag: str, kill_event: threading.Event) -> None:
-        logger.info("Starting processing for queue {}".format(queue_tag))
+    def _migrate_logs_to_internal(self, logs_queue: queue.Queue, kill_event: threading.Event) -> None:
+        logger.info("Starting _migrate_logs_to_internal")
```
khk-globus, line 581: Calling out to myself that this doesn't obfuscate log trawling as the logging setup _does_ include the thread name&nbsp;&mdash;&nbsp; the `queue_tag` is redundant.

khk-globus (approved):
Oh this lovely.  Nice find and well done!

PR #3591, opened 2024-08-16
Fallback to squeue when sacct is missing in SlurmProvider 

### LINKED ISSUE #3590: Regression: parsl fails on small slurm clusters without slurm accounting
**Describe the bug**
Although larger systems will consistently have slurm accounting, not all small clusters do (it's useful, but not required). #3422 introduced a change to retrieve the job information from the slurm accounting rather than from the scheduler. This breaks parsl on those clusters.

**To Reproduce**
On any cluster without slurm accounting, call `sacct` from a shell. It will fail

**Expected behavior**
From reading through #3422, I understand using `sacct` reduces load on the scheduler, which is likely beneficial, especially on larger systems. However, it unexpected breaks systems which previously worked. Although we don't want to add a big support burden, a fairly straightforward fix could be to conditionally try `sacct` and then fall back to the previous behavior based on `squeue`.

If sacct will be a requirement going forward, this would be useful to discuss and understand. Thanks!

**Environment**
A cluster without slurm accounting available

**Distributed Environment**
N/A

cc: @fjonasALICE

tylern4:
# Description

Adds internal check to test whether the slurm provider should use the `sacct` or `squeue` command.

# Changed Behaviour

Some slurm clusters might not use the accounting database `sacct` uses. This allows slurm clusters that use the database to use the `sacct` command which can be easier on the slurm scheduler, or if the database is not present switch to the `squeue` command which will should work on all clusters.

# Fixes

Fixes #3590

## Type of change

- Bug fix


benclifford:
Great!

Can you confirm you tested it with both sacct and squeue?

and @raymondEhlers did you test it?

If so I can avoid any testing myself and merge it for the auto-release on Monday.

raymondEhlers:
Yes, we will test it and report back, although probably not before Monday (I missed you reply, apologies)

benclifford:
> Yes, we will test it and report back, although probably not before Monday (I missed you reply, apologies)

cool thanks. There's no tight deadline on this.

benclifford (approved):

Issue #1122, opened 2019-07-08
generate pseudo-random IDs with better spread than current time


benclifford:
There are lots of places in the code where unique IDs are generated mostly based on the current time. That's a big ID space but with lots of correlation between IDs generated by the same run. At some scale these will clash and cause non-deterministic trouble that will be hard to debug.

There is plenty of experience in the world using more spread IDs (c.f. UUIDs)

Everywhere time() is used to generate unique IDs should be replaced by a call to a parsl helper function that does this more sensibly.

benclifford:
crossref #831

OfficialOzioma:
Hi @benclifford  is this issue still open, i will like to work on it, and i will need some guidance on how to get started thank you

rjmello:
Crossref: https://github.com/Parsl/parsl/pull/3639

PR #3592, opened 2024-08-19
TaskVineExecutor: write function data to temp directory

colinthomas-z80:
# Description

This moves the working directory for TaskVine executor function data to a directory in /tmp

These files were previously written adjacent to the logging in the working directory. They may be written in excess when running a large number of tasks.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/taskvine/executor.py
+++ b/parsl/executors/taskvine/executor.py
@@ -18,6 +18,7 @@ import tempfile
 import threading
 import uuid
 from concurrent.futures import Future
+from datetime import datetime
 from typing import List, Literal, Optional, Union
 
 # Import other libraries
@@ -215,7 +216,8 @@ class TaskVineExecutor(BlockProviderExecutor, putils.RepresentationMixin):
 
         # Create directories for data and results
         log_dir = os.path.join(run_dir, self.label)
-        self._function_data_dir = os.path.join(run_dir, self.label, "function_data")
+        tmp_dir = os.path.join('/tmp/', f'{self.label}-{os.getlogin()}')
+        self._function_data_dir = os.path.join(tmp_dir, datetime.now().strftime('%Y%m%d%H%M%S%f'), "function_data")
         os.makedirs(log_dir)
         os.makedirs(self._function_data_dir)
 
--- a/parsl/executors/taskvine/executor.py
+++ b/parsl/executors/taskvine/executor.py
@@ -3,6 +3,7 @@ Cooperative Computing Lab (CCL) at Notre Dame to provide a fault-tolerant,
 high-throughput system for delegating Parsl tasks to thousands of remote machines
 """
 
+import getpass
 import hashlib
 import inspect
 import itertools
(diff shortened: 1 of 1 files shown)


benclifford:
I ran in my dev container and I got this surprise.

```
>       tmp_dir = os.path.join('/tmp/', f'{self.label}-{os.getlogin()}')
E       OSError: [Errno 6] No such device or address
```

benclifford:
did some strace, and it's looking in this file which contains a decimalised unsigned 32 bit integer representation of signed 32-bit `-1`:
```
$ cat /proc/self/loginuid 
4294967295
```
and then failing after that.

The documentation for Python's `os.getlogin` says 

https://docs.python.org/3/library/os.html#os.getlogin

> For most purposes, it is more useful to use [getpass.getuser()](https://docs.python.org/3/library/getpass.html#getpass.getuser)

and in my dev environment that *is* able to figure out who I am:

```
>>> import getpass
>>> getpass.getuser()
'benc'
```

parsl/executors/taskvine/executor.py
```diff
@@ -215,7 +216,8 @@ def __create_data_and_logging_dirs(self):
 
         # Create directories for data and results
         log_dir = os.path.join(run_dir, self.label)
-        self._function_data_dir = os.path.join(run_dir, self.label, "function_data")
+        tmp_dir = os.path.join('/tmp/', f'{self.label}-{os.getlogin()}')
+        self._function_data_dir = os.path.join(tmp_dir, datetime.now().strftime('%Y%m%d%H%M%S%f'), "function_data")
```
benclifford, line 220: This is more a reference for readers of #1122 that  review comment for @colinthomas-z80 - here's another instance of using a timestamp as a unique ID that #1122 might want to address

benclifford:
Github Actions testing shows that same OS error - it's not just my wacky dev environment.

colinthomas-z80:
I do recognize the potential issues using datetime. In this instance it is specific to milliseconds so a conflict is unlikely. My rationale was to follow a logging convention in taskvine that will show the logs in a directory listing by order of creation, so it is easier to locate the most recent run. Though these files are rarely useful for a user to inspect themselves. I can change to uuid if you think it conforms better.

benclifford:
> I do recognize the potential issues using datetime. 

I don't think there's anything to be too concerned about here... mostly I was tagging it for the eventual implementer of #1122 to come fix it up.

khk-globus:
>> I do recognize the potential issues using datetime.
> 
> I don't think there's anything to be too concerned about here... mostly I was tagging it for the eventual implementer of #1122 to come fix it up.

FWIW, this is precisely the [purview of the `tempfile` module](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) [docs.python.org].  Consider:
```python
import tempfile
...
def ...:
    ...
    now_str = datetime.now().strftime(...)
    prefix = f"{self.label}-{getpass.getuser()}-{now_str}-"
    tmp_dir = tempfile.TemporaryDirectory(prefix=prefix, delete=False)
```
At which point `tmp_dir` will have the property of sortability you've stated was your goal and will also be guaranteed unique (that's `tempfile`'s guarantee).

I'm not saying it's the best solution in this precise scenario, but I do observe that temporary files are not as immediately trivial as many assume at first glance&nbsp;&mdash;&nbsp;there's a reasion `mktemp` (command) and Python's `tempfile` module exist.

benclifford:
I just merged this into my kubernetes test PR https://github.com/Parsl/parsl/pull/3482 after getting distracted making task vine work in there, to give me some more assurance about how this works when workers do not have a shared filesystem with the submit side. It works ok.

benclifford (approved):

PR #3594, opened 2024-08-19
Make deliberately scaled-in unstarted blocks not be failures

### LINKED ISSUE #3568: MISSING job detection vs htex_auto_scale strategy
This 4-minute program fails (with parsl master d8e8d4b3d99b2f034b1d5e80f89f58e1278486c2)

```
$ cat missing.py 

# attempt at reproducing something akila has reported

import parsl
from parsl.launchers import WrappedLauncher


def fresh_config():
    return Config(
        max_idletime = 6,
        strategy = 'htex_auto_scale',
        executors=[
            HighThroughputExecutor(
                label="htex_local",
                worker_debug=True,
                cores_per_worker=1,
                encrypted=True,
                provider=LocalProvider(
                    channel=LocalChannel(),
                    init_blocks=20,
                    min_blocks=0,
                    max_blocks=20,
                    launcher=WrappedLauncher(prepend="sleep 60 ; "),
                ),
            )
        ],
    )


@parsl.python_app
def task():
  return 7



with parsl.load(fresh_config()):
   import time
   time.sleep(120)
   assert task().result() == 7

print("done")
```

I've attached a log file.

This is an attempt to reproduce a failure @ravihansa3000 has seen.

Rough diagnosis:

jobs which are scaled in before they run do not register, and so are `MISSING` (despite being legitimately cancelled, not failing).

When these all accumulate, 100% of jobs being `MISSING` is treated as an error. In this example above, a large number of jobs are launched and then scaled in.
[parsl.log](https://github.com/user-attachments/files/16510912/parsl.log)

cc @yadudoc who introduced `MISSING` status in #3053 and `htex_auto_scale` in #1563

benclifford:
This PR adds a new terminal job state, SCALED_IN. None of the existing providers will return it, but the scaling layer will use it to mark a job as deliberately scaled in, so that error handling code will not regard it as failed.

Fixes #3568

### DIFF
--- a/parsl/tests/test_scaling/test_regression_3568_scaledown_vs_MISSING.py
+++ b/parsl/tests/test_scaling/test_regression_3568_scaledown_vs_MISSING.py
@@ -11,24 +11,22 @@ from parsl.providers import LocalProvider
 
 
 def local_config():
+    # see the comments inside test_regression for reasoning about why each
+    # of these parameters is set why it is.
     return Config(
         max_idletime=1,
+
         strategy='htex_auto_scale',
         strategy_period=1,
+
         executors=[
             HighThroughputExecutor(
                 label="htex_local",
-                address="127.0.0.1",
-                cores_per_worker=1,
                 encrypted=True,
                 provider=LocalProvider(
-                    channel=LocalChannel(),
                     init_blocks=1,
                     min_blocks=0,
                     max_blocks=1,
-                    # TODO: swap out the launcher during the test. make it use a much longer sleep (or infinite sleep)
-                    # to give blocks that run but never complete themselves. then swap that for a launcher which
-                    # launches the worker pool immediately.
                     launcher=WrappedLauncher(prepend="sleep inf ; "),
                 ),
             )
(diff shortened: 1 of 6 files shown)


parsl/tests/test_scaling/test_regression_3568_scaledown_vs_MISSING.py
```diff
@@ -0,0 +1,85 @@
...
+    # mentioned above, that is a necessary part of the bug being tested here.
+
+    # Give 10 strategy periods for the above to happen: each step of scale up,
+    # and scale down due to idleness isn't guaranteed to happen in exactly one
+    # scaling step.
+
+    htex = parsl.dfk().executors['htex_local']
+
+    try_assert(lambda: len(htex.status_facade) == 1 and htex.status_facade['0'].terminal,
+               timeout_ms=10000)
```
khk-globus, line 69: Consider using the `fail_msg` argument to explain that this is a block.  (Analogous to `assert ..., "fail_msg"`)

parsl/tests/test_scaling/test_regression_3568_scaledown_vs_MISSING.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Prior to issue #3568, the bug was that the scale in of the first
+    # block earlier in the test case would have incorrectly been treated as a
+    # failure, and then the block error handler would have treated that failure
+    # as a permanent htex failure, and so the task execution below would raise
+    # a BadStateException rather than attempt to run the task.
+
+    assert htex.provider.launcher.prepend != "", "Pre-req: prepend attribute should exist and be non-empty"
+    htex.provider.launcher.prepend = ""
+    assert task().result() == 7
```
khk-globus, line 85: Can you distill the comment above into the fail msg for this assert?  Not an easy task, I know, but when this _does_ fail, makes it that easier to dig in.

khk-globus (approved):
The block comment is unfortunately necessary, so thank you for including it.  The rest are window dressing suggestions for clarity.

PR #3597, opened 2024-08-21
Non-functional change: minor log call-site updates

khk-globus:
Mostly an exercise in calming my IDE and "while here" massaging a couple of `log.debug` calls to use printf-style arguments.

# Changed Behaviour

No changed behavior.

## Type of change

- Code maintenance/cleanup


benclifford (approved):

benclifford:
> Mostly an exercise in calming my IDE and "while here" massaging a couple of log.debug calls to use printf-style arguments.

i made a different commit message for this

PR #3602, opened 2024-08-27
Adding configuration example for Improv@LCRC

yadudoc:
# Description

This PR adds a configuration example for Improv at Argonne's LCRC.

## Type of change

### DIFF
--- a/parsl/configs/improv.py
+++ b/parsl/configs/improv.py
@@ -1,0 +1,34 @@
+from parsl.config import Config
+from parsl.executors import HighThroughputExecutor
+from parsl.launchers import MpiRunLauncher
+from parsl.providers import PBSProProvider
+
+config = Config(
+        executors=[
+            HighThroughputExecutor(
+                label="Improv_multinode",
+                max_workers_per_node=32,
+                provider=PBSProProvider(
+                    account="YOUR_ALLOCATION_ON_IMPROV",
+                    # PBS directives (header lines), for example:
+                    # scheduler_options='#PBS -l mem=4gb',
+                    scheduler_options='',
+
+                    queue="compute",
+
+                    # Command to be run before starting a worker:
+                    # **WARNING** Improv requires an openmpi module to be
+                    # loaded for the MpiRunLauncher. Add additional env
+                    # load commands to this multiline string.
+                    worker_init='''
+module load gcc/13.2.0;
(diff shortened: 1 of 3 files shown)


parsl/configs/improv.py
```diff
@@ -0,0 +1,33 @@
...
+                provider=PBSProProvider(
+                    account="YOUR_ALLOCATION_ON_IMPROV",
+                    # PBS directives (header lines): for array jobs pass '-J' option
+                    # scheduler_options='#PBS -J 1-10',
+                    scheduler_options='',
+
+                    queue="compute",
+                    # Command to be run before starting a worker, such as:
+                    # 'module load Anaconda; source activate parsl_env'.
+                    # **WARNING** Improv requires an openmpi module to be
```
benclifford, line 20: maybe the example here should include loading an openmpi module, either in the explicit value or in the comment on this line

parsl/configs/improv.py
```diff
@@ -0,0 +1,33 @@
...
+from parsl.providers import PBSProProvider
+
+config = Config(
+        executors=[
+            HighThroughputExecutor(
+                label="Improv_multinode",
+                max_workers_per_node=32,
+                provider=PBSProProvider(
+                    account="YOUR_ALLOCATION_ON_IMPROV",
+                    # PBS directives (header lines): for array jobs pass '-J' option
```
benclifford, line 13: i don't think we should be encouraging people to use array jobs. i've only ever seen one dubious need for them.

parsl/configs/improv.py
benclifford: oh you did, i misread `'''` as `''`

parsl/configs/improv.py
```diff
@@ -0,0 +1,33 @@
...
+                    # scheduler_options='#PBS -J 1-10',
+                    scheduler_options='',
+
+                    queue="compute",
+                    # Command to be run before starting a worker, such as:
+                    # 'module load Anaconda; source activate parsl_env'.
```
benclifford, line 19: there's already an example so you don't need this second example?

yadudoc: Fair enough, I've removed this one.

parsl/configs/improv.py
yadudoc: Removed.

benclifford (approved):

PR #3606, opened 2024-09-04
Slurm tests

### LINKED ISSUE #3579: CI testing of SLURM in a container environment
See PR #3159 which runs Parsl tests in flux environment and (unmerged prototype) PR #3482 which does similar for kubernetes.

A couple of different groups have slurm-in-container for their own tests.

So I think the pieces are broadly there.

cc @kylechard who most recently mentioned this to me

tylern4:
# Description

Adds in testing for slurm systems inside of a container which have come up from discussions with @benclifford in some emails and issue https://github.com/Parsl/parsl/issues/3579

# Changed Behaviour

More testing on slurm code.

# Fixes

Fixes #3579

## Type of change

- New feature


tylern4:
The slurm tests pass with the slurm config but it seems like it's still missing some slurm tests that could be helpful. Some helping getting a good config to test everything or the right `pytest` settings would be appreciated!

benclifford:
What you've got is a test that runs the "should work with most configs" tests against an environment using the SlurmProvider.

Are you asking about specific tests that you know exist but don't see running?

I think the only other slurm-related tests are these, which don't need a working slurm instance - they only check that the provider can set itself up correctly:

```
$ ls parsl/tests/test_providers/test_slurm_*.py
parsl/tests/test_providers/test_slurm_instantiate.py  parsl/tests/test_providers/test_slurm_template.py
```

Part of the reason there is that personally I have only really pushed on tests that are run automatically, rather than at-release testing that wasn't happening in reality. If this PR was merged, the automated tests would be in a much better place to get additional tests that might test specific `SlurmProvider` features, rather than running a full Parsl DFK.

tylern4:
Right I saw those two tests didn't run when I added the Slurm Config but if they are okay to be left off then I think this is good to go.

parsl/tests/configs/slurm_local.py
```diff
@@ -0,0 +1,29 @@
...
+                    max_blocks=1,
+                    walltime='00:10:00',
+                    launcher=SrunLauncher(),
+                ),
+            )
+        ],
+    )
+
+
+config = fresh_config()
```
benclifford, line 29: I think this line isn't needed and you got it by cut and pasting it from somewhere else that also doesn't need it... it also causes trouble in some situations by making configs be initialized at import (so every single test config gets initialised on every test run), rather than when the config is actually wanted.

benclifford:
>  I saw those two tests didn't run when I added the Slurm Config

They get run in a different part of testing, because they are marked with `pytest.mark.local` - because they aren't tied to being run with any particular configuration.

Issue #1689, opened 2020-05-12
Handling of stdout and stderr as implicit inputs


btovar:
Not really a bug, more a design question.

My (unpublished) branch of the work queue executor currently fails the test at:

https://github.com/Parsl/parsl/blob/bd9248a798c2bb5457d74e5c287ddef99310ea2e/parsl/tests/test_bash_apps/test_stdout.py#L101

This test has two tasks in sequence. The first one creates a file from stdout, and the second appends to it.  The problem in work queue is that the second task never sees the output from the first, because it was never declared as an input, and therefore not sent to the execution site.

We could treat stdout and stderr as implicit inputs if they exist when sending the task. The work queue executor currently in parsl does this implicit sending. However, this may lead to future confusion when files not declared as inputs are sent. (Specially stdout and stderr, which sound very much like outputs).

That said, if it is desired for parsl to produce as much as possible the same results regardless of the executor, there might not be a way around it.

benclifford:
I'm inclined to say:
- this behaviour is undefined in parsl in general
- happens to behave in this way for everything so far because all our tests so far have used a shared filesystem
- this doesn't make sense in other staging models: concurrent appending of staged files is messy - workqueue is the first setup that hits this undefinedness, but it isn't workqueue specific (for example, long term if we transfer stdin/out/err files using the data staging provider mechanisms, the same issue will arise)
- tests related should go away because of that un-definedness in general, and be made into `local` tests for specific configurations where we expect this more specific behaviour to hold

yadudoc:
There's an `issue363` marker in pytest to ignore tests that assume a shared-fs. I can't find an example of this in use in master, but I'm guessing passing `not issue363` to pytest might work.
It's most likely that some of these tests are straight up bad, and should not be assuming a shared-fs to start with.

benclifford:
Oh, I had forgotten about that. We should rename it. I agree this is the right thing to enable for tests that aren't on a shared fs. Add to the -k parameter of pytest when you want to filter it, like this. The existing cleannet parameter turns of tests that expect you to have a "clean" internet connection which can do FTP - which our CI does not have.

```
pytest parsl -k "not cleannet and not issue363" --config parsl/tests/configs/workqueue_ex.py
```

btovar:
The 'not issue363" works, in the sense that all tests in `test_bash_apps/test_stdout.py` are deselected. Would you be opposed if I add a test to test_stdout.py that simply writes to a file once?

[...]

PR #3607, opened 2024-09-05
Support for GlobusComputeExecutor

yadudoc:
# Description

This PR adds a new `GlobusComputeExecutor` that wraps the Globus Compute SDK to allow Parsl to execute tasks via Globus Compute. This mechanism supports remote execution of tasks similar to the functionality that `parsl.channels` enabled and is a potential replacement.

Since `GlobusCompute` often runs on remote machines that do not have a shared-filesystem with the parsl runtime, tests have been updated with a new `shared_fs` and `staging_required` pytest markers. I have not added tests and CI actions to enable executing these tests against our CI system, but you can run tests locally with these steps:

1. Install `globus-compute-sdk` with `pip install .[globus-compute]`
2. Configure and start a globus-compute-endpoint, `globus-compute-endpoint start <endpoint_name>`
3. Set an env var with the endpoint id for tests: `export GLOBUS_COMPUTE_ENDPOINT=<endpoint_id>`
4. Run tests with `pytest -v -k "not shared_fs" --config parsl/tests/configs/globus_compute.py parsl/tests/`

# Changed Behaviour

N/A

# Fixes

Fixes # (issue)

## Type of change

- New feature

### DIFF
--- a/.github/workflows/gce_test.yaml
+++ b/.github/workflows/gce_test.yaml
@@ -1,0 +1,108 @@
+name: GlobusComputeExecutor tests
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+
+env:
+  PYTHON_VERSION: 3.11
+
+jobs:
+  main-test-suite:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Set up Python Environment
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ env.PYTHON_VERSION }}
+
+    - name: Collect Job Information
+      id: job-info
+      run: |
+        echo "Python Version: ${{ env.PYTHON_VERSION }} " >> ci_job_info.txt
+        echo "CI Triggering Event: ${{ github.event_name }}" >> ci_job_info.txt
+        echo "Triggering Git Ref: ${{ github.ref }}" >> ci_job_info.txt
+        echo "Triggering Git SHA: ${{ github.sha }}" >> ci_job_info.txt
+        echo "Workflow Run: ${{ github.run_number }}" >> ci_job_info.txt
+        echo "Workflow Attempt: ${{ github.run_attempt }}" >> ci_job_info.txt
+        as_ascii="$(echo "${{ github.ref_name }}" | perl -pe "s/[^A-z0-9-]+/-/g; s/^-+|-+\$//g; s/--+/-/g;")"
+        echo "as-ascii=$as_ascii" >> $GITHUB_OUTPUT
+
+    - name: setup virtual env
+      run: |
+        make virtualenv
+        source .venv/bin/activate
+
+    - name: Non-requirements based install
+      run: |
+        # mpich: required by mpi4py which is in test-requirements for radical-pilot
+        sudo apt-get update -q
+        sudo apt-get install -qy mpich
+
+    - name: make deps clean_coverage
+      run: |
+        source .venv/bin/activate
+        make deps
+        make clean_coverage
+
+        # Temporary fix until fixes make it to a release
+        git clone -b main https://github.com/globus/globus-compute.git
+        pip3 install globus-compute/compute_sdk globus-compute/compute_endpoint
+
+    - name: start globus_compute_endpoint
+      env:
+        GLOBUS_COMPUTE_CLIENT_ID: ${{ secrets.GLOBUS_COMPUTE_CLIENT_ID }}
+        GLOBUS_COMPUTE_CLIENT_SECRET: ${{ secrets.GLOBUS_COMPUTE_SECRET_KEY }}
+      run: |
+        source /home/runner/work/parsl/parsl/.venv/bin/activate
+        globus-compute-endpoint configure default
+        which globus-compute-endpoint
+        python3 -c "import globus_compute_sdk; print(globus_compute_sdk.__version__)"
+        python3 -c "import globus_compute_endpoint; print(globus_compute_endpoint.__version__)"
+        cat << EOF > /home/runner/.globus_compute/default/config.yaml
+        engine:
+          type: ThreadPoolEngine
+          max_workers: 4
+        EOF
+        cat /home/runner/.globus_compute/default/config.yaml
+        mkdir ~/.globus_compute/default/tasks_working_dir
+        globus-compute-endpoint start default
+        globus-compute-endpoint list
+    - name: make test
+      env:
+        GLOBUS_COMPUTE_CLIENT_ID: ${{ secrets.GLOBUS_COMPUTE_CLIENT_ID }}
+        GLOBUS_COMPUTE_CLIENT_SECRET: ${{ secrets.GLOBUS_COMPUTE_SECRET_KEY }}
+      run: |
+        source .venv/bin/activate
+        export GLOBUS_COMPUTE_ENDPOINT=$(jq -r .endpoint_id < ~/.globus_compute/default/endpoint.json)
+        echo "GLOBUS_COMPUTE_ENDPOINT = $GLOBUS_COMPUTE_ENDPOINT"
+
+        export PARSL_TEST_PRESERVE_NUM_RUNS=7
+
+        make gce_test
+        ln -s pytest-parsl/parsltest-current test_runinfo
+
+    - name: stop globus_compute_endpoint
+      env:
+        GLOBUS_COMPUTE_CLIENT_ID: ${{ secrets.GLOBUS_COMPUTE_CLIENT_ID }}
+        GLOBUS_COMPUTE_CLIENT_SECRET: ${{ secrets.GLOBUS_COMPUTE_SECRET_KEY }}
+      run: |
+        source /home/runner/work/parsl/parsl/.venv/bin/activate
+        globus-compute-endpoint stop default
+
+    - name: Archive runinfo logs
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: runinfo-${{ env.PYTHON_VERSION }}-${{ steps.job-info.outputs.as-ascii }}-${{ github.sha }}
+        path: |
+          runinfo/
+          pytest-parsl/
+          ci_job_info.txt
+        compression-level: 9
--- a/parsl/executors/globus_compute.py
+++ b/parsl/executors/globus_compute.py
@@ -1,0 +1,125 @@
+from __future__ import annotations
+
+import copy
+from concurrent.futures import Future
+from typing import Any, Callable, Dict
+
+import typeguard
+
+from parsl.errors import OptionalModuleMissing
+from parsl.executors.base import ParslExecutor
+from parsl.utils import RepresentationMixin
+
+try:
+    from globus_compute_sdk import Executor
+    _globus_compute_enabled = True
+except ImportError:
+    _globus_compute_enabled = False
+
+
+class GlobusComputeExecutor(ParslExecutor, RepresentationMixin):
+    """ GlobusComputeExecutor enables remote execution on Globus Compute endpoints
+
+    GlobusComputeExecutor is a thin wrapper over globus_compute_sdk.Executor
+    Refer to `globus-compute user documentation <https://globus-compute.readthedocs.io/en/latest/executor.html>`_
+    and `reference documentation <https://globus-compute.readthedocs.io/en/latest/reference/executor.html>`_
+    for more details.
+
+    .. note::
+       As a remote execution system, Globus Compute relies on serialization to ship
+       tasks and results between the Parsl client side and the remote Globus Compute
+       Endpoint side. Serialization is unreliable across python versions, and
+       wrappers used by Parsl assume identical Parsl versions across on both sides.
+       We recommend using matching Python, Parsl and Globus Compute version on both
+       the client side and the endpoint side for stable behavior.
+
+    """
+
+    @typeguard.typechecked
+    def __init__(
+        self,
+        executor: Executor,
+        label: str = 'GlobusComputeExecutor',
+    ):
+        """
+        Parameters
+        ----------
+
+        executor: globus_compute_sdk.Executor
+            Pass a globus_compute_sdk Executor that will be used to execute
+            tasks on a globus_compute endpoint. Refer to `globus-compute docs
+            <https://globus-compute.readthedocs.io/en/latest/reference/executor.html#globus-compute-executor>`_
+
+        label:
+            a label to name the executor
+        """
+        if not _globus_compute_enabled:
+            raise OptionalModuleMissing(
+                ['globus-compute-sdk'],
+                "GlobusComputeExecutor requires globus-compute-sdk installed"
+            )
+
+        super().__init__()
+        self.executor: Executor = executor
+        self.resource_specification = self.executor.resource_specification
+        self.user_endpoint_config = self.executor.user_endpoint_config
+        self.label = label
+
+    def start(self) -> None:
+        """ Start the Globus Compute Executor """
+        pass
+
+    def submit(self, func: Callable, resource_specification: Dict[str, Any], *args: Any, **kwargs: Any) -> Future:
+        """ Submit func to globus-compute
+
+
+        Parameters
+        ----------
+
+        func: Callable
+            Python function to execute remotely
+
+        resource_specification: Dict[str, Any]
+            Resource specification can be used specify MPI resources required by MPI applications on
+            Endpoints configured to use globus compute's MPIEngine. GCE also accepts *user_endpoint_config*
+            to configure endpoints when the endpoint is a `Multi-User Endpoint
+            <https://globus-compute.readthedocs.io/en/latest/endpoints/endpoints.html#templating-endpoint-configuration>`_
+
+        args:
+            Args to pass to the function
+
+        kwargs:
+            kwargs to pass to the function
+
+        Returns
+        -------
+
+        Future
+        """
+        res_spec = copy.deepcopy(resource_specification or self.resource_specification)
+        # Pop user_endpoint_config since it is illegal in resource_spec for globus_compute
+        if res_spec:
+            user_endpoint_config = res_spec.pop('user_endpoint_config', self.user_endpoint_config)
+        else:
+            user_endpoint_config = self.user_endpoint_config
+
+        try:
+            self.executor.resource_specification = res_spec
+            self.executor.user_endpoint_config = user_endpoint_config
+            return self.executor.submit(func, *args, **kwargs)
+        finally:
+            # Reset executor state to defaults set at configuration time
+            self.executor.resource_specification = self.resource_specification
+            self.executor.user_endpoint_config = self.user_endpoint_config
+
+    def shutdown(self):
+        """Clean-up the resources associated with the Executor.
+
+        GCE.shutdown will cancel all futures that have not yet registered with
+        Globus Compute and will not wait for the launched futures to complete.
+        This method explicitly shutsdown the result_watcher thread to avoid
+        it waiting for outstanding futures at thread exit.
+        """
+        self.executor.shutdown(wait=False, cancel_futures=True)
+        result_watcher = self.executor._get_result_watcher()
+        result_watcher.shutdown(wait=False, cancel_futures=True)
--- a/parsl/tests/configs/globus_compute.py
+++ b/parsl/tests/configs/globus_compute.py
@@ -1,0 +1,20 @@
+import os
+
+from globus_compute_sdk import Executor
+
+from parsl.config import Config
+from parsl.executors import GlobusComputeExecutor
+
+
+def fresh_config():
+
+    endpoint_id = os.environ["GLOBUS_COMPUTE_ENDPOINT"]
+
+    return Config(
+        executors=[
+            GlobusComputeExecutor(
+                executor=Executor(endpoint_id=endpoint_id),
+                label="globus_compute",
+            )
+        ]
+    )
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -163,6 +163,10 @@ def pytest_configure(config):
         'markers',
         'shared_fs: Marks tests that require a shared_fs between the workers are the test client'
     )
+    config.addinivalue_line(
+        'markers',
+        'issue_3620: Marks tests that do not work correctly on GlobusComputeExecutor (ref: issue 3620)'
+    )
 
 
 @pytest.fixture(autouse=True, scope='session')
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ extras_require = {
     'flux': ['pyyaml', 'cffi', 'jsonschema'],
     'proxystore': ['proxystore'],
     'radical-pilot': ['radical.pilot==1.90', 'radical.utils==1.90'],
+    'globus_compute': ['globus_compute_sdk>=2.34.0'],
     # Disabling psi-j since github direct links are not allowed by pypi
     # 'psij': ['psi-j-parsl@git+https://github.com/ExaWorks/psi-j-parsl']
 }
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ clean_coverage:
 mypy: ## run mypy checks
 	MYPYPATH=$(CWD)/mypy-stubs mypy parsl/
 
+.PHONY: gce_test
+gce_test: ## Run tests with GlobusComputeExecutor
+	pytest -v -k "not shared_fs and not issue_3620 and not staging_required" --config parsl/tests/configs/globus_compute.py parsl/tests/ --random-order --durations 10
+
 .PHONY: local_thread_test
 local_thread_test: ## run all tests with local_thread config
 	pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/local_threads.py --random-order --durations 10
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -79,6 +79,7 @@ Executors
     parsl.executors.taskvine.TaskVineExecutor
     parsl.executors.FluxExecutor
     parsl.executors.radical.RadicalPilotExecutor
+    parsl.executors.GlobusComputeExecutor
 
 Manager Selectors
 =================
--- a/docs/userguide/configuration/examples.rst
+++ b/docs/userguide/configuration/examples.rst
@@ -250,6 +250,30 @@ running on a login node and uses the `parsl.providers.SlurmProvider` to interfac
 
 .. literalinclude:: ../../../parsl/configs/frontera.py
 
+Globus Compute (Multisite)
+--------------------------
+
+Globus Compute is a distributed Function as a Service (FaaS) platform that enables secure
+execution of functions on heterogeneous remote computers, from laptops to campus clusters, clouds, and supercomputers.
+Functions are executed on `Globus Compute Endpoints <https://globus-compute.readthedocs.io/en/latest/endpoints/endpoints.html>`_
+that can be `configured <https://globus-compute.readthedocs.io/en/latest/endpoints/endpoint_examples.html>`_
+for most clusters/HPC systems. The example configuration below allows task submission
+to Globus Compute's hosted tutorial endpoint.
+
+.. literalinclude:: ../../../parsl/configs/gc_tutorial.py
+
+.. note:: The Globus Compute tutorial endpoint runs Python 3.11. We recommend
+   using the same Python environment to avoid potential serialization errors
+   caused by environment mismatches. Globus Compute will raise a warning if any
+   environment version mismatches are detected although minor version differences
+   may not cause faults (eg, Python3.11.7 vs Python3.11.8)
+
+The configuration below specifies two remote endpoints, one at `SDSC's Expanse Supercomputer <https://www.sdsc.edu/services/hpc/expanse/>`_
+and the other at `NERSC's Perlmutter Supercomputer <https://docs.nersc.gov/systems/perlmutter/architecture/>`_.
+
+.. literalinclude:: ../../../parsl/configs/gc_multisite.py
+
+
 
 Kubernetes Clusters
 -------------------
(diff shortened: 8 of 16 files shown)


benclifford:
Aside from a lack of CI test, my main concern (from seeing people implement this before) is what goes on with different Parsl versions: there's one version in the endpoint (pinned to a specific version by Globus Compute) and another version on the submit side, and having different Parsl versions like that is out of scope for Parsl.

This might be a case of documenting what users should expect to work or not work, or might be something deeper. At the very least we should be expecting them to be able to use the same combination of versions as used in CI.

parsl/tests/conftest.py
```diff
@@ -163,6 +163,14 @@ def pytest_configure(config):
         'markers',
         'executor_supports_std_stream_tuples: Marks tests that require tuple support for stdout/stderr'
     )
+    config.addinivalue_line(
+        'markers',
+        'globus_compute: Marks tests that require a valid globus_compute target'
+    )
```
rjmello, line 169: Do any tests use this marker?

parsl/tests/configs/globus_compute.py
```diff
@@ -0,0 +1,18 @@
+import os
+
+from parsl.config import Config
+from parsl.executors import GlobusComputeExecutor
+
+
+def fresh_config():
+
+    endpoint_id = os.environ.get("GLOBUS_COMPUTE_ENDPOINT",
+                                 "4b116d3c-1703-4f8f-9f6f-39921e5864df")
```
rjmello, line 10: While anyone could potentially run tests against our tutorial EP, I don't think we should suggest doing so.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,139 @@
...
+            self,
+            endpoint_id: Optional[UUID_LIKE_T] = None,
+            task_group_id: Optional[UUID_LIKE_T] = None,
+            resource_specification: Optional[dict[str, Any]] = None,
+            user_endpoint_config: Optional[dict[str, Any]] = None,
+            label: str = "GlobusComputeExecutor",
+            batch_size: int = 128,
+            amqp_port: Optional[int] = None,
+            **kwargs,
+    ):
```
rjmello, line 31: Consider adding the `client` parameter

parsl/tests/conftest.py
yadudoc: Unused. I've removed this marker.

parsl/tests/configs/globus_compute.py
yadudoc: Good call. I've changed this to require GLOBUS_COMPUTE_ENDPOINT to be specified for tests. Failing with a KeyError should be clear enough that something is misconfigured.

parsl/executors/globus_compute.py
yadudoc: I'm hesitant to add this because it will require `globus_compute_sdk.sdk.Client` type hints when this is an Optional module. Right now, passing this in via **kwargs will still work fine. What do you think?

rjmello: I think we should add it to clearly communicate support for the parameter. I've already had an early adopter ask about this.

To add type hints, we could do something like this:
```python
try:
    from globus_compute_sdk import Client
except ImportError:
    Client = Any
 
 class GlobusComputeExecutor(ParslExecutor, RepresentationMixin):
    def __init__(
            self,
            endpoint_id: Optional[UUID_LIKE_T] = None,
            client: Optional[Client] = None,
            ...
```

or

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from globus_compute_sdk import Client

 class GlobusComputeExecutor(ParslExecutor, RepresentationMixin):
    def __init__(
            self,
            endpoint_id: Optional[UUID_LIKE_T] = None,
            client: Optional["Client"] = None,
            ...
```

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,139 @@
...
+            Args to pass to the function
+        kwargs:
+            kwargs to pass to the function
+
+        Returns
+        -------
+
+        Future
+        """
+        self._executor.resource_specification = resource_specification or self.resource_specification
```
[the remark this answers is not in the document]
benclifford: if you want to specify that per task, it maybe makes sense to make it a new entry in the task submission resource_specification - that's the mechanism for specify everything about "how" a task is to be executed in Parsl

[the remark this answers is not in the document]
christopherwharrop-noaa: Unless I'm misunderstanding something, that is how the multi-user endpoints are meant to work.  If you can't specify it at submit time, then the alternative will be to have multiple instantiations of `GlobusComputeExecutor` for the same endpoint id, but with different `user_endpoint_config`.  Maybe that's not a "bad thing", but while you would reduce number of endpoints to maintain, Parsl workflows would still have to have one `GlobusComputeExecutor` for each type of endpoint you want to interact with, even when on the same machine.  And maybe that's not so bad because even if you didn't have to do that, you'd still need to handle the multiple 'user_endpoint_config` dicts that are required for the various endpoint configs you need.  My thought was to match what is available on the Globus Compute side, which allows (I think) setting that at both Executor instantiation time and at submit time.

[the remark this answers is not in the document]
benclifford: i think per-task specification is a good option to make available. i was commenting on *where* that specification should go when submitting from Parsl.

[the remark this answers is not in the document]
christopherwharrop-noaa: Would that look something like this?:
```
self._executor.user_endpoint_config = resource_specification["user_endpoint_config"] or self.user_endpoint_config
```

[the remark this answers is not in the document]
benclifford: yes

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,112 @@
...
+      run: |
+        make virtualenv
+        source .venv/bin/activate
+
+    - name: make deps clean_coverage
+      run: |
+        source .venv/bin/activate
+        make deps
+        make clean_coverage
+        # Installing parsl into venv required for GCendpoint
```
benclifford, line 61: doesn't GC dependency handling bring in Parsl?

parsl/executors/globus_compute.py
rjmello: haha thanks Ben :metal:

.github/workflows/gce_test.yaml
yadudoc: Correct. This is now fixed.

parsl/executors/globus_compute.py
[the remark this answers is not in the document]
yadudoc: I've made some changes to allow `resource_specification` and `user_endpoint_config` to both work.

```
# Specifying MPI apps
py_app(resource_specification={'num_nodes': 2, 'ranks_per_node':4})

# Specifying user_endpoint_config
py_app(resource_specification={'user_endpoint_config': {"account_id": "user_allocation_account_id"}}

# Specifying both:
py_app(resource_specification={'num_nodes': 2, 'ranks_per_node':4,
'user_endpoint_config': {"account_id": "user_allocation_account_id"}
})

```

parsl/executors/globus_compute.py
yadudoc: Fixed now.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,162 @@
...
+        # Pop user_endpoint_config since it is illegal in resource_spec for globus_compute
+        self._executor.user_endpoint_config = resource_specification.pop('user_endpoint_config', self.user_endpoint_config)
```
christopherwharrop-noaa, line 144: Should the `pop` happen before you assign `resource_specification` to `self._executor.resource_specification`?

benclifford: I think: the `pop` live-edits the value previously assigned to self._executor.resource_specification - because it is assignment of mutable objects, not of values.

But following that same reasoning, @yadudoc does this mean this code is editing the resource_specification dict provided by the user? (at some arbitrary point in the future after the user invokes an app...?)

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,110 @@
...
+        GLOBUS_COMPUTE_CLIENT_SECRET: ${{ secrets.GLOBUS_COMPUTE_SECRET_KEY }}
+      run: |
+        source /home/runner/work/parsl/parsl/.venv/bin/activate
+        globus-compute-endpoint configure default
+        which globus-compute-endpoint
+        python3 -c "import globus_compute_sdk; print(globus_compute_sdk.__version__)"
+        python3 -c "import globus_compute_endpoint; print(globus_compute_endpoint.__version__)"
+        cat << EOF > /home/runner/.globus_compute/default/config.yaml
+        engine:
+          type: ThreadPoolEngine
```
benclifford, line 78: I feel like this is a little evasive... the thread pool engine isn't what Globus Compute is "really" intended for, right?

docs/userguide/execution.rst
```diff
@@ -89,6 +89,9 @@ Parsl currently supports the following executors:
 4. `parsl.executors.taskvine.TaskVineExecutor`: This executor uses `TaskVine <https://ccl.cse.nd.edu/software/taskvine/>`_ as the execution backend. TaskVine scales up to tens of thousands of cores and actively uses local storage on compute nodes to offer a diverse array of performance-oriented features, including: smart caching and sharing common large files between tasks and compute nodes, reliable execution of tasks, dynamic resource sizing, automatic Python environment detection and sharing.
 These executors cover a broad range of execution requirements. As with other Parsl components, there is a standard interface (ParslExecutor) that can be implemented to add support for other executors.
 
+5. `parsl.executors.globus_compute.GlobusComputeExecutor`: This executor uses `Globus Compute <https://globus-compute.readthedocs.io/en/latest/index.html>`_
+as the execution backend to run functions on remote systems.
```
benclifford, line 93: most of the other paragraphs use the word "task" here, not "function"

parsl/executors/globus_compute.py
christopherwharrop-noaa: Oh, right!  Of course. I'm mixing up my languages.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,162 @@
...
+        user_endpoint_config:
+            User endpoint configuration values as described
+            and allowed by endpoint administrators. Must be a JSON-serializable dict
+            or None. Refer docs from `globus-compute
+            <https://globus-compute.readthedocs.io/en/latest/endpoints/endpoints.html#templating-endpoint-configuration>`_
+            for more info.
+
+        label:
+            a label to name the executor; mainly utilized for
+            logging and advanced needs with multiple executors.
```
benclifford, line 64: "advanced needs" is weird phrasing that sounds like bluffing about ... something?

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,162 @@
...
+            user_endpoint_config=user_endpoint_config,
+            label=label,
+            batch_size=batch_size,
+            amqp_port=amqp_port,
+            client=self.client,
+            **kwargs
+        )
+
+    def start(self) -> None:
+        """Empty function
```
benclifford, line 111: probably don't need this docstring at all. or justify the why here if there's something interesting to talk about.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        pass
+
+    def submit(self, func: Callable, resource_specification: Dict[str, Any], *args: Any, **kwargs: Any) -> Future:
+        """ Submit fn to globus-compute
```
benclifford, line 116: Use `func` not `fn`

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,162 @@
...
+        -------
+
+        Future
+        """
+        self._executor.resource_specification = resource_specification or self.resource_specification
+        # Pop user_endpoint_config since it is illegal in resource_spec for globus_compute
+        self._executor.user_endpoint_config = resource_specification.pop('user_endpoint_config', self.user_endpoint_config)
+        return self._executor.submit(func, *args, **kwargs)
+
+    def shutdown(self, wait=True, *, cancel_futures=False):
```
benclifford, line 147: what are all these parameters for?

parsl/tests/test_bash_apps/test_basic.py
```diff
@@ -24,6 +24,7 @@ def foo(x, y, z=10, stdout=None, label=None):
     return f"echo {x} {y} {z}"
 
 
+@pytest.mark.shared_fs
```
benclifford, line 27: maybe all these shared_fs / staging_required markers should be merged as a separate PR? They're not specific to this test case (eg. they came I think originally from the kubernetes testing draft PR #3482 and that work would be slightly advanced by these being merged to master, independent of the rest of this PR being merged or not)

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,110 @@
+name: GlobusComputeExecutor tests
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+
+  workflow_dispatch:
```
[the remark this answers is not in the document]
yadudoc: Obsolete. It was used to trigger the workflow run manually for debugging.

.github/workflows/gce_test.yaml
yadudoc: Right, this was added for 2 reasons. 1) The default engine had issues spawning all the components in the same virtualenv, despite worker_init being configured. 2) Even with a single worker, tests were slow which I suspect is due to the endpoint being way more resource-intensive. 

From what I see, what I'm testing is the executor on the parsl side. Parsl tests needn't care about what engine the GC endpoint uses on its side.

parsl/executors/globus_compute.py
yadudoc: Sounds odd now that you pointed it out. Fixed.

parsl/executors/globus_compute.py
yadudoc: Removed.

parsl/tests/test_bash_apps/test_basic.py
yadudoc: Fair enough. I left this here because those changes wouldn't kick in without tests that exclude based on the marker. I've split out these changes over on this PR -> https://github.com/Parsl/parsl/pull/3669

parsl/executors/globus_compute.py
yadudoc: I've updated this section to pop `user_endpoint_config` from a copy of `resource_specification` to avoid modifying the user-supplied dict.

parsl/executors/globus_compute.py
rjmello: I could be missing something, but I don't see the `client` parameter.

yadudoc: It's here -> https://github.com/Parsl/parsl/pull/3607/files#diff-03350429687012cb1a3ed39ac9d3549f2b4c825555f7e9dee0600ff82b16164aR48

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,103 @@
...
+        source .venv/bin/activate
+
+    - name: make deps clean_coverage
+      run: |
+        source .venv/bin/activate
+        make deps
+        make clean_coverage
+        
+        # Temporary fix, until changes make it into compute releases
+        git clone -b configure_tasks_working_dir https://github.com/globus/globus-compute.git
```
benclifford, line 56: maybe its good to wait for Globus Compute to support the necessary features in a release, so that this can pin a relevant supported release?

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - name: make test
+      env:
+        GLOBUS_COMPUTE_CLIENT_ID: ${{ secrets.GLOBUS_COMPUTE_CLIENT_ID }}
+        GLOBUS_COMPUTE_CLIENT_SECRET: ${{ secrets.GLOBUS_COMPUTE_SECRET_KEY }}
+      run: |
+        source .venv/bin/activate
+        export GLOBUS_COMPUTE_ENDPOINT=$(globus-compute-endpoint list | grep default | cut -c 3-38)      
+        echo "GLOBUS_COMPUTE_ENDPOINT = $GLOBUS_COMPUTE_ENDPOINT"
+
+        # temporary; until test-matrixification
```
benclifford, line 88: nothing so temporary as boilerplate copied from someone else's unfinished business.

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,103 @@
...
+        ln -s .pytest/parsltest-current test_runinfo
+
+    - name: Archive runinfo logs
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: runinfo-${{ matrix.python-version }}-${{ steps.job-info.outputs.as-ascii }}-${{ github.sha }}
+        path: |
+          runinfo/
+          .pytest/
```
benclifford, line 101: its not called .pytest any more - see PR #3631

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,166 @@
...
+            amqp_port: Optional[int] = None,
+            client: Optional[Client] = None,
+            **kwargs,
+    ):
+        """
+        Parameters
+        ----------
+
+        endpoint_id:
+            id of the endpoint to which to submit tasks
```
benclifford, line 56: what happens when this is `None`?

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,166 @@
...
+            Port to use when connecting to results queue. Note that the
+            Compute web services only support 5671, 5672, and 443.
+
+        client:
+            instance of globus_compute_sdk.Client to be used by the executor.
+            If not provided, the executor will instantiate one with default arguments.
+
+        kwargs:
+            Other kwargs listed will be passed through to globus_compute_sdk.Executor
+            as is
```
benclifford, line 89: link to where I can find out what those are?

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,166 @@
...
+            user_endpoint_config=user_endpoint_config,
+            label=label,
+            batch_size=batch_size,
+            amqp_port=amqp_port,
+            client=self.client,
+            **kwargs
+        )
+
+    def start(self) -> None:
+        pass
```
benclifford, line 120: In the Parsl configuration model, this separation of init and start means that the code will try to connect to an executor at the point that a user defines a configuration that uses globus compute, rather than at the point that they initialize Parsl with a Globus compute.  I'm not clear on what this Executor does at the point that it is instantiated (for example, making network connections?).

That might be out of line with other code which doesn't do anything until it is `start`ed - for example, when is the interchange started in a HighThroughputExecutor configuration?

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,166 @@
...
+        res_spec = copy.deepcopy(resource_specification or self.resource_specification)
+        # Pop user_endpoint_config since it is illegal in resource_spec for globus_compute
+        if res_spec:
+            user_endpoint_config = res_spec.pop('user_endpoint_config', self.user_endpoint_config)
+        else:
+            user_endpoint_config = self.user_endpoint_config
+
+        self._executor.resource_specification = res_spec
+        self._executor.user_endpoint_config = user_endpoint_config
+        return self._executor.submit(func, *args, **kwargs)
```
benclifford, line 158: This feels a bit horribly thread unsafe, were Parsl ever to get multithreaded submission - the sort of thing DFK.submitter_lock was put in place to deal with.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def shutdown(self):
+        """Clean-up the resources associated with the Executor.
+
+        GCE.shutdown will cancel all futures that have not yet registered with
+        Globus Compute and will not wait for the launched futures to complete.
+        """
+        return self._executor.shutdown(wait=False, cancel_futures=True)
```
benclifford, line 166: shutdown returns a `None`, not whatever this returns (which is hopefully also None)

.github/workflows/gce_test.yaml
benclifford: (does this mean that right now a user shouldn't expect to be able to use an actual Globus Compute but instead needs to install this fork?)

parsl/executors/globus_compute.py
yadudoc: It returns `None`.

.github/workflows/gce_test.yaml
yadudoc: The github action uses `ThreadPoolEngine` because it is lightweight compared to running the default config that uses `GlobusComputeEngine`. The branch that I'm pointing to fixes a `ThreadPoolEngine` only bug. You can run these same tests against a local GlobusCompute endpoint, by setting `GLOBUS_COMPUTE_ENDPOINT=<ep_id>` and I've done those tests against a few configs. I believe Chris (NOAA) has also done a bunch of testing at this point.

christopherwharrop-noaa: I have been testing the `GlobusComputeExecutor` via this branch in combination with `globus-compute-endpoint==2.30.1`.  I am using that release, not a special branch/tag.  The one feature I haven't yet tested is passing in the `user_endpoint_config` at call time rather than at `GlobusComputeExecutor` instantiation time.  What I've done so far is working great!

parsl/executors/globus_compute.py
yadudoc: I'll add an assertion to force this. There's a discrepancy between the executor and the `compute_sdk.Executor` in that an endpoint_id is required at init since we don't have a mechanism in Parsl to update the `endpoint_id` dynamically at runtime.

yadudoc: I've added typeguard to `__init__` to cover this.

parsl/executors/globus_compute.py
yadudoc: Potentially yes. I don't see how this is only a GCE specific problem. Secondly, I think it's safe to wait until users report this as an issue, or ask for it before working on this.

parsl/executors/globus_compute.py
yadudoc: Done

parsl/executors/globus_compute.py
yadudoc: I've moved the actual `globus_compute_sdk.Executor` instantiation to start.

yadudoc:
@benclifford @khk-globus Thanks for the review and all the comments on this issue. I believe I've addressed all of them and at this point, we are only waiting for a release of `globus-compute-endpoint` with fixes from [ComputePR #1689](https://github.com/globus/globus-compute/pull/1689) to update the CI testing target.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,171 @@
...
+        self._executor.user_endpoint_config = user_endpoint_config
+        return self._executor.submit(func, *args, **kwargs)
+
+    def shutdown(self):
+        """Clean-up the resources associated with the Executor.
+
+        GCE.shutdown will cancel all futures that have not yet registered with
+        Globus Compute and will not wait for the launched futures to complete.
+        """
+        self._executor.shutdown(wait=False, cancel_futures=True)
```
khk-globus, line 171: If a user (or Parsl developer) mistakenly calls `.shutdown()` on a GCE instance that hasn't been started, this currently explodes.  What about being more friendly with a simple check:

```python
if not self._executor:
    raise RuntimeError("Executor not started.")
```

That's just a smidgen less cryptic than `None has no attribute .shutdown`

The same suggestion goes for `.submit()`

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,171 @@
...
+        self.batch_size = batch_size
+        self.amqp_port = amqp_port
+        self.client = client
+        self.executor_kwargs = kwargs
+
+        if not _globus_compute_enabled:
+            raise OptionalModuleMissing(
+                ['globus-compute-sdk'],
+                "GlobusComputeExecutor requires globus-compute-sdk installed"
+            )
```
khk-globus, line 110: To elide useless work in the `_globus_compute_enabled is False` case, consider making this the very first statement in `__init__()`.  No need to do the the other work if it's going to be thrown away anyway.  (I get that this change is minimal, but it's the principal of fail early and fail fast that I'm after.)

khk-globus (commented):
I haven't finished reading through the PR, but these are my first thoughts.  I'll chime in on Monday morning (-0400).

parsl/executors/globus_compute.py
benclifford: When this is used from the DFK, this is already in a lock (although as you're both well aware, there is strong community pressure to use pieces of Parsl outside of the DFK). That lock was introduced in PR #625 because executors are not expected to be thread safe on submission, and so in that context this code is not dangerous.

This is a more general backpressure against using this style of API that seems to conflate the state of the submission system as a whole with parameters to a single task execution - I've definitely fixed concurrency bugs in Parsl because of this coding style before, that lead not to Parsl errors but to subtle misexecutions that mostly look plausible.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,171 @@
...
+            If not provided, the executor will instantiate one with default arguments.
+
+        kwargs:
+            Other kwargs listed will be passed through to globus_compute_sdk.Executor
+            as is. Refer to `globus-compute docs
+            <https://globus-compute.readthedocs.io/en/latest/reference/executor.html#globus-compute-executor>`_
+        """
+        super().__init__()
+        self.endpoint_id = endpoint_id
+        self.task_group_id = task_group_id
```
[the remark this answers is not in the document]
benclifford: There's some part of me that looks at this and says " 'just' pass in a Globus Compute `executor` as the single parameter" and let users read the Globus Compute docs to understand the object they are creating. I think right now, users of this code are going to have to be understanding stuff about Globus Compute anyway.

I think it gives a better "lazy path" sustainability story.

parsl/executors/globus_compute.py
```diff
@@ -0,0 +1,171 @@
...
+            batch_size: int = 128,
+            amqp_port: Optional[int] = None,
+            client: Optional[Client] = None,
+            **kwargs,
+    ):
+        """
+        Parameters
+        ----------
+
+        endpoint_id:
```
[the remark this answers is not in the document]
benclifford: more fuel for https://github.com/Parsl/parsl/pull/3607#discussion_r1844962325

parsl/executors/globus_compute.py
benclifford: (see #3492, #1413)

parsl/executors/globus_compute.py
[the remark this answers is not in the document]
yadudoc: @benclifford I like the idea of passing in the `globus_compute_sdk.Executor` itself to the executor, but this would break the `init -> start` separation that `ParslExecutors` follow.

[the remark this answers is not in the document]
yadudoc: We just discussed this, and the consensus is that passing along the executor is the most straightforward solution here.

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+    - name: make deps clean_coverage
+      run: |
+        source .venv/bin/activate
+        make deps
+        make clean_coverage
+
+        # Temporary fix until fixes make it to a release
+        git clone -b main https://github.com/globus/globus-compute.git
+        pip3 install globus-compute/compute_sdk globus-compute/compute_endpoint
```
khk-globus, line 53: Is the `git clone` still necessary after the next GC release, come this Thursday?

yadudoc: No, this is a stopgap until the release.

parsl/executors/globus_compute.py
yadudoc: Fixed now as `self.executor` is set at init.

setup.py
```diff
@@ -45,6 +45,7 @@
     'flux': ['pyyaml', 'cffi', 'jsonschema'],
     'proxystore': ['proxystore'],
     'radical-pilot': ['radical.pilot==1.90', 'radical.utils==1.90'],
+    'globus_compute': ['globus_compute_sdk>=2.27.1'],
     # Disabling psi-j since github direct links are not allowed by pypi
```
khk-globus, line 49: It's been a minute since we started this PR; might increase this to the most up-to-date as of ~now (2.34.0)

.github/workflows/gce_test.yaml
```diff
@@ -0,0 +1,97 @@
...
+        python3 -c "import globus_compute_endpoint; print(globus_compute_endpoint.__version__)"
+        cat << EOF > /home/runner/.globus_compute/default/config.yaml
+        engine:
+          type: ThreadPoolEngine
+          max_workers: 4
+        EOF
+        cat /home/runner/.globus_compute/default/config.yaml
+        mkdir ~/.globus_compute/default/tasks_working_dir
+        globus-compute-endpoint start default
+        globus-compute-endpoint list
```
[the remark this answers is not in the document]
khk-globus: Discussed out-of-band.  This is not feasible to do under the possibility of concurrently running these tests.

setup.py
yadudoc: Fixed.

yadudoc:
@khk-globus I've added some configuration examples and squashed all the previous changes. 

I've opted to keep the `executor.shutdown` mechanism using `result_watcher.shutdown`. I agree that this isn't the best solution here, but the other approach we discussed was causing hangs. Canceling the futures to have the `result_watcher` exit when the exit handler is triggered appears to cause a hang. From the logs, it looks like the atexit handler triggered RW.shutdown, but instead of recognizing that the futures were canceled, it does a blocking wait until the task completes, which then throws an error since the future is already canceled. This smells like a bug. I've not figured out the root cause on the `globus_compute_sdk.Executor` side, but I do not think we should hold this PR up for that.

khk-globus (approved):

PR #3618, opened 2024-09-14
Raising UnsupportedFeatureError instead of InvalidResourceSpecification on passing resource_specification to HTEX

### LINKED ISSUE #3589: rationalize invalid resource specification errors
HTEX's mpi_prefix_composer defines an InvalidResourceSpecification. This is used by the non-MPI parts of htex too. WorkQueue raises an ExecutorError for the same kind of error. I have not investigated the other executors.

This should be rationalised: probably InvalidResourceSpecification should become an error in `parsl.executors.error` and all executors that validate resource specifications should use it.

**Additional context**

This came from review of PR #3582

Harichandra-Prasath:
# Description

Currently in Parsl, HTEX doesnt support resource_specification and upon passing , it will raise InvalidResourceSpecification from mpi_prefix_compose. Similar behaviour can be seen in ThreadPoolExecutor but instead of raising InvalidResourceSpecification , it raises UnsupportedFeatureError from `parsl.executors.error`.  So, Ideally HTEX should do the same,

# Changed Behaviour

From this PR, HTEX will raise the same when resource_specification is passed to it. Moreover UnsupportedFeatureError is changed slightly to support suggestions for the users when for the feature.

# Fixes

Fixes #3589

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/errors.py
+++ b/parsl/executors/errors.py
@@ -28,17 +28,18 @@ class BadStateException(ExecutorError):
 class UnsupportedFeatureError(ExecutorError):
     """Error raised when attemping to use unsupported feature in an Executor"""
 
-    def __init__(self, feature, current_executor, target_executor):
+    def __init__(self, feature, current_executor, suggestion):
         self.feature = feature
         self.current_executor = current_executor
-        self.target_executor = target_executor
+        self.suggestion = suggestion
 
     def __str__(self):
-        if self.target_executor:
+        if self.suggestion:
             return "The {} feature is unsupported in {}. \
-                    Please checkout {} for this feature".format(self.feature,
-                                                                self.current_executor,
-                                                                self.target_executor)
+                    Suggestion: \
+                    {}".format(self.feature,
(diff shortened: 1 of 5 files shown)


benclifford:
@Harichandra-Prasath my intention was that all of these errors are reported as `InvalidResourceSpecification`, not that they all become unsupported feature errors. The user gets more information by being told that this is an invalid resource specification.

Harichandra-Prasath:
So, you suggest parsl should raise `InvalidResourceSpecification` for when resource spech is passed to ThreadPoolExecutor? Any specific reasons? I believe resource_spec is not supported in HTEX and ThreadPoolExecutor. Instead of `UnSupportedFeature` , if `InvalidResourceSpecification` is raised, doesnt it implies whatever shared is invalid rather than  underlying executors doesnt support it.

benclifford:
There are upcoming uses for resource specification in the High Throughput Executor -- see PR #3575 which uses it for task priority.

PR #3625, opened 2024-09-30
Add more usage tracking to htex_local_alternate test

benclifford:
This uses the project name feature introduced in PR #3624 in one of the standard Parsl tests that already had usage turned on.

# Changed Behaviour

More usage tracking on `pytest parsl/tests/ --config parsl/tests/configs/htex_local_alternate.py` which runs in CI and often on my laptop.

## Type of change

- New feature

### DIFF
--- a/parsl/tests/configs/htex_local_alternate.py
+++ b/parsl/tests/configs/htex_local_alternate.py
@@ -66,7 +66,8 @@ def fresh_config():
                         monitoring_debug=False,
                         resource_monitoring_interval=1,
         ),
-        usage_tracking=True
+        usage_tracking=3,
+        project_name="parsl htex_local_alternate test configuration"
     )
 
 


khk-globus (approved):
Just confirming ... this (and the previous `True`) doesn't "phone home," right?

benclifford:
>  this (and the previous True) doesn't "phone home," right?

both do.

danielskatz:
> > this (and the previous True) doesn't "phone home," right?
> 
> both do.

Maybe more accurately would be that they send postcards (UDP packets)

benclifford:
I had an out-of-band conversation with @khk-globus about this. His concern was that the statistics would be in some sense corrupted by the presence of automated Parsl executions. This has been raised by a few people.

My attitude is that we should be encouraging *every* instantiation of a DFK to be reported. We should not be pushing "do we want this in our stats?" decisions onto our end users. @NishchayKarle implemented #3624 to add a user specifiable usage tracking project name. Users can describe their usage of Parsl here, if they want to, and that can help the end users of statistics make decisions about what to do with various counts - especially, if they are concerned that their automated testing will skew some perceived use of the stats, they can disclose that in the project name (for example, `project_name="Ben's automated tests"`)

PR #3627, opened 2024-10-03
Use a single exception class for all resource specification errors

### LINKED ISSUE #3589: rationalize invalid resource specification errors

Harichandra-Prasath:
# Description

There are three places where the executors raise the errors regarding resource specification.

 1.  when passing the resource_specification parameter to the htex, it raises the mpi_prefix_composer InvalidResourceSpecification from validate_resource_spec method. https://github.com/Parsl/parsl/blob/master/parsl/executors/high_throughput/executor.py#L342

  2.  ThreadPoolExecutor also raises the UnsupportedFeatureError when passing resource specification to it, https://github.com/Parsl/parsl/blob/master/parsl/executors/threads.py#L54

  3.  WorkQueueExecutor raises two errors similar to this but as a Base ExecutorError, one when there is invalid resource_specification ; https://github.com/Parsl/parsl/blob/master/parsl/executors/workqueue/executor.py#L418 and another one when the autolabel is false and resource_specification is not right, https://github.com/Parsl/parsl/blob/master/parsl/executors/workqueue/executor.py#L429

This PR rationalize these errors by creating a  `InvalidResourceSpecificationError`  in executors.error and using it across the executors. Moreover `InvalidResourceSpecification` and `MissingResourceSpecification` is removed from mpi_prefix_composer and places where the above two used were replaced with `InvalidResourceSpecificationError`.

# Changed Behaviour

From this PR, All executors will raise `InvalidResourceSpecificationError` for errors regarding resource_spec.

# Fixes

Fixes #3589 

## Type of change

- Bug fix
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/errors.py
+++ b/parsl/executors/errors.py
@@ -1,4 +1,6 @@
 """Exceptions raise by Executors."""
+from typing import Set
+
 from parsl.errors import ParslError
 from parsl.executors.base import ParslExecutor
 
@@ -42,6 +44,17 @@ class UnsupportedFeatureError(ExecutorError):
         else:
             return "The {} feature is unsupported in {}.".format(self.feature,
                                                                  self.current_executor)
+
+
+class InvalidResourceSpecificationError(ExecutorError):
+    """Error raised when Invalid input is supplied via resource Specification"""
+
+    def __init__(self, invalid_keys: Set[str], message: str = ''):
+        self.invalid_keys = invalid_keys
+        self.message = message
+
+    def __str__(self):
+        return f"Invalid Resource Specification Supplied: {self.invalid_keys}. {self.message}"
 
 
 class ScalingFailed(ExecutorError):
--- a/parsl/executors/errors.py
+++ b/parsl/executors/errors.py
@@ -46,7 +46,7 @@ class UnsupportedFeatureError(ExecutorError):
                                                                  self.current_executor)
 
 
-class InvalidResourceSpecificationError(ExecutorError):
+class InvalidResourceSpecification(ExecutorError):
     """Error raised when Invalid input is supplied via resource Specification"""
 
     def __init__(self, invalid_keys: Set[str], message: str = ''):
--- a/parsl/executors/high_throughput/mpi_prefix_composer.py
+++ b/parsl/executors/high_throughput/mpi_prefix_composer.py
@@ -1,5 +1,7 @@
 import logging
-from typing import Dict, List, Set, Tuple
+from typing import Dict, List, Tuple
+
+from parsl.executors.errors import InvalidResourceSpecificationError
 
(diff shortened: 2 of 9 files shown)


parsl/executors/high_throughput/mpi_prefix_composer.py
```diff
@@ -49,7 +31,7 @@ def validate_resource_spec(resource_spec: Dict[str, str]):
                       ))
     invalid_keys = user_keys - legal_keys
     if invalid_keys:
-        raise InvalidResourceSpecification(invalid_keys)
+        raise InvalidResourceSpecificationError(invalid_keys=invalid_keys)
```
benclifford, line 34: other places do not use keyword args `invalid_keys=` here, so it would be more consistent style to leave the parameter as a position argument

parsl/executors/errors.py
```diff
@@ -44,6 +46,17 @@ def __str__(self):
                                                                  self.current_executor)
 
 
+class InvalidResourceSpecificationError(ExecutorError):
```
benclifford, line 49: you could keep this name the same as before this PR, InvalidResourceSpecification without Error on the end. Many exception classes in Parsl omit `Error` or `Exception` on the end

parsl/tests/test_error_handling/test_resource_spec.py
```diff
@@ -26,22 +27,24 @@ def test_resource(n=2):
     fut = double(n, parsl_resource_specification=spec)
     try:
         fut.result()
-    except InvalidResourceSpecification:
-        assert isinstance(executor, HighThroughputExecutor)
+    except InvalidResourceSpecificationError:
+        assert (
+            isinstance(executor, HighThroughputExecutor) or
+            isinstance(executor, WorkQueueExecutor) or
+            isinstance(executor, ThreadPoolExecutor))
```
benclifford, line 34: Now that work queue executor exceptions are covered in this `except` block, the `except UnsupportedFeatureError` block isn't relevant any more, and you can delete it. The only exception to expect here  now is the invalid resource exception. (that is the point of this PR!)

Harichandra-Prasath: I thought it would be better to remove it if we gonna remove the `UnsupportedFeatureError`. I agree that is irrelevant from here but i believe irrespective of this Pr, that should always be true.

benclifford (approved):

PR #3629, opened 2024-10-08
Drop support for end-of-life Python 3.8

benclifford:
Python 3.8 entered the end-of-life phase on 2024-10-07. See https://peps.python.org/pep-0569/

This PR removes CI testing for Python 3.8 and relabels several occurences of 3.8 to 3.9 in the documentation.

# Changed Behaviour

Although this PR will not directly break users using Python 3.8, the removal of testing and the requirement that things work in Python 3.8 means things will likely break soon.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
   main-test-suite:
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


khk-globus (approved):

benclifford:
This needs a manual reconfiguration of github settings to no longer require 3.8 test to pass... I will do that

PR #3630, opened 2024-10-08
Remove claim that Parsl is tested on macOS

benclifford:
## Type of change


benclifford:
I saw enough trouble eg with Outreachy students trying to do things on os x that I'm not even particularly willing to say it has been tested on OS X manually... some people seem to run some pieces of Parsl on OS X, but in the last year I've also encountered plenty of badness.

Issue #3365, opened 2024-04-20
Interchange: remove OUTSTANDING_C command


benclifford:
**Is your feature request related to a problem? Please describe.**

The interchange has an RPC command `OUTSTANDING_C` which returns the number of tasks that are known to the interchange: in some of the queues and on workers.

This information is used by scaling strategy as a count of how many tasks are known to htex, to make scaling decisions and to report to the user.

I think it would be equally acceptable to use a count known on the submit side, and avoid that RPC entirely: htex's `executor.py` is aware of both task submissions and task completions and already stores data about submitted tasks in `executor.tasks`

This would remove a network interaction when a network interaction is not needed.

The value would change slightly due to a "phase shift" in counting: for example, tasks in the zmq pipe from submit side to interchange would be counted as outstanding, while they are not counted that way in current `master`. tasks which have a result in the zmq pipe from interchange to submit side would be counted as outstanding, while they are not counted that way in `master`.

PR #3637, opened 2024-10-14
Count outstanding tasks on submit side rather than in interchange

benclifford:
The count of outstanding tasks may now be higher:

A task will be counted when it enters the task submission ZMQ channel to the interchange, rather than when it reaches the end of the ZMQ channel (which could be a queue under load), and a task will be counted as complete when it arrives back on the submit side, rather than when the interchange places it into the results channel.

It's probably more desirable to count the tasks earlier like this and less desirable to count the results later like this, for scaling purposes - scaling load will be observed before it reaches the interchange, but reduction in load due to completion won't be observed until the result has got all the way through the result queue. I don't think this will affect users in real life - but I don't have any numbers for what delivery times look like across the two queues under heavy load.

This removes an RPC that could be computed locally, and removes a use of the command channel which is generally thread-unsafe.

This removes part of shutdown hangs: when pressing ctrl-c at the right point, shutdown will wait for the strategy thread to shut down, but the strategy thread can get blocked on OUTSTANDING_C, and OUTSTANDING_C will never return because the ctrl-C also made the interchange go away.

This shutdown block is hard to demonstrate in a CI test because there are other blocking actions (such as a MANAGERS rpc) involved in shutdown, and it seems to need the strategy thread to be in just the right place.

Implements #3365

# Changed Behaviour

maybe very slightly different scaling because tasks are counted differently so scaling load is computed differently.
fewer zmq vs threading race conditions around command channel actions

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -582,7 +582,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
     def outstanding(self) -> int:
         """Returns the count of tasks outstanding across the interchange
         and managers"""
-        return self.command_client.run("OUTSTANDING_C")
+        return len(self.tasks)
 
     @property
     def connected_workers(self) -> int:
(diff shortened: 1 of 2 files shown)


rjmello (approved):
Nice simplification.

khk-globus (approved):

PR #3639, opened 2024-10-17
Ensure k8s pod names/labels are RFC 1123 compliant

rjmello:
# Description

- Modified Kubernetes pod names and labels to conform to RFC 1123 for DNS subdomain names and labels, ensuring compliance with Kubernetes naming conventions.
- Modified `KubernetesProvider.submit()` to return an eight-character hex value as the job ID instead of the pod name.
- Replaced the trailing timestamp in the pod name with the job ID to improve collision avoidance.
- Replaced `app` pod label with `parsl-job-id`.
- Updated container name to use job ID.

## Type of change

- Bug fix
- New feature

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -1,10 +1,5 @@
 import logging
-import time
-
-from parsl.providers.kubernetes.template import template_string
-
-logger = logging.getLogger(__name__)
-
+import uuid
 from typing import Any, Dict, List, Optional, Tuple
 
 import typeguard
@@ -12,13 +7,16 @@ import typeguard
 from parsl.errors import OptionalModuleMissing
 from parsl.jobs.states import JobState, JobStatus
 from parsl.providers.base import ExecutionProvider
-from parsl.utils import RepresentationMixin
+from parsl.providers.kubernetes.template import template_string
+from parsl.utils import RepresentationMixin, sanitize_dns_subdomain_rfc1123
 
 try:
     from kubernetes import client, config
     _kubernetes_enabled = True
 except (ImportError, NameError, FileNotFoundError):
     _kubernetes_enabled = False
+
+logger = logging.getLogger(__name__)
 
 translate_table = {
     'Running': JobState.RUNNING,
@@ -161,7 +159,7 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         self.resources: Dict[object, Dict[str, Any]]
         self.resources = {}
 
-    def submit(self, cmd_string, tasks_per_node, job_name="parsl"):
+    def submit(self, cmd_string: str, tasks_per_node: int, job_name: str = "parsl.kube"):
         """ Submit a job
         Args:
              - cmd_string  :(String) - Name of the container to initiate
@@ -173,15 +171,19 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         Returns:
              - job_id: (string) Identifier for the job
         """
+        job_id = uuid.uuid4().hex[:8]
 
-        cur_timestamp = str(time.time() * 1000).split(".")[0]
-        job_name = "{0}-{1}".format(job_name, cur_timestamp)
-
-        if not self.pod_name:
-            pod_name = '{}'.format(job_name)
-        else:
-            pod_name = '{}-{}'.format(self.pod_name,
-                                      cur_timestamp)
+        pod_name = self.pod_name or job_name
+        try:
+            pod_name = sanitize_dns_subdomain_rfc1123(pod_name)
+        except ValueError:
+            logger.warning(
+                f"Invalid pod name '{pod_name}' for job '{job_id}', falling back to 'parsl.kube'"
+            )
+            pod_name = "parsl.kube"
+        pod_name = pod_name[:253 - 1 - len(job_id)]  # Leave room for the job ID
+        pod_name = pod_name.rstrip(".-")  # Remove trailing dot or hyphen after trim
+        pod_name = f"{pod_name}.{job_id}"
 
         formatted_cmd = template_string.format(command=cmd_string,
                                                worker_init=self.worker_init)
@@ -189,7 +191,7 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
         logger.debug("Pod name: %s", pod_name)
         self._create_pod(image=self.image,
                          pod_name=pod_name,
-                         job_name=job_name,
+                         job_id=job_id,
                          cmd_string=formatted_cmd,
                          volumes=self.persistent_volumes,
                          service_account_name=self.service_account_name,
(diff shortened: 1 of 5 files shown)


parsl/tests/test_providers/test_kubernetes_provider.py
```diff
@@ -0,0 +1,101 @@
...
+try:
+    import kubernetes
+except ImportError:
+    KUBERNETES_ENABLED = False
+else:
+    KUBERNETES_ENABLED = True
+
+require_kubernetes = pytest.mark.skipif(
+    not KUBERNETES_ENABLED, reason="kubernetes module not available"
+)
```
benclifford, line 17: this isn't working right by the looks of failing CI.

but I am massively in favour of the tests declaring "we expect this to not work" using `pytest.mark.whatever`  rather than autodetecting that the test will fail and not running the test... (for example, if this autodetection was working properly, CI would be passing without running newly introduced tests, I think?  in which case, you might as well not include them in this PR)

rjmello: Good catch; I forgot to update the `config_local_test` target of Makefile. I also removed the import check.

benclifford:
This has failed CI in the drain test a few times on Python 3.12, in what should be something unrelated. I'll dig into that a bit more and try to understand if that's a drain bug that is just started to show by unrelated chance, or if there's something going on with this PR (I suspect not).

parsl/providers/kubernetes/kube.py
```diff
@@ -161,42 +159,38 @@ def __init__(self,
         self.resources: Dict[object, Dict[str, Any]]
         self.resources = {}
 
-    def submit(self, cmd_string, tasks_per_node, job_name="parsl"):
+    def submit(self, cmd_string, tasks_per_node: int, job_name: str = "parsl.kube"):
```
benclifford, line 162: could all `:str` onto cmd_string too and a return type, for completeness?

parsl/providers/kubernetes/kube.py
```diff
@@ -161,42 +159,38 @@ def __init__(self,
...
-
-        if not self.pod_name:
-            pod_name = '{}'.format(job_name)
-        else:
-            pod_name = '{}-{}'.format(self.pod_name,
-                                      cur_timestamp)
+        pod_name = self.pod_name or job_name
+        pod_name = sanitize_dns_subdomain_rfc1123(pod_name)
+        pod_name = pod_name[:253 - 1 - len(job_id)].rstrip(".")  # Leave room for the job ID
+        pod_name = f"{pod_name}.{job_id}"
```
benclifford, line 178: this mutate after sanitization made me a bit suspicious.

if some user of the provider submits a job name of `""` and has also configured `self.pod_name = ""` then this can result in an invalid pod name, I think.

In current parsl, that won't happen because job_name is always specificed as beginning with `p`, at:

https://github.com/Parsl/parsl/blob/29f960f264c2bb959f65c07faafb07d20402ad5a/parsl/executors/status_handling.py#L249

So it's not a super big deal

benclifford:
> This has failed CI in the drain test a few times on Python 3.12, in what should be something unrelated. I'll dig into that a bit more and try to understand if that's a drain bug that is just started to show by unrelated chance, or if there's something going on with this PR (I suspect not).

I have recreated this a few times in PR #3640 independent of this PR #3639, so that test failure should not count against this PR.

parsl/providers/kubernetes/kube.py
rjmello: Good catch; I'll update the last line to handle this scenario:
```python
pod_name = f"{pod_name}.{job_id}" if pod_name else job_id
```

parsl/providers/kubernetes/kube.py
```diff
@@ -161,42 +159,38 @@ def __init__(self,
         self.resources: Dict[object, Dict[str, Any]]
         self.resources = {}
 
-    def submit(self, cmd_string, tasks_per_node, job_name="parsl"):
+    def submit(self, cmd_string: str, tasks_per_node: int, job_name: str = "parsl.kube"):
         """ Submit a job
         Args:
              - cmd_string  :(String) - Name of the container to initiate
-             - tasks_per_node (int) : command invocations to be launched per node
```
benclifford, line 168: why take this away?

parsl/providers/kubernetes/kube.py
```diff
@@ -322,7 +321,7 @@ def _create_pod(self,
                                                    claim_name=volume[0])))
 
         metadata = client.V1ObjectMeta(name=pod_name,
-                                       labels={"app": job_name},
+                                       labels={"job_id": job_id},
```
benclifford, line 324: might be more informative to get the word parsl in this label name somehow, so that users looking at a pod can see that the job ID comes from parsl rather than from some part of kubernetes - for example, so as to not confuse tech support people when they are told the "job_id" of a pod without the relevant context.

parsl/utils.py
```diff
@@ -380,3 +381,40 @@ def __exit__(
...
+    # Convert to lowercase and replace invalid characters with hyphen
+    sanitized = re.sub(r'[^a-z0-9]', '-', raw_string.lower())
+
+    # Remove consecutive hyphens
+    sanitized = re.sub(r'-+', '-', sanitized)
+
+    # DNS label cannot exceed 63 characters
+    sanitized = sanitized[:63]
+
+    return _strip_non_alphanumeric(sanitized)
```
benclifford, line 401: trim to len 63 *after* removing non-alphanumeric to allow more content to be preserved?

parsl/utils.py
```diff
@@ -380,3 +381,40 @@ def __exit__(
...
+            continue
+        sanitized_segment = sanitize_dns_label_rfc1123(segment)
+        sanitized_segments.append(sanitized_segment)
+
+    sanitized = '.'.join(sanitized_segments)
+
+    # DNS subdomain cannot exceed 253 characters
+    sanitized = sanitized[:253]
+
+    return _strip_non_alphanumeric(sanitized)
```
benclifford, line 420: sanitize_dns_label_rfc1123 should have removed all non-alphanumeric here, right? so this should always be a no-op?

parsl/providers/kubernetes/kube.py
rjmello: Good catch; that was an error.

parsl/providers/kubernetes/kube.py
rjmello: Good idea; I'll change the label key to `parsl-job-id`.

parsl/utils.py
rjmello: Similar to `sanitize_dns_label_rfc1123()`, we could end up with a trailing dot or hyphen after trimming. I included a test for this specific scenario.

parsl/providers/kubernetes/kube.py
rjmello: I just modified the code to use `f"parsl.kube.{job_id}"` if the pod name is empty after sanitization.

benclifford:
I merged this into #3482 (some of which was the base of code for this #3639 PR) so that #3482 has identical kube.py to this PR. The kubernetes test in #3482 passes, with pod names like: `parsl.taskvineexecutor.block-0.8af7938f`

benclifford (approved):

PR #3645, opened 2024-10-22
Correct docstring type to "tuple", from "triple"

benclifford:
## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Interesting; I wonder what the original thought-process was?  I'm guessing from a more mathematical background ...

PR #3647, opened 2024-10-22
Fix too-low heartbeat threshold in test

benclifford:
The threshold was the same as the heartbeat period, leading to races in CI where the manager would be expired before the heartbeat was received.

## Type of change

- Bug fix


khk-globus (approved):

khk-globus:
Urg; I wanted to :boom: react to the PR description.  Alas.

PR #3652, opened 2024-10-23
Remove Channels.abspath that is unused

benclifford:
Part of #3515 

# Changed Behaviour

none - this abstraction was never used

## Type of change

- Code maintenance/cleanup


khk-globus (approved):

yadudoc (approved):
Nice! This should be safe to remove.

PR #3656, opened 2024-10-24
Tidyup and re-enable a LocalChannel test

benclifford:
Previously this was in the /integration/ test directory, which is skipped when running automatic tests (due to a filter in conftest.py)

## Type of change

- Code maintenance/cleanup


parsl/tests/test_channels/test_local_channel.py
```diff
@@ -18,6 +21,7 @@ def test_env():
     print("RC:{} \nSTDOUT:{} \nSTDERR:{}".format(rc, stdout, stderr))
```
khk-globus, line 22: Perhaps remove the print call as part of "tidyup"?

khk-globus (approved):
Approving, but I must admit I'm mildly curious why not just remove the test entirely?  It's been unused ~forever, and channels are imminently going away.

benclifford:
>  I'm mildly curious why not just remove the test entirely? It's been unused ~forever, and channels are imminently going away.

There's a lot of complexity removing channels, because they aren't "not used" - they're tangled throughout our provider layer.

So the process of removing them is probably not a simple "merge these 74 commits in PR #3650 all in one go"

... and at the end of #3650, the `execute_wait` part of `LocalChannel` stays in the codebase, just no longer inside the channel abstraction.

So this test probably should stay around in some form or another to test that `execute_wait` continues to work as #3650 is piecewise merged, and onwards.

benclifford:
(execute_wait ends up in parsl.utils in #3650)

PR #3666, opened 2024-10-28
Remove unused source address on some monitoring messages

benclifford:
This was an extra protocol complication, the source address was never used beyond immediate logging, and the source address sometimes didn't exist with a few different filler values used instead.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Obnoxious and subtle.  Nice find and cleanup!

PR #3667, opened 2024-10-29
Remove deprecated CobaltProvider

benclifford:
This follows on from PR #3101

# Changed Behaviour

Cobalt users will not be able to use new Parsl versions.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Should there have been deprecation notice?  (Is/was there and I'm not seeing it?)

yadudoc (approved):
We had a deprecation notice already. Cobalt is long dead and gone.

PR #3669, opened 2024-10-30
Add `shared_fs` and `staging_required` test markers

yadudoc:
# Description

This PR adds a new `staging_required` marker and a marker to several tests that assume a shared filesystem to work. Similarly, a few tests that use staging are now marked with the `staging_required` marker.

This PR splits out changes from #3607.

# Changed Behaviour

These markers should not affect any test-behavior since none of the test entries in the Makefile make use of this.
These change will kick-in once they are used by `GlobusComputeExecutor` and potentially `KubernetesProvider`
for tests.

## Type of change

- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_bash_apps/test_stdout.py
+++ b/parsl/tests/test_bash_apps/test_stdout.py
@@ -34,6 +34,7 @@ testids = [
 ]
 
 
+@pytest.mark.shared_fs
 @pytest.mark.parametrize('spec', speclist, ids=testids)
 def test_bad_stdout_specs(spec):
     """Testing bad stdout spec cases"""
@@ -91,6 +92,7 @@ def test_bad_stderr_file():
 
 
 @pytest.mark.executor_supports_std_stream_tuples
+@pytest.mark.shared_fs
 def test_stdout_truncate(tmpd_cwd, caplog):
     """Testing truncation of prior content of stdout"""
 
@@ -110,6 +112,7 @@ def test_stdout_truncate(tmpd_cwd, caplog):
         assert record.levelno < logging.ERROR
 
 
+@pytest.mark.shared_fs
 def test_stdout_append(tmpd_cwd, caplog):
     """Testing appending to prior content of stdout (default open() mode)"""
 
--- a/parsl/tests/test_bash_apps/test_stdout.py
+++ b/parsl/tests/test_bash_apps/test_stdout.py
@@ -34,7 +34,6 @@ testids = [
 ]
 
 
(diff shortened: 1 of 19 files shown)


parsl/tests/test_docs/test_workflow1.py
```diff
@@ -22,6 +22,7 @@ def save(message, outputs=[]):
     return 'echo {m} &> {o}'.format(m=message, o=outputs[0])
 
 
+@pytest.mark.shared_fs
```
benclifford, line 25: I think this one doesn't need a shared filesystem: it is the job of the staging provider to bring the "output.txt" file back from the remote side onto the local filesystem, where it will be visible. (that's what the staging_required marker is indicating) ?

parsl/tests/test_bash_apps/test_stdout.py
```diff
@@ -34,6 +34,7 @@ def echo_to_streams(msg, stderr=None, stdout=None):
 ]
 
 
+@pytest.mark.shared_fs
```
benclifford, line 37: this one never gets as far as needing to see any file that was created - all the situations are meant to break

parsl/tests/test_staging/test_docs_1.py
```diff
@@ -12,6 +12,7 @@ def convert(inputs=[], outputs=[]):
 
 
 @pytest.mark.cleannet
+@pytest.mark.shared_fs
```
benclifford, line 15: this should probably be staging_required not shared_fs -- it's the job of the configured staging code to move these files to/from their worker locations, not any shared fs

parsl/tests/test_staging/test_staging_stdout.py
```diff
@@ -15,6 +15,7 @@ def output_to_stds(*, stdout=parsl.AUTO_LOGNAME, stderr=parsl.AUTO_LOGNAME):
     return "echo hello ; echo goodbye >&2"
 
 
+@pytest.mark.shared_fs
```
benclifford, line 18: this is a test of staging, not of the shared file system: the stdout file should be moved by the staging code, not across a shared filesystem

benclifford (changes requested):

parsl/tests/test_docs/test_workflow1.py
yadudoc: Fixed.

yadudoc:
@benclifford I believe all of the bits that you noted are now fixed.

benclifford (approved):

PR #3672, opened 2024-10-31
Make monitoring radio send() signature consistent with radio.send

benclifford:
This is an API tidyup that works towards all monitoring message sends taking a TaggedMonitoringMessage: callers of send are responsible for forming a tagged tuple.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -7,13 +7,12 @@ import queue
 import time
 from multiprocessing import Event, Process
 from multiprocessing.queues import Queue
-from typing import TYPE_CHECKING, Any, Literal, Optional, Tuple, Union, cast
+from typing import TYPE_CHECKING, Literal, Optional, Tuple, Union, cast
 
 import typeguard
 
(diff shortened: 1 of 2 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -202,10 +201,9 @@ def start(self, dfk_run_dir: str, config_run_dir: Union[str, os.PathLike]) -> No
 
         self.hub_zmq_port = zmq_port
 
-    # TODO: tighten the Any message format
-    def send(self, mtype: MessageType, message: Any) -> None:
-        logger.debug("Sending message type {}".format(mtype))
-        self.radio.send((mtype, message))
+    def send(self, message: TaggedMonitoringMessage) -> None:
+        logger.debug("Sending message type {}".format(message[0]))
+        self.radio.send(message)
```
khk-globus, line 206: While here, "might as well" update the `log.debug()` call.

khk-globus (approved):
Subtle change; I'm onboard.

PR #3673, opened 2024-11-01
Move all monitoring.py logging to lazy format style

benclifford:
The is based on a review comment from @khk-globus on PR #3672

# Changed Behaviour

logging formatting *should* come out the same but maybe not

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -288,6 +288,6 @@ def filesystem_receiver(logdir: str, q: "queue.Queue[TaggedMonitoringMessage]", 
                 q.put(cast(TaggedMonitoringMessage, message))
                 os.remove(full_path_filename)
             except Exception:
-                logger.exception("Exception processing %s  - probably will be retried next iteration", filename)
+                logger.exception("Exception processing %s - probably will be retried next iteration", filename)
 
         time.sleep(1)  # whats a good time for this poll?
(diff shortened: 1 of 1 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -280,15 +279,15 @@ def filesystem_receiver(logdir: str, q: "queue.Queue[TaggedMonitoringMessage]",
...
                 with open(full_path_filename, "rb") as f:
                     message = deserialize(f.read())
-                logger.debug(f"Message received is: {message}")
+                logger.debug("Message received is: %s", message)
                 assert isinstance(message, tuple)
                 q.put(cast(TaggedMonitoringMessage, message))
                 os.remove(full_path_filename)
             except Exception:
-                logger.exception(f"Exception processing {filename} - probably will be retried next iteration")
+                logger.exception("Exception processing %s  - probably will be retried next iteration", filename)
```
khk-globus, line 291: Added a space here ...

khk-globus (approved):
Heh, you're so disciplined!  Nothing in piecemeal fashion for you ...

When I'm fairly confident the log will be emitted somewhere, I often will stay with the "format now" approach&nbsp;&mdash;&nbsp;my thinking being that there is no sense in passing extra pointers and/or data structures around in memory.   In this case, that would likely mean that the `.info()`, `.error()`, and `.exception()` calls don't need to be updated.  Meh.

No biggie, just observing

benclifford:
> When I'm fairly confident the log will be emitted somewhere, I often will stay with the "format now" approach

yeah I am pretty fine with that, but I find having two different styles a bit grating.

PR #3682, opened 2024-11-07
Remove Channel.execute_wait environment wait handling, part of #3515

benclifford:
Prior to this PR, Channel.execute_wait took a parameter to modify the unix environment of the newly executed process. This is unused in current Parsl. As part of the implementation of that, LocalChannel cached the unix environment at object initialization, and used that as the base to override with any supplied execute_wait environment parameter.

Post this PR:

The `env` parameter of execute_wait is removed.

LocalChannel does not cache the environment any more. The executed process will inherit the parent process environment as of the point of execution not object initialization.

# Changed Behaviour

This is a behaviour change: if the workflow process changes its unix environment after creating the configuration objects, then prior to this PR, executed processes would not observe that change. Post this PR, executed processes will observe that change. I hope this is not a big deal.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/channels/base.py
+++ b/parsl/channels/base.py
@@ -1,5 +1,5 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
-from typing import Dict, Tuple
+from typing import Tuple
 
 
 class Channel(metaclass=ABCMeta):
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
Hurrah.  :tada: 

And hah:

> Post this PR, executed processes will observe that change. I hope this is not a big deal.

PR #3685, opened 2024-11-07
Remove channel closing at shutdown

benclifford:
LocalChannel, the only extant channel, does not have any close() behaviour.

This is part of #3515 channel removal.

## Type of change

- Code maintenance/cleanup


parsl/channels/base.py
```diff
@@ -81,12 +81,6 @@ def pull_file(self, remote_source: str, local_dir: str) -> str:
         '''
         pass
 
-    @abstractmethod
-    def close(self) -> None:
-        ''' Closes the channel.
-        '''
-        pass
-
```
khk-globus, line 89: This hunk ... is debatable.  Channels are going away anyway, so "whatever."

khk-globus (approved):

parsl/channels/base.py
benclifford: removing it helps flush out where it was called for the typechecker. sometimes.

my thinking mostly is that I would like to remove all the bits of the base channel class piecewise, eventually leaving an empty class that can be removed.

Issue #2554, opened 2023-01-11
Parsl Components for maturity and usage assessment


benclifford:
This issue is an attempt to begin identifying distinct "components" or pieces of functionality within parsl that might be individually assessed for maturity (both by gathering a consensus opinion on component maturity, and through changes in the parsl usage tracking system).

Plugin modules (such as executor and providers) should probably each count as their own component.
In addition, functionality not separated out in that way might sometimes be counted as a component (for example "memoization/checkpointing").

## executors

| Component | Technical sponsor | Maturity comments/rating | 
| ------------- | ------------- | --------- | 
| htex  | -  | commonly used - some options might be individually identified |
| htex | - | scaledown | uncommon |

PR #3690, opened 2024-11-08
Remove provider/channel push/pull of files

benclifford:
This removes push/pull support from channels.

All the HPC providers used channel.push_file to push their batch scripts to the (no-longer) remote system. This has been a basically dead code path since removal of non-remote channels:

The channel is always a LocalChannel now (PR #3677)
The "remote" script directory is always the local script directory (PR #3688) and so `LocalChannel.push_file` always skips making a copy and returns the path it was given without further action.

So all the removed code is a no-op, and this PR simplifies that into nothing.

# Changed Behaviour

Some providers had an option to let users decide if scripts (and other files) would be pushed/pulled to the remote system. Those options are removed by this PR.

## Type of change

- New feature
- Code maintenance/cleanup


benclifford:
@ryanchard any chance you could test the PBS bit of this? (specifically, verify that you can make parsl submit PBS jobs ok still?)

@yadudoc you're listed as technical sponsor (#2554) for a bunch of these changed providers - can you review those?

ryanchard:
I tried this on Polaris to test PBS and it all worked great.

benclifford:
@yadudoc wakey wakey - I'd like to get the rest of channel removal done before the Parsl maintenance work ends in May.

PR #3692, opened 2024-11-08
TaskVine Executor: add vine tune to manager config

colinthomas-z80:
# Description
Make vine_tune config parameters accessible through the manager config while using the TaskVine executor

## Type of change

- New feature

### DIFF
--- a/parsl/executors/taskvine/manager.py
+++ b/parsl/executors/taskvine/manager.py
@@ -44,10 +44,16 @@ def _set_manager_attributes(m, config):
     # Enable peer transfer feature between workers if specified
     if config.enable_peer_transfers:
         m.enable_peer_transfers()
+    else:
+        m.disable_peer_transfers()
 
     # Set catalog report to parsl if project name exists
     if m.name:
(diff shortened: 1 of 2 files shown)


parsl/executors/taskvine/manager.py
```diff
@@ -44,11 +44,17 @@ def _set_manager_attributes(m, config):
     # Enable peer transfer feature between workers if specified
     if config.enable_peer_transfers:
         m.enable_peer_transfers()
+    else:
+        m.disable_peer_transfers()
```
benclifford, line 48: is this an unrelated bugfix?

colinthomas-z80: It is arguably related. This feature is on by default, so default behavior has not changed. However previously a user was not able to "disable" peer transfers. This small fix, and the addition of vine_tune is allowing the same configuration abilities you would get using the base taskvine api

colinthomas-z80: I guess the comment in the code there is a bit misleading to suggest it is not on by default

PR #3694, opened 2024-11-11
Adds clusters option for slurm

### LINKED ISSUE #3675: SlurmProvider cannot find jobs submitted with "#SBATCH --clusters=some_cluster"
**Describe the bug**

Our local HPC ([HPC-UGent Tier-2](https://docs.vscentrum.be/gent/tier2_hardware.html)) groups its nodes in Slurm clusters rather than partitions. Luckily, it is possible to submit jobs from any cluster (e.g., the default login one) to any other cluster using the `#SBATCH --clusters=other_cluster` option. However, this leads to Parsl not recognising submitted blocks and endlessly requesting new resources.

**What I think is happening**
_Bear in mind that my experience with Parsl internals is lackluster at best. The archive attached below contains full runinfo dirs._

1. Job submission works fine, blocks are requested from the cluster(s) specified in `#SBATCH --clusters=..`

2. In the [SlurmProvider._status()](https://github.com/Parsl/parsl/blob/6dd5845baff745ae012b6c681926e3ca20278033/parsl/providers/slurm/slurm.py#L190C1-L190C23) call, it queries the system using `sacct -X --noheader --format=jobid,state%20 --job '{0}'` (by default). This only returns jobs/jobsteps running on the cluster where the main Parsl process lives.

3. As a result, Parsl does not find the submitted jobs and periodically requests scaling out indefinitely (until something triggers shutdown).

In the Parsl log, we find entries like:
```
parsl.providers.slurm.slurm:207 _status DEBUG: Executing sacct -X --noheader --format=jobid,state%20 --job '29835197'
parsl.providers.slurm.slurm:209 _status DEBUG: sacct/squeue returned  
parsl.providers.slurm.slurm:238 _status DEBUG: Updating missing job 29835197 to completed status
```
and
```
parsl.jobs.strategy:217 _general_strategy DEBUG: Executor HighThroughputExecutor has 1 active tasks, 0/0 running/pending blocks, and 1 connected workers
parsl.jobs.strategy:275 _general_strategy DEBUG: Strategy case 2b: active_blocks 0 < max_blocks 1 so scaling out
parsl.jobs.strategy:279 _general_strategy DEBUG: Requesting 1 more blocks
```

Executor-side, things seem to work as expected: a task is only passed to a single worker. Excess job allocations initialise and simply idle.

**A silly patch**

I could fix the issue by simply overwriting `SlurmProvider._cmd = "sacct -X -L --noheader --format=jobid,state%20 --job '{0}'"` (see [--allclusters](https://slurm.schedmd.com/sacct.html#OPT_allclusters)). However, I have no idea of the potential side effects this might cause elsewhere. Also, it is slightly cumbersome to adapt provider behaviour manually.

[test_cases.zip](https://github.com/user-attachments/files/17622999/test_cases.zip)

tylern4:
# Description

Adds the `clusters` option from slurm for sites with [multi-cluster slurm](https://slurm.schedmd.com/multi_cluster.html) setups. 

# Changed Behaviour

Users on federated Slurm clusters will be able to schedule tasks between different clusters within the same slurm instance.

# Fixes

Fixes #3675 

## Type of change

- New feature

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -70,6 +70,8 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         Slurm queue to place job in. If unspecified or ``None``, no queue slurm directive will be specified.
     constraint : str
         Slurm job constraint, often used to choose cpu or gpu type. If unspecified or ``None``, no constraint slurm directive will be added.
+    clusters : str
+        Slurm cluster name, or comma seperated cluster list, used to choose between different clusters in a federated Slurm instance. If unspecified or ``None``, no slurm directive for clusters will be added.
     channel : Channel
         Channel for accessing this provider.
     nodes_per_block : int
@@ -116,6 +118,7 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
                  account: Optional[str] = None,
                  qos: Optional[str] = None,
                  constraint: Optional[str] = None,
+                 clusters: Optional[str] = None,
                  channel: Channel = LocalChannel(),
                  nodes_per_block: int = 1,
                  cores_per_node: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


benclifford (approved):
see review discussion in #3675

Issue #3697, opened 2024-11-14
Interchange SIGABRT on kill_event driven exit (only cosmetic?)


benclifford:
**Describe the bug**
There are a few paths through which the interchange exits. The regular shutdown path, driven by the DFK, is to send a SIGTERM which immediately kills the process.

Another rare path is using `kill_event` which is polled every 10ms, and is set when a particular form of incorrect worker registration is received.

When that kill_event path is taken, the interchange exits with a SIGABRT, placing this (or a variant) on stderr:

```
Exception in thread Interchange-Task-Puller:
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.12/threading.py", line 1012, in run
Exception in thread Interchange-Command:
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self._target(*self._args, **self._kwargs)
  File "/home/benc/parsl/src/parsl/parsl/process_loggers.py", line 26, in wrapped
    self.run()
  File "/usr/local/lib/python3.12/threading.py", line 1012, in run
    r = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/interchange.py", line 213, in task_puller
    self._target(*self._args, **self._kwargs)
    msg = self.task_incoming.recv_pyobj()
  File "/home/benc/parsl/src/parsl/parsl/process_loggers.py", line 26, in wrapped
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = func(*args, **kwargs)
  File "/home/benc/parsl/virtualenv-3.12/lib/python3.12/site-packages/zmq/sugar/socket.py", line 975, in recv_pyobj
        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/benc/parsl/src/parsl/parsl/executors/high_throughput/interchange.py", line 251, in _command_server
    command_req = self.command_channel.recv_pyobj()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    msg = self.recv(flags)
  File "/home/benc/parsl/virtualenv-3.12/lib/python3.12/site-packages/zmq/sugar/socket.py", line 975, in recv_pyobj
          ^^^^^^^^^^^^^^^^
[...]

PR #3698, opened 2024-11-14
Test interchange exit on bad registration message

benclifford:
On certain bad registration messages, the interchange should exit immediately. This tests that.

See #3697 for some bad (cosmetic?) behaviour here - the interchange SIGABRTs on this code path rather than exiting cleanly, and this test includes a commented out assert that could check for clean exit (in addition to checking that the interchange process exits at all)

## Type of change

- Code maintenance/cleanup


parsl/tests/test_htex/test_interchange_exit_bad_registration.py
```diff
@@ -0,0 +1,120 @@
...
+    task_channel.set_hwm(0)
+    task_channel.setsockopt(zmq.SNDTIMEO, channel_timeout)
+    task_channel.connect(f"tcp://127.0.0.1:{task_port}")
+
+    b_msg = json.dumps(msg).encode('utf-8')
+
+    task_channel.send(b_msg)
+
+    # check that the interchange exits within some reasonable time
+    try_assert(lambda: interchange_proc.poll() is not None, "Interchange did not exit after killing watched client process", timeout_ms=5000)
```
khk-globus, line 114: `timeout_ms=5000` is the default, I think?

khk-globus: Intentional, per discussion.

PR #3699, opened 2024-11-14
Remove ability to set monitoring log directory separate from rundir

benclifford:
Prior to this PR, MonitoringHub had a logdir parameter which let the log directory be set separately from the DFK-level run directory.

Other Parsl components generally don't let the user set this unless there is a specific reason.

So this PR removes that feature, reducing the amount of state to be threaded around.

When reading this patch, note that what the DFK calls the rundir is a different directory vs what a Config object calls the rundir.

# Changed Behaviour

This removes a parameter from user facing MonitoringHub configuration.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -279,7 +279,7 @@ class Database:
 class DatabaseManager:
     def __init__(self,
                  db_url: str = 'sqlite:///runinfo/monitoring.db',
-                 logdir: str = '.',
+                 run_dir: str = '.',
                  logging_level: int = logging.INFO,
                  batching_interval: float = 1,
                  batching_threshold: float = 99999,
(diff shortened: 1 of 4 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -279,20 +279,20 @@ class Resource(Base):
...
         self.workflow_start_message: Optional[MonitoringMessage] = None
-        self.logdir = logdir
-        os.makedirs(self.logdir, exist_ok=True)
+        self.run_dir = run_dir
+        os.makedirs(self.run_dir, exist_ok=True)
 
         logger.propagate = False
 
-        set_file_logger("{}/database_manager.log".format(self.logdir), level=logging_level,
+        set_file_logger("{}/database_manager.log".format(self.run_dir), level=logging_level,
```
khk-globus, line 295: In at least one other spot in this PR, this style was transformed to f-string syntax.  A consideration for here as well?  Or vice-versa, and undo the other hunk?

khk-globus (approved):

PR #3701, opened 2024-11-19
Move `execute_task()` to a dedicated module

rjmello:
# Description

Multiple executors use the `execute_task()` function, so moving it to its own module improves code organization and reusability.

Also removed MPI-related code from `execute_task()`, as it's specific to the HTEX.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -23,6 +23,7 @@ import zmq
 
 from parsl import curvezmq
 from parsl.app.errors import RemoteExceptionWrapper
+from parsl.executors.execute_task import execute_task
 from parsl.executors.high_throughput.errors import WorkerLost
 from parsl.executors.high_throughput.mpi_prefix_composer import (
     VALID_LAUNCHERS,
@@ -35,7 +36,7 @@ from parsl.executors.high_throughput.mpi_resource_management import (
 from parsl.executors.high_throughput.probe import probe_addresses
 from parsl.multiprocessing import SpawnContext
 from parsl.process_loggers import wrap_with_logs
-from parsl.serialize import serialize, unpack_res_spec_apply_message
+from parsl.serialize import serialize
 from parsl.version import VERSION as PARSL_VERSION
 
 HEARTBEAT_CODE = (2 ** 32) - 1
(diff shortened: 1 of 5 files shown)


parsl/tests/test_execute_task.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+import pytest
+
+from parsl.executors.execute_task import execute_task
+from parsl.serialize.facade import pack_res_spec_apply_message
+
+
+def addemup(*args: int, name: str = "apples"):
+    total = sum(args)
```
khk-globus, line 10: Pedantically, I observe that `int` is not iterable.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -590,45 +591,13 @@ def update_resource_spec_env_vars(mpi_launcher: str, resource_spec: Dict, node_i
...
-                                           argname, kwargname)
-    exec(code, user_ns, user_ns)
-    return user_ns.get(resultname)
+def _init_mpi_env(mpi_launcher: str, resource_spec: Dict):
+    node_list = resource_spec.get("MPI_NODELIST")
+    if node_list is None:
+        return
+    nodes_for_task = node_list.split(',')
+    logger.info(f"Launching task on provisioned nodes: {nodes_for_task}")
+    update_resource_spec_env_vars(mpi_launcher=mpi_launcher, resource_spec=resource_spec, node_info=nodes_for_task)
```
khk-globus, line 600: I'm in favor of this addition, but I would like to see this in it's own PR.  That is, moving of `execute_task` is the nominal name-sake of this PR (and I think it's a good call to do it!), but this hunk here is an additional change.

Then, seeing precisely when this logic moved in history is more plain.  (FWIW, @benclifford set the [precedent for strict separation](https://youtu.be/WdJwvXOw7Fw?t=61) this past year.)

khk-globus (commented):
A split-up request, but otherwise, looks good!

parsl/tests/test_execute_task.py
benclifford: my interpretation of https://typing.readthedocs.io/en/latest/spec/callables.html#annotating-args-and-kwargs is that `int` here means that this function takes an arbitrary number of positional parameters of type `int`.

khk-globus: Ah; I've been educated.  Thank you; I wasn't aware of that syntax.  "Cheers to visual analysis," as opposed to _trusting the tools_.  :facepalm:

parsl/executors/high_throughput/process_worker_pool.py
rjmello: To avoid temporarily moving `update_resource_spec_env_vars()` to `execute_task.py`, only to move it back to `process_worker_pool.py`, I created a [preliminary PR](https://github.com/Parsl/parsl/pull/3702) to extract the MPI code.

khk-globus (approved):
Awesome.  Thanks for the precursor PR.  Much clearer exactly what the intended change was for this PR.

Thanks for this cleanup!

PR #3702, opened 2024-11-19
Extract MPI code from `execute_task()`

rjmello:
# Description

The `execute_task()` function is used by multiple executors, but the MPI code is specific to HTEX.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -590,28 +590,25 @@ def update_resource_spec_env_vars(mpi_launcher: str, resource_spec: Dict, node_i
         os.environ[key] = prefix_table[key]
 
 
-def execute_task(bufs, mpi_launcher: Optional[str] = None):
-    """Deserialize the buffer and execute the task.
+def _init_mpi_env(mpi_launcher: str, resource_spec: Dict):
+    node_list = resource_spec.get("MPI_NODELIST")
+    if node_list is None:
+        return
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -590,28 +590,28 @@ def update_resource_spec_env_vars(mpi_launcher: str, resource_spec: Dict, node_i
...
 
+def execute_task(bufs: bytes):
+    """Deserialize the buffer and execute the task.
     Returns the result or throws exception.
     """
     user_ns = locals()
     user_ns.update({'__builtins__': __builtins__})
 
-    f, args, kwargs, resource_spec = unpack_res_spec_apply_message(bufs, user_ns, copy=False)
+    f, args, kwargs, resource_spec = unpack_res_spec_apply_message(bufs, copy=False)
```
khk-globus, line 609: As you've just identified that `user_ns` isn't used here, might as well move the creation of `user_ns` closer to where it's needed, and then can remove one of the `.update()` calls.

khk-globus (approved):
Nice cleanup.

parsl/executors/high_throughput/process_worker_pool.py
rjmello: Good call.

rjmello: Agreed; I can do it in a follow up PR.

PR #3704, opened 2024-11-22
Adding support for IPV6 addresses

yadudoc:
# Description

This PR adds a new option: `HighThoughputExecutor(loopback_address: str = "127.0.0.1")` which can be used to specify the internal address used by HTEX for communication between the executor and the interchange. In addition, all ZMQ sockets are now are set to having IPv6 enabled.

The test config `htex_local` has been updated to use `loopback_address="::1"` for testing.

# Changed Behaviour

* IPv6 support is enabled on all HTEX ZMQ components.
* HTEX now supports a `loopback_address` which allows configuring the address used for internal communication

Fixes # (issue)

## Type of change

- New feature

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -6,6 +6,7 @@ The helper to use depends on the network environment around the submitter,
 so some experimentation will probably be needed to choose the correct one.
 """
 
+import ipaddress
 import logging
 import platform
 import socket
@@ -17,7 +18,7 @@ try:
 except ImportError:
     fcntl = None  # type: ignore[assignment]
 import struct
-from typing import Callable, List, Set
+from typing import Callable, List, Set, Union
 
 import psutil
 import typeguard
@@ -156,3 +157,18 @@ def get_any_address() -> str:
     if addr == '':
         raise Exception('Cannot find address of the local machine.')
     return addr
+
+
+def tcp_url(address: str, port: Union[str, int, None] = None) -> str:
+    """Construct a tcp url safe for IPv4 and IPv6"""
+    stripped_address = address.strip('[]')
+    ip_addr = ipaddress.ip_address(stripped_address)
+
+    port_suffix = f":{port}" if port else ""
+
+    if ip_addr.version == 6 and port_suffix:
+        url = f"tcp://[{stripped_address}]{port_suffix}"
+    else:
+        url = f"tcp://{stripped_address}{port_suffix}"
+
+    return url
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -162,6 +162,9 @@ def get_any_address() -> str:
 def tcp_url(address: str, port: Union[str, int, None] = None) -> str:
     """Construct a tcp url safe for IPv4 and IPv6"""
     stripped_address = address.strip('[]')
+    if address == "*":
+        return "tcp://*"
+
     ip_addr = ipaddress.ip_address(stripped_address)
 
     port_suffix = f":{port}" if port else ""
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -161,17 +161,16 @@ def get_any_address() -> str:
 
 def tcp_url(address: str, port: Union[str, int, None] = None) -> str:
     """Construct a tcp url safe for IPv4 and IPv6"""
-    stripped_address = address.strip('[]')
     if address == "*":
         return "tcp://*"
 
-    ip_addr = ipaddress.ip_address(stripped_address)
+    ip_addr = ipaddress.ip_address(address)
 
     port_suffix = f":{port}" if port else ""
 
     if ip_addr.version == 6 and port_suffix:
-        url = f"tcp://[{stripped_address}]{port_suffix}"
+        url = f"tcp://[{address}]{port_suffix}"
     else:
-        url = f"tcp://{stripped_address}{port_suffix}"
+        url = f"tcp://{address}{port_suffix}"
 
     return url
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -86,15 +86,18 @@ GENERAL_HTEX_PARAM_DOCS = """provider : :class:`~parsl.providers.base.ExecutionP
 
     address : string
         An address to connect to the main Parsl process which is reachable from the network in which
-        workers will be running. This field expects an IPv4 address (xxx.xxx.xxx.xxx) (when
-        ``enable_ipv6`` is not set)
+        workers will be running. This field expects an IPv4 or IPv6 address.
         Most login nodes on clusters have several network interfaces available, only some of which
         can be reached from the compute nodes. This field can be used to limit the executor to listen
         only on a specific interface, and limiting connections to the internal network.
         By default, the executor will attempt to enumerate and connect through all possible addresses.
         Setting an address here overrides the default behavior.
-        If ``enable_ipv6`` is set, specify an IPv6 address in brackets, for eg: ``address=[::1]``
         default=None
+
+    loopback_address: string
+        Specify address used for internal communication between executor and interchange.
+        Supports IPv4 and IPv6 addresses
+        default=127.0.0.1
 
     worker_ports : (int, int)
         Specify the ports to be used by workers to connect to Parsl. If this option is specified,
@@ -227,9 +230,6 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
 
         default: empty list
 
-    enable_ipv6: bool
-        Set this flag to enable communication over IPV6 in addition to IPV4.
-        default: False
     """
 
     @typeguard.typechecked
@@ -239,6 +239,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
                  launch_cmd: Optional[str] = None,
                  interchange_launch_cmd: Optional[Sequence[str]] = None,
                  address: Optional[str] = None,
+                 loopback_address: str = "127.0.0.1",
                  worker_ports: Optional[Tuple[int, int]] = None,
                  worker_port_range: Optional[Tuple[int, int]] = (54000, 55000),
                  interchange_port_range: Optional[Tuple[int, int]] = (55000, 56000),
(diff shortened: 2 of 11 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -86,12 +86,14 @@
...
         An address to connect to the main Parsl process which is reachable from the network in which
-        workers will be running. This field expects an IPv4 address (xxx.xxx.xxx.xxx).
+        workers will be running. This field expects an IPv4 address (xxx.xxx.xxx.xxx) (when
+        ``enable_ipv6`` is not set)
         Most login nodes on clusters have several network interfaces available, only some of which
         can be reached from the compute nodes. This field can be used to limit the executor to listen
         only on a specific interface, and limiting connections to the internal network.
         By default, the executor will attempt to enumerate and connect through all possible addresses.
         Setting an address here overrides the default behavior.
+        If ``enable_ipv6`` is set, specify an IPv6 address in brackets, for eg: ``address=[::1]``
```
rjmello, line 96: This PR sets the `ZMQ_IPV6` option for every socket, regardless of the value of `enable_ipv6`. Thus, users can always specify an IPv6 address for the `address` argument, which is used for communication between the interchange and worker worker managers. The `enable_ipv6` argument only affects communication between the executor and interchange.

parsl/executors/high_throughput/executor.py
```diff
@@ -224,6 +226,10 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         Parsl will create names as integers starting with 0.
 
         default: empty list
+
+    enable_ipv6: bool
+        Set this flag to enable communication over IPV6 in addition to IPV4.
+        default: False
```
rjmello, line 232: We may want to name this `enable_ipv6_loopback`, or something else that more accurately describes what the flag does. Alternatively, we could create a `loopback_address` argument and allow users to specify any address in `127.0.0.0/8` or `::1/128`. I prefer the latter option.

parsl/executors/high_throughput/executor.py
```diff
@@ -268,6 +275,13 @@ def __init__(self,
         self.address = address
         self.address_probe_timeout = address_probe_timeout
         self.manager_selector = manager_selector
+        self.enable_ipv6 = enable_ipv6
+        if self.enable_ipv6:
+            # This will force usage of IPV6 for internal communication
+            self._internal_address = "[::1]"
```
rjmello, line 281: Consider renaming this attribute to `self._loopback_address`.

parsl/executors/high_throughput/executor.py
benclifford: this maybe doesn't need "enable" on the start. Name the feature thats being turned on or off, like `encrypted` or `worker_debug`.

use a lowercase `v` in IPv6 - see https://datatracker.ietf.org/doc/html/rfc8200

parsl/tests/test_htex/test_zmq_binding.py
```diff
@@ -103,4 +103,5 @@ def test_limited_interface_binding(cert_dir: Optional[str]):
 
     matched_conns = [conn for conn in conns if conn.laddr.port == ix.worker_result_port]
     assert len(matched_conns) == 1
-    assert matched_conns[0].laddr.ip == address
+    # laddr.ip can return ::ffff:127.0.0.1 when using IPv6
+    assert address in matched_conns[0].laddr.ip
```
benclifford, line 107: is this always exactly the string "127.0.0.1" or the string "::ffff:127.0.0.1"?

parsl/executors/high_throughput/executor.py
```diff
@@ -86,12 +86,14 @@
 
     address : string
         An address to connect to the main Parsl process which is reachable from the network in which
-        workers will be running. This field expects an IPv4 address (xxx.xxx.xxx.xxx).
+        workers will be running. This field expects an IPv4 address (xxx.xxx.xxx.xxx) (when
+        ``enable_ipv6`` is not set)
```
benclifford, line 90: how about describe what is expected in all situations?

parsl/executors/high_throughput/executor.py
benclifford: This style of using brackets here feels a bit awkward to me: it's a piece of URL from https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2 not the standard RFC4291 format of addresses https://datatracker.ietf.org/doc/html/rfc4291#section-2.2

Looking into the future, does that mean some address fields are going to end up taking RFC4291 standard addressing and some are going to be "paste this URL fragment" style addresses? (for example the parameter that sets UDP address that monitoring reports to?)

benclifford:
in my ipv6 only environment (without even a 127.0.0.1 loopback address), the default address detection detects no addresses and gives this error:

```parsl.executors.errors.BadStateException: Executor htex_local failed due to: Error 1:
        EXIT CODE: 2
        STDERR: /home/benc/parsl/virtualenv-3.12/bin/process_worker_pool.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').require('parsl==2024.11.11+desc.2024.11.14a')
usage: process_worker_pool.py [-h] [-d] [-a ADDRESSES] --cert_dir CERT_DIR
                              [-l LOGDIR] [-u UID] [-b BLOCK_ID]
                              [-c CORES_PER_WORKER] [-m MEM_PER_WORKER] -t
                              TASK_PORT
                              [--max_workers_per_node MAX_WORKERS_PER_NODE]
                              [-p PREFETCH_CAPACITY] [--hb_period HB_PERIOD]
                              [--hb_threshold HB_THRESHOLD]
                              [--drain_period DRAIN_PERIOD]
                              [--address_probe_timeout ADDRESS_PROBE_TIMEOUT]
                              [--poll POLL] -r RESULT_PORT --cpu-affinity
                              CPU_AFFINITY
                              [--available-accelerators [AVAILABLE_ACCELERATORS ...]]
                              [--enable_mpi_mode]
                              [--mpi-launcher {srun,aprun,mpiexec}]
process_worker_pool.py: error: argument -a/--addresses: expected one argument
```

(I wanted to force that the workers would not be connecting over 127.0.0.1)  -- i'll try this test in a different way but I think the address detection will still fail to detect the IPv6 addresses in address auto-detection.

parsl/executors/high_throughput/executor.py
benclifford: the presence of these brackets seems to cause some trouble in my environment: this shell command

```
$ cat pytest-parsl/parsltest-current/runinfo/000/submit_scripts/parsl.htex_local.block-0.1732487389.7241297.sh

export JOBNAME=parsl.htex_local.block-0.1732487389.7241297
process_worker_pool.py --debug  -a [2001:db8::2] -p 0 -c 1 -m None --poll 10 --task_port=54325 --result_port=54619 --cert_dir /home/benc/parsl/src/parsl/pytest-parsl/parsltest-current/runinfo/000/htex_local/certificates --logdir=/home/benc/parsl/src/parsl/pytest-parsl/parsltest-current/runinfo/000/htex_local --block_id=0 --hb_period=30  --hb_threshold=120 --drain_period=None --cpu-affinity none  --mpi-launcher=mpiexec --available-accelerators
```

results in a parsed address argument list of:

```
$ grep address pytest-parsl/parsltest-current/runinfo/000/htex_local/block-0/b3cb42e760c3/manager.log 
2024-11-24 22:29:50.151 parsl:887 456 MainThread [INFO]  addresses: :
```

Should I be expecting this to work?

benclifford: Hilariously, because this PR asks the user to specify shell globs and then doesn't make the shell not interpret these as globs, the actual behaviour of what gets passed to process_worker_pool.py is dependent on what particular files you have in your working directory!

For example:

```
benc@parsl-dev-3-12-9310:~/parsl/src/parsl$ echo [2001:db8::2]
[2001:db8::2]
benc@parsl-dev-3-12-9310:~/parsl/src/parsl$ touch 2
benc@parsl-dev-3-12-9310:~/parsl/src/parsl$ echo [2001:db8::2]
2
```

rjmello: The brackets should be not necessary, as pyzmq does not require them.

Re: shell globs, we should add single quotes around the `--addresses` launch command option value.

benclifford: ok, without brackets I can run a test successfully:

```
                address="2001:db8::2",
```

benclifford: if you're going to require valid RFC4291 IP addresses on the submit side, maybe don't need to do any quotation against shell globs - at least, not any more than IPv4 addresses need the same protections.

benclifford:
I didn't try very hard to make it work, but I couldn't get scoped addresses (https://en.wikipedia.org/wiki/IPv6_address#zone_index) to work. I think this probably isn't a big deal, but reporting it here anyway in case you want to document/investigate it more.

parsl/addresses.py
```diff
@@ -156,3 +157,21 @@ def get_any_address() -> str:
     if addr == '':
         raise Exception('Cannot find address of the local machine.')
     return addr
+
+
+def tcp_url(address: str, port: Union[str, int, None] = None) -> str:
+    """Construct a tcp url safe for IPv4 and IPv6"""
+    stripped_address = address.strip('[]')
```
benclifford, line 164: this code is cruft  already maybe? - IP addresses don't have `[]` in or around them

yadudoc: Fair enough, I'll remove that.

parsl/executors/high_throughput/executor.py
yadudoc: I've made changes here to this effect.

rjmello (approved):

parsl/executors/high_throughput/executor.py
benclifford: > What's weird to me (and maybe what you were observing?) is requiring the brackets when it's strictly the IP address.

right: 3986 is a standard for *URLs*. RFC 4291 is a standard for how to write *IPv6 addresses*. RFC 3986 incorporates (the predecessor to) RFC 4291 as to what goes inside the brackets. The brackets come along when you start doing URLs, not addresses.

benclifford: (or I guess, URL-like things... eg SSH targets too sometimes)

LeiGlobus:
> I didn't try very hard to make it work, but I couldn't get scoped addresses (https://en.wikipedia.org/wiki/IPv6_address#zone_index) to work. I think this probably isn't a big deal, but reporting it here anyway in case you want to document/investigate it more.

I think our goal is to get *some* basic variation of ipv6 addressing to work, not '100% support of ipv6', so that should be ok.

In other news, looking at [socket.py in zmq](https://github.com/zeromq/pyzmq/blob/163acca57058ef0dafd3c6bde87c68083c1379f6/zmq/sugar/socket.py#L505), it just does "%s:*" and f'{addr}:{port}', which isn't confidence boosting, making me think I should bracket ipv6 addr to be safe.  However, I'll test both with and without brackets (from the compute side), to possibly eliminate one of the two options (bracket or not to bracket).  

Assuming both work, I'll go with bracketing to be safe, as that's the 'standard' and should be future-proof.

PR #3707, opened 2024-11-29
Move monitoring radios to own modules

benclifford:
In subsequent PRs, these modules will get more radio-specific code, as part of PR #3315 monitoring radio plugin work: for example, the receiving code for each radio should move here too.

I used `git show --color-moved` to check that the moved RadioSender definitions were not changed.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


parsl/monitoring/radios/udp.py
```diff
@@ -0,0 +1,56 @@
...
+        monitoring_url : str
+            URL of the form <scheme>://<IP>:<PORT>
+        timeout : int
+            timeout, default=10s
+        """
+        self.monitoring_url = monitoring_url
+        self.sock_timeout = timeout
+        try:
+            self.scheme, self.ip, port = (x.strip('/') for x in monitoring_url.split(':'))
+            self.port = int(port)
```
khk-globus, line 24: I suppose simpler than using `urllib.parse`.

parsl/monitoring/radios/udp.py
```diff
@@ -0,0 +1,56 @@
...
+            buffer = pickle.dumps(message)
+        except Exception:
+            logging.exception("Exception during pickling", exc_info=True)
+            return
+
+        try:
+            self.sock.sendto(buffer, (self.ip, self.port))
+        except socket.timeout:
+            logging.error("Could not send message within timeout limit")
+            return
```
khk-globus, line 55: These are two different operations, so this organization makes sense, but could also combine them:
```python
try:
    buffer = pickle.dumps(message)
    self.sock.sendto(buffer, (self.ip, self.port))
except socket.timeout:
    logging.error("Could not send message within timeout limit")
except Exception:
    # exc_info=True not necessary, given .exception()?
    logging.exception("Failed to pickle or send")
    # no return necessary ...
```

parsl/monitoring/remote.py
```diff
@@ -7,12 +7,10 @@
...
-from parsl.monitoring.radios import (
-    FilesystemRadioSender,
-    HTEXRadioSender,
-    MonitoringRadioSender,
-    UDPRadioSender,
-)
+from parsl.monitoring.radios.base import MonitoringRadioSender
+from parsl.monitoring.radios.filesystem import FilesystemRadioSender
+from parsl.monitoring.radios.htex import HTEXRadioSender
+from parsl.monitoring.radios.udp import UDPRadioSender
```
khk-globus, line 13: Is there any (ergonomic) value in importing these to `radios/__init__.py` and/or sticking them in the `__all__`?

khk-globus (approved):

parsl/monitoring/radios/udp.py
benclifford: it's probably better to use urllib. I've been hoping to delete the UDP radio entirely though so I've been concentrating on figuring out how to arrange things to put in an ok-ish replacement rather than on making it nicer.

parsl/monitoring/remote.py
benclifford: They should go away entirely Real Soon -- they're needed for an `if`-statement that turns a text string radio name into a `..RadioSender`. That enumerates all known radios - both at the imports and in the `if` statement.

Part of my ongoing work here is to turn that into more object style pluggability so that the remote code will receive an object (probably a `...RadioSender`) as part of the task submission, rather than receiving a text string.

parsl/monitoring/radios/udp.py
benclifford: same comment as above

PR #3709, opened 2024-11-29
Make result queue poll for shutdown, and tidy up at shutdown

benclifford:
This poll happens at the configured htex poll period, which defaults to 10ms.

Under heavy result load, this shoudn't result in much additional load: the poll loop will already be looping a lot to process the results.

Under lower result load, there is a slight observable increase in CPU usage: a 30second sleep task shows this before this PR:

before:
real    0m37.451s
user    0m2.160s
sys     0m0.376s

run 2, user 2.160s
run 3, user 2.116s

and this after this PR:

real    0m37.473s
user    0m2.400s
sys     0m0.557s

Run 2, 2.457s
Run 3, 2.452s

At shutdown, the ZMQ socket for incoming results is closed.

This reduces both the number of threads and number of file descriptors left behind by the `--config local` tests. For example:

$ pytest parsl/tests/test_monitoring/ --config local

Before this PR, at end of test: 32 threads, 451 fds open.

After this PR, at end of test: 1 thread, 48 fds open.

This is part of PR #3397 shutdown tidyup.

# Description

# Changed Behaviour

nothing should be really visible to normal users. Increased CPU usage in the above documented situations.

## Type of change

- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -331,6 +331,9 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
             interchange_launch_cmd = DEFAULT_INTERCHANGE_LAUNCH_CMD
         self.interchange_launch_cmd = interchange_launch_cmd
 
+        self._result_queue_thread_exit = threading.Event()
+        self._result_queue_thread: Optional[threading.Thread] = None
+
     radio_mode = "htex"
     enable_mpi_mode: bool = False
     mpi_launcher: str = "mpiexec"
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -455,9 +458,11 @@ def _result_queue_worker(self):
         """
         logger.debug("Result queue worker starting")
 
-        while not self.bad_state_is_set:
+        while not self.bad_state_is_set and not self._result_queue_thread_exit.is_set():
             try:
-                msgs = self.incoming_q.get()
+                msgs = self.incoming_q.get(timeout_ms=self.poll_period)
+                if msgs is None:  # timeout
+                    continue
```
khk-globus, line 465: Oh good.  This exact procedure has been on my mind for quite awhile, so thank you for this!

I'm thinking at some point we can go farther and simply block completely, but I haven't fully gotten through the shutdown logic for this route yet.  But "on my mind," at some point.

khk-globus (approved):
Love the `.close()` and `.join()` additions.  Good call.

PR #3713, opened 2024-12-03
Remove now-unused channel errors

benclifford:
The final use of channel errors was removed in PR #3690

This is part of #3515 channel removal project

# Changed Behaviour

Any user catching these errors explicitly will now get an error about the exceptions being undefined. Because these don't exist any more, it should be safe to remove those catches.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
:heart:

PR #3714, opened 2024-12-03
Remove script_dir state from channels

benclifford:
This leaves channels as a stateless (and useless) object that can be removed in a subsequent PR. That removal will change/break a lot of config files so I want it to keep it separate from this attribute removal.

# Changed Behaviour

Batch providers will now use their own script_dir attribute, rather than the channel script_dir attribute. Prior to PR #3688 these attributes were paths on different file systems: the submit side (for providers) and the batch job execution side (for channels). PR #3688 removed support for batch job commands to run somewhere that isn't the workflow submit side, and so since then, these two attributes have been roughly equivalent.

In some (obscure seeming cases) they might be different: if a channel is shared between DFKs, then the script_dir used by providers in one DFK will now no longer use the channel script dir set by the other DFK. This was probably a bug anyway but I am noting it because this PR isn't completely behaviour preserving.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/channels/base.py
+++ b/parsl/channels/base.py
@@ -1,22 +1,5 @@
-from abc import ABCMeta, abstractproperty
+from abc import ABCMeta
 
 
 class Channel(metaclass=ABCMeta):
-    @abstractproperty
-    def script_dir(self) -> str:
-        ''' This is a property. Returns the directory assigned for storing all internal scripts such as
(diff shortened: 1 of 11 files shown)


khk-globus (approved):
This is quite a nice tidy-up.  I like it a lot.

PR #3716, opened 2024-12-03
Remove channel classes and parameters

benclifford:
Before this PR, all channel functionality was removed and the channel classes and config parameters remained as stubs waiting to be removed. This PR removes them.

# Changed Behaviour

Any place that a user is specifying `channel=LocalChannel()` in their configuration will now break. That option can be removed by the user because that option should not have any effect in the codebase immediately prior to this PR.

## Type of change

- Code maintenance/cleanup


parsl/tests/test_providers/test_pbspro_template.py
```diff
@@ -12,7 +11,7 @@ def test_submit_script_basic(tmp_path):
     """Test slurm resources table"""
 
     provider = PBSProProvider(
-        queue="debug", channel=LocalChannel()
+        queue="debug"
```
khk-globus, line 14: Couldn't put these style hunks into one line?!  Heh.

khk-globus (approved):

PR #3721, opened 2024-12-11
Fix rounding error in htex block scale in

### LINKED ISSUE #3696: Infinite loop of scaling in and out with HTEX
**Describe the bug**
I've encountered an infinite loop of scaling in and out with the high throughput executor. Blocks get scaled out only to be immediately scaled in as idle, followed by blocks getting scaled out again. 

I believe the issue is due to the block info in the high throughput executor scale in logic having infinite idle time (per their initialization). This seems related to #3353 where I can imagine un-started blocks may not report an idle time. 

The logic is here: https://github.com/Parsl/parsl/blob/2024.11.11/parsl/executors/high_throughput/executor.py#L706-L747
```
        @dataclass
        class BlockInfo:
            tasks: int  # sum of tasks in this block
            idle: float  # shortest idle time of any manager in this block

        # block_info will be populated from two sources:
        # the Job Status Poller mutable block list, and the list of blocks
        # which have connected to the interchange.

        def new_block_info():
            return BlockInfo(tasks=0, idle=float('inf'))

        block_info: Dict[str, BlockInfo] = defaultdict(new_block_info)

        for block_id, job_status in self._status.items():
            if job_status.state not in TERMINAL_STATES:
                block_info[block_id] = new_block_info()

        managers = self.connected_managers()
        for manager in managers:
            if not manager['active']:
                continue
            b_id = manager['block_id']
            block_info[b_id].tasks += manager['tasks']
            block_info[b_id].idle = min(block_info[b_id].idle, manager['idle_duration'])

        # The scaling policy is that longest idle blocks should be scaled down
        # in preference to least idle (most recently used) blocks.
        # Other policies could be implemented here.

        sorted_blocks = sorted(block_info.items(), key=lambda item: (-item[1].idle, item[1].tasks))

        logger.debug(f"Scale in selecting from {len(sorted_blocks)} blocks")
        if max_idletime is None:
            block_ids_to_kill = [x[0] for x in sorted_blocks[:blocks]]
        else:
            block_ids_to_kill = []
            for x in sorted_blocks:
                if x[1].idle > max_idletime and x[1].tasks == 0:
                    block_ids_to_kill.append(x[0])
                    if len(block_ids_to_kill) == blocks:
                        break
```

and this line in particular is the culprit: https://github.com/Parsl/parsl/blob/2024.11.11/parsl/executors/high_throughput/executor.py#L744
```
                if x[1].idle > max_idletime and x[1].tasks == 0:
```

Note that `new_block_info()` initializes `idle` to `float('inf')` and if that default value is used in comparison with `max_idletime` it will always return true and be appended to the block ids to kill.

A one-line fix to
```
                if x[1].idle != float('inf') and x[1].idle > max_idletime and x[1].tasks == 0:
```
has fixed the behavior for me.

**To Reproduce**
I'm not sure how to reproduce this consistently, as it only appeared when moving to a new computing environment.

**Expected behavior**
I expect scaling to behave as normal.

**Environment**
 - OS: Rocky Linux 9
 - Python version: 3.11.8
 - Parsl version: 1.3.0-dev

**Distributed Environment**
- Compute node
- Compute nodes

benclifford:
# Description

PR #2196 calculates a number of blocks to scale in, in the htex strategy, rather than scaling in one block per strategy iteration. However, it rounds the wrong way: it scales in a rounded up, rather than rounded down, number of blocks.

Issue #3696 shows that then resulting in oscillating behaviour: With 14 tasks and 48 workers per block, on alternating strategy runs, the code will either scale up to the rounded up number of needed blocks (14/48 => 1), or scale down to the rounded down number of needed blocks (14/48 => 0).

This PR changes the rounding introduced in #2196 to be consistent: rounding up the number of blocks to scale up, and rounding down the number of blocks to scale down.

# Changed Behaviour

HTEX scale down should oscillate less

# Fixes

Fixes #3696 

## Type of change

- Bug fix


benclifford:
@jrueb not sure if you're still interested, but this PR affects code you introduced so you might want to look at it

parsl/jobs/strategy.py
```diff
@@ -298,8 +298,8 @@ def _general_strategy(self, executors: List[BlockProviderExecutor], *, strategy_
                     # Scale in for htex
                     if isinstance(executor, HighThroughputExecutor):
                         if active_blocks > min_blocks:
-                            excess_slots = math.ceil(active_slots - (active_tasks * parallelism))
-                            excess_blocks = math.ceil(float(excess_slots) / (tasks_per_node * nodes_per_block))
+                            excess_slots = math.floor(active_slots - (active_tasks * parallelism))
+                            excess_blocks = math.floor(float(excess_slots) / (tasks_per_node * nodes_per_block))
                             excess_blocks = min(excess_blocks, active_blocks - min_blocks)
```
khk-globus, line 303: I'm sure it's accounted for prior to this if-branch&nbsp;&mdash;&nbsp;and it exists prior to this hunk&nbsp;&mdash;&nbsp;but the first thing that comes to mind is "is it possible for excess_blocks to become negative?)

The if-condition guarantees that `active_blocks - min_blocks` is at least 1, but I'm not clear on the guarantees of `active_slots - (active_tasks * parallelism)`

khk-globus (approved):

parsl/jobs/strategy.py
benclifford: probably many of these values are assumed to be "sensible" and this code will go wrong if not. see for example this recent issue https://github.com/Parsl/parsl/issues/3726

PR #3724, opened 2024-12-16
Introduce Function Context Feature to TaskVineExecutor

tphung3:
# Description

This PR introduces the function context feature in TaskVine to the TaskVineExecutor. In short, a traditional function can now specify its computational context to be shared across multiple invocations of the same function, allowing drastic improvements in execution performance.

For example, machine learning models, especially LLMs, have a large overhead of model creation to do one inference. Instead of coupling model creation and inferences in the same function, a user now can specify the model creation as the context of the actual inference function, allowing the de-duplication of the model creation cost.

Helpful blog: https://cclnd.blogspot.com/2025/10/reducing-overhead-of-llm-integrated.html.

Tests are added to make sure the feature works as intented.

# Changed Behaviour

TaskVineExecutor now has a new feature allowing functions to specify computational contexts to be shared. 

## Type of change

- New feature


parsl/executors/taskvine/executor.py
```diff
@@ -318,11 +335,35 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+        # If `_parsl_monitoring_task_id` is in kwargs, Parsl monitoring code is enabled.
+        is_monitoring_enabled = '_parsl_monitoring_task_id' in kwargs
+
         # Default execution mode of apps is regular
         exec_mode = resource_specification.get('exec_mode', self.function_exec_mode)
 
+        # Fall back to regular execution if a function is Parsl-monitored as a monitored function is invocation-specific.
+        # Note that it is possible to get the wrapped function by calling the `__wrapped__` attribute when monitoring is enabled.
+        # It will disable the monitoring wrapper code however.
+        if exec_mode == 'serverless' and is_monitoring_enabled:
```
benclifford, line 353: this is a bit suspicious to me -- what is it about monitoring that is special here?

is it that there is a new function definition for every submission? If so, there are other parsl modes like some file staging scenarios, where that can happen too.

Although I think that is also what the block below, with function equality checking and fallback to regular is trying to do too?

benclifford: people don't use file staging much, but it might not be none - for example, I know one user who uses work queue plus zip file staging in Parsl, and that should translate directly to tv + zip file staging.

does that monitoring test detect anything that the line below it doesn't already detect? or is it for getting a nicer error message?

tphung3: It is for getting a nicer message so users may know what's wrong specifically with each case. The monitoring detection can be removed without affecting functionality.

tphung3: Actually the first monitoring check removes the case where the first function runs serverless-ly and subsequent functions run regularly.

parsl/tests/test_vineex/test_function_context.py
```diff
@@ -0,0 +1,75 @@
...
+    y = load_variable_in_serverless('y')
+    return y + x + 1
+
+
+def f_local_compute(x):
+    return x + 1
+
+
+@require_taskvine
+@pytest.mark.taskvine
```
benclifford: this `mark` here is what lets you specify you don't want to test taskvine

parsl/tests/test_vineex/test_function_context.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.mark.parametrize('num_tasks', (1, 50))
+def test_function_context_computation(num_tasks, current_config_name):
+    if current_config_name != 'taskvine_ex':
```
benclifford: if you want to test against a specific configuration, have a look at tests that are `@pytest.mark.local` and have their own `with parsl.load()` in them - rather than using some ambient environment we don't expect the feature to work in.

That would be more consistent with existing tests. parsl/tests/test_monitoring/test_basic.py is a complicated example. or parsl/tests/test_htex/test_priority_queue.py

parsl/executors/taskvine/utils.py
```diff
@@ -83,3 +90,8 @@ def run_parsl_function(map_file, function_file, argument_file, result_file):
     """
     from parsl.executors.taskvine.exec_parsl_function import run
     run(map_file, function_file, argument_file, result_file)
+
+
+def load_variable_in_serverless(var_name):
```
benclifford: docstring if users are meant to be calling this to tell me what it is/why its useful

parsl/executors/taskvine/manager.py
```diff
@@ -405,26 +450,37 @@ def _taskvine_submit_wait(ready_task_queue=None,
...
             logger.debug(f"completed executor task info: {executor_task_id}, {t.category}, {t.command}, {t.std_output}")
 
+            # This serializes the result to a file again.
+            # Also a double serialization, mark as TODO.
+            is_serverless_output_ok = True
+            if exec_mode == "serverless":
+                try:
+                    if t.successful():
+                        _serialize_object_to_file(result_file, t.output)
+                except Exception:
```
benclifford: i think this conceals the exception message/object from the user completely?

parsl/executors/taskvine/executor.py
```diff
@@ -493,6 +589,12 @@ def _serialize_object_to_file(self, path, obj):
             while written < len(serialized_obj):
                 written += f_out.write(serialized_obj[written:])
 
+    def _cloudpickle_serialize_object_to_file(self, path, obj):
```
benclifford: we talked about this somewhere before but I can't remember where: you should be using the parsl serialization libraries not cloudpickle unless you have a specific reason that needs different serialization.

parsl/tests/test_vineex/test_function_context.py
benclifford: I just pushed 3a079dd424 onto this branch to demonstrate what I think this should look like

parsl/executors/taskvine/executor.py
```diff
@@ -330,11 +347,25 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
         logger.debug(f'Got resource specification: {resource_specification}')
 
         # Default execution mode of apps is regular
         exec_mode = resource_specification.get('exec_mode', self.function_exec_mode)
 
+        if exec_mode == 'serverless':
+            if func.__name__ not in self._map_func_names_to_func_details:
+                self._map_func_names_to_func_details[func.__name__] = {'func_obj': func}
+            else:
+                if id(func) != id(self._map_func_names_to_func_details[func.__name__]['func_obj']):
```
benclifford: I think it gives exactly the same comparison but I feel like if you want to compare two objects for actually being the same object, then:

`if func is self._map_func_names_to_func_details[func.__name__]['func_obj']`

might be a more fundamental test?

Although not an issue as the code is written here, it is stylistically resistant to id reuse over time, in case people later on get the idea that they can cache the result of `id()`.

parsl/executors/taskvine/executor.py
```diff
@@ -398,22 +429,74 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+                function_context_input_files = resource_specification.get('function_context_input_files', {})
+
+        logger.debug("Creating executor task {} with function at: {}, argument at: {}, and result to be found at: {}".format(executor_task_id,
+                                                                                                                             function_file,
+                                                                                                                             argument_file,
+                                                                                                                             result_file))
 
         # Serialize function object and arguments, separately
-        self._serialize_object_to_file(function_file, func)
+        if exec_mode == 'regular' or not self._map_func_names_to_func_details[func.__name__]['is_serialized']:
```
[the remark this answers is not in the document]
tphung3: Not true, ignored.

parsl/executors/taskvine/utils.py
```diff
@@ -1,19 +1,23 @@
...
                  category: str,                    # category of Parsl function
-                 input_files: list,                # list of input files to this function
-                 output_files: list,               # list of output files to this function
+                 input_files: List[str],                # list of input files to this function
+                 output_files: List[str],               # list of output files to this function
                  map_file: Optional[str],          # pickled file containing mapping of local to remote names of files
                  function_file: Optional[str],     # pickled file containing the function information
                  argument_file: Optional[str],     # pickled file containing the arguments to the function call
                  result_file: Optional[str],       # path to the pickled result object of the function execution
+                 function_context_file: Optional[List[Any]],  # path to the pickled list of function context details for serverless functions
```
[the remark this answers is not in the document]
tphung3: fix added.

parsl/executors/taskvine/executor.py
```diff
@@ -398,22 +429,74 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
         self._serialize_object_to_file(argument_file, args_dict)
 
+        if exec_mode == 'serverless':
+            if function_context:
+                kwargs['parsl_resource_specification']['function_context'] = function_context
+            if function_context_args:
+                kwargs['parsl_resource_specification']['function_context_args'] = function_context_args
+            if function_context_kwargs:
+                kwargs['parsl_resource_specification']['function_context_kwargs'] = function_context_kwargs
+            if function_context_input_files:
```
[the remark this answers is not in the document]
tphung3: correct, fix added.

parsl/executors/taskvine/executor.py
```diff
@@ -398,22 +429,74 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
-        logger.debug("Creating executor task {} with function at: {}, argument at: {}, \
-                and result to be found at: {}".format(executor_task_id, function_file, argument_file, result_file))
+        if exec_mode == 'serverless':
+            if 'function_context' in resource_specification:
+                if 'function_context_file' not in self._map_func_names_to_func_details[func.__name__]:
+                    function_context = resource_specification.get('function_context')
+                    function_context_args = resource_specification.get('function_context_args', [])
+                    function_context_kwargs = resource_specification.get('function_context_kwargs', {})
+                    function_context_file = os.path.join(self._function_data_dir, func.__name__, 'function_context')
+
```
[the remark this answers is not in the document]
tphung3: the directory creation is already added, also tasks aren't submitted concurrently so ignored.

parsl/tests/test_vineex/test_function_context.py
```diff
@@ -0,0 +1,62 @@
...
+                                      'exec_mode': 'serverless',
+                                      'function_context': f_context,
+                                      'function_context_args': [y, input_file_names],
+                                      'function_context_input_files': function_context_input_files}))
+        total_sum = 0
+        for i in range(num_tasks):
+            total_sum += futs[i].result()
+
+        expected = (f_local_compute(x) + f_context(y, input_file_names)['y']) * num_tasks
+        assert total_sum == expected
```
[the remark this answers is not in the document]
tphung3: future work. comment added to test file.

parsl/executors/taskvine/executor.py
```diff
@@ -398,22 +429,74 @@ def submit(self, func, resource_specification, *args, **kwargs):
...
+        function_context_file = None
+        function_context_input_files = {}
 
         # Get path to files that will contain the pickled function,
         # arguments, result, and map of input and output files
-        function_file = self._path_in_task(executor_task_id, "function")
+        if exec_mode == 'serverless':
+            if 'function_file' not in self._map_func_names_to_func_details[func.__name__]:
+                function_file = os.path.join(self._function_data_dir, func.__name__, 'function')
+                os.makedirs(os.path.join(self._function_data_dir, func.__name__))
```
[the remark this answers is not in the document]
tphung3: no concurrency so ignored.

parsl/tests/test_vineex/test_function_context.py
```diff
@@ -0,0 +1,62 @@
+import os
+
+import pytest
+
+import parsl
```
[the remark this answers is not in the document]
tphung3: fix added.

tphung3:
Copilot PR review is enabled by default on my account, sorry for the unneeded clutter :)

parsl/executors/taskvine/executor.py
benclifford: yeah the `__main__` module is one (maybe the only?) place where pickle doesn't behave as a strict subset of dill.

I'm kinda inclined to change the parsl.serialize behaviour to not do pickle first - because behaviour like you report above with `__main__` is usually wrong anyway, as far as a Parsl-style distributed environment.

tphung3:
> The new test is being ignored at the moment and only ThreadPoolExecutor is testing it. For example, `pytest parsl/tests/ -k "not cleannet" --config parsl/tests/configs/taskvine_ex.py --random-order --durations 10` says [this](https://github.com/Parsl/parsl/actions/runs/20077416213/job/57595367584?pr=3724#step:7:969). Removing @pytest.mark.local causes other Github actions to run this test and fails. Do you have any suggestions?

@benclifford What do you think about this?

tphung3:
This PR is ready to merge. The test runs as shown in the latest TaskVine test suite (https://github.com/Parsl/parsl/actions/runs/28272398110/job/83772232325?pr=3724#step:7:1166)

PR #3727, opened 2024-12-27
Add priority queue to HTEX interchange

matthewc2003:
# Description

Replaces pending_task_queue's queue.Queue() as a SortedList which functions as a priority queue in the interchange. Used to prioritize certain user-labeled tasks when distributing tasks to managers.

# Changed Behaviour

Users can now use the parsl_resource_specification keyword in python apps to specify a 'priority'. Lower value is higher priority, tasks with higher priority will be distributed to workers first.

# Fixes

Part of #3323 work.

## Type of change

- New feature


benclifford:
switching this to draft

PR #3732, opened 2024-12-31
Group user guide pages by subject

WardLT:
# Description

Split user guide documentation based on which part of Parsl they deal with: App definition, Config, Workflow. Also includes an "advanced" section for things that I couldn't figure out how to group and weren't specific enough to get their own header.

1. Divided into four sections, placing simple introductions where necessary
1. Did not reduce any contents. Only reshaping
1. Split long documents into multiple sections so they show up better on the table of contents

# Changed Behaviour

N/A

# Fixes

N/A

## Type of change


benclifford:
renames of rsts turns into broken 404 on inbound hyperlinks/search results, I think. I don't know how much traffic goes into the docs but I would hope that at least search engines are sending people deep in.

A style for some deprecated stuff in the codebase has been labelling something as "this is deprecated. this warning will be removed in 6 months." - and maybe the equivalent here is put almost-empty RSTs that describe where the user's search result has been moved to, with a stated June 2025 removal date.

WardLT:
I also need to update links within the docs to those moved documents.n

WardLT:
I also need to update links within the docs to those moved documents.

WardLT:
Good call about putting in those stubs. I put redirects into the HTML via `<meta>` tags, which Google [interprets as permanent redirects](https://developers.google.com/search/docs/crawling-indexing/301-redirects) in their crawler so long as I make them instant.

I've also left myself a reminder to remove those stubs in July.

benclifford (approved):

PR #3739, opened 2025-01-11
Remove a deliberately non-deterministic test

benclifford:
This test would pass if execution either succeeded or failed with a `DependencyError`. It did not assert any relationship between those two possibilities and whether earlier tasks actually succeeded or failed.

`test_fail_sequence` in the same file tests more strongly whether a failure is correctly propagated.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
:heart:

PR #2025, opened 2021-04-22
Rework __repr__ and __str__ for OptionalModuleMissing

benclifford:
__repr__ should be quasi-machine-readable, and __str__ human readable
    
See PR #1966, commit a423955f4a9e03cf6986a6e21d285cf46fa3bc88, for further context.
    
    Before:
    
    >>> str(e)
    "(['mymod'], 'this test needs demonstrating')"
    >>> repr(e)
    "The functionality requested requires a missing optional module:['mymod'],  Reason:this test needs demonstrating"
    
    After:
    
    >>> str(e)
    "The functionality requested requires missing optional modules ['mymod'], because: this test needs demonstrating"
    >>> repr(e)
    "OptionalModuleMissing(['mymod'], 'this test needs demonstrating')"

## Type of change

- Code maintentance/cleanup

PR #3741, opened 2025-01-13
Change some exceptions to put human readable text in __str__ not __repr__

benclifford:
# Description

See #2025 for more context.

# Changed Behaviour

Rendering of exceptions made with `__repr__` will change to a more machine readable format. Generally Python itself reports exceptions using `__str__`, but DFK task failure history (which goes into the monitoring database) is rendered with `__repr__`. This is probably better behaviour for that history field, which is quasi-machine-readable.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/app/errors.py
+++ b/parsl/app/errors.py
@@ -70,8 +70,8 @@ class MissingOutputs(ParslError):
         self.reason = reason
         self.outputs = outputs
 
-    def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
I was idly wondering about those a couple of months ago.  Question answer: an inconsistency.  Thanks for this!

PR #3742, opened 2025-01-13
Tidy DataFuture argument types

benclifford:
The task ID is never optional: a DataFuture always comes from a Parsl task. (in the same way that an AppFuture always has a task record)

Long ago, Files could be specified as strings. This behaviour was removed slightly less long ago. Right now, typeguard should be detecting if a string is passed, because of the type signature, and so the explicit type check for strings should be unnecessary. This PR removes that explicit type check.

The future being tracked by the data future is not necessarily an AppFuture: it can be any kind of future that completes to indicate the represented file is now available. That is already reflected in the type annotations and code, but this PR fixes the docstring.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/app/futures.py
+++ b/parsl/app/futures.py
@@ -45,7 +45,7 @@ class DataFuture(Future):
 
         Args:
             - fut (Future) : Future that this DataFuture will track.
-                             Completion of `fut` indicates that the data is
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
I very-often forget about typeguard.  Useful functionality.

PR #3743, opened 2025-01-13
Refactor 4 different renderings of task description

benclifford:
Prior to this PR, there were 4 variant ways to render a future (which is usually but not always a Parsl task) as a text string:

1. describing failure of inner future in a join_app with 1 future to join

 - rendered task ID if the inner future has a task record attribute, without checking it is a parsl.dataflow.taskrecord.TaskRecord or that that attribute contains a task id entry.
 - otherwise render as None

2. describing failure of inner future in a join_app with a list of futures to join

 - is meant to do the same as case 1, but is buggy and always renders as None (the wrong object is checked for having a task_record attribute)

3. describing failure of a dependency future

 - rendered task ID if there was a task_record, and the
   future is from the same DFK as is rendering the code
   (so that a task number from a different DFK is not
   rendered, in the corner case of a task from one DFK
   being used as a dependency for a task in another DFK)

 - otherwise, renders the repr of the future

4. describing which dependencies will be waited on when submitting a task to the DFK

 - rendered task ID if the future is an instance of AppFuture or DataFuture
 - otherwise renders the repr of the future

This PR makes a single render method. It is a member of the DFK, because rendering is done in the context of the DFK: two parsl task AppFutures will be rendered differently if they are from this DFK or a different DFK.

This PR implements render_future_description which tries to combine all of the above:

 - if the future is an AppFuture, and is from the same
   DFK, render the task ID
 - otherwise render the repr of the future

# Changed Behaviour

human readable descriptions of dependencies/join inner futures will be changed

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_python_apps/test_fail.py
+++ b/parsl/tests/test_python_apps/test_fail.py
@@ -27,17 +27,26 @@ def test_no_deps():
             pass
 
 
-@pytest.mark.parametrize("fail_probs", ((1, 0), (0, 1)))
-def test_fail_sequence(fail_probs):
-    """Test failure in a sequence of dependencies
-
-    App1 -> App2 ... -> AppN
-    """
-
-    t1_fail_prob, t2_fail_prob = fail_probs
-    t1 = random_fail(fail_prob=t1_fail_prob)
-    t2 = random_fail(fail_prob=t2_fail_prob, inputs=[t1])
+def test_fail_sequence_first():
+    t1 = random_fail(fail_prob=1)
+    t2 = random_fail(fail_prob=0, inputs=[t1])
     t_final = random_fail(fail_prob=0, inputs=[t2])
 
     with pytest.raises(DependencyError):
         t_final.result()
+
+    assert len(t_final.exception().dependent_exceptions_tids) == 1
+    assert isinstance(t_final.exception().dependent_exceptions_tids[0][0], DependencyError)
+    assert t_final.exception().dependent_exceptions_tids[0][1].startswith("task ")
+
+
+def test_fail_sequence_middle():
+    t1 = random_fail(fail_prob=0)
+    t2 = random_fail(fail_prob=1, inputs=[t1])
+    t_final = random_fail(fail_prob=0, inputs=[t2])
+
+    with pytest.raises(DependencyError):
+        t_final.result()
+
+    assert len(t_final.exception().dependent_exceptions_tids) == 1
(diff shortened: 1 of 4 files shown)


parsl/tests/test_python_apps/test_fail.py
```diff
@@ -27,17 +27,26 @@ def test_no_deps():
...
-    t1 = random_fail(fail_prob=t1_fail_prob)
-    t2 = random_fail(fail_prob=t2_fail_prob, inputs=[t1])
+    assert len(t_final.exception().dependent_exceptions_tids) == 1
+    assert isinstance(t_final.exception().dependent_exceptions_tids[0][0], DependencyError)
+    assert t_final.exception().dependent_exceptions_tids[0][1].startswith("task ")
+
+
+def test_fail_sequence_middle():
+    t1 = random_fail(fail_prob=0)
+    t2 = random_fail(fail_prob=1, inputs=[t1])
```
khk-globus, line 45: Much better inductive thinking.  I approve.

parsl/tests/test_python_apps/test_join.py
```diff
@@ -97,7 +97,10 @@ def test_error():
     f = outer_error()
     e = f.exception()
     assert isinstance(e, JoinError)
+
+    assert len(e.dependent_exceptions_tids) == 1
     assert isinstance(e.dependent_exceptions_tids[0][0], InnerError)
+    assert e.dependent_exceptions_tids[0][1].startswith("task ")
```
khk-globus, line 104: From a human perspective, why is this `.startswith` important?  I get _what_ it's verifying, but I'm not clear on the why.

parsl/tests/test_python_apps/test_join.py
```diff
@@ -97,7 +97,10 @@ def test_error():
     f = outer_error()
     e = f.exception()
     assert isinstance(e, JoinError)
+
+    assert len(e.dependent_exceptions_tids) == 1
```
khk-globus, line 101: This makes me wonder if at least 2 and 3 dependent exception tids would also be correct.  Missing test or parameterization opportunity?

khk-globus (approved):
Nice improvement.

parsl/tests/test_python_apps/test_join.py
benclifford: that is the behaviour tested in test_two_errors right after

khk-globus: Ah, indeed.  Already tested.  Should have expanded the file in GH.

PR #3745, opened 2025-01-14
Remove over-verbose debug message

benclifford:
This was introduced in PR #2965 to help debug ZMQ hangs.

PR #3709 changed this loop to run every poll period (by default, every 10ms) rather than every result message.

That turns this log line into a huge source of log noise/volume.

This PR removes that log line - ZMQ debugging must proceed without it.

# Changed Behaviour

parsl.log size from:
pytest parsl/tests/ --config parsl/tests/configs/htex_local.py

before this PR: 7 megabytes
after this PR: 2 megabytes

## Type of change

- Bug fix


khk-globus (approved):
Damnit; I was too slow yesterday!  Other issues cropped up all at the same time.  Alas.  Glory is yours.  Heh.

PR #3747, opened 2025-01-14
Remove description of exception handling from exception text

benclifford:
An exception should describe the exception, not other associated behaviour.

## Type of change


khk-globus (approved):
Hah!  Approved with pleasure.

PR #3748, opened 2025-01-15
Usage Tracking: Only log messages when tracking is enabled

### LINKED ISSUE #3740: Usage tracking code logs that it is sending data even when it isn't
**Describe the bug**
The DFK logs that it is entering usage tracking code in a way that suggests it is logging usage tracking data, even when turned off.

This is a user experience bug.

parsl version 2025.01.06

**Expected behavior**
non-scary logging

NishchayKarle:
Fix the bug where the DFK logs that it is entering usage tracking code in a way that suggests it is logging usage tracking data, even when turned off.

# Changed Behaviour

The DFK will only log messages about usage tracking data being sent when usage tracking is enabled.

# Fixes

Fixes #3740 

## Type of change

- Bug fix

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -47,6 +47,7 @@ from parsl.monitoring import MonitoringHub
 from parsl.monitoring.message_type import MessageType
 from parsl.monitoring.remote import monitor_wrapper
 from parsl.process_loggers import wrap_with_logs
+from parsl.usage_tracking.levels import DISABLED as USAGE_TRACKING_DISABLED
 from parsl.usage_tracking.usage import UsageTracker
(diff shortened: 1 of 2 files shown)


benclifford:
I think I would approach this very differently: I'd move all the logging inside the usage tracking module, rather than expose more usage tracking internals outwards into the DFK. (the DFK historically knows too much stuff -- https://en.wikipedia.org/wiki/God_object -- and I'm slowly chipping away at that over the years)

benclifford (approved):

benclifford:
@yadudoc @khk-globus you might be interested in this test failure (which i'm going to ignore as it's not a required test pass here) https://github.com/Parsl/parsl/actions/runs/12867506582/job/35872290547?pr=3748#step:8:48

PR #3752, opened 2025-01-18
Move htex interchange tasks incoming thread into main thread

benclifford:
This removes one of two non-main threads in the interchange - the task puller thread - and moves the behaviour there (receive a message and put it in an in-process queue) into the main thread (where that in-process queue is ultimately dequeued, anyway)

This is aimed at helping with ZMQ-vs-threads issues within the interchange -- most immediately, clean shutdown #3697

performance notes:

parsl-perf -t 30, my laptop, no logging

before this PR, 2320 tasks/second
post this PR, 2344 tasks/second

cc @rjmello who expressed especial interest in this

# Changed Behaviour

Some performance difference, although the brief measurements above are not concerning.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -132,6 +132,11 @@ class Interchange:
         self.hub_zmq_port = hub_zmq_port
 
         self.pending_task_queue: queue.Queue[Any] = queue.Queue(maxsize=10 ** 6)
+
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
This looks great.  Thank you soooo much for this.

rjmello (approved):
🎸 🚀

PR #3753, opened 2025-01-18
Remove HTEX interchange control channel thread into main thread

benclifford:
# Description

This is a follow on from #3752 - see that PR for some motivation

# Changed Behaviour

some performance reduction

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -9,7 +9,7 @@ import queue
 import sys
 import threading
 import time
-from typing import Any, Dict, List, NoReturn, Optional, Sequence, Set, Tuple, cast
+from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, cast
 
 import zmq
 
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -217,79 +217,68 @@ def _send_monitoring_info(self, monitoring_radio: Optional[MonitoringRadioSender
...
-                    manager_id = s_manager.encode('utf-8')
-                    logger.info("Received HOLD_WORKER for {!r}".format(manager_id))
-                    if manager_id in self._ready_managers:
-                        m = self._ready_managers[manager_id]
-                        m['active'] = False
-                        self._send_monitoring_info(monitoring_radio, m)
+        if self.command_channel in self.socks and self.socks[self.command_channel] == zmq.POLLIN:
+
+            command_req = self.command_channel.recv_pyobj()
+            logger.debug("Received command request: {}".format(command_req))
```
khk-globus, line 230: The previous iteration had this in a try-except; is `zmq.Again` no longer a concern?  (Perhaps it never was?)

Of course, the previous iteration also wrapped _everything_ in the try-except, so :vomiting_face: to that.

parsl/executors/high_throughput/interchange.py
```diff
@@ -217,79 +217,68 @@ def _send_monitoring_info(self, monitoring_radio: Optional[MonitoringRadioSender
...
+                for manager_id in self._ready_managers:
+                    m = self._ready_managers[manager_id]
+                    idle_since = m['idle_since']
+                    if idle_since is not None:
+                        idle_duration = time.time() - idle_since
                     else:
-                        logger.warning("Worker to hold was not in ready managers list")
-
-                    reply = None
+                        idle_duration = 0.0
```
khk-globus, line 248: From practice, I assume that `m['idle_since']` is guaranteed to exist here (given that this hasn't changed), but mildly odd to check for `None`ness then.  (Not a comment on you, just, "hmm.")

Moving on, a clarity suggestion/bikeshed to inline the if.  "The point is that `idle_duration` is set":
```python
idle_since = time.time() - idle_since if idle_since else 0.0
```
Truthy comparison is fine here?

khk-globus (approved):
Looks good; small non-blocker comments for consideration.

parsl/executors/high_throughput/interchange.py
benclifford: `None` is a meaningful value here, meaning not idle.

parsl/executors/high_throughput/interchange.py
benclifford: this was part of polling-to-exit handling which was mostly removed by #2429 and #2438 - it could have removed this try/except series, I think, but did not.

parsl/executors/high_throughput/interchange.py
```diff
@@ -217,79 +217,68 @@ def _send_monitoring_info(self, monitoring_radio: Optional[MonitoringRadioSender
...
-
-                elif command_req.startswith("HOLD_WORKER"):
-                    cmd, s_manager = command_req.split(';')
-                    manager_id = s_manager.encode('utf-8')
-                    logger.info("Received HOLD_WORKER for {!r}".format(manager_id))
-                    if manager_id in self._ready_managers:
-                        m = self._ready_managers[manager_id]
-                        m['active'] = False
-                        self._send_monitoring_info(monitoring_radio, m)
+        if self.command_channel in self.socks and self.socks[self.command_channel] == zmq.POLLIN:
```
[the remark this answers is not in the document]
benclifford: its not necessary, but more a stylistic thing

parsl/executors/high_throughput/interchange.py
khk-globus: Yes; but I was speaking pragmatically.  As in `idle_since` is only otherwise "falsey" if it's 0.0, but presumably idle_since is an absolute point in time, and thus will never be less than `time.time()` (+NTP drift) when the script started.

No matter, just a minor suggestion.

PR #3754, opened 2025-01-21
Adding dynamic files

astro-friedel:
# Description

Parsl has an issue with files that are created by an app when they are not specified in the arguments when the app is called. For example:

```python
@python_app
def process(inputs=[], outputs=[]):
    with open(inputs[0], 'r') as fh:
        lines = fh.readlines()
    for i in [1,2,3]:
        with open(f"dat.{i}.log", 'w') as wh:
            wh.write("xyz\n")
        outputs.append(File("dat.{i}.log"))

@python_app
def compact(inputs=[], outputs=[]):
    with open(outputs[0], 'w') as fh:
        for f in inputs:
            lines = open(inputs[i]).readlines()
            fh.write(lines)

outs = []
p = process(inputs=File("input.dat"), outputs=outs)
c = compact(inputs=p.outputs, outputs=File("compact.log"))
```
While `process` will properly write the log files to the list. `compact` is unlikely to see them. This is because Parsl sees the `outputs` from `process` as an empty list and does not know that any files are actually created. This causes the constructed DAG to not create any connection between `process` and `compact`, allowing them to run in parallel, instead of the expected serial.

To fix this I have created the `DynamicFileList` class. This class behaves just like a list, but is also a future. If `outp` in the above example is an instance of this class, then Parsl will know that there are files being created and will make a dependency in the DAG for `compact`, forcing it to not execute until `process` is complete.

There is also a new wrapper in apps called the `bash_watcher`. This wrapper was created as a way to implement using `DynamicFileLists` with `bash_apps`.

# Changed Behaviour

Using the `DynamicFileList` class will allow for the user to write code which has an unknown number of outputs be properly tracked and linked, and run as expected.

## Type of change

- New feature


parsl/__init__.py
```diff
@@ -46,6 +46,7 @@
     'bash_app',
     'python_app',
     'join_app',
+    'watch_app',
```
benclifford, line 49: I don't see where this is defined

parsl/app/futures.py
```diff
@@ -30,15 +38,23 @@ def parent_callback(self, parent_fu):
...
+                    self.file_obj.timestamp = datetime.fromtimestamp(stat(self.file_obj.filepath).st_ctime, tz=timezone.utc)
+                if not self.file_obj.size:
+                    self.file_obj.size = stat(self.file_obj.filepath).st_size
+                if not self.file_obj.md5sum:
+                    self.file_obj.md5sum = md5(open(self.file_obj, 'rb').read()).hexdigest()
+            self.data_flow_kernel.register_as_output(self.file_obj, self.app_fut.task_record)
 
     @typeguard.typechecked
-    def __init__(self, fut: Future, file_obj: File, tid: int) -> None:
+    def __init__(self, fut: Future, file_obj: File, dfk: "DataFlowKernel", tid: Optional[int] = None, app_fut: Optional[Future] = None) -> None:
```
benclifford, line 57: PR #3742 made tid non-optional because it is always specified - this looks like a regression of that PR?

If so, it might be true that app_fut is also always available and doesn't need to be optional? Although skimming the rest of the PR, app_fut is never specified / always `None`?

parsl/app/futures.py
```diff
@@ -56,8 +72,18 @@ def __init__(self, fut: Future, file_obj: File, tid: int) -> None:
         self._tid = tid
         self.file_obj = file_obj
         self.parent = fut
-
+        if app_fut:
```
benclifford, line 75: see other comment about app_fut always being `None` - if so, this `if` doesn't need to exist.

parsl/app/futures.py
```diff
@@ -56,8 +72,18 @@ def __init__(self, fut: Future, file_obj: File, tid: int) -> None:
...
+            self.app_fut = app_fut
+        else:
+            self.app_fut = fut
+        self.data_flow_kernel = dfk
         self.parent.add_done_callback(self.parent_callback)
+        # only capture this if needed
+        if self.data_flow_kernel.file_provenance and self.file_obj.scheme == 'file' and os.path.exists(file_obj.path):
+            file_stat = os.stat(file_obj.path)
```
benclifford, line 83: this could be refactored with the code at line 48 https://github.com/Parsl/parsl/pull/3754/files#diff-8ddcd7398ccac61957e74e4e419aafc09215aa3723bcb845503481bcc6edd4c9R48 to have a single provenance update on a DataFuture?

parsl/app/futures.py
```diff
@@ -76,6 +102,30 @@ def filename(self):
...
         return self.filepath
 
+    @property
+    def uuid(self):
+        """UUID of the File object this datafuture represents."""
+        return self.file_obj.uuid
+
+    @property
+    def timestamp(self):
+        """Timestamp when the future was marked done."""
```
benclifford, line 112: this is coming from the files st_ctime, I think? which is the first "last modification time" that was observed by Parsl for this file? not coming from future completion with the python process?

parsl/app/watcher.py
```diff
@@ -0,0 +1,96 @@
...
+    *args : list
+        Arbitrary arguments to pass to the bash_app.
+    **kwargs : dict
+        Arbitrary keyword arguments to pass to the bash_app.
+
+    Returns
+    -------
+    The return code of the wrapped `bash_app`
+    """
+    from watchdog.events import FileSystemEventHandler
```
benclifford, line 40: I'm not clear about the semantics here:

this is running concurrently with the bash app, but I think the end effect is that you end up with a list that is the equivalent of the list of files at the end of the bash app minus the list of files already present at the start of the bash app, with any dynamic behaviour during the execution of the app being otherwise irrelevant?

will this give weird results if several watcher-wrapped concurrent apps run watching the same directory?

parsl/app/watcher.py
```diff
@@ -0,0 +1,96 @@
+"""This module provides the bash_watch_app function. This function wraps a bash_app and captures
```
benclifford, line 1: is this anything bash_app specific? (vs wrapping a python app?)

parsl/curvezmq.py
```diff
@@ -144,7 +144,7 @@ def _start_auth_thread(self) -> ThreadAuthenticator:
         auth_thread.start()
         # Only allow certs that are in the cert dir
         assert self.cert_dir  # For mypy
-        auth_thread.configure_curve(domain="*", location=self.cert_dir)
+        auth_thread.configure_curve(domain="*", location=str(self.cert_dir))
```
benclifford, line 147: what's this for?

parsl/data_provider/files.py
```diff
@@ -96,3 +109,8 @@ def filepath(self) -> str:
             return self.path
         else:
             raise ValueError("No local_path set for {}".format(repr(self)))
+
+    @property
+    def timesatmp(self) -> Optional[str]:
```
benclifford, line 114: typo in name?

parsl/monitoring/monitoring.py
```diff
@@ -46,7 +46,8 @@ def __init__(self,
                  logging_endpoint: Optional[str] = None,
                  monitoring_debug: bool = False,
                  resource_monitoring_enabled: bool = True,
-                 resource_monitoring_interval: float = 30):  # in seconds
+                 resource_monitoring_interval: float = 30,
+                 file_provenance: bool = False):  # in seconds
```
benclifford, line 50: that comment belongs to the original line it was on!

parsl/__init__.py
astro-friedel: This was a left over, it has been removed.

parsl/app/futures.py
astro-friedel: The regression appears to be caused by me merging master into my branch and my git deciding to keep my version since there were other changes to the line. For the app_fut there is an instance where it is specifically needed (the reason why I added the parameter). The code in lines 969-974 of dataflow/dflow.py

parsl/app/futures.py
astro-friedel: There is a least one instance where it is (see above comment)

parsl/app/futures.py
astro-friedel: I put that in the init so that data about an actual file on disk (as opposed to a true future) would be captured as soon as possible. But I think we could defer that until the parent_callback is called at the end of the app.

parsl/app/futures.py
astro-friedel: Yes, the timestamp is meant to be related to the file, not the future. With the change in code from above it is the only "last modification time" Parsl observes.

parsl/app/watcher.py
astro-friedel: Yes, you end up with the file list you specified. Any dynamic behavior is irrelevant with regard to constructing a list of files created by the app. Yes, it could give "weird"/redundant results if multiple watchers are monitoring the same set of directories. I will add that as a warning/caveat.

parsl/app/watcher.py
astro-friedel: It is not bash_app specific and could be used by a python_app. However, it is less efficient, prone to potential issues (as you mentioned above) and would be redundant (as it attempts to do what the python_app will do in the background anyway).

parsl/data_provider/files.py
astro-friedel: yes, thanks for catching that

parsl/monitoring/monitoring.py
astro-friedel: Yes it does.

parsl/app/futures.py
```diff
@@ -30,15 +38,23 @@ def parent_callback(self, parent_fu):
         Returns:
             - None
         """
-
         e = parent_fu._exception
         if e:
             self.set_exception(e)
         else:
             self.set_result(self.file_obj)
+            # only update the file object if it is a file
```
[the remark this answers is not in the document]
astro-friedel: Good point. I have re-ordered the code.

parsl/curvezmq.py
astro-friedel: I was getting an error from mypysince we allow cert_dir to be Pathlike, but configure_curve wants a string. I could make this a separate PR if needed.

benclifford: ok. that definitely should be separate from this PR - @rjmello wrote that bit and might have some thoughts.

Are you using a different version of mypy compared to parsl CI?

rjmello: The `location` argument accepts a `str` or `PathLike` object (see [pyzmq code](https://github.com/zeromq/pyzmq/blob/01bd01c77277f16f714807a3ae4769f5b726710a/zmq/auth/base.py#L134)). This sounds like a mypyism.

astro-friedel: This is not true for all versions of pyzmq. I am running Ubuntu 24.04 with Python 3.12.6, with a system installed pyzmq (python3-zmq for Ubuntu) 24.0.1 In this version the `configure_curve` call maps to [here](https://github.com/zeromq/pyzmq/blob/6ff49f843f1f4932422aa68f4cc286fbd7e9e203/zmq/auth/thread.py#L215) where the location argument accepts only a `str`

I will remove this change from this PR, but I think this will need to be looked at more deeply.

rjmello: That's a fair point since we officially support [pyzmq>=17.1.2](https://github.com/Parsl/parsl/blob/7241f29a293c23453ef1880691f90a2bda4c58c3/requirements.txt#L1), which is older than your version and accepts the [same type](https://github.com/zeromq/pyzmq/blob/027c800ddf63135a6d706841edb3c2b44ac43a37/zmq/auth/thread.py#L177). I'm personally fine with this change and maybe a brief comment.

PR #3758, opened 2025-01-27
Delete redundant checkpoint test

benclifford:
This test checks that the checkpoint directory exists.

test_python_checkpoint_1.py checks more thoroughly that the actual checkpoint files exist, inside that directory

# Changed Behaviour

none - no user facing change, and test coverage should not be reduced here

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
And removes a silly test `print()` as well.  :magic_wand:

PR #3759, opened 2025-01-27
Load checkpoints in memoizer code, rather than DFK

benclifford:
This is first of a set of PRs to move checkpoint handling code out of the DFK in eventually-pluggable memoizers.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -28,7 +28,7 @@ from parsl.config import Config
 from parsl.data_provider.data_manager import DataManager
 from parsl.data_provider.files import File
 from parsl.dataflow.dependency_resolvers import SHALLOW_DEPENDENCY_RESOLVER
-from parsl.dataflow.errors import BadCheckpoint, DependencyError, JoinError
+from parsl.dataflow.errors import DependencyError, JoinError
 from parsl.dataflow.futures import AppFuture
 from parsl.dataflow.memoization import Memoizer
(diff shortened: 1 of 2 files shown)


parsl/dataflow/dflow.py
```diff
@@ -161,13 +161,13 @@ def __init__(self, config: Config) -> None:
                                  workflow_info))
 
         if config.checkpoint_files is not None:
-            checkpoints = self.load_checkpoints(config.checkpoint_files)
+            checkpoint_files = config.checkpoint_files
         elif config.checkpoint_files is None and config.checkpoint_mode is not None:
-            checkpoints = self.load_checkpoints(get_all_checkpoints(self.run_dir))
+            checkpoint_files = get_all_checkpoints(self.run_dir)
         else:
```
khk-globus, line 167: Not the focus of this PR per se, but I observe that given the first `if` condition, the first `elif` clause is redundant.  Of _course_ `checkpoint_files is None`.

khk-globus (approved):
A subtly nice alteration.  No need to update any tests?  That's perhaps odd.

benclifford:
The checkpointing testing is all blackbox testing that doesn't need changing for the same reason that this code doesn't change the user-facing interface.

As an end product of this sequence of PRs, the checkpoint/memoization code might end up modularised in a way that makes it testable as a module separately from having a full DFK.

PR #3765, opened 2025-02-12
Update globus_compute.py

noelppp:
Add two arguments `storage_access` and `working_dir`.
@yadudoc 

# Description

These arguments are needed to allow custom staging providers.
https://parsl.readthedocs.io/en/stable/userguide/configuration/data.html 

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/globus_compute.py
+++ b/parsl/executors/globus_compute.py
@@ -58,7 +58,7 @@ class GlobusComputeExecutor(ParslExecutor, RepresentationMixin):
 
         storage_access:
             a list of staging providers that will be used for file staging
-            
+
         working_dir:
             The working dir to be used for file staging
         """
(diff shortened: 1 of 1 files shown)


yadudoc (commented):
Hi @noelppp, thanks for taking the time to push the PR. 
Adding the `storage_access` attribute makes sense.  

I'm concerned about `working_dir` though. If `working_dir` were set here I would expect all functions submitted to run in that directory, while I do not see any calls to the GCExecutor to set this up. Have you tested `working_dir` by itself?

benclifford:
@yadudoc that's not what working_dir means in other executors -- it's only used for file staging placement.

yadudoc:
@noelppp, I'm so sorry! My response on this PR sat unsent for a week! 

I see how `working_dir` in this context is only related to staging. I'd recommend adding a line documenting these two new arguments to avoid confusion with the `working_dir` that could be set on the `globus-compute-endpoint` configurations.

yadudoc (approved):

PR #3767, opened 2025-02-12
Update plotly usage

khk-globus:
Plotly released 6.0.0 a couple of weeks ago, and dropped a deprecated item we use.  Accordingly, alter usage to the updated API.

# Changed Behaviour

No changed behavior.

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
Hopefully, this is the extent of breaking changes affecting us from plotly's `6.0.0` update.

PR #3777, opened 2025-02-17
Update kube.py

noelppp:
# Description

Add `extra_requests` and `extra_limits` dicts for generic support of Kubernetes resource specifications like `n_gpus`
https://github.com/Parsl/parsl/pull/1798

## Type of change
- Code maintenance/cleanup

### DIFF
--- a/parsl/providers/kubernetes/kube.py
+++ b/parsl/providers/kubernetes/kube.py
@@ -102,8 +102,8 @@ class KubernetesProvider(ExecutionProvider, RepresentationMixin):
                  max_mem: str = "500Mi",
                  init_cpu: float = 1,
                  init_mem: str = "250Mi",
-                 extra_requests: Optional[Dict[str, str]] = {},
-                 extra_limits: Optional[Dict[str, str]] = {},
+                 extra_requests: Optional[Dict[str, str]] = None,
+                 extra_limits: Optional[Dict[str, str]] = None,
                  parallelism: float = 1,
                  worker_init: str = "",
                  pod_name: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


parsl/providers/kubernetes/kube.py
```diff
@@ -98,6 +102,8 @@ def __init__(self,
                  max_mem: str = "500Mi",
                  init_cpu: float = 1,
                  init_mem: str = "250Mi",
+                 extra_requests: Optional[Dict[str, str]] = {},
+                 extra_limits: Optional[Dict[str, str]] = {},
```
rjmello, line 106: I suggest adding requests and limits that you'd like to configure as init parameters, like we do with `max_cpu` and `init_mem`, rather than through an open-ended dict object.

rjmello (changes requested):

parsl/providers/kubernetes/kube.py
rjmello: Ah, makes sense. Another option might be something like `hugepages` and `hugepages_size` in addition to a `custom_requests` and `custom_limits` parameters, but I'm cool with either route.

rjmello (commented):
This should be all set once you address the mypy errors.

parsl/providers/kubernetes/kube.py
rjmello, line 106:
```suggestion
                 extra_requests: Optional[Dict[str, str]] = None,
                 extra_limits: Optional[Dict[str, str]] = None,
```

parsl/providers/kubernetes/kube.py
```diff
@@ -142,6 +148,8 @@ def __init__(self,
         self.max_mem = max_mem
         self.init_cpu = init_cpu
         self.init_mem = init_mem
+        self.extra_requests = extra_requests
+        self.extra_limits = extra_limits
```
rjmello, line 152:
```suggestion
        self.extra_requests = extra_requests if extra_requests else {}
        self.extra_limits = extra_limits if extra_limits else {}
```

rjmello (approved):
Nice!

PR #3778, opened 2025-02-17
Refactor result loop, remove timeout

khk-globus:
There is no need to iterate the loop to send a heartbeat.  Instead, use a dedicated thread (heartbeater), and no longer send messages to ZMQ in bulk.  ZMQ already handles messages messages performantly and `_multipart()` is for an _atomic_ send of multiple messages or message parts.

This reduces the complexity of the loop, and also reduces the idle CPU load.  On my local hardware, it reduces the idle load by about 40% (not scientific; just looking at htop).

In future work, can replace `result_outgoing.send_multipart()` with just `result_outgoing.send()`

# Changed Behaviour

No changed user-facing behavior.  Potentially visible is less idle CPU use.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -403,44 +403,27 @@ class Manager:
         result_outgoing.connect(self._result_q_url)
         logger.info("Manager result pipe connected to interchange")
 
-        push_poll_period = max(10, self.poll_period) / 1000    # push_poll_period must be atleast 10 ms
-        logger.debug("push poll period: {}".format(push_poll_period))
-
-        last_beat = time.time()
-        last_result_beat = time.time()
-        items = []
-
         while not self._stop_event.is_set():
+            logger.debug("Starting pending_result_queue get")
             try:
-                logger.debug("Starting pending_result_queue get")
-                r = self.task_scheduler.get_result(block=True, timeout=push_poll_period)
-                logger.debug("Got a result item")
-                items.append(r)
-            except queue.Empty:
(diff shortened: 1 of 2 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -524,18 +507,24 @@ def start(self):
...
         logger.info("Manager threads started")
 
         # This might need a multiprocessing event to signal back.
         self._stop_event.wait()
         logger.critical("Received kill event, terminating worker processes")
 
+        # Invite blocking threads to quit
+        self.pending_result_queue.put(None)
+
+        thr_heartbeater.join()
```
benclifford, line 527: i haven't verified that this exit really happens - in real life worker pools are usually killed by external termination and I'm not sure how well tested this other path is before this PR.

benclifford (approved):

parsl/executors/high_throughput/process_worker_pool.py
khk-globus: Meaning, add a test?  I can do that.  However, for that precise reason, I also wasn't immediately worried about this exit dance, especially in light of the flux that is currently happening in this file.  The larger point of interest for me was the unnecessary iteration during idle.  If there's nothing to do ... then don't do anything.

benclifford: meaning i don't particularly care but was noting that i don't care

khk-globus: So ... we're of similar mind.  You just chose to make that known.  :stuck_out_tongue:

PR #3783, opened 2025-02-24
[Draft] Fixes TypeError  and infinite looping in MPITaskScheduler

yadudoc:
# Description

This PR attempts to fix the following bugs in the MPITaskScheduler:

1. Currently the MPITaskScheduler's `schedule_backlog_tasks` method takes tasks from the backlog and attempts to schedule them until the queue is empty. However since calling put_task pops the task back onto the backlog queue, this ends up in an infinite loop if there's at least 1 task that cannot be scheduled.
2. Putting multiple tasks with the same priority into the internal PriorityQueue results in attempts to sort using the task dict which fails with `TypeError` unhashable type: dict.
3. PriorityQueue using increasing order for sorting queue items. This currently results in smaller tasks getting scheduler first while scheduling large tasks is generally preferred.

# Changed Behaviour

* Larger MPI tasks will be scheduled for execution on the manager.

# Fixes

1. `schedule_backlog_tasks` is now updated to fetch all tasks in the backlog_queue and then attempt to schedule them avoiding the infinite loop.
2. A new `PrioritizedTask` dataclass is added that disable comparison on the `task: dict` element.
3. The priority is set `num_nodes * -1` to ensure that larger jobs get prioritized.

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup


parsl/executors/high_throughput/mpi_resource_management.py
```diff
@@ -182,14 +192,16 @@ def put_task(self, task_package: dict):
...
+
+        # Separate fetching tasks from the _backlog_queue and scheduling them
+        # since tasks that failed to schedule will be pushed to the _backlog_queue
+        backlogged_tasks = []
+        while not self._backlog_queue.empty():
+            prioritized_task = self._backlog_queue.get(block=False)
+            backlogged_tasks.append(prioritized_task.task)
+
+        for backlogged_task in backlogged_tasks:
+            self.put_task(backlogged_task)
```
khk-globus, line 205: From a static analysis, this looks better to me.  No more infinite loop potential, but I do observe that this could mean a lot of unpacking and then repacking.  "It works," so I'm not going to fuss about it, but a different data structure might help with that.

More actionably, however, this looks like it would lose tasks?  What happens when `.get(block=False)` raises `queue.Empty`?

WardLT: Good catch, Kevin! These few lines do make me worry about race conditions.

Additionally, will the very aggressive scheduling here (always attempt to schedule everything) will still result in large tasks being continually delayed? If there are small tasks, they'll get scheduled before the big one still.

That might be ok with some users, but what about a simple "run in the order of execution" strategy as our baseline?

benclifford: @WardLT it's issue https://github.com/Parsl/parsl/issues/3323 but one layer deeper into the dispatch logic!

benclifford: (we shouldn't overtrivialise this or assume there's a universal solution or try to make a comprehensive set of here are the options that will satisfy *everyone*)

benclifford:
there's enough interesting stuff here that the "if your PR description is an itemised list, there should be one PR per item" rule probably applies.

parsl/executors/high_throughput/mpi_resource_management.py
yadudoc: @khk-globus Thanks for the review, this is a good catch!
* Unpacking-repacking: Yep, we shouldn't have to do this if we store the resource_spec
* `queue.Empty`: I was working with the idea that since only this function can pop an item from the queue, checking for `empty()` is sufficient to guarantee that `get` will not raise a `queue.Empty`. I can rework this to avoid this. 

@WardLT I share your concern. There's no notion of fairness here, and as @benclifford pointed out coming up with scheduling logic that'll work for everyone is hard. Right now, I expect larger tasks to end up getting delayed. Like @benclifford mentioned we could move this logic to the interchange (https://github.com/Parsl/parsl/issues/3323), but we still need to implement these alternative scheduling algorithms but I'm hesitant to do so without user feedback.

yadudoc:
@benclifford Your comment on splitting the PR is fair, I can get that sorted.

PR #3794, opened 2025-03-07
Fix bug in MPITaskScheduler where tasks with identical priorities fail with `TypeError`

yadudoc:
# Description

The `MPITaskScheduler` uses Python's PriorityQueue to prioritize tasks based on the number of nodes requested.
When items with identical priorities are are submitted to the PriorityQueue, they attempt to sort based on the task dict which fails with TypeError unhashable type: dict. This PR adds a new `PrioritizedTask` dataclass that sets the task element to `field(compare=False)`.

I'm splitting changes in #3783 to keep the PR concise. This is split 1 of 3.

# Changed Behaviour

Fixes the bug described above.

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_mpi_apps/test_mpi_scheduler.py
+++ b/parsl/tests/test_mpi_apps/test_mpi_scheduler.py
@@ -161,3 +161,28 @@ def test_MPISched_contention():
     assert task_on_worker_side['task_id'] == 2
     _, _, _, resource_spec = unpack_res_spec_apply_message(task_on_worker_side['buffer'])
     assert len(resource_spec['MPI_NODELIST'].split(',')) == 8
+
+
+@pytest.mark.local
+def test_hashable_backlog_queue():
+    """Run multiple large tasks that to force entry into backlog_queue
+    where queue.PriorityQueue expects hashability/comparability
+    """
+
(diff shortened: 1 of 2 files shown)


parsl/tests/test_mpi_apps/test_mpi_scheduler.py
```diff
@@ -161,3 +161,28 @@ def test_MPISched_contention():
     assert task_on_worker_side['task_id'] == 2
     _, _, _, resource_spec = unpack_res_spec_apply_message(task_on_worker_side['buffer'])
     assert len(resource_spec['MPI_NODELIST'].split(',')) == 8
+
+
+@pytest.mark.local
+def test_hashable_backlog_queue():
```
benclifford, line 167: if this is a regression test, I would expect it to fail, but it doesn't with master 07a4efb14e4f0863a1b28f01a5d163e06f30c131 on my laptop

benclifford (approved):

parsl/tests/test_mpi_apps/test_mpi_scheduler.py
yadudoc: I was surprised. It definitely fails for me! Happy we got this sorted :)

PR #3796, opened 2025-03-10
Split MonitoringRouter into ZMQ and UDP processes

benclifford:
This is to facilitate further work where the ZMQ and UDP router processes can have different lifetimes (along with the already separated filesystem router process and any future plugins).

The ZMQ router process is only needed when HTEX is being used, to route messages from the interchange.

The UDP router process is only needed when resource monitoring is enabled, and an executor is configured with the udp radio mode.

This PR does not change when these two processes run, but is only intended to help future PRs make this change.

# Changed Behaviour

one more process at runtime when monitoring is enabled

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -15,7 +15,8 @@ import typeguard
 from parsl.log_utils import set_file_logger
 from parsl.monitoring.errors import MonitoringHubStartError
 from parsl.monitoring.radios.multiprocessing import MultiprocessingQueueRadioSender
-from parsl.monitoring.router import router_starter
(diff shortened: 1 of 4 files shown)


khk-globus (approved):
The immediate apparent duplication smells a little off with my squinty eyes on, but I see the reasoning for it going forward.

benclifford:
> The immediate apparent duplication smells a little off with my squinty eyes on, but I see the reasoning for it going forward.

right. I don't like this either - but it's my intention that these two pieces of code go off in fairly different directions, so I was aiming for a minimal split, to start building on.

PR #3804, opened 2025-03-11
Make PPW monitoring thread blocking

khk-globus:
No sense in uselessly iterating this thread when there's no incoming message, just to handle the "time to quit" event.  Instead, just inject a poison pill when it's time to go away.

Before, if the workers were in debug mode, whether or not there were any messages to handle the block's `manager.log` file would fill with lines like:

```
... parsl:475 ... Monitoring-Handler [DEBUG]  Starting monitor_queue.get()
... parsl:478 ... Monitoring-Handler [DEBUG]  monitoring_queue.get() has timed out
```

Now, the lines are only emitted when an event happens (blocking for a new item in the queue, or an item is received):
```
... parsl:472 ... Monitoring-Handler [DEBUG]  Starting monitor_queue.get()
... ...
... parsl:476 ... Monitoring-Handler [DEBUG]  Got a monitoring message
```

This change simultaneously reduces log noise and CPU cycles.

# Changed Behaviour

Should be no user-facing change, other than less logs to trawl through if the workers are `debug==True`, and less CPU burned.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -468,20 +468,17 @@ class Manager:
         """
         logger.debug("Starting monitoring handler thread")
 
-        poll_period_s = max(10, self.poll_period) / 1000    # Must be at least 10 ms
-
         while not self._stop_event.is_set():
             try:
                 logger.debug("Starting monitor_queue.get()")
-                msg = self.monitoring_queue.get(block=True, timeout=poll_period_s)
-            except queue.Empty:
-                logger.debug("monitoring_queue.get() has timed out")
-            except Exception as e:
-                logger.exception(f"Got an exception: {e}")
-            else:
+                msg = self.monitoring_queue.get(block=True)
+                if msg is None:
+                    continue
                 logger.debug("Got a monitoring message")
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
This makes sense. The major change in behavior I see is that the Monitoring thread no longer can guarantee closing within `self.poll_period`. Instead, the thread can wait till the `None` message is reached. This looks good to me.

khk-globus:
> The major change in behavior I see is that the Monitoring thread no longer can guarantee closing within self.poll_period. Instead, the thread can wait till the None message is reached.

Actually, I believe this will be, on average, `poll_period / 2` _faster_.  That is, `None` is sent as soon as the shutdown event is set:
```python
# the main thread / start()
...
self._stop_event.wait()   # <----- the main thread blocks here until time to shutdown
logger.info("--- Stop event set; terminating worker processes")

# Invite blocking threads to quit
self.monitoring_queue.put(None)
...
```
So I would word that differently.  "The Monitoring thread _will_ guarantee closing within `self.poll_period` of the event being set.  In fact, near instantaneously, pending normal overhead of the queue.

PR #3805, opened 2025-03-11
Make `MPITaskScheduler` prioritize large tasks

yadudoc:
# Description

The python `PriorityQueue` prioritizes items with smaller priority values, which leads us to our current `MPITaskScheduler` prioritizing MPI tasks with fewer nodes requested since the nodes requested is used as the priority. This PR changes the ordering to prefer larger jobs. 

# Changed Behaviour

MPI Tasks with larger node requests are prioritized by the manager. Please note that as @WardLT pointed out in the comments on #3783, the greedy scheduling that we use can lead to larger tasks getting delayed in favor of smaller tasks that can get scheduled immediately.

These are change split from https://github.com/Parsl/parsl/pull/3783 to keep the PR concise.
This is split 3 of 3.

## Type of change

### DIFF
--- a/parsl/tests/test_mpi_apps/test_mpi_scheduler.py
+++ b/parsl/tests/test_mpi_apps/test_mpi_scheduler.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import pickle
+import random
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


parsl/tests/test_mpi_apps/test_mpi_scheduler.py
```diff
@@ -218,3 +218,33 @@ def test_tiny_large_loop():
...
+    for task_id, num_nodes in enumerate(node_request_list):
+        mock_task_buffer = pack_res_spec_apply_message("func", "args", "kwargs",
+                                                       resource_specification={
+                                                           "num_nodes": num_nodes,
+                                                           "ranks_per_node": 2
+                                                       })
+        task_package = {"task_id": task_id, "buffer": mock_task_buffer}
+        scheduler.put_task(task_package)
+
+    # Confirm that the tasks are sorted in decreasing order
```
benclifford, line 245: this test checks that tests coming out of the queue are sorted in priority order (that is, is the `priority` part of PrioritizedTask implemented correctly).

It doesn't check that they are sorted in num_nodes order, I think? (for example, introduce a bug into the calculation at  line 200 and I think this test will still pass, even though the code is buggy?)

yadudoc: Good point, I'll update this test.

yadudoc: I just pushed the updated test that tests for the backlog queue returning tasks in decreasing order of nodes requested.

benclifford (approved):

benclifford:
what is happening with the base of this PR? this should be going to `master`, right?

benclifford:
>     Update to human readable text: Documentation/error messages/comments

I think this is not the right change description.

PR #3806, opened 2025-03-11
Refactor worker shutdown procedure

khk-globus:
The detail of interest is to follow up after sending SIGTERM to each worker process.  This implementation now waits for up to 1s for each child process, before forcibly killing it (SIGKILL).

While here, move the storage to a non-class variable.

# Changed Behaviour

No user workflows will change; just make intended shutdown of process pool worker more robust in the errant case.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -15,6 +15,7 @@ import threading
 import time
 import uuid
 from importlib.metadata import distributions
+from multiprocessing.context import SpawnProcess
 from multiprocessing.managers import DictProxy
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -516,7 +514,7 @@ def start(self):
 
         # This might need a multiprocessing event to signal back.
         self._stop_event.wait()
-        logger.critical("Received kill event, terminating worker processes")
+        logger.info("--- Stop event set; terminating worker processes")
```
benclifford, line 517: what is this `---` for?

benclifford (approved):

PR #3809, opened 2025-03-12
Use ZMQ's `.send()` over `.send_multipart()`

khk-globus:
`send_multipart()` is for structured data.  In our case, we're sending one-off results, with no other related message bits.

This is a follow-up to the teaser in 9d2a97c7d658ebe26b6cbab96e51940f0a4ba41b

# Changed Behaviour

None.

## Type of change

- Code maintenance/cleanup


parsl/executors/high_throughput/interchange.py
```diff
@@ -499,7 +499,7 @@ def process_results_incoming(self, interesting_managers: Set[bytes], monitoring_
         # Receive any results and forward to client
         if self.results_incoming in self.socks and self.socks[self.results_incoming] == zmq.POLLIN:
             logger.debug("entering results_incoming section")
-            manager_id, *all_messages = self.results_incoming.recv_multipart()
+            manager_id, *all_messages = self.results_incoming.recv()
```
benclifford, line 502: this might still need to be recv_multipart: there is stuff where for some socket types a message appears as a multipart message with the remote connection ID as the first message (that I am not going to look up right now). (so the "multipartness" of here and the other piece of this PR  maybe don't have to align)

khk-globus: Yeah; that dawned on me a bit later.  I haven't quite nailed the ZMQ connection types; should have run the tests locally first, clearly.

benclifford (approved):

PR #3811, opened 2025-03-13
Make monitoring shutdown always event-terminate-kill-close

benclifford:
Prior to this PR, the monitoring proceses were shut down in two ways:

* In the happy path, processes would be notified that it was shutdown time via a multiprocessing.Event(). They would be joined on, with no timeout, assuming that they would eventually shut down based on the Event. In the unhappy case that one does not shut down, DFK shutdown would hang here.

* In the unhappy path, processes each have an opportunity to notify that an exception has happened via the exception_q. This is used for two purposes: to transcribe the exception into the main Parsl log, and to change how shutdown happens: if any exception has been delivered this way by any of the monitoring processes, then shutdown will happen by calling .terminate() on each process, without event notification.

This PR changes that so the exception_q is ignored for termination purposes (so it becomes purely a remote-logging system).

In both situations, processes will be ended using a sequence:

* notify by event

* join with timeout

* .terminate()

* join with timeout

* .kill()

The timeout for join is set quite high: monitoring processes sometimes take a while to shut down, and the timeout path is an error path not the fast happy path.

# Changed Behaviour

Monitoring processes which take longer than the timeout but would still otherwise successfully shut down will now be .terminate()ed, meaning whatever they were usefully doing will not happen.

Shutdown in the failing case might take longer: if a process reports an exception and then hangs, it will now take the timeout period to be terminated, rather than immediately.

If a process hangs without reporting an exception, it will now eventually be terminated rather than shutdown hanging.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/monitoring.py
+++ b/parsl/monitoring/monitoring.py
@@ -5,6 +5,7 @@ import multiprocessing.synchronize as ms
 import os
 import queue
 from multiprocessing import Event
+from multiprocessing.context import ForkProcess as ForkProcessType
 from multiprocessing.queues import Queue
 from typing import TYPE_CHECKING, Optional, Tuple, Union
 
@@ -247,31 +248,23 @@ class MonitoringHub(RepresentationMixin):
                         exception_msg[0],
                         exception_msg[1]
                     )
-                self.zmq_router_proc.terminate()
-                self.udp_router_proc.terminate()
-                self.dbm_proc.terminate()
-                self.filesystem_proc.terminate()
             logger.info("Setting router termination event")
             self.router_exit_event.set()
 
             logger.info("Waiting for ZMQ router to terminate")
-            self.zmq_router_proc.join()
-            self.zmq_router_proc.close()
+            join_terminate_close_proc(self.zmq_router_proc)
 
             logger.info("Waiting for UDP router to terminate")
-            self.udp_router_proc.join()
-            self.udp_router_proc.close()
+            join_terminate_close_proc(self.udp_router_proc)
 
             logger.debug("Finished waiting for router termination")
             logger.debug("Waiting for DB termination")
             self.dbm_exit_event.set()
-            self.dbm_proc.join()
-            self.dbm_proc.close()
+            join_terminate_close_proc(self.dbm_proc)
             logger.debug("Finished waiting for DBM termination")
 
             logger.info("Terminating filesystem radio receiver process")
-            self.filesystem_proc.join()
-            self.filesystem_proc.close()
+            join_terminate_close_proc(self.filesystem_proc)
 
             logger.info("Closing monitoring multiprocessing queues")
             self.exception_q.close()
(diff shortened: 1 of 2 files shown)


parsl/monitoring/monitoring.py
```diff
@@ -247,35 +248,61 @@ def close(self) -> None:
...
+            process.kill()
+            process.join(timeout)
+            # This kill should not be caught by any signal handlers so it is
+            # unlikely that this join will timeout. If it does, there isn't
+            # anything further to do except log an error in the next if-block.
+
+    if process.is_alive():
+        logger.error("Process failed to end")
+        # don't call close if the process hasn't ended:
+        # process.close() doesn't work on a running process.
```
khk-globus, line 306: Wow; I've only seen `SIGKILL` fail in case of kernel bugs (2.4 days, process stuck in state `D`, internal kernel lock, and no access actually happening).  I suppose it's good to log this scenario, but man I'd expect this to "not ever happen."  FLW, I know.

khk-globus: Yeah; like 20+ years ago.  "But sure."  Heh.

parsl/tests/test_monitoring/test_exit_helper.py
```diff
@@ -0,0 +1,56 @@
...
+from parsl.multiprocessing import ForkProcess
+
+
+def noop():
+    pass
+
+
+@pytest.mark.local
+def test_pre_exit():
+    """Test calling against a process that has already exited."""
```
khk-globus, line 17: FWIW, you might consider using the test name itself as the documentation.  Consider:
```python
@pytest.mark.local
def test_join_helper_returns_on_already_finished_process():
```
Now the intention is clear in the test output, as for example when one runs `pytest` with arguments like `-vsx`.

Not a blocker.

parsl/tests/test_monitoring/test_exit_helper.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@pytest.mark.local
+def test_hang():
+    """Test calling against a process that will not exit itself."""
+    p = ForkProcess(target=hang)
+    p.start()
+    pid = p.pid
+    join_terminate_close_proc(p, timeout=1)
+    assert not psutil.pid_exists(pid), "process should not exist any more"
```
khk-globus, line 36: If you don't want to use `psutil` ("another lib"), I believe the ForkProcess class can do equivalently via:
```python
assert p.exitcode is not None, ... # one option
assert not p.is_alive(), ... # another
```
Similarly, the return code should be the negation of the signal, so could ensure that it was closed by the signal (as opposed to crashing for, say, a SIGBUS :wink:) via:
```python
assert p.exitcode == -signal.SIGTERM, ...
```

parsl/tests/test_monitoring/test_exit_helper.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.mark.local
+def test_hang_no_sigint():
+    """Test calling against a process that will not exit itself."""
+    e = multiprocessing.Event()
+    p = ForkProcess(target=hang_no_sigint, args=(e,))
+    p.start()
+    pid = p.pid
+    join_terminate_close_proc(p, timeout=1)
+    assert not psutil.pid_exists(pid), "process should not exist any more"
+    assert e.is_set(), "hung process should have set event on signal"
```
khk-globus, line 56: I see _what_ `e.is_set()` is verifying, but I don't get the utility?  Is it to ensure that the signal was received&nbsb;&mdash;&nbsp;and summarily ignored&nbsb;&mdash;&nbsp;by the process?

parsl/monitoring/monitoring.py
```diff
@@ -247,35 +248,61 @@ def close(self) -> None:
...
+    This function assumes that the process is likely to exit before
+    the join timeout, driven by some other means, such as the
+    MonitoringHub router_exit_event. If the process does not exit, then
+    first terminate() and then kill() will be used to end the process.
+
+    In the case of a very mis-behaving process, this function might take
+    up to 3*timeout to exhaust all termination methods and return.
+    """
+    process.join(timeout)
+
```
khk-globus, line 289: If potentially waiting for a few seconds, would a `logger.debug()` call be warranted before the `.join()`?

khk-globus (approved):
A few suggestions, no blockers.

parsl/tests/test_monitoring/test_exit_helper.py
benclifford: not after the process object has been closed by join_terminate_close_proc! that was my first implementation.

parsl/tests/test_monitoring/test_exit_helper.py
benclifford: yes - it's a lets gather some more bits of evidence that this behaved how we expected rather than exiting in some other way and so still successfully passing the join-terminate-close-proc routine. I also thought about making assertions about the emitted log messages to also get introspection into the code paths taken inside that routine, or making the return value be some "we succeeded, and this is how!" value

parsl/tests/test_monitoring/test_exit_helper.py
khk-globus: Ah; the `.close()`.  Hmm; I wonder if `.join_()` should consider returning anything then?  Like the exit code?  Not a big detail, but "perhaps."

parsl/monitoring/monitoring.py
benclifford: yeah, I was relying on that message happening right before this is called -- with more context about what the process is, which is a bit unmodular but has more context.
eg

```
            logger.debug("Waiting for DB termination")
            self.dbm_exit_event.set()
            join_terminate_close_proc(self.dbm_proc)
```

It's probably good to be explicit about going into the join, though, given the evidence decade of comedy exit hangs that more logging around exit is better than less.

PR #3812, opened 2025-03-14
Remove exception_q from monitoring

benclifford:
After PR #3811 removed exeception_q's role in exit handling, this code remains only as a rudimentary distributed logging system to convey some specific error messages back to the main process for logging in parsl.log.

No other part of Parsl does this in this way, and instead each process logs to a file with the user expected to look in individual log files.

This PR makes monitoring consistent with that expectation.

# Changed Behaviour

Error messages from monitoring failures will appear in different places, compared to before this PR.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/radios/zmq_router.py
+++ b/parsl/monitoring/radios/zmq_router.py
@@ -107,7 +107,6 @@ class MonitoringRouter:
 @typeguard.typechecked
 def zmq_router_starter(*,
                        comm_q: mpq.Queue,
-                       exception_q: mpq.Queue,
                        resource_msgs: mpq.Queue,
                        exit_event: Event,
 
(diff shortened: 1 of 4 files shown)


parsl/monitoring/radios/zmq_router.py
```diff
@@ -129,10 +128,4 @@ def zmq_router_starter(*,
         comm_q.put(f"Monitoring router construction failed: {e}")
     else:
         comm_q.put(router.zmq_receiver_port)
-
-        router.logger.info("Starting MonitoringRouter in router_starter")
```
khk-globus, line 133: Removing this logging message was intentional?

khk-globus (approved):
Weird.  Thank your for this find and cleanup.

PR #3813, opened 2025-03-14
Move ZMQ monitoring radio initialization into HTEX

benclifford:
This is part of an ongoing project to move monitoring message forwarding into relevant executors, with executors only initializing whichever routers are needed.

In this step of the project:

The ZMQ radio/router is only used by the high throughput executor to receive messages from the interchange. This PR makes the initialization of that look more like other interchange<->submit side ZMQ initialiation, rather than look like monitoring system initialization.

This does not stop other future components which want to use monitoring-over-ZMQ: those components would initialize ZMQ routing themselves.

API-wise:

Parsl executors lose the hub_zmq_port attribute which was configured by the DFK after initialising monitoring.

As a replacement, Parsl executors gain a monitoring_messages attribute, which is the multiprocessing.Queue where executors should send monitoring messages when they have received them from their own components. This queue already existed between monitoring components, but is now exposed as an interface for other Parsl components.

Configuration-wise:

Several ZMQ-related monitoring configuration options now come from HTEX configuration, rather than from MonitoringHub configuration. There are no new configuration options.

* Choice of ZMQ port now comes from the same range as other HTEX ZMQ ports, rather than from the port range in MonitoringHub. This means the minimum number of available ports there is now 4, instead of 3.

* ZMQ router log files now go into the High Throughput Executor's log directory, usually a subdirectory of the main log directory.

* ZMQ log verbosity is configured by High Throughput Executor's worker_debug configuration parameter, rather than by MonitoringHub.monitoring_debug.

* The ZMQ router will now listen on the interchange loopback address, configured by HighThroughputExecutor.loopback_address, rather than on all addresses.

## Type of change

- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/radios/zmq_router.py
+++ b/parsl/monitoring/radios/zmq_router.py
@@ -3,16 +3,27 @@ from __future__ import annotations
 import logging
 import multiprocessing.queues as mpq
 import os
+import queue
 import time
-from multiprocessing.synchronize import Event
+from multiprocessing.context import SpawnProcess as SpawnProcessType
+from multiprocessing.queues import Queue as QueueType
+from multiprocessing.synchronize import Event as EventType
 from typing import Tuple
 
 import typeguard
 import zmq
 
+from parsl.addresses import tcp_url
 from parsl.log_utils import set_file_logger
+from parsl.monitoring.errors import MonitoringRouterStartError
 from parsl.monitoring.radios.multiprocessing import MultiprocessingQueueRadioSender
 from parsl.monitoring.types import TaggedMonitoringMessage
+from parsl.multiprocessing import (
+    SizedQueue,
+    SpawnEvent,
+    SpawnProcess,
+    join_terminate_close_proc,
+)
 from parsl.process_loggers import wrap_with_logs
 from parsl.utils import setproctitle
 
@@ -23,21 +34,21 @@ class MonitoringRouter:
 
     def __init__(self,
                  *,
-                 hub_address: str,
-                 zmq_port_range: Tuple[int, int] = (55050, 56000),
+                 address: str,
+                 port_range: Tuple[int, int] = (55050, 56000),
 
                  run_dir: str = ".",
                  logging_level: int = logging.INFO,
                  resource_msgs: mpq.Queue,
-                 exit_event: Event,
+                 exit_event: EventType,
                  ):
         """ Initializes a monitoring configuration class.
 
         Parameters
         ----------
-        hub_address : str
+        address : str
              The ip address at which the workers will be able to reach the Hub.
-        zmq_port_range : tuple(int, int)
+        port_range : tuple(int, int)
              The MonitoringHub picks ports at random from the range which will be used by Hub.
              Default: (55050, 56000)
         run_dir : str
@@ -51,11 +62,11 @@ class MonitoringRouter:
         """
         os.makedirs(run_dir, exist_ok=True)
         self.logger = set_file_logger(f"{run_dir}/monitoring_zmq_router.log",
-                                      name="monitoring_router",
+                                      name="zmq_monitoring_router",
                                       level=logging_level)
         self.logger.debug("Monitoring router starting")
 
-        self.hub_address = hub_address
+        self.address = address
 
         self.loop_freq = 10.0  # milliseconds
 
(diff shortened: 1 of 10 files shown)


parsl/tests/test_monitoring/test_radio_zmq.py
```diff
@@ -0,0 +1,27 @@
...
+from parsl.multiprocessing import SpawnQueue
+
+
+@pytest.mark.local
+def test_send_recv_message(tmpd_cwd, try_assert):
+    q = SpawnQueue()
+    loopback = "127.0.0.1"
+    r = start_zmq_receiver(monitoring_messages=q,
+                           loopback_address=loopback,
+                           port_range=(49152, 65535),
```
khk-globus, line 14: For clarity, for those who may not be inclined toward networks, should we consider making this specific tuple/range a constant within Parsl?  I'm thinking both "perhaps it doesn't matter, this is only likely to be used by developers" and "this is how students and younger developers learn of these seemingly random details."

I'm not sure where it would live if we did include it.  Perhaps arbitrarily in the top-level Parsl namespace?  Or `addresses.py`?
```python
# parsl/__init__.py or parsl/addresses.py?
...
DYNAMIC_PORT_RANGE = (0xc000, 0xffff)
...
```

parsl/monitoring/radios/zmq_router.py
```diff
@@ -129,3 +140,53 @@ def zmq_router_starter(*,
...
+                                       },
+                               name="Monitoring-ZMQ-Router-Process",
+                               daemon=True,
+                               )
+    router_proc.start()
+
+    try:
+        comm_q_result = comm_q.get(block=True, timeout=120)
+        comm_q.close()
+        comm_q.join_thread()
```
khk-globus, line 183: Ought these be in the `finally` block?  Or otherwise no skipped in the exceptional case?

parsl/monitoring/radios/zmq_router.py
```diff
@@ -129,3 +140,53 @@ def zmq_router_starter(*,
...
+                                       "run_dir": logdir,
+                                       "logging_level": logging.DEBUG if worker_debug else logging.INFO,
+                                       },
+                               name="Monitoring-ZMQ-Router-Process",
+                               daemon=True,
+                               )
+    router_proc.start()
+
+    try:
+        comm_q_result = comm_q.get(block=True, timeout=120)
```
khk-globus, line 181: Suggest some kind of log line prior to blocking here.  In the happy path we won't care, but when tracking something down in the unhappy case ...

parsl/monitoring/radios/zmq_router.py
```diff
@@ -108,17 +119,17 @@ def start(self) -> None:
...
 
-                       hub_address: str,
-                       zmq_port_range: Tuple[int, int],
+                       address: str,
+                       port_range: Tuple[int, int],
 
                        run_dir: str,
                        logging_level: int) -> None:
     setproctitle("parsl: monitoring zmq router")
     try:
```
khk-globus, line 130: Is there any debugging or human utility in including some (terse) details in the process title?  Like here:
```python
setproctitle("parsl: monitoring zmq router (booting up)")
```
and then below, after acquiring the port:
```python
setproctitle(f"parsl: monitoring zmq router (port: {router.zmq_receiver_port})")
```
Probably TMI, but thinking generally in terms of visibility and trying to think what a user or administrator might care about as easy reference.

khk-globus (approved):
Looks generally good; a couple of inline thoughts, questions, and/or suggestions.

parsl/tests/test_monitoring/test_radio_zmq.py
benclifford: theres an RFC for ephemeral ports https://datatracker.ietf.org/doc/html/rfc6056#section-2.1 and I separately had been thinking about making a parsl.utils constant for that and moving everything over. seeing as you have also gone along roughly the same path, you or me can make that tidyup.

parsl/monitoring/radios/zmq_router.py
benclifford: not in this PR. maybe interesting to include more info here - a runid/workflow ID would be another thing, for example, to help people associate different processes together in `ps`

parsl/monitoring/radios/zmq_router.py
khk-globus: Done in the followup #3820.

PR #3815, opened 2025-03-17
Use itertools batch to get long jobs lists

### LINKED ISSUE #3814: Calling `sacct` on many jobs Slurm hangs causing Globus Compute tasks to timeout
**Describe the bug**
A user saw that Slurm calls for `_status` hang when there are more than 200 jobs on Perlmutter. This could also be the case on other systems as well but haven't tested.

**To Reproduce**
Start many long running jobs on Perlmutter or other Slurm cluster through Parsl.

**Expected behavior**
Calls to `_status` should return quickly and not hang when interacting with Slurm. This should also work the same for a small job list as well as large a large job list.

**Environment**
- Not sure of the users exact environment
- Globus Compute on Perlmutter

**Distributed Environment**
- This is coming from many long running Globus Compute jobs managed by Parsl

tylern4:
# Description

When there are many jobs being tracked by parsl `sacct` can become unresponsive and timeout if there are too many jobs. A users workflow was experiencing this while running through Globus Compute on Perlmutter.

# Changed Behaviour

This should batch calls in groups to Slurm

# Fixes

Fixes #3814

## Type of change

- Bug fix

### DIFF
--- a/parsl/providers/slurm/slurm.py
+++ b/parsl/providers/slurm/slurm.py
@@ -4,6 +4,7 @@ import os
 import re
 import time
 from typing import Any, Dict, Optional
+import itertools
 
 import typeguard
 
@@ -148,6 +149,8 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         self.qos = qos
         self.constraint = constraint
         self.clusters = clusters
+        # Used to batch requests to sacct/squeue for long jobs lists
+        self._job_batch_size = 50
         self.scheduler_options = scheduler_options + '\n'
         if exclusive:
             self.scheduler_options += "#SBATCH --exclusive\n"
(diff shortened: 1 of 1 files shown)


parsl/providers/slurm/slurm.py
```diff
@@ -116,6 +114,12 @@ class SlurmProvider(ClusterProvider, RepresentationMixin):
         symbolic group for the job ID.
     worker_init : str
         Command to be run before starting a worker, such as 'module load Anaconda; source activate env'.
+    cmd_timeout : int (Default = 10)
+        Number of seconds to wait for slurm commands to finish. For schedulers with many this
+        may need to be increased to wait longer for scheduler information.
+    status_batch_size: ine (Default = 50)
```
benclifford, line 120: int

benclifford:
it would be better if you do code reformatting in a separate PR that claims to not change behaviour -- the diff for this PR is mostly that and nothing to do with the PR topic, and it hurts me to go look at this stuff in `git bisect` in the future.

parsl/providers/slurm/slurm.py
```diff
@@ -199,22 +223,24 @@ def _status(self):
...
+        job_list_batches = batched(active_jobs.items(), self.status_batch_size)
+        stdout = ""
+        for job_batch in job_list_batches:
+            job_id_list = ",".join([jid for jid, job in job_batch if not job["status"].terminal])
+            cmd = self._cmd.format(job_id_list)
+            logger.debug("Executing %s", cmd)
+            retcode, _stdout, stderr = self.execute_wait(cmd)
+            logger.debug("sacct/squeue returned %s %s", stdout, stderr)
+            stdout += _stdout
+            # Execute_wait failed. Do no update
```
khk-globus, line 240: - `_stdout` - the prefix implies unused or private.  Fine, but stylistically, consider maybe `chunk_stdout` or `batch_stdout`
- `logger.debug()` currently uses `stdout`, the increasing variable.  That's [Shlemiel in nature](https://www.joelonsoftware.com/2001/12/11/back-to-basics/).  I'm guessing this should be the `batch_stdout` instead?
- Since nominally modifying `logger.debug()`, consider including the `retcode` in the debug line.

khk-globus (commented):
A couple of inline comments to address, but we've also recently unbroken CI so please rebase on `master`.

parsl/providers/slurm/slurm.py
```diff
@@ -50,6 +51,20 @@
...
+    Turns a list into a batch of size n. This code is from
+    https://docs.python.org/3.12/library/itertools.html#itertools.batched
+    and in versions 3.12+ this can be replaced with
+    itertools.batched
+    """
+    if n < 1:
+        raise ValueError("n must be at least one")
+    iterator = iter(iterable)
+    while batch := tuple(islice(iterator, n)):
+        yield batch
```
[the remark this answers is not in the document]
tylern4: Thanks for the pyupgrade block docs. I had thought about this originally and was going to use a,
```python
try:
   from itertools import batched
except ImportError:
   ...
```
but this seems much cleaner.

parsl/providers/slurm/slurm.py
tylern4: Fixed the name and have it added to the debug log line with return node now.

parsl/providers/slurm/slurm.py
```diff
@@ -199,22 +223,24 @@ def _status(self):
...
-        retcode, stdout, stderr = self.execute_wait(cmd)
-        logger.debug("sacct/squeue returned %s %s", stdout, stderr)
-
-        # Execute_wait failed. Do no update
-        if retcode != 0:
-            logger.warning("sacct/squeue failed with non-zero exit code {}".format(retcode))
-            return
+        job_list_batches = batched(active_jobs.items(), self.status_batch_size)
+        stdout = ""
+        for job_batch in job_list_batches:
```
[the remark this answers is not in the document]
tylern4: Yeah I have no preference so I moved it.

khk-globus (approved):
Thanks for your staying power!  :muscle:

PR #3833, opened 2025-03-28
Add monitoring info for tasks given to a manager

khk-globus:
# Description

Update any interested listeners to the state of the manager when new tasks are assigned.  Based on `.start()`, this now covers most (all?) relevant changes to a manager's info block that will be shared.

# Changed Behaviour

Adds a monitoring message per manager in the interchange loop, for any outstanding tasks that move from the interchange to a worker manager.  I don't think any existing workflows will be impacted, but there will now be an extra NODEINFO record in the sqlite database per task.

## Type of change

- New feature

### DIFF
--- a/parsl/tests/test_monitoring/test_basic.py
+++ b/parsl/tests/test_monitoring/test_basic.py
@@ -120,7 +120,7 @@ def test_row_counts(tmpd_cwd, fresh_config):
             # Two entries: one showing manager active, one inactive
             result = connection.execute(text("SELECT COUNT(*) FROM node"))
             (c, ) = result.first()
-            assert c == 2
+            assert c == 3
 
         # There should be one block polling status
         # local provider has a status_polling_interval of 5s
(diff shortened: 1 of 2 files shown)


parsl/tests/test_monitoring/test_basic.py
```diff
@@ -120,7 +120,7 @@ def test_row_counts(tmpd_cwd, fresh_config):
             # Two entries: one showing manager active, one inactive
             result = connection.execute(text("SELECT COUNT(*) FROM node"))
             (c, ) = result.first()
-            assert c == 2
+            assert c == 3
```
benclifford, line 123: i'm a little surprised that this isn't 4 -- there's a message when a task is sent to a worker, but not when a result comes back. If this PR is to get more accurate "what's on my managers" info in the monitoring system, does that mean you don't see the reported load decrease properly?

khk-globus: That's correct; I made this before I made the other PR, and I have the change locally to do just this.  But I didn't want to muddle #3834 after I put together this missing detail in my mind.  Since that's now merged, I'll this PR presently.

benclifford (approved):

parsl/tests/test_monitoring/test_basic.py
khk-globus: Address in #3837; had meant to add it to this PR, but missed word in my previous message in this thread: _update_.  Heh.

PR #3838, opened 2025-04-07
Fix binding when user sets HTEx worker ports manually

WardLT:
# Description

Fixes the creation of the TCP url when the ports used by HTEx workers are fixed.

The `tcp_url` helper function from addresses.py was only appending the port number when the address was not "*".

# Changed Behaviour

HTEx works with the ports fixed. Not sure when it was broken

I added a unit test which shows that the current code fails when launching an HTEx interchange with user-set ports.

# Fixes

None (yet? should I make one?)

## Type of change

- Bug fix

### DIFF
--- a/parsl/addresses.py
+++ b/parsl/addresses.py
@@ -161,12 +161,11 @@ def get_any_address() -> str:
 
 def tcp_url(address: str, port: Union[str, int, None] = None) -> str:
     """Construct a tcp url safe for IPv4 and IPv6"""
+    port_suffix = f":{port}" if port else ""
     if address == "*":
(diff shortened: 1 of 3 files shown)


WardLT:
Might I suggest @yadudoc for the review? I altered a function he wrote a few months ago

benclifford:
you might add a case into `parsl/tests/unit/test_address.py` which more directly tests the `tcp_url` function

khk-globus (approved):
Thanks for adding the test.  We're working through a build error at the moment, but I'll merge this once that gets sorted.  Thanks!

Issue #3839, opened 2025-04-14
rename htex HOLD_WORKER internal command to HOLD_MANAGER


benclifford:
**Describe the bug**
This interchange command confuses worker pools/managers (which are already confused amongst themselves) with workers - of which there can be many in a worker pool attached to a single manager.

**To Reproduce**
Read the source code

**Expected behavior**
HOLD_whatever should suggest to the reader that a `whatever` is being held, not a `something else`.

**Environment**
Parsl `master` 2025-04

Issue #3368, opened 2024-04-20
Interchange: HOLD_WORKER command should not parse its own private protocol


benclifford:
**Is your feature request related to a problem? Please describe.**
Interchange commands and responses are sent in structured objects (using pickle).

Nevertheless, the HOLD_WORKERS command contains an ad-hoc UTF-8/ASCII protocol to pass its parameter, which is treated as a byte string elsewhere in htex, not UTF-8:

2024-04-20 08:36:13.204 interchange:282 HTEX-Interchange(131272) Interchange-Command _command_server [DEBUG] Received command request: HOLD_WORKER;d9052d12878b

This command should use the existing data structures available for conveying a separate command and byte sequence argument, such as a tagged dictionary `{"type": "hold_worker", "manager_id", bytes(d9052d12878b)}` or a tuple.

This would remove some construction/parsing complexity and potentially give some cross-network type safety.

benclifford:
crossref #3839 for another HOLD_WORKER tidyup

benclifford:
see more protocol structure comment here https://github.com/Parsl/parsl/pull/3840#discussion_r2044530685

PR #3840, opened 2025-04-14
Rearrange HTEX hold_worker because it holds managers, not workers

benclifford:
See issue #3839

This changes the Python API side of this method to reflect that the action is on managers, not workers.

A separate PR will change the wire protocol command.

# Changed Behaviour

this is internal to htex - nothing except stack traces should change for users

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -599,20 +599,20 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         self._result_queue_thread.start()
         logger.debug("Started result queue thread: %r", self._result_queue_thread)
 
-    def hold_worker(self, worker_id: str) -> None:
-        """Puts a worker on hold, preventing scheduling of additional tasks to it.
+    def _hold_manager(self, manager_id: str) -> None:
+        """Puts a manager on hold, preventing scheduling of additional tasks to it.
 
         This is called "hold" mostly because this only stops scheduling of tasks,
-        and does not actually kill the worker.
+        and does not actually kill the manager or workers.
 
         Parameters
         ----------
 
-        worker_id : str
-            Worker id to be put on hold
+        manager_id : str
+            Manager id to be put on hold
         """
-        self.command_client.run("HOLD_WORKER;{}".format(worker_id))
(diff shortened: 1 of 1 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -599,20 +599,20 @@ def _start_result_queue_thread(self):
...
 
-        worker_id : str
-            Worker id to be put on hold
+        manager_id : str
+            Manager id to be put on hold
         """
-        self.command_client.run("HOLD_WORKER;{}".format(worker_id))
-        logger.debug("Sent hold request to manager: {}".format(worker_id))
+        self.command_client.run("HOLD_WORKER;{}".format(manager_id))
+        logger.debug("Sent hold request to manager: {}".format(manager_id))
```
khk-globus, line 615: For completeness, should we change the text of the command as well?  `%s/HOLD_WORKER/HOLD_MANAGER/g` ...

parsl/executors/high_throughput/executor.py
```diff
@@ -656,7 +656,7 @@ def _hold_block(self, block_id):
         for manager in managers:
             if manager['block_id'] == block_id:
                 logger.debug("Sending hold to manager: {}".format(manager['manager']))
-                self.hold_worker(manager['manager'])
+                self._hold_manager(manager['manager'])
```
khk-globus, line 659: Unrelated to the thrust of this PR, but I see it now so observing it: since this is a blocking request, too bad this loop isn't in bulk.  Or that `_hold_manager` is blocking.

khk-globus (approved):

parsl/executors/high_throughput/executor.py
benclifford: yeah I have a separate PR for that  - see 

> A separate PR will change the wire protocol command.

in the original PR text

parsl/executors/high_throughput/executor.py
benclifford: yeah, that's probably quite interesting especially for large-node-count blocks (eg 1000 nodes). I'll link this comment from #3368 which talks about reworking this protocol.

PR #3841, opened 2025-04-15
Remove reassignment of interchange global logger

benclifford:
Since PR #2828, this has been assigned at the top level as part of module import, and the reassignment has always been reassigning to the same Logger value - effectively a no-op

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


benclifford:
added @yadudoc who made PR #2828 in case there's some logging subtlety here: that PR is focused around internet protocol work, not logging.

khk-globus (approved):
Looks correct to me; will wait for @yadudoc to chime in before clicking the merge button, however.

PR #3842, opened 2025-04-15
Remove interchange logging for sections where nothing happens

benclifford:
The interchange tight loops through several sections. Two of those sections log that they are doing nothing. Other sections only log when they do something.

This PR makes those two sections no longer log when they do nothing.

As an example, an interchange idle for 20 seconds in a test case on my laptop logs:

before this PR: 5699 lines
after this PR: 1925 lines

The bulk of the remaining 1925 lines is another per-iteration message that is interchange status summary information, rather than per-section activity logging. Rather than implement something specific for this message to log it less frequently, this is left for a future PR which will make the interchange loop only when something happens, rather than in a tight poll.

# Description

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -220,11 +220,11 @@ class Interchange:
     def process_command(self, monitoring_radio: Optional[MonitoringRadioSender]) -> None:
         """ Command server to run async command to the interchange
         """
-        logger.debug("entering command_server section")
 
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
I have commented these out locally for months.  I should have discussed this with you back in ~September.  :facepalm:

khk-globus:
Trying merge a second time; let's see if GH Actions has repaired itself since 60m ago ...

Issue #3544, opened 2024-07-25
`make test` tries to download non-existing file


giordano:
**Describe the bug**

Trying to run `make test` on 71d9c711cee30211aaadb1725490d1ff0c7f194a I get
```
parsl/executors/taskvine/install-taskvine.sh
+ [[ -z 7.8.0 ]]
+ TARBALL=cctools-7.8.0-x86_64-ubuntu20.04.tar.gz
+ [[ -f /etc/redhat-release ]]
+ TARBALL=cctools-7.8.0-x86_64-centos8.tar.gz
+ [[ ! -t 1 ]]
+ wget '' '' -O /tmp/cctools.tar.gz https://github.com/cooperative-computing-lab/cctools/releases/download/release/7.8.0/cctools-7.8.0-x86_64-centos8.tar.gz
http://: Invalid host name.
http://: Invalid host name.
--2024-07-25 20:58:39--  https://github.com/cooperative-computing-lab/cctools/releases/download/release/7.8.0/cctools-7.8.0-x86_64-centos8.tar.gz
Resolving github.com (github.com)... 20.26.156.215
Connecting to github.com (github.com)|20.26.156.215|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2024-07-25 20:58:39 ERROR 404: Not Found.

make: *** [/tmp/cctools] Error 8
```

The file doesn't exist among the assets of https://github.com/cooperative-computing-lab/cctools/releases/tag/release%2F7.8.0

**To Reproduce**

Follow instructions at https://github.com/Parsl/parsl/blob/71d9c711cee30211aaadb1725490d1ff0c7f194a/README.rst?plain=1#L95-L107

**Expected behavior**

Tests don't fail because they try to download non-existing files.

**Environment**

 - OS: RedHat 7.8
 - Python version: 3.11.3
 - Parsl version: 71d9c711cee30211aaadb1725490d1ff0c7f194a

**Distributed Environment**

N/A

dthain:
Hmm, it would appear that the centos version for 7.8.0 was never generated:
https://github.com/cooperative-computing-lab/cctools/releases/tag/release%2F7.8.0

Let me see what I can do to make it...

benclifford:
[...]

PR #3844, opened 2025-04-15
Update GH actions runner per GH image retirement

khk-globus:
# Description

GitHub retired the Ubuntu 20.04 image per their warning that we had been ignoring.  In order to unbreak CI, update to use the current LTS image, Ubuntu 24.04.

In so doing, the Radical Pilot tests broke, so take the opportunity to do some housekeeping in two key changes:

- Install MPI and other external dependencies from Conda -- no more manual download and installation

- Move the Radical, WorkQueue, and TaskVine tests to separate GitHub actions so that optional extra packages can be tested separately.  This allows us to 1) identify test failures originating from core parsl vs external packages and 2) test different components in parallel.

# Changed Behaviour

No user-facing behavior, but CI tests are now a bit more decoupled.

# Fixes

- Broken CI
- Fixes: #3544

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
Finally! It looks we've got all the pieces to get our CI system updated and unblocked. Thanks to @khk-globus, @matthewc2003, and @AymenFJA for getting all the issues ironed out.

PR #3846, opened 2025-04-21
Move WorkQueue and TaskVine tests to a separate GH Action

yadudoc:
# Description

This PR splits out the testing for WorkQueue and TaskVine into a separate Github action.
The current mechanism that relies on pulling binaries for cctools is broken in Ubuntu 24.04, and based on the discussion here: https://github.com/Parsl/parsl/issues/3544, the recommended solution is to switch to using binaries distributed through conda. 

* Split out TaskVine and WorkQueue test to a separate GH Action
* Update GH Action to use conda instead of virtualenv.

# Changed Behaviour

The default Parsl CI test action will no longer test WorkQueue and TaskVine. A new action is added for these tests.

# Fixes

Fixes # (issue)

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/.github/workflows/parsl+cctools.yaml
+++ b/.github/workflows/parsl+cctools.yaml
@@ -1,0 +1,50 @@
+name: WorkQueue and TaskVine tests
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    strategy:
+      max-parallel: 5
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v3
+      with:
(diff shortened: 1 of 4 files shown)


.github/workflows/parsl+cctools.yaml
```diff
@@ -0,0 +1,50 @@
...
+        python3 -c "from ndcctools import work_queue; print(work_queue.__version__)"
+        python3 -c "import parsl; print(parsl.__version__)"
+        which parsl_coprocess.py
+        which python3
+        find /usr/share/miniconda | grep libssl
+        
+    - name: Run pytest suite for WorkQueue and TaskVine
+      run: |
+        # make wqex_local_test
+        make vineex_local_test
```
khk-globus, line 50: Need to uncomment?

khk-globus (approved):
Looks good to me, but I'm guessing you want to uncomment line 49?

PR #3848, opened 2025-05-05
Add Add priority queue to HTEX interchange v4

matthewc2003:
# Description

Replaces pending_task_queue's queue.Queue() as a SortedList which functions as a priority queue in the interchange. Used to prioritize certain user-labeled tasks when distributing tasks to managers

# Changed Behaviour

Users can now use the parsl_resource_specification keyword in python apps to specify a 'priority'. Lower value is higher priority, tasks with higher priority will be distributed to workers first. If no priority is given, tasks default to the lowest priority and the FIFO queue behavior of the original queue.Queue() is retained.

# Fixes

Part of #3323, reverts changes in #3750 

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -5,13 +5,13 @@ import logging
 import os
 import pickle
 import platform
-import queue
 import sys
 import threading
 import time
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, cast
 
 import zmq
+from sortedcontainers import SortedList
 
 from parsl import curvezmq
 from parsl.addresses import tcp_url
@@ -131,7 +131,7 @@ class Interchange:
         self.hub_address = hub_address
         self.hub_zmq_port = hub_zmq_port
 
-        self.pending_task_queue: queue.Queue[Any] = queue.Queue(maxsize=10 ** 6)
+        self.pending_task_queue: SortedList[Any] = SortedList(key=lambda tup: (tup[0], tup[1]))
 
         # count of tasks that have been received from the submit side
         self.task_counter = 0
@@ -197,12 +197,11 @@ class Interchange:
         """
         tasks = []
         for _ in range(0, count):
-            try:
-                x = self.pending_task_queue.get(block=False)
-            except queue.Empty:
-                break
+            if len(self.pending_task_queue) > 0:
+                _, _, task = self.pending_task_queue.pop(0)
+                tasks.append(task)
             else:
-                tasks.append(x)
+                break
 
         return tasks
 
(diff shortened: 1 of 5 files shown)


parsl/executors/high_throughput/interchange.py
```diff
@@ -342,8 +341,17 @@ def process_task_incoming(self) -> None:
         if self.task_incoming in self.socks and self.socks[self.task_incoming] == zmq.POLLIN:
             logger.debug("start task_incoming section")
             msg = self.task_incoming.recv_pyobj()
+
+            # Process priority, higher number = lower priority
+            resource_spec = msg.get('resource_spec', {})
+            resource_spec.setdefault("priority", float('inf'))
+            msg['resource_spec'] = resource_spec
+            priority = resource_spec['priority']
```
khk-globus, line 349: It looks like `priority` is strictly utilized within the interchange?  If so, there's no need to set a default value on the spec, and then modify the message.  Just get the default value directly via `.get()`.  Consider:
```python
resource_spec = msg.get('resource_spec', {})
priority = resource_spec.get('priority', float('inf'))
```

parsl/executors/high_throughput/interchange.py
```diff
@@ -342,8 +341,17 @@ def process_task_incoming(self) -> None:
...
+            queue_entry = (priority, self.task_counter, msg)
```
khk-globus, line 350: If these are negative values, then the task retrieval can pop from the end of the list:
```
queue_entry = (-priority, -self.task_counter, msg)
```
```diff
- *_, task = self.pending_task_queue.pop(0)
+ *_, task = self.pending_task_queue.pop()
```
The latter is a more typical pattern.  The usual reason is that it's more performant to pop from the end of a list than the beginning.  Given `SortedList`'s performance, the argument between `.pop(0)` and `.pop()` is not strong (though it does inherently exist), but `.pop(0)` will make most readers double check the implementation.  Consequently, I encourage use of `.pop()` to reduce future reader mental question marks.

parsl/executors/high_throughput/interchange.py
```diff
@@ -475,10 +483,10 @@ def process_tasks_to_send(self, interesting_managers: Set[bytes], monitoring_rad
             len(self._ready_managers)
         )
 
-        if interesting_managers and not self.pending_task_queue.empty():
+        if interesting_managers and (len(self.pending_task_queue) != 0):
             shuffled_managers = self.manager_selector.sort_managers(self._ready_managers, interesting_managers)
 
-            while shuffled_managers and not self.pending_task_queue.empty():  # cf. the if statement above...
+            while shuffled_managers and (len(self.pending_task_queue) != 0):  # cf. the if statement above...
```
khk-globus, line 489: `SortedList`, like many other Python datastructures, implements `__bool__()`.  Thus, can use truthiness of the data structure directly.  Consider:
```python
if interesting_managers and self.pending_task_queue:
    ...
    while shuffled_managers and self.pending_task_queue:
```

test-requirements.txt
```diff
@@ -16,6 +16,7 @@ globus-compute-sdk>=2.34.0
 # as well as at runtime for optional monitoring execution
 # (where it's specified in setup.py)
 sqlalchemy>=2,<2.1
+sortedcontainers-stubs
```
khk-globus, line 19: Does `sortedcontainers` not need to be in `requirements.txt`

parsl/tests/test_htex/test_priority_queue.py
```diff
@@ -0,0 +1,74 @@
...
+            key: future.result() for key, future in futures.items()
+        }
+
+        # Sort by finish time
+        sorted_by_completion = sorted(results.items(), key=lambda item: item[1])
+        execution_order = [key for key, _ in sorted_by_completion]
+
+        # print("\nExecution Order (priority, submission index):")
+        # for key in execution_order:
+        #     print(f"  {key}")
```
khk-globus, line 61: Left over debugging.  Please remove.

parsl/executors/high_throughput/interchange.py
matthewc2003: Implemented

parsl/executors/high_throughput/interchange.py
matthewc2003: Implemented

khk-globus (approved):

Issue #3849, opened 2025-05-06
Radical executor hangs in test_mpi_funcs.py


matthewc2003:
**Describe the bug**
When running the test_mpi_funcs.py, the test will hang infinitely. 

**To Reproduce**
Steps to reproduce the behavior, for e.g:
1. Clone/setup parsl from repo 
2. Run `pytest -v --log-cli-level=DEBUG parsl/tests/test_radical/test_mpi_funcs.py --config local`
3. Test does not finish, just hangs

**Expected behavior**
Test doesn't hang

**Environment**
 - OS: Ubuntu 24.04 LTS
 - Python version: 3.12.3
 - Parsl version: 1.3.0.dev0

**Distributed Environment**
- Where are you running the Parsl script from ? Standalone Google Cloud Platform VM
- Where do you need the workers to run ? Local

[runinfo_000.tar.gz](https://github.com/user-attachments/files/20052506/runinfo_000.tar.gz)

AymenFJA:
Hey @yadudoc, @matthewc2003 sorry for the late response. I have 2 questions: 

1- Are the Google instances both the same size? i.e. the number of the `vcpus`/`cores`
2-Do you see a `radical.pilot.sandbox` in your `/home/mchun078`? if so can you also attach the session folders (should be with exact same as the one generated in `/home/mchun078/parsl/pytest-parsl/parsltest-current/runinfo/000/rpex.session.instance-20250505-230515.c.ucr-ursa-major-hpc-club.internal.mchun078.020214.0012`). 

I do not see any error in the logs that you attached, besides a `KeyboardInterrupt` signal. Most likely, the issue is on the `agent` side (i.e., the folder generated under `/home/mchun078/radical.pilot.sandbox`)

matthewc2003:
Hi @AymenFJA, the vCPU count (8) and architecture (AMD Genoa) are identical between the two instances. I have attached the session folder for 0012. I hope this helps!

[rpex_session_20250505.tar.gz](https://github.com/user-attachments/files/20312430/rpex_session_20250505.tar.gz)

AymenFJA:
@matthewc2003 indeed helpful thanks. Check the following: 
```log
746514707.447 : rpex.worker.000000   : 15396 : 130104424396480 : DEBUG    : wtq tasks: 1
1746514707.447 : rpex.worker.000000   : 15396 : 130104424396480 : DEBUG    : wtq task.000000 0 - task pulled
[...]

PR #3850, opened 2025-05-06
Fixes CI with Ubuntu 24.04 image

matthewc2003:
# Description

Identifies test_mpi_funcs.py as hanging test in CI, reported in #3849. Skipping this test enables CI to pass under Ubuntu 24.04 GH action runner image. 

# Changed Behaviour

test_mpi_funcs.py now gets skipped during CI

# Fixes

Fixes #3844 

## Type of change

- Bug fix


benclifford:
this test was previously passing -- I don't think it is right to disable it without properly understanding why it hangs now and doesn't in previous parsl CI.

benclifford:
(a different perspective: are you withdrawing support in Parsl for the features that are testing in this test?)

PR #3868, opened 2025-05-29
Move try_assert to exponential backoff

khk-globus:
# Description

In networking contexts, the general advice for polling for specific events is "exponential backoff."  I don't think this is related to our recent CI test failures as 20ms is a super long time to wait (the standard game frame time is 1/60s or ~16.7ms, for example, and considered a low-bar to beat) and all of the test infrastructure is on the same host.  Regardless, exponential backoff is good hygiene in general, so "let's see what happens."

# Changed Behaviour

No user-facing change.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -408,13 +408,13 @@ def try_assert():
         check_period_ms: int = 20,
     ):
         tb = create_traceback(start=1)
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
The exponential backoff looks good to me.

PR #3869, opened 2025-05-29
Only run GCE test if setup exists

khk-globus:
# Description

In practice, this means only PRs created by maintainers will have this test run.  This should hopefully stop the scary failure for external contributors since they will not have access to the repository secrets.

# Changed Behaviour

No product behavior change.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/.github/workflows/gce_test.yaml
+++ b/.github/workflows/gce_test.yaml
@@ -10,8 +10,25 @@ env:
   PYTHON_VERSION: 3.11
 
 jobs:
+  check-secret:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


yadudoc (commented):
I believe that the `if` condition  with `needs` will force the action to the `skipped` state. @khk-globus can you confirm that, please?

khk-globus:
Having talked out of band ... yes, it will show as a gray circle on the left where the green checkmark or red x typically goes.  "Skipped"

yadudoc (approved):
All good now 👍

PR #3871, opened 2025-05-30
Refactor managers to a single ZMQ socket

### LINKED ISSUE #3022: Reduce number of file descriptors used on interchange from manager connections

khk-globus:
# Description

ZMQ ROUTER and DEALER sockets are two way pipes.  I'm not sure why Parsl has largely ignored that.  This commit removes one of the ports.

The two main pieces to address for this change are `interchange.py` and `process_worker_pool.py`.  In the interchange, it's largely to merge the result handling into what was the "outgoing" method.  To disambiguate what message is what, use the first message of a multipart send to say what the message contents are (a simple dictionary).

The change in the manager takes the protocol change into account, but also has to jump through a quick hoop to move the incoming results to the proper thread for upstream communication: that's the `inproc` PAIR socket.

After the above two main changes, the rest is "just cleanup," like:

- removing the two ports (task/result) and replacing with `port` or `worker_port` (singular)

- modifying slightly the command protocol.  BINDS is arbitrary, but is making way for an upcoming second refactor.

- related mypy and test changes

# Changed Behaviour

No workflow should have to change, but with the reduction of a socket per node manager, the interchange should be able to support larger swarms.

# Fixes

Fixes #3022 

## Type of change

- Bug fix
- Code maintenance/cleanup


yadudoc:
It is exactly as @benclifford pointed out. `HTEX` started out with the `interchange` being mostly a passthrough with very limited bookkeeping logic. This did help initially with splitting responsibilities for tasks and results into separate threads/processes. However, this hasn't really worked out as we wanted more functionality and added more bloat to the interchange.

parsl/executors/high_throughput/interchange.py
```diff
@@ -353,116 +342,176 @@ def process_task_incoming(self) -> None:
...
-                    python_version=msg['python_v'],
+        if not self.socks.get(self.worker_sock) == zmq.POLLIN:
+            return
+
+        logger.debug('starting worker message section')
+        msg_parts = self.worker_sock.recv_multipart()
+        try:
+            manager_id, meta_b, *msgs = msg_parts
+            meta = pickle.loads(meta_b)
+            mtype = meta.get('type')
```
yadudoc, line 363: We should fail if we can't get `type`

parsl/executors/high_throughput/interchange.py
```diff
@@ -353,116 +342,176 @@ def process_task_incoming(self) -> None:
...
+                manager['last_heartbeat'] = time.time()
+                logger.debug("Manager %r sent heartbeat via tasks connection", manager_id)
+                self.worker_sock.send_multipart([manager_id, b'', PKL_HEARTBEAT_CODE])
             else:
-                logger.error(f"Unexpected message type received from manager: {msg['type']}")
-            logger.debug("leaving task_outgoing section")
+                logger.warning("Received heartbeat via tasks connection for not-registered manager %r", manager_id)
+
+        elif mtype == 'drain':
+            self._ready_managers[manager_id]['draining'] = True
```
yadudoc, line 508: We already have the manager dict available.

yadudoc (approved):
Following the code walkthrough with @khk-globus, change here look reasonable. I have added some minor cleanup suggestions. We'll have to do a fair bit of cleaning in a separate PR to handle the worker/manager naming ambiguity.

PR #3879, opened 2025-06-16
Move more UDP receiver code to own class/file

benclifford:
This is part of making radios more pluggable.

After this PR, there is a UDPRadioReceiver, patterned after ZMQRadioReceiver which contains process state and can be asked to close.

One MonitoringHub shutdown test no longer tests the UDP receiver process being killed -- the MonitoringHub is no longer directly aware of that process because of the new UDPRadioReceiver class.

There is a new test that is more focused on the UDP code starting and exiting correctly, without involving the MonitoringHub, and which runs much faster.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/radios/udp_router.py
+++ b/parsl/monitoring/radios/udp_router.py
@@ -4,15 +4,26 @@ import logging
 import multiprocessing.queues as mpq
 import os
 import pickle
+import queue
 import socket
 import time
+from multiprocessing.context import SpawnProcess as SpawnProcessType
+from multiprocessing.queues import Queue
 from multiprocessing.synchronize import Event
-from typing import Optional
+from multiprocessing.synchronize import Event as EventType
+from typing import Optional, Union
 
 import typeguard
 
 from parsl.log_utils import set_file_logger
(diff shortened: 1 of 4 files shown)


parsl/monitoring/radios/udp_router.py
```diff
@@ -137,3 +148,53 @@ def udp_router_starter(*,
...
+                               name="Monitoring-UDP-Router-Process",
+                               daemon=True,
+                               )
+    router_proc.start()
+
+    try:
+        udp_comm_q_result = udp_comm_q.get(block=True, timeout=120)
+        udp_comm_q.close()
+        udp_comm_q.join_thread()
+    except queue.Empty:
```
khk-globus, line 192: Should there be any provision for if `q_result` is not an `int`eger?  Or an integer in the valid range?  In a future type case, would reduce a debugging session to reading a log message.

khk-globus (approved):

parsl/monitoring/radios/udp_router.py
khk-globus: Should've been more direct in my question.  "That."

Issue #3350, opened 2024-04-11
worker draining test failure


benclifford:
**Describe the bug**

I've seen this problem in the htex worker draining test occasionally:

```
parsl/tests/test_htex/test_drain.py F

=================================== FAILURES ===================================
__________________________________ test_drain __________________________________

try_assert = <function try_assert.<locals>._impl at 0x7f2dc81bfeb0>

    @pytest.mark.local
    def test_drain(try_assert):
    
        htex = parsl.dfk().executors['htex_local']
    
        # wait till we have a block running...
    
        try_assert(lambda: len(htex.connected_managers()) == 1)
    
        managers = htex.connected_managers()
        assert managers[0]['active'], "The manager should be active"
        assert not managers[0]['draining'], "The manager should not be draining"
    
        fut = f(TIME_CONST * 2)
    
        time.sleep(TIME_CONST)
    
        # this assert should happen *very fast* after the above delay...
>       try_assert(lambda: htex.connected_managers()[0]['draining'], timeout_ms=500)

parsl/tests/test_htex/test_drain.py:66: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
parsl/tests/conftest.py:410: in _impl
    if test_func():
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   try_assert(lambda: htex.connected_managers()[0]['draining'], timeout_ms=500)
E   IndexError: list index out of range
```

That list index out of range makes me think there isn't a worker connected at that point - perhaps it has drained and exited?

**To Reproduce**
heisenbug in CI

**Environment**
CI, PR against master 9378a4d2c64661a019d09f98b8b02c379de53e15

benclifford:
I got this a bunch of times in one day in the last few days, so I dug into log files for one of those runs a bit more to see if I could get more information. Here's a timeline (pulled from various log files). The zip file is attached too, from whence I pulled all the timings (manually).

[runinfo-3.10-3878-merge-12d9e555885027d65b0a3dc8494cafa1f8eda11d.zip](https://github.com/user-attachments/files/20759361/runinfo-3.10-3878-merge-12d9e555885027d65b0a3dc8494cafa1f8eda11d.zip)

[...]

PR #3880, opened 2025-06-16
Make try_assert exponential backoff smoother

benclifford:
Before this PR, a sequence of small random number choices could make doublings happen quite fast, pushing the delay time to be quite high faster than exponential rate and then causing later iterations to wait for longer than might be expected.

This PR makes the delay time increase according to the chosen delay for each iteration.

See issue #3350 for an example of where this happened and was part of a non-deterministic test failure.

This PR should make (by default) try_assert wait up to 2x as long as it takes for the assert condition to become true. This is now parameterised by `factor`, so that it might be set to 1.1 to make the assert wait for up to 1.1x the time it takes for the assert condition to become true.

# Changed Behaviour

Exponential backoff should now be more smoothly exponential.

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/conftest.py
+++ b/parsl/tests/conftest.py
@@ -406,13 +406,15 @@ def try_assert():
         timeout_ms: float = 5000,
         attempts: int = 0,
         check_period_ms: int = 20,
+        factor: float = 2,
     ):
         tb = create_traceback(start=1)
         check_period_s = abs(check_period_ms) / 1000.0
         if attempts > 0:
             for _attempt_no in range(attempts):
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Ran the numbers in a spreadsheet as opposed to strictly in my head; much smoother, still random, but also doesn't increase nearly as fast.  For example, given a 10s timeout, the original algorithm would timeout, on average, after 10 tries.  This approach is less jittery with smaller guaranteed increases per iteration, leading to, on average, 18 tries, and each one, as suggested in the PR description, more closely follow the exponential curve.

I like it; let see it in action.

PR #3887, opened 2025-06-18
Revisit per-task resource monitoring enabling

benclifford:
This PR disables MPIExecutor per task monitoring because it doesn't make sense, and rephrases documentation around this option for my updated understanding of the situation.

# Changed Behaviour

MPI executor won't monitor resource usage of its task process. But this was probably useless anyway: it would have been monitoring the launcher process, not the actual work processes.

## Type of change

- New feature
- Code maintenance/cleanup


benclifford:
tagging @yadudoc for MPI executor review

khk-globus (approved):
Look good to me, so approving.  I'll let you push the merge button when ready.

Issue #3723, opened 2024-12-15
Move Parsl away from multiprocessing `Fork`


benclifford:
Issue #2343 brought up that using multiprocessing's `fork` start method is not safe in the presence of threads. More recently, Python has been more aggressively pushing that line too - https://docs.python.org/3/library/multiprocessing.html states that:

> The default start method will change away from fork in Python 3.14. Code that requires fork should explicitly specify that[.]

Using `fork` in Parsl leads to practical race conditions resulting in mysterious, non-deterministic hangs - this is not a theoretical problem.

PR #3463 switches one use of multiprocessing fork (the HTEX interchange) to a separately launched process. This was quite a bit of effort, and the interchange was already mostly set up to be a separate process due to earlier remote-interchange work that @yadudoc did a long time ago.

The experience there suggests that it would be useful to keep using multiprocessing where it is currently used, but with a different spawn method.

The main user-facing problem I see there is that it would become mandatory to wrap your workflow scripts with this idiom, which https://docs.python.org/3/library/multiprocessing.html#multiprocessing-programming justifies:

```
if __name__ == "__main__":
    your_workflow_here()
```

This new requirement has discouraged me previously from pursuing that approach. However, I pasted this to the #parsl-help slack channel and got largely positive responses that this would not be a serious problem:

```
Here is a thing that is awkward and I would like user input:

* the current way Parsl does multiple processes is fundamentally broken, and results in random-seeming deadlocks
* there is a different Python-recommended way, but that will *require* all user workflow scripts to use the common Python idiom of protecting your main code with:
  if __name__ == "__main__":
    your code here

otherwise it will break. That is the main reason we haven't made that change before.
* there are other more invasive changes that could be made to Parsl to not use this recommended way but that is a lot of work
* the payoff is fewer random-seeming deadlocks in Parsl code (that actually have an explanation)

I am wondering how much requiring this protection idiom will cause trouble for users. I know a lot of people already write their main code this way anyway.
```

In a brief technical test of using `forkserver` or `spawn`, I got at least one problem related to object serialization so this is not an entirely straightforward switch internally either.

I think this switch will make things nicer for people trying to run Parsl on other platforms: for example, see issue #1878 where some errors are related to the absence of the `fork` start method on Windows; and PR #2076 which introduces unpleasant global state changes when running on macOS.

benclifford:
I wrote a blog post about this here https://parsl-project.org/2025/02/26/main-protection.html

PR #3891, opened 2025-06-19
Rationalise monitoring radio logging in a post-fork world

benclifford:
Prior to PR #3817, router processes were launched with multiprocessing fork-no-exec. This carried log configuration into the new process, and made the direction of logs to a log file for the new process vs the other process awkward and confusing.

Post-#3817, router processes start with a clean logging environment. So, this PR makes those processes log to `parsl.*`, following logging conventions, and directs all of `parsl.*` in those processes to the relevant per-process log file.

See issue #3723 for the overarching issue about moving away from fork-no-exec.

# Changed Behaviour

This will make log files longer per-entry because, like parsl.log, the fully qualified module name is now recorded.

This will allow parsl.* log lines from any helper functions to be recorded into the relevant process log - prior to this PR, in that situation, those lines would be discarded - for example, there is now one more log line in the zmq router log, from `wrap_with_logs`.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/radios/filesystem_router.py
+++ b/parsl/monitoring/radios/filesystem_router.py
@@ -16,12 +16,13 @@ from parsl.multiprocessing import SpawnEvent, join_terminate_close_proc
 from parsl.process_loggers import wrap_with_logs
 from parsl.utils import setproctitle
 
+logger = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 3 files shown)


khk-globus (approved):
Awesome; have made basically this change locally as well on the way to something else.

PR #3894, opened 2025-06-24
Teach LocalProvider to use os.kill() directly

khk-globus:
# Description

For the local provider, teach `_is_alive()` to use `os.kill()` directly from in-process, rather than outsourcing to multiple external processes and parsing. In context, this is negligibly more performant, but also removes some unnecessary noise in logs.  The biggest detail is just "simpler."

# Changed Behaviour

There are no user-facing or functionality changes.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/providers/local/local.py
+++ b/parsl/providers/local/local.py
@@ -114,17 +114,15 @@ class LocalProvider(ExecutionProvider, RepresentationMixin):
 
         return [self.resources[jid]['status'] for jid in job_ids]
 
-    def _is_alive(self, job_dict):
-        retcode, stdout, stderr = execute_wait(
-            'ps -p {} > /dev/null 2> /dev/null; echo "STATUS:$?" '.format(
-                job_dict['remote_pid']), self.cmd_timeout)
-        for line in stdout.split('\n'):
-            if line.startswith("STATUS:"):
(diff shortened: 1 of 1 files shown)


yadudoc (approved):
Nice! It looks like this works on MacOS as well.

parsl/providers/local/local.py
```diff
@@ -230,8 +228,9 @@ def submit(self, command, tasks_per_node, job_name="parsl.localprovider"):
                                   stdout, stderr)
         for line in stdout.split('\n'):
             if line.startswith("PID:"):
-                remote_pid = line.split("PID:")[1].strip()
+                remote_pid = int(line.split("PID:")[1].strip())
                 job_id = remote_pid
```
LeiGlobus, line 232: Where does this get used?  I didn't search all the places but the docstring specifies job_id is a string, but you've changed it to int.  I assume the int() is because os.kill expects an int.

LeiGlobus (commented):
Nice cleanup of _is_alive

parsl/providers/local/local.py
khk-globus: That's a good point.  Yes, `remote_pid` should be an integer given `os.kill()`, but the external interface of `job_id` should be a string.  Thanks.

PR #3895, opened 2025-06-24
Update variable names to match concept

khk-globus:
# Description

As a follow-up to the just-landed ZMQ socket reduction, update a couple of variable names.  Even though we currently have `worker_port`, the _workers_ aren't talking to the interchange, the managers are.  Similarly, on the manager side, the connection is no longer task-only.  It's for general communication with the interchange.

# Changed Behaviour

None.

## Type of change

- Code maintenance/cleanup


LeiGlobus (approved):
👍 to more appropriate names

PR #3896, opened 2025-06-24
Probe addresses with app-level connection

khk-globus:
# Description

Two changes of notes:

  - take advantage of `zmq.Poller()` to do the waiting for us; no need manually poll

  - More strongly verify the connection to a manager by receiving an application-level response.  The connection will be guaranteed if encryption is in play.

# Changed Behaviour

No user-facing change; a slight internal, non-mandatory protocol change.

## Type of change

- Code maintenance/cleanup


parsl/executors/high_throughput/probe.py
```diff
@@ -1,62 +1,70 @@
...
-
-    for addr in addr_map:
-        addr_map[addr]['sock'].close()
-
-    return first_connected
+    if not addresses:
+        raise ValueError("No address to probe!")
+
+    sock_map = {}
+    urls = addresses.split(",")
```
rjmello, line 46: Since we split the CSV here, why not accept a list of strings for the argument?

khk-globus: Because this PR is making way for an upcoming change that will no longer pass around the `host` and `port` as two separate entities.  Instead, the executor will start the managers with a precise list of full TCP URLs to attempt.  The core detail of this function is to return a known-good address ... so could move that to outside of the probe function, but it's just as easy to keep the string manipulation isolated here.

parsl/executors/high_throughput/probe.py
```diff
@@ -1,62 +1,70 @@
...
-
-    start_t = time.time()
-
-    first_connected = None
-    while time.time() < start_t + timeout and not first_connected:
-        for addr in addr_map:
-            try:
-                recv_monitor_message(addr_map[addr]['mon_sock'], zmq.NOBLOCK)
-                first_connected = addr
-                logger.info("Connected to interchange on {}".format(first_connected))
```
rjmello, line 50: This change drops a number of useful log messages that I think we should retain in some form. Notably, when the connection is successful or unsuccessful.

rjmello: The former code emits the following logs:

```python
# Success
logger.info(f"Connection to Interchange successful on {ix_address}")
logger.info(f"Interchange url: {ix_url}")

# Fail
logger.exception("Caught exception while trying to determine viable address to interchange")
```

The new code only emits the following:

```python
# Success
logger.info(f"Interchange url: {ix_url}")
```

I suggest we add some wording to clarify the success log and include a failure log.

rjmello (approved):

parsl/executors/high_throughput/probe.py
khk-globus: In the most recent push, we worked through the requested logging changes in-(virtual)person.  The notable changes are:

- In the failure case, add a non-ZMQ language message that speaks directly about the interchange -- the thing of most interest to a non-dev debugging this issue.
- In the happy path case, update the message to say "Probe discovered interchange url"
- After opening the final/main connection the interchange, add a second connection probe to verify the connection is open, and then log _that_ as proof that the manager is connected to the interchange.

PR #3900, opened 2025-06-26
Add HMAC to UDP monitoring messages

benclifford:
This is intended to stop spoofing and other misdelivery of UDP messages.

Each UDP receiver generates a new HMAC key, which is then conveyed to the relevant senders using the config mechanism introduced in PR #3892. That makes key lifetime roughly as long as the DFK/executor lifetime - or around the length of the workflow run.

This PR adds two tests around failing validation:
 i) when an incorrect HMAC key is used
 ii) when the message is so short it cannot be parsed into an HMAC+data

## Type of change

- New feature

### DIFF
--- a/parsl/monitoring/radios/udp.py
+++ b/parsl/monitoring/radios/udp.py
@@ -1,3 +1,4 @@
+import hashlib
 import hmac
 import logging
 import pickle
@@ -29,15 +30,13 @@ class UDPRadio(RadioConfig):
         return UDPRadioSender(self.address, self.port, self.hmac_key, self.hmac_digest)
 
     def create_receiver(self, run_dir: str, resource_msgs: Queue) -> MonitoringRadioReceiver:
-        # 128 byte key length is chosen based on Microsoft's recommendation
-        # here:
-        # https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha512.-ctor
-        # and on comments in RFC 2104 section 2 that the key length be at
+        # RFC 2104 section 2 recommends that the key length be at
         # least as long as the hash output (64 bytes in the case of SHA512).
         # RFC 2014 section 3 talks about periodic key refreshing. This key is
         # not refreshed inside a workflow run, but each separate workflow run
         # uses a new key.
-        self.hmac_key = secrets.token_bytes(128)
+        keysize = hashlib.new(self.hmac_digest).digest_size
+        self.hmac_key = secrets.token_bytes(keysize)
 
         udp_receiver = start_udp_receiver(logdir=run_dir,
                                           monitoring_messages=resource_msgs,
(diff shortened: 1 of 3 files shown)


parsl/monitoring/radios/udp.py
```diff
@@ -23,25 +25,37 @@ def __init__(self, *, port: Optional[int] = None, atexit_timeout: int = 3, addre
...
+        # 128 byte key length is chosen based on Microsoft's recommendation
+        # here:
+        # https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha512.-ctor
+        # and on comments in RFC 2104 section 2 that the key length be at
+        # least as long as the hash output (64 bytes in the case of SHA512).
+        # RFC 2014 section 3 talks about periodic key refreshing. This key is
+        # not refreshed inside a workflow run, but each separate workflow run
+        # uses a new key.
+        self.hmac_key = secrets.token_bytes(128)
+
```
khk-globus, line 40: From a single-point-of-authority standpoint, what about defining the algorithm on the instance?  I'm not suggesting to allow it be specified (but also a possibility), but define it once, here, and then similarly pass it through the start-up chain to `udp_router_starter()`.  Perhaps:

```python
self.hmac_key = secrets.token_bytes(128)
self.digest_function = 'sha512'

udp_receiver = start_udp_receiver(..., hmac_digest=self.digest_function, ...)
```

Then the `origin_hmac = ` and `recomputed_hmac =` lines would not have a hard-coded-and-lucky-we're-looking-now setup.

parsl/monitoring/radios/udp_router.py
```diff
@@ -118,6 +116,25 @@ def start(self) -> None:
         finally:
             logger.info("UDP listener finished")
 
+    def process_message(self) -> None:
+        hmdata, addr = self.udp_sock.recvfrom(2048)
+        origin_hmac = hmdata[0:64]
+        data = hmdata[64:]
```
khk-globus, line 122: 64 is such a specific number.  I'm not thinking we'll ever change the digest function, but all the same, I'm questioning the hard-coding of it.  The HMAC instance returns the digest size, so could do something like:

```python
h = hmac.HMAC(key=self.hmac_key, digestmod=self.digest_function)
origin_hmac = hmdata[:h.digest_size]
data = hmdata[h.digest_size:]
```

I don't know if this still uses the C-optimized digest function though.

khk-globus (approved):
The basic thrust looks good, so I won't quibble of the SPoA suggestions.

parsl/monitoring/radios/udp.py
benclifford: dc608c3dbb adds a new config value for the digest function. It wasn't much extra to expose that through the initializer.

parsl/monitoring/radios/udp_router.py
benclifford: 110bfeda8f removes this 64 and also removes a hard coded 128 byte key length (replacing it with a dynamically computer 64 byte key length based on RFC2014 section 2)

PR #3901, opened 2025-06-26
Flesh out multiprocessing queue monitoring radio

benclifford:
Previously, it was not a proper RadioConfig style radio, but was used ad-hoc to send messages into a multiprocessing queue.

This PR makes it into a real radio which is then used by the ThreadPoolExecutor - allowing monitoring messages to be sent from there within process (rather than the previous UDPRadio default)

# Changed Behaviour

ThreadPoolExecutor will no longer open a UDP listening socket for monitoring. Monitoring messages should flow the same as before.

## Type of change

- New feature

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -191,6 +191,7 @@ Monitoring
     parsl.monitoring.radios.filesystem.FilesystemRadio
     parsl.monitoring.radios.htex.HTEXRadio
     parsl.monitoring.radios.udp.UDPRadio
+    parsl.monitoring.radios.multiprocessing.MultiprocessingRadio
 
 
 Internal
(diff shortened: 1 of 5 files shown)


khk-globus (approved):
I still find it odd that a _thread_ pool uses a mp.Queue.  Alas.

benclifford:
> I still find it odd that a thread pool uses a mp.Queue. Alas.

The radio, at least as far as it is intended to work with ThreadPoolExecutor, could send messages into a non-multiprocessing queue. where they would then be restricted to only being passed to ... another thread in the same process. which would then need to exist and gateway things into the multiprocessing queue towards the database process anyway.

PR #3903, opened 2025-06-30
Restore part of #3801 that was regressed in #3892

benclifford:
#3892 did some code movement, which is hard to review via a diff.

In #3801, the exit mechanism for filesystem radio was switched to event based, rather than hard process termination. Part of that was lost by this bad #3892 code movement.

This PR restores that.

## Type of change

- Bug fix


khk-globus (approved):
Och; good catch.

PR #3917, opened 2025-08-09
Make executor.connected_workers a regular method

benclifford:
See PR #3915 for related justification on the same change for executor.outstanding()

In the connected_workers case, for High Throughput Executor, an RPC to the interchange is made when this attribute is accessed, which is execptionally high failure risk and has personally caused me a lot of trouble in understanding previously.

# Changed Behaviour

User code that is accessing connected workers will have to switch to method syntax, from attribute syntax.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -626,7 +626,6 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         and managers"""
         return len(self.tasks)
 
-    @property
     def connected_workers(self) -> int:
         """Returns the count of workers across all connected managers"""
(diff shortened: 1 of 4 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -626,7 +626,6 @@ def outstanding(self) -> int:
         and managers"""
         return len(self.tasks)
 
-    @property
     def connected_workers(self) -> int:
```
khk-globus, line 629: Thank you!  I wasn't sure how to handle this case, and had it on a local stack to get back to.  In particular, I was observing that even the `hasattr(..., "connected_workers")` was invoking the logic, which is not, I think, what the call-site expects.

PR #3924, opened 2025-08-11
Add Python 3.14 to CI

benclifford:
This PR begins testing against upcoming Python 3.14 to flush out issues ahead of time.

### DIFF
--- a/parsl/concurrent/__init__.py
+++ b/parsl/concurrent/__init__.py
@@ -99,9 +99,10 @@ class ParslPoolExecutor(Executor, AbstractContextManager):
                 passed iterables.
             timeout: The maximum number of seconds to wait. If None, then there
                 is no limit on the wait time.
-            chunksize: If greater than one, the iterables will be chopped into
-                chunks of size chunksize and submitted to the process pool.
-                If set to one, the items in the list will be sent one at a time.
+            chunksize: This parameter is ignored. Caution should be exercised
+                if expecting behaviour as documented in the base `concurrent.futures.Executor` class.
+            buffersize: This parameter is ignored. Caution should be exercised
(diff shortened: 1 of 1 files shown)


benclifford:
proxystore also depends on pydantic so i've removed it from this branch, pending new pydantic

benclifford:
dill release 0.4.0 contains this python 3.14 support. https://github.com/uqfoundation/dill/pull/691 which I take as implying earlier dills will not work properly with 3.14

globus compute pins an earlier dill (and I have noted that to the globus compute dev team on slack) and I wonder if that means the pin will propagate to the parsl test environment -- the test that passed ok, way above here ^, used dill 0.4.0 because GC was disabled in that build.

benclifford:
Now there's a type checking error:

```
parsl/concurrent/__init__.py:91: error: Signature of "map" incompatible with supertype "concurrent.futures._base.Executor"  [override]
parsl/concurrent/__init__.py:91: note:      Superclass:
parsl/concurrent/__init__.py:91: note:          def [_T] map(self, fn: Callable[..., _T], *iterables: Iterable[Any], timeout: float | None = ..., chunksize: int = ..., buffersize: int | None = ...) -> Iterator[_T]
parsl/concurrent/__init__.py:91: note:      Subclass:
parsl/concurrent/__init__.py:91: note:          def map(self, fn: Callable[..., Any], *iterables: Iterable[Any], timeout: float | None = ..., chunksize: int = ...) -> Iterator[Any]
```

The type checker is now more strict about the signature of `map` to be aware of the type correspondence between the output of the callable it is given, and the output of `map` itself.

I'll fix this up in a separate PR.

cc @WardLT who implemented this originally

WardLT:
> cc @WardLT who implemented this originally

Thanks for the heads up! Good to know about the new typing stuff

Issue #3856, opened 2025-05-13
`AttributeError: 'MacSafeQueue' object has no attribute '_counter'` when using MonitoringHub on MacOS with HTEX


d33bs:
**Describe the bug**
Parsl experiences a exception when attempting to run MonitoringHub on MacOS: `AttributeError: 'MacSafeQueue' object has no attribute '_counter'. This results in a hung process which never seems to exit and has to be manually terminated. The `monitoring.db` database seems to be created but not populated afterwards. I tested this quickly within a Ubuntu 22.04.4 environment and the same does not occur (runs successfully and populates the monitoring database).

**To Reproduce**
- MacOS 15
- Python 3.10
- Parsl 2025.5.12
- Install deps: `pip install "parsl[monitoring] numpy`
- Run the following code using `python test.py`

```python
import parsl

from parsl.config import Config
from parsl.executors import HighThroughputExecutor
from parsl.monitoring.monitoring import MonitoringHub
from parsl.addresses import address_by_route

from parsl.app.app import python_app

if __name__ == '__main__':
    # Load a Parsl configuration with the HTEX executor
    parsl.load(Config(
        executors=[
            HighThroughputExecutor(
                label="local_htex",
                cores_per_worker=1,
                max_workers_per_node=4,
                address=address_by_route(),
            )
        ],
        monitoring=MonitoringHub(
            hub_address=address_by_route(),
            monitoring_debug=True,
            resource_monitoring_interval=0.0001,
        ),
        strategy='none'
        ))

    @python_app
    def allocate_memory(task_id):
        import numpy as np
        arr = np.ones((40_000_000,), dtype=np.float64)
        return f"Task {task_id} allocated memory successfully"

    # Launch 3 memory-using tasks
    futures = [allocate_memory(0)]

    # Collect and print results
    for fut in futures:
        print(fut.result())

    parsl.dfk().cleanup()
```

**Expected behavior**
I'd hope that the process is able to complete successfully or provide helpful information on how to troubleshoot or diagnose the challenge.

**Distributed Environment**
- Parsl was launched on a local MacOS machine for testing.

benclifford:
@d33bs I just came across approximately this in PR #3924 which is moving towards making Parsl work on Python 3.14.

[...]

PR #3932, opened 2025-09-02
Remove broken(?) mac/darwin specific code

benclifford:
# Description

MacSafeQueue seems to be broken - on macs (issue #3856) , and as of Python 3.14, increasingly more so on Linux (where it is tested) - (PR #3924). 

The only use of the features of MacSafeQueue is for queue length, and there only to determine if it is non-zero. This PR changes that to a test for the queue being empty, at every occurrence, meaning the same behaviour will be used on all platforms. I am hoping this fixes @d33b's problem on PR #3856.

This PR then removes the now-unused MacSafeQueue entirely, which should fix the test failure in PR #3924.

I saw a couple of other Darwin related changes which make multiprocessing use `fork` by default across the whole Python environment. All of Parsl explicitly launches either a Fork process or a Spawn process now, so this default probably isn't used by Parsl on Darwin any more, and so the code is gratuitously messing with global state. This PR removes that.

I intend to merge this PR in stages - basically the commits on this branch one by one.

# Changed Behaviour

Monitoring might work on mac now. Asking @d33bs.

If you're used to Parsl forcing your own multiprocessing code to use `fork`, that might break.

## Type of change

- Code maintenance/cleanup


benclifford:
heads up @yadudoc that I have this in the pipeline as you made a bunch of the original functionality

d33bs:
@benclifford - apologies for the delay! The monitoring database created on the test MacOS environment looked to be sensible. I didn't do a deep dive on the data but all tables appeared and were populated with various records. Are there any tests I could run on the data to ensure integrity for the Parsl changes you're making?

benclifford:
@d33bs the monitoring test suite isn't really set up for running on an environment that is very different from the GitHub Actions linux CI environment - so I'm wary of getting bogged down figuring out how dependencies should be installed / ignored / ...

you could try this command which will run one specific monitoring test that uses htex:

pytest 'parsl/tests/test_monitoring/test_basic.py::test_row_counts_base[htex_config]' --config local

PR #3936, opened 2025-09-03
Reduce SizedQueue to a regular Queue in monitoring radios

benclifford:
The radios do not make use of the .qsize() functionality of queues, and so th (broken, see #3856) platform-specific sized queue functionality is not necessary.

This is part of a series of PRs to remove that platform-specific functionality entirely - see PR #3932.

# Changed Behaviour

This code is part of broken behaviour on Macs. The behaviour will be slightly better but still broken after this PR on that platform.

Otherwise, the queue should not behave any differently.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Oof; I should start following through on some of the along-the-way details I run across.  I had done exactly this PR's work locally for tangential reasons, but didn't get back to verify that it was a valid change for beyond just my laptop.

Thanks for this and the related verification research.

PR #3952, opened 2025-09-15
Refactor packing and unpacking with resource spec

khk-globus:
# Description

Refactor packing and unpacking with resource spec

The fundamental change is:

```diff
- f, a, k, resource_spec = unpack_res_spec_apply_message(...)
+ f, a, k = unpack_apply_message(...)
```

in `execute_task()`.  This re-generalizes the `execute_task()` implementation for all executors. From there, everything else is "cleanup" to not pack the resource specification within the function buffer, or in the HTEX case, to use the `resource_spec` passed along-side.

This removes the need entirely for `res_spec` variants of `pack_res_spec_apply_message` and `unpack_res_spec_apply_message`.  The requisite functionality (`for`-loop) is now in the HTEX's worker (`process_worker_pool.py`).

# Changed Behaviour

No user workflows should change; this is an internal API refactor.

## Type of change

- Code maintenance/cleanup


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -753,8 +753,8 @@ def manager_is_alive():
             worker_enqueued = True
 
         try:
-            # The worker will receive {'task_id':<tid>, 'buffer':<buf>}
             req = task_queue.get(timeout=task_queue_timeout)
+            # req is {'task_id':<tid>, 'buffer':<buf>, 'context':<dict>}
```
benclifford, line 757: in a few other places i've pushed towards TypedDict for this. I don't really know how much enforcement you get on the receiving end (none?) but it's at least something on the sending side.

benclifford: i don't have a problem with this in this PR, it a passing comment on the theme of "never write comments as `# comments`".

khk-globus: Heh.  Believe it or not, that comment was a straw on the motivation-camel for my thought-process on the MessagePack route.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -603,6 +603,10 @@ def update_resource_spec_env_vars(mpi_launcher: str, resource_spec: Dict, node_i
 
 
 def _init_mpi_env(mpi_launcher: str, resource_spec: Dict):
+    for varname in resource_spec:
```
yadudoc, line 606: I suspect that env vars could persist over to the following tasks because we do not explicitly clear them. This is unrelated to this PR.

parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -766,7 +766,13 @@ def manager_is_alive():
             ready_worker_count.value -= 1
         worker_enqueued = False
 
-        _init_mpi_env(mpi_launcher=mpi_launcher, resource_spec=req["resource_spec"])
+        ctxt = req['context']
+        res_spec = ctxt['resource_spec']
+
+        _init_mpi_env(mpi_launcher=mpi_launcher, resource_spec=res_spec)
+        for varname in res_spec:
+            envname = "PARSL_" + str(varname).upper()
```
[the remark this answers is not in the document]
yadudoc: There's definitely room for some cleanup with having one method update the env vars. I also see that the env vars could easily bleed from one task to the next since we do not clear then. I'll clean this up as a follow up PR.

By [convention](https://google.github.io/styleguide/shellguide.html#constants-environment-variables-and-readonly-variables
), environment variables are written in uppercase.

Correct, these env-vars were added in case the user is calling an application that wants to use the info directly rather than Parsl's MPI prefix, like @WardLT 's chemistry libraries.

yadudoc (approved):
I've looked closely at the changes on the `MPIExecutor` side, and everything looks clean. I've added some minor notes on a potential bug unrelated to this PR that needs to be addressed separately.

PR #3958, opened 2025-09-22
Move import to dunder main

khk-globus:
# Description

No practical difference, but remove one import for any library-approach uses of interchange.

# Changed Behaviour

No changed behavior

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
This looks reasonable.

PR #3959, opened 2025-09-22
Allow for a quicker GC

khk-globus:
# Description

No sense in holding on to the possibly-large data structures after they're no longer utilized.  Tell the runtime that it's free to GC them at its convenience.

# Changed Behaviour

No functional change; negligibly more memory aware HTEX workers.

## Type of change

- Code maintenance/cleanup


parsl/executors/high_throughput/process_worker_pool.py
```diff
@@ -792,6 +795,7 @@ def manager_is_alive():
                                         })
 
         result_queue.put(pkl_package)
+        del pkl_package, result_package
```
yadudoc, line 798: If these are explicitly deleted here, why not delete `req` as well?

yadudoc (commented):
I do not have a sense for how far this goes towards garbage collection, but being explicit here does not hurt. Have you instrumented the benefits from these changes? 

I've added a note about another chunky object that you might want to consider deleting.

parsl/executors/high_throughput/process_worker_pool.py
khk-globus: Good thought.  Just added it.

khk-globus:
No, I have not instrumented this.  This change is pittance and was more to help me while thinking through something else.  After the fact, I realized it was good to just remove them anyway in the larger data cases.  Intuitively, I think the effect is going to be negligible.  This is mostly about hygiene and helping me see something else.

yadudoc:
Fair enough.

yadudoc (approved):

PR #3963, opened 2025-09-24
Don't hide original exception

khk-globus:
# Description

The caught exception provided negligible value.  Just remove it -- which is, pragmatically speaking, the same behavior (we were already throwing!), but now let the raw exception bubble rather hiding the origin.

# Changed Behaviour

Pragmatically no difference, but in the case of inter-component serde mismatch or bitrot, now emit a full traceback rather than only a partial one with an unhelpful summarization.

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -501,10 +501,7 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
             else:
 
                 for serialized_msg in msgs:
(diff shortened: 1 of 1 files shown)


benclifford:
out of band i talked with kevin about whether this should be a `raise BadMessage ... from`. But whatever exception is raised here, it crashes the result loop, causing an irrecoverable Parsl hang. So let's give as much gruesome debugging info as possible here without needing to wrap it up too much.

PR #3965, opened 2025-09-24
Introduce a task context in HTEX

khk-globus:
# Description

Introduce a `context` side-along variable for HTEX tasks.  `resource_spec` is an entry within this new dictionary.  For the moment, they serve the same purpose, but I expect `context` to become a more general concept.

Functionally, the only difference currently is a one-layer-deeper nest of dictionaries:

```diff
- msg = {..., "resource_spec": {...}}
+ msg = {..., "context": {"resource_spec": {...}}}
```

# Changed Behaviour

None.

## Type of change

- (Infrastructure for a) new feature

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -709,7 +709,11 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         except TypeError:
             raise SerializationError(func.__name__)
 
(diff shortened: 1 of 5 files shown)


yadudoc (approved):
Neat. This looks reasonable. Do you see any other bits that might go into the `context` for HTEX besides `resource_specification`?

khk-globus:
> Do you see any other bits that might go into the context for HTEX besides `resource_specification`?

Yes.  But I wanted to separate the motivating feature (importable task-specific `execute_task` function) from the required infrastructure, for clarity of PR reviewing.

PR #3966, opened 2025-09-25
Remove support for Python 3.9

benclifford:
# Description

Python 3.9 becomes end-of-life in the next month - see https://devguide.python.org/versions/

This PR executes Parsl's informal policy to drop Python version support around the time that a Python version becomes end-of-life.

At the same time and on a similar topic, PR #3924 investigates adding support for Python 3.14 which has an initial release scheduled soon.

# Changed Behaviour

If you are installing on Python 3.9, you will probably now get package install errors that your Python is too old.
Post this PR, you will start finding the codebase begin to use features which are not supported on Python 3.9 without any further notification.

## Type of change

- Code maintenance/cleanup


tylern4:
I updated the [slurm container](https://github.com/tylern4/slurm-docker-cluster/blob/single-container/Dockerfiles/ubuntu.Dockerfile) to use ubuntu which seems to keep it's system python more up to date.

benclifford:
@tylern4 thanks for updating the container.  This test was still failing after that as the packaging system was now getting upset about installing stuff as root without a Python virtualenv. That was a straightforward change, and is something the main CI workflow already does. Now the slurm test is passing.

benclifford:
slurm fix is separated out into #3970 to merge before this one.

benclifford:
I just reconfigured GitHub in its GUI to no longer require Python 3.9 CI as a mandatory test pass. That's necessary before this PR can be merged.

Issue #3874, opened 2025-06-03
Parsl HTEX hangs due to incorrect `run_dir` references when running multiple executors in sequence within same process


d33bs:
**Describe the bug**

Hello - thanks again for the great work on Parsl! I noticed recently when attempting to use multiple Parsl executors within the same Python process that the 2nd executor configured after clearing the first appears to reference the 1st executor's `run_dir` information. This causes an exception preventing multiple executors from being leveraged. This appears to occur when using customized `run_dir` settings and HTEX. The result is that Parsl hangs without exception (an error message appears in the `run_dir` logs about a missing file).

**To Reproduce**

See this [notebook](https://colab.research.google.com/drive/1ETRvsgVV4NgdSIq9Zj14S24YENfbHjnk?authuser=0#scrollTo=tkLq2Ei8sekO) (or [gist backup](https://gist.github.com/d33bs/4681c7426a81f5142de1aef2e15d8e88)) as a reference.

```python
import shutil
from parsl.app.app import python_app
import parsl
from parsl.config import Config
from parsl.executors import HighThroughputExecutor


@python_app
def add(x, y):
    return x + y

# HTEX run 1

config = Config(executors=[HighThroughputExecutor()])
config.run_dir = "./runinfo_1"

parsl.load(config)

futs = [add(i, i) for i in range(5)]
results = [f.result() for f in futs]
print(results)

parsl.dfk().cleanup()
parsl.clear()

shutil.rmtree("./runinfo_1")

# HTEX run 2

config = Config(executors=[HighThroughputExecutor()])
config.run_dir = "./runinfo_2"

parsl.load(config)

futs = [add(i, i) for i in range(5)]
results = [f.result() for f in futs]
print(results)

parsl.dfk().cleanup()
parsl.clear()

shutil.rmtree("./runinfo_2")
```

**Expected behavior**

I'd expect that after performing a `dkf().cleanup()` and `parsl.clear()` that the executors and their workers would be fully cleaned up and not reference previous configurations.

**Environment**
 - OS: MacOS, Linux
 - Python version: >= 3.9
 - Parsl version: 2025.6.2

**Distributed Environment**
- Where are you running the Parsl script from ? Laptop, Google Colab
- Where do you need the workers to run ? Same as Parsl script

benclifford:
Hi. I'm just getting to this. It looks like a fairly straightforward "mutable default argument" bug in Parsl (the same reason you shouldn't use whatever=[] as a default argument).

[...]

Issue #1662, opened 2020-04-30
Optional arg `if x=None: x=[]` may be in indicator of other mutation bugs


benclifford:
**Describe the bug**
This is a code hygene issue I'd like to investigate further, so I'm noting here.

Sometimes we declare code like this:
def f(x=[]):
  x.append("foo")
  return x

which might be intended to return "foo" appended onto the end of the supplied list, with an empty list by default.

However, the code doesn't work this way:

```
>>> print(f())
['hi']
>>> print(f())
['hi', 'hi']
>>> print(f())
['hi', 'hi', 'hi']
>>> print(f())
['hi', 'hi', 'hi', 'hi']
```

Instead, the parsl codebase often uses a different pattern:
```
def f(x=None):
  if x==None:
    x=[]
  x.append("foo")
  return x
```

This does work, when `x` is defaulted. There is nothing magic about using `None` here as the default. What is important is that the default value is not mutated, and a new default value is constructed at each invocation, by being constructed inside the function body.

The important thing is that the function body is not allowed to the value being passed in via defaulting.

More generally, in cases where this is done, it might be the case that The important thing is that the function body is not allowed to the value being passed in, wherever that value comes from - from a caller or from defaulting.

But the protecting `if` statement above doesn't protect against a caller's values being mutated - in some cases, some kind of protection (copying a mutable input value) might need to happen in all cases, not only in the defaulting case.

PR #3969, opened 2025-09-26
Fix bad global default provider in several executors

benclifford:
This affects the High Throughput Executor, the MPI Executor and Task Vine.

Mutable objects should not be used as default values, and in this case, a mutable LocalProvider is used as the default provider.

This PR changes that to use a `None` default value and makes a new LocalProvider whenever a new High Throughput Executor is made.

See issue #1662 for broader context around mutable defaults.

This PR adds a regression test based on the reproducer supplied by @d33bs in issue #3874.

# Fixes

Fixes issue #3874.

## Type of change

- Bug fix


khk-globus (approved):
Assuming tests pass this time, still approved.

PR #3981, opened 2025-10-01
HTEX submit returns type-aware future

khk-globus:
# Description

Rather than a stock Future, that typing doesn't know will have a `.parsl_executor_task_id` attached, formalize the attribute with a subclass.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup


yadudoc (approved):
Nice, this is definitely cleaner.

PR #3983, opened 2025-10-03
HTEX handles pre-serialized submissions

khk-globus:
# Description

Allow an interested party to utilize a custom task executor with HTEX, rather than the one provided by Parsl.  One example might be if a task is pre-serialized in a custom manner, the default `execute_task()` function will not know how to deserialize and run it.  As of this PR, the HTEX implements the context-variable keys `resource_spec` and `task_executor`.  The `resource_spec` is nominally well-understood at this point.  The `task_executor` is documented in the `submit_payload` docstring.  An example, borrowed from that documentation:

```python
>>> htex: HighThroughputExecutor  # setup prior to this example
>>> ctxt = {
...   "task_executor": {
...     "f": "full.import.path.of.custom_execute_task",
...     "a": ("additional", "arguments"),
...     "k": {"some": "keyword", "args": "here"}
...   }
... }
>>> fn_buf = custom_serialize(task_func, *task_args, **task_kwargs)
>>> fut = htex.submit_payload(ctxt, fn_buf)
```

The custom ``custom_execute_task`` would be dynamically imported, and invoked within the `process_worker_pool.py` worker as if:

```python
from full.import.path.of import custom_execute_task

args = ("additional", "arguments")
kwargs = {"some": "keyword", "args": "here"}
result = custom_execute_task(fn_buf, *args, **kwargs)
```

# Changed Behaviour

There should be no change to existing workflows, but new workflows may be able to use `submit_payload` to fine-tune how tasks are executed within the worker.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -697,19 +697,10 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
 
         self.validate_resource_spec(resource_specification)
 
-        if self.bad_state_is_set:
-            raise self.executor_exception
-
-        self._task_counter += 1
-        task_id = self._task_counter
-
         # handle people sending blobs gracefully
         if logger.getEffectiveLevel() <= logging.DEBUG:
             args_to_print = tuple([ar if len(ar := repr(arg)) < 100 else (ar[:100] + '...') for arg in args])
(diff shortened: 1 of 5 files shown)


yadudoc (approved):
@khk-globus The changes here look reasonable. The only bit that I'd point is that the commit message would benefit a note that the `process_worker_pool` can load a custom serializer on a per-task basis.

Issue #3992, opened 2025-10-14
Should job status poller or executors stop scaling blocks when an executor is in a bad state?


benclifford:
Currently (0d4a261c00ce2e78d6f11786f1772f652dc18d93) and I think since its introduction in #1519, the job status poller continues to scale out blocks even when an executor is in a bad state (in the sense of `set_bad_state_and_fail_all`):

the job status poller doesn't stop scaling up if the executor reports active tasks, and the HighThroughputExecutor  continues to count bad-state-failed tasks as active in htex.outstanding, making those tasks continue to contribute to scaling load.

Generally this would go mostly un-noticed except for occasional reports of spurious block launches.

PR #3991 experienced a race condition in testing which is sensitive to this, though, and which I was able to recreate in my dev environment with a suitably placed 10ms delay -- see https://github.com/Parsl/parsl/pull/3991#issuecomment-3401053135

The main point of this issue is to think about what the behaviour should be here:

if htex is in bad state, perhaps it should not be reporting any outstanding tasks -- perhaps by removing them as they are cancelled or perhaps by reporting 0 tasks.

if an executor is in a bad state, perhaps the job status poller should never try to scale up a block, no matter what spurious scaling load is reported: after all, the executor is in a "bad state" now so we perhaps should be ignoring its state as much as possible?

PR #3994, opened 2025-10-14
Don't strategize executors in bad state

benclifford:
This is consistent with JobStatusPoller.close, which does not scale in executors in bad state.

See issue #3992 for more context. This fixes a race condition that I can create by adding fuzz delays (as described in #3992) and that appeared (without adding fuzz timing) in PR #3991 probably due to reduced time taken per strategy iteration in that PR.

The new use of `bad_state_is_set` is prone to race conditions - that's part of the model of how bad_state_is_set works everywhere in the codebase, and is briefly discussed in PR #3995.

# Changed Behaviour

Scaling will not happen for bad-state executors. I think this is always the right thing to do.

## Type of change

- Bug fix


khk-globus (approved):
Another that I'll approve in distaste; and I still can't bring myself to hit the merge button.

PR #3997, opened 2025-10-14
Fix GCE tests in CI

yadudoc:
# Description

The GCE testing CI action was not setup to install parsl from the local repo but was installing the Parsl version pinned to `globus-compute-endpoint`. With the latest parsl not installed, running tests against the endpoint would trigger errors as @benclifford mentioned [here](https://github.com/Parsl/parsl/pull/3980#issuecomment-3356156807): 
```
E                ModuleNotFoundError: No module named 'parsl.tests.test_python_apps.test_exception'

```

## Type of change

- Bug fix

### DIFF
--- a/.github/workflows/gce_test.yaml
+++ b/.github/workflows/gce_test.yaml
@@ -67,7 +67,8 @@ jobs:
         source .venv/bin/activate
         make deps
         make clean_coverage
-
+        pip install .
+        
         # Temporary fix until fixes make it to a release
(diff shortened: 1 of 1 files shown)


benclifford:
this is going to break in the same but different way when the Parsl API or internal not-API changes in a way that the GC endpoint hasn't tracked yet, right? (as an example, JobStatusPoller going away as an independent entity and merging into BlockProviderExecutor would be a likely thing to happen in the next two years)

PR #3998, opened 2025-10-15
Address race condition in htex priority test

benclifford:
This race condition is in the test, not in the main Parsl codebase, and is that the test was not waiting for tasks to all arrive in the interchange before launching a worker.

This was happening fast enough most of the time, but under some extreme race condition testing, this was breaking the test - I'm working towards that testing being automated in CI, so this is not a theoretical concern.

If the tasks are not all in the interchange when a worker connects, they won't necessarily be run in priority order: the highest priority task of the subset of tasks that has already arrived will be run first, even if it is not the highest priority task of the entire task set.

# Changed Behaviour

this is a test behaviour change, should not be user-exposed

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_htex/test_priority_queue.py
+++ b/parsl/tests/test_htex/test_priority_queue.py
@@ -18,7 +18,7 @@ def fake_task(parsl_resource_specification=None):
 
 
 @pytest.mark.local
-def test_priority_queue():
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
This is an awful hack.  But no one has thought of a better approach given the current infrastructure.  🤮

PR #4026, opened 2025-11-11
Move checkpointing earlier in task completion

benclifford:
Prior to this PR, checkpointing happened in the inherently race-prone handle_app_update where the user has already been informed that the task has completed. This means the user can observe a situation where checkpointing has not happened even though the task appears to have completed.

Because the user-facing app future is now not set by the time of checkpointing, this PR changes the checkpointing code to pass results and exceptions explicitly as a CheckpointCommand object. This aligns with the on-disk checkpoint format and with the explicit separation of results and exceptions in earlier work - eg PR #4017

This new CheckpointCommand object contains support for representing exceptions because coming up in ~6 PRs time is support for choosing to checkpoint exceptions as well as results, as a user-configurable policy.

See #1279 for summary issue on task completion, broader than checkpointing.

# Changed Behaviour

checkpoint now happens at a different time. previously in task_exit mode, users could observe tasks that were completed but not checkpointed. that should now not be observable.

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/memoization.py
+++ b/parsl/dataflow/memoization.py
@@ -316,29 +316,13 @@ class Memoizer:
 
     def update_memo_result(self, task: TaskRecord, r: Any) -> None:
         self._update_memo(task)
-
-        if self.checkpoint_mode == 'task_exit':
-            self.checkpoint_one(CheckpointCommand(task, result=r))
-        elif self.checkpoint_mode in ('manual', 'periodic', 'dfk_exit'):
-            with self._checkpoint_lock:
-                self.checkpointable_tasks.append(CheckpointCommand(task, result=r))
-        elif self.checkpoint_mode is None:
-            pass
-        else:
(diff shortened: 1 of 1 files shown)


parsl/dataflow/memoization.py
```diff
@@ -286,9 +314,29 @@ def check_memo(self, task: TaskRecord) -> Optional[Future[Any]]:
     def update_memo_result(self, task: TaskRecord, r: Any) -> None:
         self._update_memo(task)
 
+        if self.checkpoint_mode == 'task_exit':
+            self.checkpoint_one(CheckpointCommand(task, result=r))
+        elif self.checkpoint_mode in ('manual', 'periodic', 'dfk_exit'):
+            # with self._modify_checkpointable_tasks_lock:  # TODO: sort out use of this lock
+            self.checkpointable_tasks.append(CheckpointCommand(task, result=r))
```
khk-globus: Is this comment a leftover thought-process?

PR #4030, opened 2025-11-26
Make choice of memoizer configurable

benclifford:
# Description

This PR moves memoizer/checkpoint configuration into a class-based plugin mechanism, so that the DFK now takes a `Memoizer` instance, and the user is responsible for passing in a suitable `BasicMemoizer`. The DFK memoization and checkpoint configuration parameters are moved to be arguments of `BasicMemoizer`.

This provides the plugin interface for alternate memoizer/checkpointer interfaces, and provides an opportunity for `BasicMemoizer` parameters to be added/removed without touching the DFK.

This is the main breaking change of PR #3535 checkpoint plugin prototype: it will require any user providing checkpointing or memoization configuration to change their configuration.

# Changed Behaviour

Checkpointing and memoization should behave the same when configured the same, but the configuration interface is rearranged.

## Type of change

- New feature

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -358,10 +358,10 @@ class DataFlowKernel:
                 self._send_task_info(task_record)
 
                 task_record['try_id'] += 1
-                self._update_task_state(task_record, States.pending)
                 task_record['try_time_launched'] = None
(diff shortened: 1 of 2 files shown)


benclifford:
PR #4031 fixes a test that shows that there is a bug in this PR

Issue #43, opened 2017-12-07
Workflow-level checkpointing


kylechard:
We need to provide the ability to record workflow state (e.g., task status, input/output, task command) and export it to an external store (e.g., database, file). Allow for restart from the checkpoint.

yadudoc:
Breaking this issue into concrete tasks : 

- [ ] Enable checkpointing functionality in the dfk.
- [ ] Enable option to auto-checkpoint at defined time-intervals and events such as cleanup
- [ ] Restarts from checkpoints. Restarting could be complex for non-deterministic applications.

For workflows with no-nondeterministic pieces such as conditionals or loops on non-deterministic outputs, can be quite hard and the best we can do is, use the checkpoint as a memoization lookup table.
This approach would work correctly for deterministic workflows. However it might be expensive to track, input file timestamps, hashes of function body, etc to know if a the task and its inputs has not changed.

Alternatively, for purely deterministic workflows we could use task_ids (which are assigned sequentially) to track what apps have completed, and allow for restarting from the checkpoint.

salmanhabib:
Also, don't forget having the run info in a database (this part of both the provenance tracking and checkpointing requirements).

yadudoc:
Adding couple more task to track requirements from @salmanhabib  : 

- [ ]  Checkpoint periodically to a database 
- [ ]  Streaming state to a database

danielskatz:
@salmanhabib - can you please define what you mean by "provenance tracking" here?  What specific information do you want to be stored?

salmanhabib:
@danielskatz Things like run parameters, status of outputs, versions of software/compute modules, lists of failed tasks (if relevant), etc. Basically we want to have the notion of reproducibility baked into the workflow so we can rerun things and get the same answer plus compare against runs where we know certain changes/modifications have been made.

kylechard:
@salmanhabib We also have an issue related to capturing broad provenance information: https://github.com/Parsl/parsl/issues/42.

salmanhabib:
@kylechard Yes, I know, originally I was just saying that there is some overlap between the two.

yadudoc:
Here's what I propose : 

The DFK will have two methods:
1. checkpoint( ) - write all non-checkpointed info to a checkpoint object
2. resume_from( ) that loads all checkpointed information from a checkpoint object

We will support three modes for automatic checkpointing:
1.  (eager) checkpoint at the completion of every task or 
[...]

PR #4031, opened 2025-11-26
Fix memoization-disabled test

benclifford:
Prior to this PR, this test would still (usually) pass when modified to set app_cache=True. It is supposed to fail in such situations.

This PR fixes this in two ways:

- the python_app under test was not set to cache its results, so its results would be ignored for caching.
- the test runs all the apps with the same parameter potentially simultaneously. Those apps would not be considered for memoization reuse, because results are only memoized after completion.

After this PR, manually changing app_cache=True in this test causes the test to fail, as I expect.

I think this test was probably broken by
f0f592ac56bdef42fac899e041eef66fb2d8ee3c as part of issue #43
which introduced the per-app cache parameter, set off by default.

# Changed Behaviour

test should test better. no user facing changes.

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_python_apps/test_memoize_2.py
+++ b/parsl/tests/test_python_apps/test_memoize_2.py
@@ -17,7 +17,7 @@ def local_config():
     )
 
 
-@python_app
+@python_app(cache=True)
 def random_uuid(x):
     import uuid
     return str(uuid.uuid4())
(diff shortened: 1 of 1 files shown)


benclifford:
@yadudoc The mentioned 2018-era commit is yours, so I'm adding you for original context.

khk-globus (approved):

PR #4032, opened 2025-11-26
Move DFK task state update next to monitoring info send

benclifford:
This makes every `_update_task_state()` and `_send_task_info()` appear as a pair right next to each other, in preparation for refactoring.

There should be no calls to one without the other.

Work towards this PR identified one such situation, which drove PR #4023 to remove the unsched state entirely.

This doesn't break any of the tests but does slightly change the order in which related pieces will see states change: for example, the memoizer is now informed about an exception before rather than after the state change. I'm noting that there may be subtle issues here, for when you arrive here from the future via git bisect.

# Changed Behaviour

hopefully none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -362,10 +362,10 @@ class DataFlowKernel:
                 self._send_task_info(task_record)
 
                 task_record['try_id'] += 1
-                self._update_task_state(task_record, States.pending)
                 task_record['try_time_launched'] = None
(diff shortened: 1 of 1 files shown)


khk-globus (approved):
Yeah, I'm looking at the order of state change with a quizzical eye.  I am not in the weeds enough to know how this will affect the future state of the race-machine, but I trust your general instinct.

I suppose YOLO; let's see what happens.

PR #4034, opened 2025-11-27
Send task state monitoring record as part of general state update

benclifford:
Previous PR #4032 should make this PR a straightforward refactoring.

After this PR, _send_task_info invocations and monitoring TASK_INFO message sends now happen in only one place.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -359,14 +359,12 @@ class DataFlowKernel:
 
                 # record the final state for this try before we mutate for retries
                 self._update_task_state(task_record, States.fail_retryable)
-                self._send_task_info(task_record)
 
                 task_record['try_id'] += 1
                 task_record['try_time_launched'] = None
                 task_record['try_time_returned'] = None
                 task_record['fail_history'] = []
                 self._update_task_state(task_record, States.pending)
-                self._send_task_info(task_record)
 
                 logger.info("Task {} marked for retry".format(task_id))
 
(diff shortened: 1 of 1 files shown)


benclifford:
this failed in AppTimeout test in a test that on first observation I think is a race condition in AppTimeout handling:

```
parsl/tests/test_python_apps/test_timeout.py .F

______________________ test_timeout_after_exception[0.01] ______________________

timeout_dur = 0.01

    @pytest.mark.parametrize("timeout_dur", (0.005, 0.01))
    def test_timeout_after_exception(timeout_dur):
        with pytest.raises(TimeoutTestSpecificException):
>           timeout(kernel, timeout_dur)(True)

parsl/tests/test_python_apps/test_timeout.py:35: 
parsl/app/python.py:32: in wrapper
    with AutoCancelTimer(seconds, inject_exception, args=[thread]):
parsl/utils.py:384: in __enter__
    self.start()
/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/threading.py:999: in start
    self._started.wait()
/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/threading.py:655: in wait
    signaled = self._cond.wait(timeout)

self = <Condition(<unlocked _thread.lock object at 0x7ff623814f40>, 0)>
timeout = None

    def wait(self, timeout=None):
        """Wait until notified or until a timeout occurs.
    
        If the calling thread has not acquired the lock when this method is
        called, a RuntimeError is raised.
    
        This method releases the underlying lock, and then blocks until it is
        awakened by a notify() or notify_all() call for the same condition
        variable in another thread, or until the optional timeout occurs. Once
        awakened or timed out, it re-acquires the lock and returns.
    
        When the timeout argument is present and not None, it should be a
        floating-point number specifying a timeout for the operation in seconds
        (or fractions thereof).
    
        When the underlying lock is an RLock, it is not released using its
        release() method, since this may not actually unlock the lock when it
        was acquired multiple times recursively. Instead, an internal interface
        of the RLock class is used, which really unlocks it even when it has
        been recursively acquired several times. Another internal interface is
        then used to restore the recursion level when the lock is reacquired.
    
        """
        if not self._is_owned():
            raise RuntimeError("cannot wait on un-acquired lock")
        waiter = _allocate_lock()
        waiter.acquire()
        self._waiters.append(waiter)
        saved_state = self._release_save()
        gotit = False
        try:    # restore state no matter what (e.g., KeyboardInterrupt)
            if timeout is None:
>               waiter.acquire()
E               parsl.app.errors.AppTimeout

/opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/threading.py:355: AppTimeout
```

I'll re-run to see what happens

PR #4049, opened 2025-12-29
Switch two visualization views to safer SQL parameter style

benclifford:
Prior to this PR, these two SQL statements were formed by direct string substitution, which allowed arbitrary text injection into the SQL string.

I think in the default configuration this isn't a security vulnerability, because whatever is injected at these points is limited in what it can do: the database is already public because thats what parsl-visualize does, and sqlite will not allow other commands to be executed alongside the query.

This was reported by @viralvaghela in
https://github.com/Parsl/parsl/security/advisories/GHSA-f2mf-q878-gh58 (unpublished at time of commit)

# Changed Behaviour

In the happy path, nothing. In error paths, malformed (malicious or not) workflow identifiers will not leak into SQL.

# Fixes

https://github.com/Parsl/parsl/security/advisories/GHSA-f2mf-q878-gh58

## Type of change

- Bug fix

### DIFF
--- a/parsl/monitoring/visualization/views.py
+++ b/parsl/monitoring/visualization/views.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import sqlalchemy
 from flask import current_app as app
 from flask import render_template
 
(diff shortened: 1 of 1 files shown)


viralvaghela (commented):
I have reviewed the PR, the patch looks correct.

PR #4056, opened 2026-01-16
Make BasicMemoizer load checkpoints from config-level run dir

### LINKED ISSUE #4040: Memoizer searches in run-specific sub-folder for checkpoints and ignores previous runs
**Describe the bug**
When I initialize the Memoizer with a checkpoint_mode but no checkpoint_files, parsl looks for checkpoint files in the run_dir. However, before doing this, parsl automatically appends the run-specific sub-folder (002, for example) which is incremented each run. So it doesn't ever find any checkpoints in there.
Relevant lines:
https://github.com/Parsl/parsl/blob/2571116fcf4c869bab760d934c45a8f474d1dcdf/parsl/dataflow/memoization.py#L214

- A simple workaround is to pass checkpoint_files by first finding them externally using get_checkpoint_files from the parent run_dir.
- (Note that I'm on benc-checkpoint-plugins, but the author @benclifford mentioned the behavior is copied from master, so I'm reporting here.)
- Side note: run_dir sometimes refers to the runinfo folder and sometimes to the sub-folders, i.e. runinfo/000? More consistent naming could help mitigate this issue.

**To Reproduce**
1. Do not pass checkpoint_files to the Memoizer
2. Run twice

**Expected behavior**
I would expect the memoizer to load checkpoints from the previous run, i.e. runinfo/000/checkpints, however, it only searches in runinfo/001

**Environment**
 - OS: [ubuntu]
 - Python version: 3.10
 - Parsl version: master

**Distributed Environment**
- Where are you running the Parsl script from ? Slurm cluster
- Where do you need the workers to run ? Slurm cluster

benclifford:
Prior to this, default checkpoint loading loaded from the run-specific run dir. This will not contain any checkpoints from previous runs, as those go to their own different run-specific run dirs.

This PR makes checkpoints be loaded from the config-level run dir, one level up.

This was the intended behaviour of PR #1918 which introduced this auto-loaded feature but it looks like it has never worked and that bug #4040 has existed ever since that PR.

As with other places where the two kinds of run dir are juxtaposed, the naming of parameters to Memoizer.start now looks awkward, but it is consistent with the naming used in a similar situation for MonitoringHub.start, which also requires both.

This PR introduces a test which fails before this PR and passes afterwards.

This PR modifies existing tests that are broken by this bug fix:

* some tests were setting the config.run_dir to tmpd_cwd. This is a violation of the type of that attribute, which is a str, vs tmpd_cwd which is a pathlib.Path. These tests are modified to convert that path to a str.

* conftest.py was making the same type violation and is similarly modified by this PR

* one monitoring test was reliant on checkpoints not being properly reloaded (this test). With this PR, checkpoints from previous test runs are reloaded and the test was not passing. This PR makes that test use tmpd_cwd as its config.run_dir, rather than using the default working directory runinfo/

# Changed Behaviour

more checkpoints will be reloaded. this might surprise users who thought checkpoints were being reloaded.

# Fixes

Fixes #4040

## Type of change

- Bug fix


benclifford:
Python 3.11 test failed with most recent commit with this:

```
 parsl/tests/test_serialization/test_3495_deserialize_managerlost.py F

_______________________ test_manager_lost_system_failure _______________________

tmpd_cwd = PosixPath('/home/runner/work/parsl/parsl/pytest-parsl/parsltest-current/test_manager_lost_system_failure-1_k70r1r')

    @pytest.mark.local
    def test_manager_lost_system_failure(tmpd_cwd):
        hte = HighThroughputExecutor(
            label="htex_local",
            address="127.0.0.1",
            max_workers_per_node=2,
            cores_per_worker=1,
            worker_logdir_root=str(tmpd_cwd),
            heartbeat_period=1,
            heartbeat_threshold=3,
        )
        c = Config(executors=[hte], strategy='simple', strategy_period=0.1)
    
        with parsl.load(c):
            manager_pgid = get_manager_pgid().result()
            try:
>               with pytest.raises(ManagerLost):
E               Failed: DID NOT RAISE <class 'parsl.executors.high_throughput.errors.ManagerLost'>
```

I can't immediately see how this PR would cause that. I also don't recognise it as a standard Parsl test suite race condition. So logging it here for later reference.

PR #4063, opened 2026-01-26
Add an out-of-memory sqlite backed memoizer

benclifford:
This memoizer does not make a distinction between memoization and checkpointing (which the default BasicMemoizer does: memoization is the first stage in memory, and checkpointing is later on disk).

Instead, all operations about whether a task is memoized are presented to an sqlite database. This can make lookup operations slower, but the tradeoff is that no memory is occupied by the memoization table. The main usecase for this is checkpoint restore in situations where a lot of memory is used by the default memoizer's restore of all tasks every run.

The BasicMemoizer handles exceptions differently between memoization and checkpointing: exceptions are memoized (within a run) but not checkpointed.

As this new sqlite memoizer does not have that distinction, this code makes an arbitrary choice to not record exceptions, even in memory.

The code is open to implementing exception memoization/checkpointing in a future PR at the same time as it would be implemented for BasicMemoizer.

# Changed Behaviour

Nothing unless you configure this new component.

## Type of change

- New feature


parsl/dataflow/memoizers/sqlite.py
```diff
@@ -0,0 +1,102 @@
...
+        self.config_run_dir = config_run_dir
+
+        dir = self.checkpoint_dir if self.checkpoint_dir is not None else self.config_run_dir
+
+        self.db_path = Path(dir) / "checkpoint.sqlite3"
+        logger.info("Starting with db_path %r", self.db_path)
+
+        assert sqlite3.threadsafety == 3, "sqlite3 was not built with the right thread safety level"
+        self._connection = sqlite3.connect(self.db_path, check_same_thread=False, autocommit=True)
+        self._cursor = self._connection.cursor()
```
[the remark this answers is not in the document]
benclifford: the parsl configuration style tends to have people define configurations and not necessarily use them, and that pushes for things to start when they are started, not when the object is created in configuration style.
this is not formally defined anywhere but see the different between `htex .start` and `htex.__init__` for the common example.

benclifford:
I did some work to get backwards compatibility pre Python 3.12.

but now thats revealed that at least the sqlite3 used by github's python 3.10 doesn't seem to have the expected thread safety level.

>       assert sqlite3.threadsafety == 3, "sqlite3 was not built with the right thread safety level"
E       AssertionError: sqlite3 was not built with the right thread safety level

ugh

benclifford:
python 3.10 will be EOL in a few months and removed -- it's tempting to make this untested/unsupported on python 3.10, if it tests in github CI ok on other versions, rather than doing more release engineering.

parsl/dataflow/memoizers/sqlite.py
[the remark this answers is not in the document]
benclifford: what I've done is try a style of: type those two variables to be optional, set them to be `None` in __init__ so they always exist. add asserts at the start of each method that requires the stateful calling protocol to have advanced beyond start. that allows the type checker to know that its someone else's responsibility to have made these variables be not-none. (someone else being the checkpoint driving code in the DataFlowKernel)

parsl/dataflow/memoizers/sqlite.py
```diff
@@ -0,0 +1,115 @@
...
+        self.run_dir = run_dir
+        self.config_run_dir = config_run_dir
+
+        dir = self.checkpoint_dir if self.checkpoint_dir is not None else self.config_run_dir
+
+        self.db_path = Path(dir) / "checkpoint.sqlite3"
+        logger.info("Starting with db_path %r", self.db_path)
+
+        assert sqlite3.threadsafety == 3, \
+            f"sqlite3 was not built with the right thread safety level: required level 3, this environment is level {sqlite3.threadsafety}"
```
khk-globus: If this was in `__init__`, it would hit/fail sooner.  Since this is not imported into the `memoizers` namespace, you might consider putting it after the imports at the module-level.  Would die earlier, as soon as it's imported.

parsl/dataflow/memoizers/sqlite.py
```diff
@@ -0,0 +1,115 @@
...
+        self._connection = sqlite3.connect(self.db_path, check_same_thread=False)
+
+        # >= python 3.12, can use autocommit=True in this self._connection call, as recommended in
+        # https://docs.python.org/3/library/sqlite3.html#sqlite3-transaction-control-autocommit
+        # but until then, use legacy transaction control via isolation level.
+        assert hasattr(self._connection, "isolation_level")
+        self._connection.isolation_level = None
```
khk-globus: I'm not clear on this specifics of this, but is there utility in letting [PyUpgrade](https://github.com/asottile/pyupgrade) do the memory and maintenance work for us in 18 months?  That is, if this is written as a `sys.version_info` conditional, `pyupgrade` will do the correct thing when Py311 support is dropped:
```python
if sys.version_info < (3, 12):
    # this branch will be removed entirely when --py312-plus runs:
    ...
else:
    # this branch will be promoted when --py312-plus runs
    ...
```

parsl/dataflow/memoizers/sqlite.py
```diff
@@ -0,0 +1,115 @@
...
+
+            memo_fu: Future = Future()
+
+            if data['exception'] is None:
+                memo_fu.set_result(data['result'])
+            else:
+                assert data['result'] is None
+                memo_fu.set_exception(data['exception'])
+
+            return memo_fu
```
khk-globus, line 92: Style: no need for the else-branch.

khk-globus (approved):

benclifford:
> but now thats revealed that at least the sqlite3 used by github's python 3.10 doesn't seem to have the expected thread safety level.

... well on a re-run it did. hurrah for deterministic build environments.

parsl/dataflow/memoizers/sqlite.py
benclifford: stylistically I've tried to: avoid if statements because its more to reason about. and avoid different behaviours on different versions because thats even more to reason about... and I've seen that in the past turn into "well we really only support version N but I put in some thing I don't really imagine ever being used, for version N-1" (in the same way that a lot of error handling code has traditionally been written).

Seeing as I was fine with dropping 3.10 due to build system not working, it would also be fine to say this component doesn't work on <= Python 3.12 at all...

benclifford:
> > but now thats revealed that at least the sqlite3 used by github's python 3.10 doesn't seem to have the expected thread safety level.
> 
> ... well on a re-run it did. hurrah for deterministic build environments.

... and on another re-run it did not.

So I think I have to do version specific fiddling here, at least in CI, and if I'm doing that fiddling I have a preference for using the python >=3.12 features for this: it's a new feature so this isn't restricting any existing user of older Pythons from existing functionality.

benclifford:
argh. mypy still wants to check on early versions of python, even if the tests don't run.

PR #4082, opened 2026-02-24
strategy refactoring

Arthur-Prince:
# Description

This PR refactors `_general_strategy` to reduce technical debt and improve readability.

Each scaling case (Case 1, 2, 4a, 4b) was extracted into its own helper function, and the initialization logic was consolidated into a dedicated `init_strategy` helper to remove duplication.

This is a structural refactor only. The goal is to make the strategy easier to understand. I believe this will also simplify future work if the strategy becomes a parsl module.

# Changed Behaviour

No behavior changes. Scaling decisions and thresholds remain exactly the same.

# Fixes

N/A

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/jobs/strategy.py
+++ b/parsl/jobs/strategy.py
@@ -148,14 +148,21 @@ class Strategy:
         for executor in executors:
             self.executors[executor.label] = {'idle_since': None, 'first': True}
 
+    def _init_strategy(self, executor: BlockProviderExecutor, prefix: str) -> None:
+        """Run one-time init_blocks scaling for an executor (first poll only)."""
+        label = executor.label
+        if self.executors[label]["first"]:
+            init_blocks = executor.provider.init_blocks
+            logger.debug("%s Scaling out %d initial blocks", prefix, init_blocks)
+            executor.scale_out_facade(init_blocks)
+            self.executors[label]["first"] = False
+
     def _strategy_init_only(self, executors: List[BlockProviderExecutor]) -> None:
-        """Scale up to init_blocks at the start, then nothing more.
-        """
+        """Scale up to init_blocks at the start, then do nothing more."""
         for executor in executors:
             if self.executors[executor.label]['first']:
-                logger.debug(f"strategy_init_only: scaling out {executor.provider.init_blocks} initial blocks for {executor.label}")
-                executor.scale_out_facade(executor.provider.init_blocks)
-                self.executors[executor.label]['first'] = False
+                prefix = f"[Scaling executor {executor.label}]"
+                self._init_strategy(executor, prefix)
             else:
                 logger.debug("strategy_init_only: doing nothing")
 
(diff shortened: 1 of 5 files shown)


benclifford:
I feel like this makes the code less understandable to me. Look at all these new functions which take some of their state from `self` in an object oriented style but some of their state from long parameter lists in a more functional style.

Why can't all these new methods be functions and not have access to any kind of `self`?  vs Why can't all the state be in `self` and no-body has parameters?

Arthur-Prince:
I agree that it looks a bit awkward for the functions to have so many parameters.

My initial idea for structuring it this way was thinking ahead to a future refactor where strategy becomes a Parsl plugin. I thought it would be easier to move the functions around if they were more decoupled from object state.

For the remaining parameters, the main reason I avoided putting them in self is that their state changes on every strategy execution. One alternative could be to introduce something like an ExecutorState object that encapsulates this evolving state, with an update_state() function returning the refreshed snapshot. I think that would be better.

My main motivation overall was that, while writing the unit tests, it was quite difficult to reason about the possible strategy cases, so I tried to make it more explicit.

I’ll make these changes and then you can let me know what you think.

WardLT:
Thanks for taking a look at this, @Arthur-Prince ! I'll aim to get some time to review it this weekend

Arthur-Prince:
Did you look? what do you think?

WardLT:
Thanks for following up. Sorry, I haven't got to it yet

parsl/jobs/strategy.py
```diff
@@ -179,6 +197,117 @@ def _strategy_htex_auto_scale(self, executors: List[BlockProviderExecutor]) -> N
...
+        active_blocks = running + pending
+        self.executors[label].active_slots = active_blocks * tasks_per_node * nodes_per_block
+
+        self.executors[label].active_blocks = active_blocks
+        self.executors[label].tasks_per_node = tasks_per_node
+        self.executors[label].nodes_per_block = nodes_per_block
+        logger.debug('%s Executor %d active tasks, %d active slots, and %d/%d running/pending blocks',
+                     prefix, self.executors[label].active_tasks, self.executors[label].active_slots, running, pending)
+
+    def _case_1_no_tasks(self, executor: BlockProviderExecutor, prefix: str,) -> None:
```
WardLT: Would you mind adding docstring to these functions? Something brief is fine.

WardLT (commented):
I think it's good to go, but do have another request if you have bandwidth and because you just reviewed this code's internals:

Could you document what the options are and how they differ in the [elasticity documentation](https://github.com/Parsl/parsl/blob/master/docs/userguide/configuration/providers/elasticity.rst)? A brief section after Parallelism would be excellent.

Does that work for you? If not, I'll just merge as-is.

WardLT:
Also, I do agree that the class doesn't make much sense, but that might be a large change and the plugins will need some more thought. Adjusting the functions and - ideally - documenting expected behavior moves us towards that direction.

Arthur-Prince:
> Could you document what the options are and how they differ in the elasticity documentation? A brief section after Parallelism would be excellent.

I can do that, but it might take me about a week.

> Would you mind adding docstring to these functions? Something brief is fine.

Sure, I'll add them.

> Also, I do agree that the class doesn't make much sense, but that might be a large change and the plugins will need some more thought. Adjusting the functions and - ideally - documenting expected behavior moves us towards that direction.

Yes, that was also the direction we were planning on PR #4075, but in smaller and easier-to-review steps.

Arthur-Prince:
I forgot to mention one behavior change related to case 1.

Previously, scale-in would only happen when the executor idle duration
was strictly greater than `max_idletime`. I changed this to `>=`
because when `max_idletime == 0` the scale-in would sometimes not
trigger.

In practice this edge case was previously masked by the
`assert idle_since is not None`, which made the idle duration rarely
evaluate exactly to zero. After removing that assert during the
refactoring in PR #4075, some tests started failing because scale-in
was not triggered as expected.

parsl/jobs/strategy.py
```diff
@@ -179,6 +197,147 @@ def _strategy_htex_auto_scale(self, executors: List[BlockProviderExecutor]) -> N
...
+        """
+        Handle the case where the executor has more slots than active tasks.
+
+        This strategy branch is specific to the HighThroughputExecutor (HTEX),
+        which supports a smart scale-in mechanism. Given that capability,
+        this method computes how many blocks correspond to the excess execution
+        capacity and requests scaling in by that amount, while respecting the
+        executor's minimum block limit.
+        """
+        executor_state = self.executors[executor.label]
```
WardLT: I'm with @benclifford on making executor state an argument. The first line of each of these functions is always to lookup the state. So, accessing the state should be pulled out of these classes and used elsewhere.

parsl/jobs/strategy.py
```diff
@@ -179,6 +197,147 @@ def _strategy_htex_auto_scale(self, executors: List[BlockProviderExecutor]) -> N
...
+
+        logger.debug("%s Strategy case 4a: No active slots but some active tasks - could scale out by a single block", prefix)
+
+        if executor_state.active_blocks < executor_state.max_blocks:
+            logger.debug("%s Requesting single block", prefix)
+            executor.scale_out_facade(1)
+        else:
+            logger.debug("%s Not requesting any blocks, because at maxblocks already", prefix)
+
+    def _case_4b_more_slots_than_tasks(self, executor: BlockProviderExecutor, prefix: str, strategy_type: str) -> None:
```
WardLT: If this PR's a step towards #4075, let's make it very clear what the arguments to these functions are. Each are slightly different. Some w/ and w/o the strategy_type.

parsl/jobs/strategy.py
WardLT, line 21: Regarding refactoring the docs, all of this seems unrelated to what a "Strategy" class does and how to use it.

WardLT (changes requested):
I'd like to propose a different direction for refactoring: getting rid of the logic where we implicitly define the `strategy_type` being passed between function by picking which of the wrapper functions (e.g., `_strategy_simple`) we choice, and instead making `strategy_type` a class attribute. 

Setting `strategy_type` as a attribute will have the setting be controlled in the same way as any of the settings (like `max_idletime`).

That would also make `strategize` a formal class method able to be reflected in the API docs.

Arthur-Prince:
I'll try to clarify the direction I had in mind.

My current understanding of the long-term goal is that **Strategy should eventually become a plugin-like component**. In that model I was imagining something like:

- a `Strategy` interface
- `InitOnlyStrategy` implementing `Strategy`
- `SimpleStrategy` extending `InitOnlyStrategy`
- `HtexStrategy` extending `SimpleStrategy`

One complication is that today `strategy`, `max_idletime`, and `strategy_period` are parameters of the DFK. In practice, not every strategy needs them. For example:

- `InitOnlyStrategy` does not use any of those parameters.
- `SimpleStrategy` might not need `max_idletime`, since it only performs scale-in when there are no tasks running on the executor.

Because of that, my idea was that those parameters would eventually be passed to the **constructors of each concrete strategy class** instead of being handled centrally.

PR #4075 proposes migrating the strategy from the DFK to the `BlockProviderExecutor`. That migration already helps move toward that design. Another possibility might be migrating the strategy into the **provider itself**, but that would likely require reimplementing things such as `scale_in`, `scale_out`, `poll_facade`, and `handle_errors` in the provider, and I'm not sure if that is feasible.

So for now, the goal of this PR is only to clean up the existing code without changing behavior. The intention is to make it easier to later introduce the plugin-style strategy classes (`InitOnlyStrategy`, `SimpleStrategy`, `HtexStrategy`).

@benclifford mentioned that smaller PRs are easier to review, which is why some changes that I think should eventually happen are not included yet. For example:

- the logging messages could be improved
- function names could be clearer
- the following dispatch structure should eventually disappear:

```python
self.strategies = {
    None: self._strategy_init_only,
    "none": self._strategy_init_only,
    "simple": self._strategy_simple,
    "htex_auto_scale": self._strategy_htex_auto_scale,
}
```
since those functions would become separate strategy classes.

There are also a few issues that I plan to address later:

- Case **4a** should probably be merged with **case 2**  
- The `prefix` parameter likely should not be passed around everywhere (it could become a class attribute, especially if we end up with one strategy object per executor).

The next PR I was planning is to **remove the logic where Strategy owns the list of executors**, and instead have `JobStatusPoller` manage that list. This change would require further refactoring of the strategy code.

After that is done, it should become clearer how the `Strategy` class should actually look. After that step, the strategy will likely still need to be refactored into **three classes and one interface**.

---

> I'm with @benclifford on making executor state an argument. The first line of each of these functions is always to lookup the state. So, accessing the state should be pulled out of these classes and used elsewhere.

I agree this makes sense. My initial plan was to do that when implementing the PR that removes the executor list from `Strategy`. At that stage I was also planning to remove `executor: BlockProviderExecutor` from several function parameters.

> If this PR's a step towards https://github.com/Parsl/parsl/pull/4075, let's make it very clear what the arguments to these functions are. Each are slightly different. Some w/ and w/o the strategy_type.

Currently `strategy_type` is only used in **case 4b**, because that branch performs an additional check to see whether the executor is an instance of `HighThroughputExecutor`.

The reason is that **only HTEX supports a "smart" scale-in**, where the block that has been idle the longest (based on `max_idletime`) can be removed.

> Regarding refactoring the docs, all of this seems unrelated to what a "Strategy" class does and how to use it.

Those comments were mostly useful for helping me understand how the class works internally. I agree they are not particularly strong as API documentation.

The only actual mistake in the example is that it is missing `max_workers_per_node=2`. But i think we can change it when we create the module.

---

It might be clearer if I first submit a PR that **moves toward having one executor per strategy object**, and then perform the deeper refactoring of the class afterwards.

If you prefer, I can follow that order instead.

For this PR specifically, I think the improvements I can make are:

- rename some functions  
- improve the logging messages  
- merge case 2 with case 4a

All while keeping behavior unchanged.

WardLT:
Thanks for the explanation and working with us to figure out how to make the Strategy class better.

I'm generally in favor of moving to one strategy per executor, and might suggest moving towards so that we're not debating over code that you'll eventually make irrelevant. 

How does that sound to you?

Arthur-Prince:
Benclifford mentioned earlier that he prefers splitting changes into multiple PRs to make review easier, so I’ve been trying to follow that approach. That said, I also like the idea of going back to discuss PR #4075, since that’s where those change were implemented.

In that PR, the migration of Strategy from the DFK to the BlockProviderExecutor has already been done, so it might be a better place to continue the discussion and refine the design. However, the strategy plugin still needs to be discussed and implemented. For now, the idea is as I mentioned before:

> - a `Strategy` interface  
> - `InitOnlyStrategy` implementing `Strategy`  
> - `SimpleStrategy` extending `InitOnlyStrategy`  
> - `HtexStrategy` extending `SimpleStrategy`  

The idea is that `InitOnlyStrategy` would not take any parameters, `SimpleStrategy` would take only `strategy_period`, and `HtexStrategy` would take both `strategy_period` and `max_idletime`. It would also require something like a `set_executor` method, to pass the executor to the strategy (since it needs access to `scale_in`/`scale_out`).

That said, `max_idletime` is currently used in two slightly different ways:
- as the maximum time an executor can stay idle
- as the maximum time a block can stay idle

I’m not sure it makes sense to keep the executor-level idle timeout. My current thinking is that if an executor is idle, it could immediately scale in to `min_blocks` (after `strategy_period`).

If you agree, I can update PR #4075 with this change and we can continue the discussion there. Of course, feel free to suggest any changes as well.

Just to make it explicit: this PR will break config compatibility with previous Parsl versions, and will introduce a `strategy` parameter in the `BlockProviderExecutor` that receives a strategy object.

WardLT:
[Pardon the slow reply. Was at conferences the last ~2 weeks]

> Just to make it explicit: this PR will break config compatibility with previous Parsl versions, and will introduce a strategy parameter in the BlockProviderExecutor that receives a strategy object.

I like the direction, but we would want to go through a deprecation period before removing the strategy parameter. 

Can you implement those strategy classes and then use them as part of the current "strategy" class? That would allow us to both test out the new code and avoid a hard break in capability.

Arthur-Prince:
yeah that makes sense, I hadn’t really thought about keeping backward compatibility, but I agree a deprecation period is a good idea

do you still prefer that I split this into multiple smaller PRs? something like:
1. first refactor `Strategy` into separate classes  
2. then make each strategy instance manage a single executor  
3. then move the strategy thread from the DFK to the executor  

all while keeping compatibility during the transition

or would you rather I put everything together in a single PR?

I will refactor `Strategy` into separate classes in the best way that I can in this PR while I wait your anwser

[I was busy those weeks too, so dont worry]

Arthur-Prince:
I initially considered removing `max_idletime` from `SimpleStrategy`, so that whenever the executor is idle it would immediately scale in down to `min_blocks`.

The idea was that `max_idletime` should only be used to control scale-in of individual idle blocks. At the executor level, the effective behavior corresponds to something like `ceil(max_idletime / strategy_period)`.

However, I ran into issues with `parsl/tests/test_scaling/test_scale_down_htex_unregistered.py`. I was able to make it pass by setting `strategy_period = 0.01`, but I didn’t fully understand the purpose of that test or why this change affected it.

Because of that uncertainty, I decided to keep `max_idletime` in `SimpleStrategy` for now.

If you think it would be better to remove the idle time behavior from `SimpleStrategy`, I can revisit this and update the test accordingly.

Arthur-Prince:
Hi, Just checking in on this PR. Do you have any feedback or updates?

PR #4084, opened 2026-02-26
Create submit cmd scripts in the Provider's script_dir

christopherwharrop-noaa:
# Description

This PR does five things:

1. Adds a `script_dir` argument to all launcher `__call__` methods.
2. Modifies all providers to supply the above `script_dir` argument when calling `self.launcher` when they submit their submit scripts. (`self.script_dir` is initialized to `None` for cloud-based providers)
3. Updates all launcher submit scripts that create command scripts on-the-fly to write those command scripts to the `script_dir` location instead of the CWD.
4. If `script_dir` does not exist on the host where the submit script executes (the filesystem is not shared, for example) then the command script cannot be created in `script_dir` and the original behavior is preserved (`script_dir` is ignored).
5. Adds a series of tests for the launchers to ensure that the scripts are valid bash syntax, that they run, and that they work with and without the new `script_dir` parameter.

This PR closes #4081.

The purpose of the above modification is to change the location where submit command scripts are written.  Instead of being created in the current working directory, a command script will be created in the same directory as the submit script which created it.

For example, instead of having a script called `cmd_parsl.blah.block-0.1772144087.5482771.sh` appear in one's working directory, it will instead appear in the `runinfo` directory alongside the `parsl.blah.block-0.1772144087.5482771` script that created it:

```
$ ls -1 runinfo/001/submit_scripts/
cmd_parsl.blah.block-0.1772144087.5482771.sh
parsl.blah.block-0.1772144087.5482771
parsl.blah.block-0.1772144087.5482771.stderr
parsl.blah.block-0.1772144087.5482771.stdout

```

This does two things:

1. Prevents accumulation of command scripts of unknown origin in directories where Parsl scripts are run
2. Properly associates the command script with the submit script that created it inside the `runinfo` directory.

# Changed Behaviour

The location of command scripts (i.e. `cmd_parsl.blah.block-blah.sh`) will change.  Instead of appearing in the current working directory, they will be written to the appropriate `runinfo` directory.  This is a change in behavior that users may notice.

# Fixes

Fixes #4081 

## Type of change

- Bug fix


parsl/launchers/launchers.py
```diff
@@ -142,6 +142,7 @@ def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int, scri
 [[ "{debug}" == "1" ]] && echo "Found cores : $CORES"
 WORKERCOUNT={task_blocks}
 SCRIPT_DIR="{script_dir}"
+[ -d "$SCRIPT_DIR" ] || SCRIPT_DIR=""
```
khk-globus, line 146: FWIW, we've established by the syntax on line 142 that we're using Bash, and so can use the built-in operator `[[` instead.  Consider:
```shellscript
[[ -d $SCRIPT_DIR ]] || SCRIPT_DIR=""
```
Note that because using `[[` instead, there's also no need to quote the variable.

christopherwharrop-noaa: Fixed.

christopherwharrop-noaa:
I made an attempt at adding tests for the launchers to make sure the scripts work with and without a `script_dir`.  Let me know what you think of the approach.  From what I can tell the latest CI test ran the new tests successfully.

christopherwharrop-noaa:
I am not planning to push any more updates.  This ready for review.

parsl/launchers/launchers.py
```diff
@@ -139,6 +141,8 @@ def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> s
 export CORES=$(getconf _NPROCESSORS_ONLN)
 [[ "{debug}" == "1" ]] && echo "Found cores : $CORES"
 WORKERCOUNT={task_blocks}
+SCRIPT_DIR="{script_dir}"
+[[ -d $SCRIPT_DIR ]] || SCRIPT_DIR=""
```
WardLT, line 145: Does this mean the code will behave differently if the directory does not exist? Why not just have it fail?

parsl/launchers/launchers.py
```diff
@@ -148,20 +152,20 @@ def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> s
     sort -u $PBS_NODEFILE > $SSHLOGINFILE
 fi
 
-cat << PARALLEL_CMD_EOF > cmd_$JOBNAME.sh
+cat << PARALLEL_CMD_EOF > ${{SCRIPT_DIR:+$SCRIPT_DIR/}}cmd_$JOBNAME.sh
```
WardLT, line 155: I'm making sure I understand by Bash. Does this just add a "/" to the end of the Path if it exists?

I was going to suggest having the check in Python, but - if I have it right - I like this solution of ensuring we have that "/" better in Bash.

parsl/launchers/launchers.py
```diff
@@ -11,7 +11,7 @@ class SimpleLauncher(Launcher):
     def __init__(self, debug: bool = True) -> None:
         super().__init__(debug=debug)
 
-    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int) -> str:
```
WardLT, line 14: Could you add the script_dir command into the base class as well?

WardLT (changes requested):

parsl/launchers/launchers.py
christopherwharrop-noaa: The intent behind this is to account for the possibility that the script_dir may not exist on the host where the launcher script actually runs.  And if it does end up using cwd for creation of the cmd script in a remote location, then there's nothing we can do about it anyway.  I think the safest thing to do is to revert to the original behavior if the script_dir doesn't exist in the location where the launcher script is running.

parsl/launchers/launchers.py
christopherwharrop-noaa: Yes.  Exactly.  If SCRIPT_DIR is defined and not empty, then it inserts the value of SCRIPT_DIR with a / appended to it.  This helps prevent a bare "/" from being appended if SCRIPT_DIR isn't set.

parsl/launchers/launchers.py
christopherwharrop-noaa: I forgot that!  Thanks.  I'll add and push again as soon as I can.

parsl/launchers/launchers.py
christopherwharrop-noaa: Sorry, I meant prepended, not appended.  Of course.

khk-globus (approved):
Thank you for this work.

parsl/launchers/launchers.py
christopherwharrop-noaa: @WardLT - I just added this and the test immediately failed because  WrappedLauncher has a "debug" argument that also is not in the base class.  And so now the arguments don't match. 

WrappedLauncher has:
```
    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int, debug: bool = True, script_dir: str = "") -> str:
```

But now the base class has:
```
    def __call__(self, command: str, tasks_per_node: int, nodes_per_block: int, script_dir: str = "") -> str:
```

And that won't work.  It looks like WrappedLauncher is the only one that has the extra debug argument. Should I revert this?  Or would you like me to do something different?

WardLT: It seems you can remove can remove the `debug` input from the WrappedLauncher. I don't believe it's used anywhere and not even clear why it was added (6 years ago)

christopherwharrop-noaa: I'm not sure if just reverting is best or if adding the `debug` argument to all the launcher `__call__` methods and the base class is best.  The `debug` is already present everywhere in the `__init__` methods.  Another possibility is to remove the `debug` arg from WrappedLauncher as I can't see how it gets used.  It seems out of place.

WardLT: I'd remove the debug from the `__call__` of the launcher. As you note, it duplicates the one from `__init__` and isn't defined in the function specification.

christopherwharrop-noaa: Fixed!

WardLT: Nice!

parsl/launchers/launchers.py
WardLT: Great. Thanks for clarifying!

parsl/launchers/launchers.py
WardLT: I would have chose to just have it fail, but falling back to the old behavior is good enough for me too.

WardLT (approved):
LGTM. Thanks, @christopherwharrop-noaa !

PR #4087, opened 2026-04-13
Tighten up typing of `addresses` in HTEX

chris-janidlo:
# Description

Cleans up a couple instances of "stringly-typed" variables in the HTEX internal code, which previously passed the list of candidate addresses as a comma-separated string. While that can't be avoided in between `executor.py` and `process_worker_pool.py` (since the latter is started up as a shell command), it *can* be avoided within the executor class and in the `get_all_addresses` function, freeing us of one call each of `str.split` and `str.join`.

# Changed Behaviour

None

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -289,9 +289,9 @@ class HighThroughputExecutor(BlockProviderExecutor, RepresentationMixin, UsageIn
         self.loopback_address = loopback_address
 
         if self.address:
-            self.all_addresses = address
+            self.all_addresses = {self.address}
         else:
-            self.all_addresses = ','.join(get_all_addresses())
+            self.all_addresses = get_all_addresses()
 
         self.max_workers_per_node = max_workers_per_node or float("inf")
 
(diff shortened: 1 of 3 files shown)


parsl/executors/high_throughput/executor.py
```diff
@@ -289,9 +289,9 @@ def __init__(self,
         self.loopback_address = loopback_address
 
         if self.address:
-            self.all_addresses = address
+            self.all_addresses = {self.address}
```
benclifford, line 292: is this a behaviour change that means users could previous supply multiple addresses in the htex executor init field separated by "," and now they cannot?

(I don't think this is really a problem because it's documented as a single address and there isn't a strong usecase for a user to supply multiple addresses - just trying to get my head around expected changes)

chris-janidlo: Yeah, that would be a behavior change. I'm on the same page that this is an undocumented feature and should be safe to remove.

That said, should I add anything like a changelog entry or a dedicated test in this PR to handle that change more explicitly/robustly?

benclifford: no need for further docs or tests for that, i think

benclifford (approved):

PR #4091, opened 2026-04-24
Introduce LogConfig class and use it for the DFK

benclifford:
# Description

This is the first of a series of patches introducing a log configuration mechanism across various Parsl Python processes, intended to eventually allow richer configuration than the current ad-hoc mechanisms (such as `worker_debug` in the High Throughput Executor, and `initialize_logging` in `Config`)

This PR adds a new log configuration base class, `LogConfig`. Instances of this class can be passed to the DFK as part of Parsl configuration, and the DFK will use that instance to initialize logging rather than the previous hard coded `parsl.log`.

This log configuration is structured to be pickled to other processes, with the configured `LogConfig` retained to be used by other Parsl components at initialization. Future PRs will introduce those further uses.

This is quite similar to the logging system introduced into Academy in https://github.com/academy-agents/academy/pull/295

# Changed Behaviour

Nothing should change if not using the new configuration mechanism. Otherwise `parsl.log` equivalents will be initialized differently, according to the supplied configuration.

## Type of change

- New feature

### DIFF
--- a/docs/userguide/advanced/plugins.rst
+++ b/docs/userguide/advanced/plugins.rst
@@ -102,6 +102,14 @@ This plugin interface might be used to interface other task-like or future-like
 objects to the Parsl dependency mechanism, by describing how they can be
 interpreted as a Future.
 
+Log configuration
+-----------------
+
+The `LogConfig` class provides a place for log configuration, which is planned
+to extend across the various Python processes involved in a workflow
+execution.  Right now, it is only used by the initialize_logging parameter of
+the `DataFlowKernel`.
+
 Removed interfaces
 ------------------
 
--- a/docs/userguide/advanced/plugins.rst
+++ b/docs/userguide/advanced/plugins.rst
@@ -107,7 +107,7 @@ Log configuration
 
 The `LogConfig` class provides a place for log configuration, which is planned
 to extend across the various Python processes involved in a workflow
-execution.  Right now, it is only used by the initialize_logging parameter of
+execution.  Right now, it is only used by the ``initialize_logging`` parameter of
 the `DataFlowKernel`.
 
 Removed interfaces
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -42,6 +42,8 @@ from parsl.executors.base import ParslExecutor
 from parsl.executors.status_handling import BlockProviderExecutor
 from parsl.executors.threads import ThreadPoolExecutor
 from parsl.jobs.job_status_poller import JobStatusPoller
+from parsl.logconfigs.base import LogConfig
+from parsl.logconfigs.file import FileLogging
 from parsl.monitoring import MonitoringHub
 from parsl.monitoring.errors import RadioRequiredError
 from parsl.monitoring.message_type import MessageType
@@ -91,10 +93,22 @@ class DataFlowKernel:
         self.run_dir = make_rundir(config.run_dir)
 
         self._logging_unregister_callback: Optional[Callable[[], None]]
-        if config.initialize_logging:
-            self._logging_unregister_callback = parsl.set_file_logger("{}/parsl.log".format(self.run_dir), level=logging.DEBUG)
-        else:
+
+        self.log_config: Optional[LogConfig]
+
+        if config.initialize_logging is True:
+            # This legacy behaviour deliberately does not pass log configuration
+            # to other components, so that they can preserve their own legacy
+            # logging behaviour.
+            self.log_config = None
+            dfk_log_config = FileLogging(level=logging.DEBUG)
+            self._logging_unregister_callback = dfk_log_config.initialize_logging(log_dir=self.run_dir, log_name="parsl")
+        elif config.initialize_logging is False:
+            self.log_config = None
             self._logging_unregister_callback = None
+        else:
+            self.log_config = config.initialize_logging
+            self._logging_unregister_callback = self.log_config.initialize_logging(log_dir=self.run_dir, log_name="parsl")
 
         logger.info("Starting DataFlowKernel with config\n{}".format(config))
 
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -6,6 +6,7 @@ import datetime
 import inspect
 import logging
 import os
+import pathlib
 import random
 import sys
 import threading
@@ -102,13 +103,13 @@ class DataFlowKernel:
             # logging behaviour.
             self.log_config = None
             dfk_log_config = FileLogging(level=logging.DEBUG)
-            self._logging_unregister_callback = dfk_log_config.initialize_logging(log_dir=self.run_dir, log_name="parsl")
+            self._logging_unregister_callback = dfk_log_config.initialize_logging(log_dir=pathlib.Path(self.run_dir), log_name="parsl")
         elif config.initialize_logging is False:
             self.log_config = None
             self._logging_unregister_callback = None
         else:
             self.log_config = config.initialize_logging
-            self._logging_unregister_callback = self.log_config.initialize_logging(log_dir=self.run_dir, log_name="parsl")
+            self._logging_unregister_callback = self.log_config.initialize_logging(log_dir=pathlib.Path(self.run_dir), log_name="parsl")
 
         logger.info("Starting DataFlowKernel with config\n{}".format(config))
 
--- a/parsl/logconfigs/base.py
+++ b/parsl/logconfigs/base.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import abc
+import uuid
+from collections.abc import Callable
+
+
+class LogConfig(abc.ABC):
+    """Implementations of this class can initialize Parsl logging.
+
+    Parsl logging is built around Python's `logging` system, with
+    an Parsl-provided configuration system that allows multiple
+    configurations to be initialized and deinitialized within the
+    same process.
+
+    Configurations which might be shared between processes should expect to
+    be pickled/unpickled multiple times as they move around a Parsl
+    distributed system.
+
+    A log configuration is identified across the distributed system by
+    the `uuid` attribute. This is intended to allow code calling the
+    log_configuration class to only initialise a particular log configuration
+    once per process.
+    """
+
+    def __init__(self) -> None:
+        self.uuid = str(uuid.uuid4())
+
+    @abc.abstractmethod
+    def initialize_logging(self, *, log_dir: str, log_name: str) -> Callable[[], None]:
+        """Initialize logging in current process.
+
+        This should be implemented by users wanting to define their own log
+        configuration policies.
+
+        This should be called by Parsl components which which to initialize
+        logging according to a user supplied policy, rather than a hard-coded
+        log policy.
+
+        This should return a callback to uninitialize the logging initialized
+        by this call.
+        """
+        ...
--- a/parsl/logconfigs/base.py
+++ b/parsl/logconfigs/base.py
@@ -33,9 +33,9 @@ class LogConfig(abc.ABC):
         This should be implemented by users wanting to define their own log
         configuration policies.
 
-        This should be called by Parsl components which which to initialize
-        logging according to a user supplied policy, rather than a hard-coded
-        log policy.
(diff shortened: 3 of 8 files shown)


docs/userguide/advanced/plugins.rst
```diff
@@ -102,6 +102,14 @@ This plugin interface might be used to interface other task-like or future-like
 objects to the Parsl dependency mechanism, by describing how they can be
 interpreted as a Future.
 
+Log configuration
+-----------------
+
+The `LogConfig` class provides a place for log configuration, which is planned
+to extend across the various Python processes involved in a workflow
+execution.  Right now, it is only used by the initialize_logging parameter of
+the `DataFlowKernel`.
```
khk-globus, line 111: - I can't recall ... the backtick syntax here will automatically linkify those?
- Do you want to monotype `initialize_logging`?  (```... the ``initialize_logging`` parame...```)

parsl/dataflow/dflow.py
```diff
@@ -91,10 +93,22 @@ def __init__(self, config: Config) -> None:
...
 
         self._logging_unregister_callback: Optional[Callable[[], None]]
-        if config.initialize_logging:
-            self._logging_unregister_callback = parsl.set_file_logger("{}/parsl.log".format(self.run_dir), level=logging.DEBUG)
-        else:
+
+        self.log_config: Optional[LogConfig]
+
+        if config.initialize_logging is True:
+            # This legacy behaviour deliberately does not pass log configuration
```
khk-globus, line 101: Stylistically, could treat `config.initialize_logging` as a truthy value.  Not that it could be `None` (per defaults and typing), but there doesn't seem to be a need to check this strictly.  Not a requirement, just a suggestion.

parsl/logconfigs/base.py
```diff
@@ -0,0 +1,43 @@
...
+import uuid
+from collections.abc import Callable
+
+
+class LogConfig(abc.ABC):
+    """Implementations of this class can initialize Parsl logging.
+
+    Parsl logging is built around Python's `logging` system, with
+    an Parsl-provided configuration system that allows multiple
+    configurations to be initialized and deinitialized within the
```
khk-globus, line 13:
```diff
- "... with an Parsl-provided ..."
+ "... with a Parsl-provided ..."
```

parsl/logconfigs/base.py
```diff
@@ -0,0 +1,43 @@
...
+    distributed system.
+
+    A log configuration is identified across the distributed system by
+    the `uuid` attribute. This is intended to allow code calling the
+    log_configuration class to only initialise a particular log configuration
+    once per process.
+    """
+
+    def __init__(self) -> None:
+        self.uuid = str(uuid.uuid4())
```
khk-globus: - `uuid` is a bit overloaded here and specific.
    - Overloaded because `uuid` is already the `uuid` module name.  Fine, this is an attribute on the instance, so "no problem," but seems like we could do better.
    - Meanwhile, I observe that the identifier of the instance happens to be a uuid, but I don't know that it _has to be_.  Consider a name that isn't tied to its type.  It doesn't look like `.uuid` is currently used anywhere?  Is it expected to be used externally to this class?  If not, could do `._id`.  If so, then perhaps `.ident` or ... even `.id`.  It's a weaker argument, but `id()` is utilized rarely enough that I'm less concerned about that overload.
- Also, not that it matters in the least for total functionality — this identifier appears to be ephemeral? — but I wonder if making use of `UUIDv7` could be of debugging help someday?  Just a thought.

parsl/logconfigs/file.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, *, format_string: str = DEFAULT_FORMAT, level: int = logging.DEBUG):
+        super().__init__()
+        self.format_string = DEFAULT_FORMAT
+        self.level = level
+
+    def initialize_logging(self, *, log_dir: str, log_name: str) -> Callable[[], None]:
+
+        os.makedirs(log_dir, exist_ok=True)
+        filename = log_dir + "/" + log_name + ".log"
```
khk-globus: I recognize that Parsl is basically *nix specific, and so this may be a moot point, but this is an opportunity to use `pathlib.Path`:
```python
ldir = pathlib.Path(log_dir)
ldir.mkdir(parents=True, exist_ok=True)
log_path = ldir / f"{log_name}.log"

handler = logging.Filehandler(log_path)
...
```
**Not** arguably simpler, but isolated, and _starts_ the process of thinking about using `pathlib`.  I do like the win that the `__div__` operator gives where it handles the `os.sep` OS differences automatically.  Another suggestion, not a requirement.

parsl/logconfigs/file.py
```diff
@@ -0,0 +1,50 @@
...
+
+        handler = logging.FileHandler(filename)
+        handler.setLevel(self.level)
+        formatter = logging.Formatter(self.format_string, datefmt='%Y-%m-%d %H:%M:%S')
+        handler.setFormatter(formatter)
```
khk-globus, line 25: Given _some_ reports of logs getting unruly, is this an opportunity to think in terms of [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler)?

I don't have the total-view of this yet, so perhaps not and perhaps we could think about that at a higher-level.  Stepping back, I've also heard from a number of folks that they'd like all Parsl logs funneled into a single file.  So ... I know that I don't have the full picture in mind.

parsl/logconfigs/base.py
```diff
@@ -0,0 +1,43 @@
...
+    def initialize_logging(self, *, log_dir: str, log_name: str) -> Callable[[], None]:
+        """Initialize logging in current process.
+
+        This should be implemented by users wanting to define their own log
+        configuration policies.
+
+        This should be called by Parsl components which which to initialize
+        logging according to a user supplied policy, rather than a hard-coded
+        log policy.
+
```
khk-globus: "which which"

parsl/logconfigs/base.py
```diff
@@ -0,0 +1,43 @@
...
+        This should return a callback to uninitialize the logging initialized
+        by this call.
+        """
```
khk-globus, line 34: I don't have a sense of this yet, but this strikes me as C-think.  Not a problem, but I'm also put in mind of a contextmanager.  It seems like `__enter__`/`__exit__` would be a natural fit here, and that the logging instances could take care of their own shutdown.

parsl/logconfigs/file.py
```diff
@@ -0,0 +1,50 @@
...
+            logger.setLevel(self.level)
+        else:
+            logger.setLevel(min(logger.level, self.level))
+
+        logger.addHandler(handler)
+
+        # Concurrent.futures errors are also of interest, as exceptions
+        # which propagate out of the top of a callback are logged this way
+        # and then discarded. (see #240)
+        futures_logger = logging.getLogger("concurrent.futures")
```
khk-globus, line 39: Oh interesting!  I hadn't clocked this aspect.  Thanks for that tidbit.

parsl/logconfigs/file.py
```diff
@@ -0,0 +1,50 @@
...
+            futures_logger.setLevel(self.level)
+        else:
+            futures_logger.setLevel(min(futures_logger.level, self.level))
+        futures_logger.addHandler(handler)
+
+        def unregister_callback():
+            logger.removeHandler(handler)
+            futures_logger.removeHandler(handler)
+
+        return unregister_callback
```
khk-globus, line 50: Where is this called, currently?  Am I missing where the DFK uses this?

khk-globus (commented):
This overall looks like a nice new addition, but I've a number of questions.  Nothing blocking per se, but given the number of comments I had, thought we'd converse first.

parsl/dataflow/dflow.py
benclifford: stylistically I try to stay away from "truthy"-style when the value is a union type: True, the bool, specifically means something different from LogConfig objects which should hit the third (`else`) case of this switch statement - and those LogConfig objects can (and probably will) be truthy-true.

This is basically backwards compatibility: I'm expecting users to already be passing in True or False here. Otherwise, the whole if statement could be changed to force `LogConfig` style configuration.

parsl/logconfigs/base.py
benclifford: fixed in b6e8d7355c

parsl/logconfigs/base.py
benclifford: yeah this naming has bothered me before. ident is probably better.

parsl/logconfigs/file.py
benclifford: RotatingFileHandler should probably never be used in this context: in the bigger picture, multiple processes will all be initializing the same LogConfig, and I think that would work badly with RotatingFileHandler at rotation time. (Globus Compute had some related weirdness but I don't remember).

There is scope for plugging in a rotating file handler that also separates out by process (or some other suitable ID) to stop multiple processes rotating the same log file -- and that is pretty much exactly the sort of thing this plugin mechanism is designed to facilitate.


But in the context of this PR, this particular handler is meant to be a boring implementation of the existing log functionality - I'll add a note in my patch stack about this, though.

> they'd like all Parsl logs funneled into a single file

yes the opposite direction is something more like a CloudWatch style single location where you interact with your logs in a more query-like style rather than pre-determined process buckets. Over in Academy-land, myself and Kyle have a student looking at how that might interact with Diaspora Octopus, one of @ryanchard 's projects -- and I hope that work will be directly applicable here in Parsl too. (and again out of scope for this PR, except that LogConfig is the mechanism that would configure it)

parsl/logconfigs/base.py
benclifford: corrected in fe0e761314

parsl/logconfigs/file.py
benclifford: The call to this already existed before this PR, to uninitialize the previous implementation of log configuration. here after everything else is shut down https://github.com/Parsl/parsl/pull/4091/changes#diff-925607d59cf08d406591135640088fe4e3f372dec7769adc4dc6a49565cfac6fR1196

parsl/logconfigs/base.py
benclifford: I think we had a discussion like this before.

Parsl's startup/shutdown isn't implemented in a context manager style: there's a pile of initializations in DFK start, and a pile of uninitializations in DFK cleanup().

Syntactically, its fairly unlikely end users will be using `with` lexical style to start/end logging because this is intended to fit into the existing log initialization points.

(you *could* rearchitect the DFK startup/shutdown to look like a bunch of stacked context managers but that is not how it is)

This log config object is intended to be copied around between processes - which doesn't happen in this PR, but you can see it in the enclosing PR https://github.com/Parsl/parsl/pull/4089 - and so any context to be remembered for exiting/shutdown would have to go into a new object.

... which is what's happening here: the new object is the callback function object, rather than an instance of an explicitly named class that contains the same information in explicit fields.

parsl/logconfigs/file.py
benclifford: This is mostly a copy and paste from other log initialization, but I'll take the opportunity to port to pathlib.

parsl/logconfigs/base.py
benclifford: > Is it expected to be used externally to this class?

Yes. The point is to give the object an identity which survives pickling by multiple routes. It's not super urgent in Parsl but in the related Academy code, it is used to track when a log configuration has already been initialized in a process and not reinitialize it, no matter which route the configuration arrived in the process. There isn't actually much need for that in Parsl and I could probably remove this from this PR and add it later if/when needed.

docs/userguide/advanced/plugins.rst
benclifford: single backticks turns into a link to the relevant API documentation -- I just check these two in a local doc build on my laptop.

I just added double backticks around initialize_logging

parsl/logconfigs/file.py
khk-globus: Oh, I see.  Clearly there, in retrospect.  Thanks.

khk-globus (approved):

PR #4095, opened 2026-04-29
Tidyup references and asserts around decoding tagged message

benclifford:
The message is already asserted to be length 2 so remove another duplicate assert.

In this two-element case, x[-1] is the same as x[1] so make that consistent across all cases.

One case uses an intermediate variable where others don't. Make that consistent across all cases too.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -576,16 +576,13 @@ class DatabaseManager:
         if x[0] in [MessageType.WORKFLOW_INFO, MessageType.TASK_INFO]:
             self.pending_priority_queue.put(cast(Any, x))
         elif x[0] == MessageType.RESOURCE_INFO:
-            body = x[1]
-            self.pending_resource_queue.put(body)
+            self.pending_resource_queue.put(x[1])
         elif x[0] == MessageType.NODE_INFO:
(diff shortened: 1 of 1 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -576,16 +576,13 @@ def _dispatch_to_internal(self, x: Tuple) -> None:
...
-            assert len(x) == 2, "expected NODE_INFO tuple to have exactly two elements"
-
             logger.debug("Will put {} to pending node queue".format(x[1]))
             self.pending_node_queue.put(x[1])
         elif x[0] == MessageType.BLOCK_INFO:
             logger.debug("Will put {} to pending block queue".format(x[1]))
-            self.pending_block_queue.put(x[-1])
+            self.pending_block_queue.put(x[1])
         else:
             logger.error("Discarding message of unknown type {}".format(x[0]))
```
khk-globus, line 587: I guess it saves a destructure operation, but I am naively looking for:
```python
msg_type, body = x
```
after the assert.  Meh.

khk-globus (approved):

parsl/monitoring/db_manager.py
benclifford: me too, but one case doesn't use that pattern and I didn't want to go off poking in that direction

PR #4097, opened 2026-05-04
Split RESOURCE_INFO monitoring messages into two types

benclifford:
Prior to this PR, there were two types of message sharing the same tag, RESOURCE_INFO. These were:

Task execution start/end messages, with a first_msg or last_msg flag set. These do not contain resource usage information. These messages indicate that a task has started execution on a worker, or ended execution on a worker. In this PR, these messages become WORKER_TASK_INFO messages.

Messages with neither of those flags set must contain resource usage information.  These messages are send periodically during execution, and don't indicate anything about the start of end of a task. In this PR, these messages remain as RESOURCE_INFO messages.

Most of this PR is variable renaming. The logic changes are:

Split handling of received RESOURCE_INFO messages into two cases, the two cases above, which makes them less intertwined in the database manager now.

Remove first_msg and last_msg flags from resource usage messages as these messages are now identified by the RESOURCE_INFO tag rather than needing both first_msg and last_msg to be set to false.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/monitoring/db_manager.py
+++ b/parsl/monitoring/db_manager.py
@@ -306,6 +306,7 @@ class DatabaseManager:
         self.pending_node_queue: queue.Queue[MonitoringMessage] = queue.Queue()
         self.pending_block_queue: queue.Queue[MonitoringMessage] = queue.Queue()
         self.pending_resource_queue: queue.Queue[MonitoringMessage] = queue.Queue()
+        self.pending_worker_task_queue: queue.Queue[MonitoringMessage] = queue.Queue()
 
         self.external_exit_event = exit_event
 
@@ -337,15 +338,15 @@ class DatabaseManager:
         # Exactly one message per task/try can be deferred, which may not be
         # enough when a try happens several times (for example when an
         # executor performs its own Parsl-try unaware retries.
-
         inserted_tasks: Set[object] = set()
         inserted_tries: Set[object] = set()
-        deferred_resource_messages: dict[str, MonitoringMessage] = {}
+        deferred_worker_task_messages: dict[str, MonitoringMessage] = {}
 
         exception_happened = False
 
         while (not self._kill_event.is_set() or
                not self.pending_priority_queue.empty() or not self.pending_resource_queue.empty() or
+               not self.pending_worker_task_queue.empty() or
                not self.pending_node_queue.empty() or not self.pending_block_queue.empty() or
                not resource_queue.empty()):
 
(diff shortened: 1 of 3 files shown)


parsl/monitoring/db_manager.py
```diff
@@ -487,62 +488,74 @@ def start(self,
...
+                if resource_messages:
+                    logger.debug("Got {} messages from resource queue".format(len(resource_messages)))
+
+                    insert_resource_messages = []
+                    for msg in resource_messages:
+                        task_try_id = str(msg['task_id']) + "." + str(msg['try_id'])
+                        insert_resource_messages.append(msg)
+
+                    if insert_resource_messages:
+                        self._insert(table=RESOURCE, messages=insert_resource_messages)
```
khk-globus: What's happening with `task_try_id`?  Is that used elsewhere that ... I'm not seeing?  If not, this could reduce to a list comprehension or potentially even just a list:
```python
insert_resource_messages = list(resources_messages)
```
or
```python
# no need for internal if-block, if I understand the truthy condition on line 548?
self._insert(table=RESOURCE, messages=list(resources_messages))
```

benclifford: That `if` should always fire now -- previously there might be resource messages that don't go here because they are (what is now) WORKER_TASK_INFO messages. But not any more.

benclifford: that task_try_id is left behind from this loop being split into two -- it's needed in the other `if` cases in the sibling/copy/split of this loop, here https://github.com/Parsl/parsl/pull/4097/changes#diff-cc882bfcdc4c84a015b83c68e077903be0fa64d561632953f29f7c28c7e6991bR504 and isn't needed in the case that remains in this sibling/copy/split.

benclifford: `resource_messages` can be passed directly to `self._insert` now, like NODE_INFO handling a little bit further up. https://github.com/Parsl/parsl/pull/4097/changes#diff-cc882bfcdc4c84a015b83c68e077903be0fa64d561632953f29f7c28c7e6991bR469

resource_messages is already a list, and with all the conditionals removed by this PR, this is now just a list to list copy.

khk-globus (approved):
Awesome; that looks on target.

parsl/monitoring/db_manager.py
khk-globus: Looks all addressed.  Good deal.

PR #4098, opened 2026-05-07
Allow disabling of both parsl and python version checks between IX and managers.

yadudoc:
# Description

Reids work in #3951 added support for disabling the Python version checks between the interchange and manager using a private flag `_check_python_version`. This flag does not disable parsl version checks which means strict version matches are required between the components, which is troublesome in Globus Compute deployments.

This PR expands the scope of the `_check_python_version` flag to toggle version checks for both python and parsl versions. I considered adding an additional `_check_parsl_version` flag but decided against it due to the additional backwards compatibility headaches.

Here are two options:
1. Overload the existing `_check_python_mismatch` to also ignore parsl version mismatches.
```
MEP updated   Worker updated    Compatibility
       No                     No                  Status quo
       No                     Yes                 Status quo
       Yes                    No                  New behavior, version checks ignored
       Yes                    Yes                 New behavior, version checks ignored
```
Cons: Flag does more than what the name implies, (but it is hidden from users)

2. Add a new `_check_parsl_mismatch` flag.
```
MEP updated, Worker updated,  Compatibility
       No                     No                  Status quo
       No                     Yes                 Status quo
       Yes                    No                  Worker lanch fails: Unknown _check_parsl_mismatch flag
       Yes                    Yes                 New behavior, version checks ignored
```
Cons: This approach will be disruptive, because we'll need both admins and users to update the EP envs

# Changed Behaviour

This change does not affect Parsl users. Only GC uses this private flag AFAIK.
For GC users, when EP administrators update the parsl versions, users will no longer see VersionMismatch errors that terminate endpoints.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/interchange.py
+++ b/parsl/executors/high_throughput/interchange.py
@@ -102,8 +102,9 @@ class Interchange:
 
         _check_python_mismatch : bool
             If True, the interchange and worker managers must run the same version of
-            Python. Running different versions can cause inter-process communication
-            errors, so proceed with caution.
+            Python and Parsl. Running different versions can cause inter-process
+            communication errors, so proceed with caution.
+            Default: True
         """
         self.cert_dir = cert_dir
         self.logdir = logdir
@@ -404,7 +405,10 @@ class Interchange:
 
             python_mismatch: bool = ix_minor_py != mgr_minor_py
             parsl_mismatch: bool = ix_parsl_v != mgr_parsl_v
-            if parsl_mismatch or (self._check_python_mismatch and python_mismatch):
+            logger.warning(f"Yadu : {self._check_python_mismatch=}")
+            logger.warning(f"Yadu : {python_mismatch=} {ix_minor_py=} {mgr_minor_py=}")
+            logger.warning(f"Yadu : {parsl_mismatch=}  {ix_parsl_v=} {mgr_parsl_v=}")
+            if self._check_python_mismatch and (parsl_mismatch or python_mismatch):
                 kill_event.set()
                 vm_exc = VersionMismatch(
                     f"py.v={ix_minor_py} parsl.v={ix_parsl_v}",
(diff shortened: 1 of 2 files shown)


yadudoc:
Adding a reference to an older issue for background: https://github.com/Parsl/parsl/issues/2957

parsl/executors/high_throughput/interchange.py
```diff
@@ -404,7 +405,10 @@ def process_manager_socket_message(
 
             python_mismatch: bool = ix_minor_py != mgr_minor_py
             parsl_mismatch: bool = ix_parsl_v != mgr_parsl_v
-            if parsl_mismatch or (self._check_python_mismatch and python_mismatch):
+            logger.warning(f"Yadu : {self._check_python_mismatch=}")
```
benclifford: maybe tidy up this log block so it looks ok for regular parsl users

benclifford (approved):
i've privately expressed concerns to Globus Compute about the sensibility of this emergency measure, but those shouldn't affect regular Parsl users so no need to delay the hilarity.

parsl/tests/test_htex/test_interchange_exit_bad_registration.py
```diff
@@ -78,41 +201,41 @@ def test_exit_with_bad_registration(tmpd_cwd, try_assert):
...
-           'start_time': time.time(),
-           'prefetch_capacity': 0,
-           'max_capacity': 1,
-           'os': platform.system(),
-           'hostname': platform.node(),
-           'dir': os.getcwd(),
-           'cpu_count': psutil.cpu_count(logical=False),
-           'total_memory': psutil.virtual_memory().total,
-           }
+    msg = {
```
benclifford: i feel like this test is modified way more than it needs to be - changing ' to " shouldn't be part of a version matching test, for example? (or at least the PR should describe why?). I'd prefer you make a more readable diff before merging this as it isn't clear to me without putting in a bunch more effort what you intended to change - so good luck to readers from the future.

parsl/executors/high_throughput/interchange.py
yadudoc: Removed.

parsl/tests/test_htex/test_interchange_exit_bad_registration.py
yadudoc: My bad, I used `ruff` to auto-lint. I've rolled these back to match the existing linting style.

PR #4107, opened 2026-05-21
Use LogConfig mechanism for HighThroughputExecutor worker pools

benclifford:
This follows on from the same for the interchange in PR #4103

Log configurations are passed via file, in the same way as certificate configurations, and so in situations where filesystems force encrypted=False, you should expect to not be able to use log configurations.

This PR modifies the directory creation behaviour of the two file-based configurations, so that they will create parent directories too. This is needed because workers put logs in a deeper hierarchy than other places where log configuration is used.

# Changed Behaviour

DFK-level log configuration introduced in PR #4091 will be respected inside htex workers.

## Type of change

- New feature

### DIFF
--- a/parsl/executors/high_throughput/executor.py
+++ b/parsl/executors/high_throughput/executor.py
@@ -1,5 +1,6 @@
 import logging
 import math
+import os
 import pickle
 import subprocess
 import threading
@@ -51,6 +52,7 @@ DEFAULT_LAUNCH_CMD = ("process_worker_pool.py {debug} {max_workers_per_node} "
                       "--port={worker_port} "
                       "--cert_dir {cert_dir} "
                       "--logdir={logdir} "
+                      "--logconf={logconf} "
                       "--block_id={{block_id}} "
                       "--hb_period={heartbeat_period} "
                       "{address_probe_timeout_string} "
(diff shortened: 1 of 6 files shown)


benclifford:
@yadudoc you probably want to review this in the context of  Globus Compute cross-parsl compatibility: I can see that the additional pool argument is likely to cause problems, although there is a default and so in one direction that should be ok. To handle the other direction, I looked at making the templating not include the new parameter if would be the default value, but the commandline template isn't really set up for that. You should be able to fix that separately from this PR though as GC doesn't need *every* version of Parsl to be compatible, only the versions it chooses to pin.

 If/when GC wants to start using this feature, unpickling of configurations is also something to pay attention to.

parsl/executors/high_throughput/executor.py
```diff
@@ -446,6 +449,16 @@ def start(self):
                 "None or encrypted to True."
             )
 
+        if self.log_config:
+            log_config_dir = os.path.join(self.logdir, "log_config")
+            os.makedirs(log_config_dir, mode=0o700, exist_ok=True)
+
+            self.logconf_path = os.path.join(log_config_dir, "log_config.pkl")
+            with open(self.logconf_path, "wb") as f:
+                pickle.dump(self.log_config, f)
```
khk-globus, line 458: As a hard-coded `log_config.pkl` basename, should there be any concern for which configuration is related to which run for debugging?

khk-globus (approved):

parsl/executors/high_throughput/executor.py
benclifford: the log_config_dir is in a run directory and then a per-executor directory within that:

> runinfo/000/htex_Local/log_config/log_config.pkl

PR #4115, opened 2026-05-29
Type-related on worker count restrictions in htex worker pool

benclifford:
Previous to this, the worker count variables were a combination of integers and floats, but the only float used was +inf, as a marker to indicate a particular restriction wasn't applied.

Added individually, the type checker does not care about type annotations added onto max workers per node and cores per worker (because there remain enough implicit Anys) by this PR.

But when both added, there is/was a conflict:

parsl/executors/high_throughput/process_worker_pool.py:212: error: Incompatible types in assignment (expression has type "int | float", variable has type "int")  [assignment]

The revealed types for the three parameters of this min statement are:

        self.worker_count: int = min(max_workers_per_node,
                                     mem_slots,
                                     math.floor(cores_on_node / cores_per_worker))

parsl/executors/high_throughput/process_worker_pool.py:212: note: Revealed type is "builtins.int | builtins.float"
parsl/executors/high_throughput/process_worker_pool.py:213: note: Revealed type is "builtins.int | builtins.float"
parsl/executors/high_throughput/process_worker_pool.py:214: note: Revealed type is "builtins.int"

When calculating the worker count, this minimum should probably be over ints not floats.

So why are the first two int|float?

i) max_workers_per_node can be a float here solely for the +infinity value that is used to flag that it is unspecified.

ii) mem_slots can be a float solely because it is initialized to the value of max_workers_per_node and then potentially adjusted to an int after computation.

The cores_on_node/cores_per_worker is the only non-optional piece of this minimization.

The other two components of the min either do nothing (if they're +inf) or they reduce the worker count further. (that's what happens to *any* additional parameter to a min!)

This PR reworks the calculation to something more like a semi-group with `min` as the operation:

Start with the only definite limit (cores_on_node / cores_per_worker) and then optionally apply further limits (there are three further restrictions) when they apply using `min`.

When previously a float +inf was passed to min, that overall had a no-op effect (because it will always be larger than cores_on_node/cores_per_worker, for sensible values of that calculation). Those two cases are now handled by `if` statements: in one case, mem_slots, the `if` statement already existed!

TODO: ^ validate cores_on_node/cores_per_worker?

TODO: how to test?

# Description

# Changed Behaviour

# Fixes

Fixes # (issue)

## Type of change

- Bug fix
- New feature
- Code maintenance/cleanup

### DIFF
--- a/parsl/executors/high_throughput/process_worker_pool.py
+++ b/parsl/executors/high_throughput/process_worker_pool.py
@@ -75,9 +75,9 @@ class Manager:
                  addresses: str,
                  address_probe_timeout,
                  port,
-                 cores_per_worker,
+                 cores_per_worker: float,
                  mem_per_worker: Optional[float],
-                 max_workers_per_node,
+                 max_workers_per_node: Optional[int],
                  prefetch_capacity,
                  uid,
                  block_id,
@@ -117,7 +117,7 @@ class Manager:
              the there's sufficient memory for each worker. If set to None, memory on node is not
              considered in the determination of workers to be launched on node by the manager.
 
-        max_workers_per_node : int | float
+        max_workers_per_node : optional int
              Caps the maximum number of workers that can be launched.
 
         prefetch_capacity : int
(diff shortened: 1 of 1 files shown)


benclifford:
I did a more serious refactor of this code, in PR #2119. closing this PR in favour of that PR.

PR #4118, opened 2026-05-29
Move some HTEX specific tests into test_htex

benclifford:
Two tests move from sites/ - they are testing htex-worker specific behaviour that other Parsl executors would not provide.

All of test_mpi moves. test_mpi doesn't test Parsl's general MPI capabilities, but only the HTEX specific pieces. Compare with the test of Radical's MPI pieces under parsl/tests/test_radical/

# Changed Behaviour

none for users. but if you're developing htex and only running test_htex tests, you'll get more coverage now.

## Type of change

- Code maintenance/cleanup


khk-globus (approved):
Well that was an easy review ...

PR #4122, opened 2026-06-29
Make log initialization ref-counted

benclifford:
This PR brings in a feature that I didn't include in the initial Parsl log configuration work, but that I did over in the Academy equivalent of this work.

This PR introduces a distributed identity to log config objects (using a UUID) so that they may be serialised and deserialised by various paths and retain their identity.

This identity is then used to reference count the same configuration being initialized in a Python interpreter by multiple paths, so that a particular configuration is only initialized once, but kept alive for as along as someone needs it.

The motivating usecase in Parsl is:

When running tasks via htex, the htex code is LogConfig aware and will initialize each worker with the DFK log configuration (PR #4107), and in the thread pool executor, tasks will run in the main process which will already have the DFK log configuration initialized. But when running in some arbitrary not-LogConfig aware executor, Parsl will need to wrap tasks in the same way that it wraps tasks for monitoring (a future PR).

This reference counting is the mechanism by which both of those initializations can happen without actually duplicating the initializations.

# Changed Behaviour

none

## Type of change

- New feature

### DIFF
--- a/parsl/logconfigs/base.py
+++ b/parsl/logconfigs/base.py
@@ -1,8 +1,13 @@
 from __future__ import annotations
 
 import abc
+import logging
 import pathlib
+import threading
+import uuid
 from collections.abc import Callable
+
+logger = logging.getLogger(__name__)
 
 
 class LogConfig(abc.ABC):
@@ -17,6 +22,9 @@ class LogConfig(abc.ABC):
     be pickled/unpickled multiple times as they move around a Parsl
     distributed system.
     """
+
+    def __init__(self) -> None:
+        self.uuid = uuid.uuid4()
 
     @abc.abstractmethod
     def initialize_logging(self, *, log_dir: pathlib.Path, log_name: str) -> Callable[[], None]:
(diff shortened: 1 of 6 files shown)


parsl/logconfigs/base.py
```diff
@@ -33,3 +41,77 @@ def initialize_logging(self, *, log_dir: pathlib.Path, log_name: str) -> Callabl
...
+
+    There is an ambiguity on the use of log_name and log_dir across multiple
+    initializations: the name and directory of the first initialization
+    will be used and the name and directory of subsequent initializations
+    will be ignored.
+    """
+    with _initialized_log_context_lock:
+        if log_config.uuid in _initialized_log_contexts:
+            _initialized_log_contexts[log_config.uuid].count += 1
+            logger.debug('Configuration %r already initialized', log_config)
```
khk-globus: - Might also add the refcount value, since this is at DEBUG level.

parsl/logconfigs/base.py
```diff
@@ -33,3 +41,77 @@ def initialize_logging(self, *, log_dir: pathlib.Path, log_name: str) -> Callabl
...
+    def uninit() -> None:
+        with _initialized_log_context_lock:
+            assert log_config.uuid in _initialized_log_contexts, f"log context for {log_config} not found, perhaps too many uninitializations"
+            assert _initialized_log_contexts[log_config.uuid].count >= 1, \
+                f"log context for {log_config} has bad reference count {_initialized_log_contexts[log_config.uuid].count}"
+            _initialized_log_contexts[log_config.uuid].count -= 1
+            if _initialized_log_contexts[log_config.uuid].count < 1:
+                logger.debug('All references removed for %r', log_config)
+                _initialized_log_contexts[log_config.uuid].uninit_callback()
+                del _initialized_log_contexts[log_config.uuid]
```
khk-globus: With `_initialized_log_context_lock` and `_initialized_log_contexts` being module-global, I find myself wondering if this function couldn't also be a module-level function.  The only datastructure of consequence enclosed in the closure is `log_config.uuid`, a UUID.

I also observe that whether to invoke the callback is up to the caller.  If it doesn't store the returned closure, the closure is thrown away (no lazy evaluation here!), and the caller already has all that's the necessary: the UUID value (in the `log_config` object).

parsl/tests/test_logging/test_oneshot.py
```diff
@@ -0,0 +1,139 @@
...
+    cb.assert_not_called()
+    uninit()
+    cb.assert_called()
+
+    assert b._initialized_log_contexts == {}, "back to satisfying the precondition"
+
+
+@pytest.mark.local
+def test_nested(tmpd_cwd):
+    lc = Mock()
```
khk-globus: If you want to make sure that the LogConfig object is accessed "correctly", you might appreciate `spec`:

```python
lc = Mock(spec=LogConfig)
```

Not always warranted, but good to keep in mind.

parsl/tests/test_logging/test_oneshot.py
```diff
@@ -0,0 +1,139 @@
...
+    cb.assert_not_called()
+    uninit2()
+    cb.assert_called()
+
+    assert b._initialized_log_contexts == {}, "back to satisfying the precondition"
+
+
+@pytest.mark.local
+def test_sequential(tmpd_cwd):
+    """Test that if a log context can be released completely, and then reinitialized."""
```
khk-globus: Grammar?

parsl/tests/test_logging/test_oneshot.py
```diff
@@ -0,0 +1,139 @@
...
+    lc.initialize_logging.assert_called_once()
+
+    uninit2 = b.oneshot_initialize_logging(log_config=lc, log_dir=tmpd_cwd, log_name="test_overlap_2")
+    lc.initialize_logging.assert_called_once()
+
+    cb.assert_not_called()
+    uninit()
+    cb.assert_not_called()
+    uninit2()
+    cb.assert_called()
```
khk-globus: If it's not super expensive, one approach I like to take in tests as a measure of "nothing up my sleeve" is to use a random number of objects in places like this.  Consider:

```python
uninits = [
    b.oneshot_initialize_logging(
        log_config=lc,
        log_dir=tmpd_cwd,
        log_name=f"test_overlap_{num}"
    )
    for num in range(1, random.randint(1, 15))  # some number; maybe parameterized
]
cb = b._initialized_log_contexts[lc.uuid].uninit_callback

# Pending the other comment, currently all returned closures are unique, so could assert that
# assert len(set(uninits)) == len(uninits)

while len(uninits) > 1:
    uninits.pop(0)()  # In this case, could even take randomly, right?
    assert not cb.called, "Refcount > 0; expect not cleaned up!"
uninits.pop(0)()
assert cb.called, "At refcount == 0, should clean up!"
```

parsl/tests/test_logging/test_oneshot.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.mark.local
+def test_overlap(tmpd_cwd):
+    lc = Mock()
+    lc.uuid = uuid.uuid4()
+
+    assert b._initialized_log_contexts == {}, "precondition"
+
+    uninit = b.oneshot_initialize_logging(log_config=lc, log_dir=tmpd_cwd, log_name="test_overlap_1")
+    cb = b._initialized_log_contexts[lc.uuid].uninit_callback
+    lc.initialize_logging.assert_called_once()
```
khk-globus: This, the first assert, is perhaps redundant?  Maybe more plain as to intent to assert simply that it was called at all?
```python
assert lc.initialize_logging.called
```
The next `assert_called_once()` or `assert ...call_count == 1` confirms the intent.

khk-globus (approved):
Makes sense.

A couple of inline comments and suggestions to take or leave, but not blockers.

parsl/logconfigs/base.py
benclifford: done

parsl/logconfigs/base.py
benclifford: yeah i think that makes sense to change.

The history is that this was exposed as a "we don't know how to uninitialize the logging, but the thing that initialized it does, through this callback" callback (in the style of LogConfig), and this uninit call kept that same API. But no one should ever be calling any of those callbacks except the reference count manager...

benclifford: I've moved callbacks out of the DFK, in f6094db2c4. The DFK code still uses callbacks for its log management but maybe that can be tidied a bit. This is basically to support previous behaviour which isn't expressable with the new log config code (with a log configuration only used by the DFK)

parsl/tests/test_logging/test_oneshot.py
benclifford: done in e133f51b52

parsl/tests/test_logging/test_oneshot.py
benclifford: fixed in a33dbae6de

PR #4123, opened 2026-06-29
Log workflow monitoring metadata via logging

benclifford:
This is part of an observability theme of making sure that all of the data available through parsl monitoring will also be available as structured extra fields in the logging system.

This PR makes a pretty straightforward transcription of the WORKFLOW_INFO message (which then goes into the monitoring workflows table) directly into extra= fields of a workflow start/end message.

# Changed Behaviour

more logs

## Type of change

- New feature

### DIFF
--- a/parsl/dataflow/dflow.py
+++ b/parsl/dataflow/dflow.py
@@ -174,6 +174,8 @@ class DataFlowKernel:
                 'host': gethostname(),
         }
 
+        loggable_workflow_info = {"parsl.workflow_info." + k: v for k, v in workflow_info.items()}
+        logger.info("Workflow startup information: %r", workflow_info, extra={"parsl.dfk": self.run_id} | loggable_workflow_info)
         if self.monitoring_radio:
             self.monitoring_radio.send((MessageType.WORKFLOW_INFO,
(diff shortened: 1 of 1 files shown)


parsl/dataflow/dflow.py
```diff
@@ -174,6 +174,8 @@ def __init__(self, config: Config) -> None:
                 'host': gethostname(),
         }
 
+        loggable_workflow_info = {"parsl.workflow_info." + k: v for k, v in workflow_info.items()}
+        logger.info("Workflow startup information: %r", workflow_info, extra={"parsl.dfk": self.run_id} | loggable_workflow_info)
```
khk-globus, line 178: `k` is clearly always a string, given construction just above, so this is just a stylistic choice and doesn't matter, but I'm curious the choice to use the plus (`+`) binary operator:

```python
"parsl.workflow_info." + k: v
```

to make the key, rather than either of an f-string or modulo-string approach:

```python
f"parsl.workflow_info.{k}": v
"parsl.workflow_info.%s" % s: v
```

khk-globus (approved):
LGTM

PR #4125, opened 2026-06-29
Add structured logging for add_executor

benclifford:
This is intended to help log analysis code understand what other information it might find (for example, the output of the work queue executor is very different to the output of the high throughput executor).

Although this call is exposed to the user for dynamic execution, it is more commonly used implicitly as part of initial DFK configuration. The test modified in this PR checks both of those paths.

# Changed Behaviour

more logs

## Type of change

- New feature


khk-globus (approved):
Looks reasonable.

I'm struck by `type(executor)` invoked 4 times, but I'm not fussed about it.

PR #4126, opened 2026-07-03
Add some ad-hoc htex max worker calculation tests

benclifford:
This PR is intended to sit alongside PR #4119 which refactors max worker calculations.

These tests are testing the executor side of max worker calculation, which has no access to the worker environment and works entirely from user supplied parameters - so there is no need to be mocking anything here.

This set of tests does not test the worker pool side of the calculation.

These tests are not particularly thorough but making them more complex ends up mostly reimplementing the calculation code a second time.

# Changed Behaviour

none

## Type of change

- Code maintenance/cleanup

### DIFF
--- a/parsl/tests/test_htex/test_maxworker_computation_executor.py
+++ b/parsl/tests/test_htex/test_maxworker_computation_executor.py
@@ -1,0 +1,89 @@
+import random
+
+import pytest
+
+import parsl
+from parsl.config import Config
+from parsl.executors.high_throughput.executor import HighThroughputExecutor
+from parsl.providers import LocalProvider
+
+
+@pytest.mark.local
+def test_max_workers_per_node():
+
+    w = random.randint(1, 8)
+    h = HighThroughputExecutor(max_workers_per_node=w,
+                               provider=LocalProvider(max_blocks=0, init_blocks=0, min_blocks=0)
+                               )
+    c = Config(executors=[h])
+
+    with parsl.load(c):
+        # tests the parameter was wired all the way through
+        assert h.max_workers_per_node == w
+
+        # tests that the calculation chose the same number
+        # in the absence of other restrictions
+        assert h.workers_per_node == w
+
+
+@pytest.mark.local
+def test_cores():
+
+    p = LocalProvider(max_blocks=0, init_blocks=0, min_blocks=0)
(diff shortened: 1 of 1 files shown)


parsl/tests/test_htex/test_maxworker_computation_executor.py
```diff
@@ -0,0 +1,89 @@
...
+                               )
+    c = Config(executors=[h])
+
+    with parsl.load(c):
+        # tests the parameter was wired all the way through
+        assert h.max_workers_per_node == w
+
+        # tests that the calculation chose the same number
+        # in the absence of other restrictions
+        assert h.workers_per_node == w
```
khk-globus, line 26: Props if you move the comments to the assertion message:
```python
assert h.max_workers_per_node == w, "Verify setting properly wired in"
```
I have this same comment for all of the `assert`ions, so I'll not leave this comment again.

parsl/tests/test_htex/test_maxworker_computation_executor.py
```diff
@@ -0,0 +1,89 @@
...
+                               cores_per_worker=4
+                               )
+    c = Config(executors=[h])
+
+    with parsl.load(c):
+        # tests the parameter was wired all the way through
+        assert h.cores_per_worker == 4
+
+        # 8 cores divided out at 4-per-worker
+        assert h.workers_per_node == 2
```
khk-globus, line 46: You might consider parameterizing this test to verify a number of calculations.  Assuming the HTEX is fast to initialize, perhaps something like:
```python
@pytest.mark.local
@pytest.mark.parametrize("mwpn", range(1, 16))  # for discussion ...
@pytest.mark.parametrize("cpn", range(1, 8))    # ... 8*16 is probably too many
def test_cores(mwpn: int, cpn: int):
    p = LocalProvider(...)
    p.cores_per_node = cpn
    for cpw in range(1, cpn):
        h = HighThroughputExecugtor(
              max_workers_per_node=mwpn,
              provider=p,
              cores_per_worker=cpw,
        )
    ...
    assert h.workers_per_node == ...
```

khk-globus (approved):
I'm not going to harp on it, but I think some parameterization would make these much stronger tests.

PR #4131, opened 2026-07-06
Test HTCondor provider in CI

benclifford:
# Description

This adds a CI test for the Condor provider against an HTCondor-supplied docker image, in the same style as the slurm CI test.

# Changed Behaviour

more testing

## Type of change

- New feature


khk-globus (approved):
Yay!  Good job on finding a CI testing route.

PR #4136, opened 2026-07-12
Fix case-typo in test

benclifford:
# Changed Behaviour

tests will test slightly differently now -- the managers packages command should be properly invoked

## Type of change

- Bug fix

### DIFF
--- a/parsl/tests/test_htex/test_command_concurrency_regression_1321.py
+++ b/parsl/tests/test_htex/test_command_concurrency_regression_1321.py
@@ -44,7 +44,7 @@ def blast(cc, e):
     target_end = time.monotonic() + DURATION_S
 
     while time.monotonic() < target_end:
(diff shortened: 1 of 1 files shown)


parsl/tests/test_htex/test_command_concurrency_regression_1321.py
```diff
@@ -44,7 +44,7 @@ def blast(cc, e):
     target_end = time.monotonic() + DURATION_S
 
     while time.monotonic() < target_end:
-        cc.run("MANGERs_PACKAGES")
+        cc.run("MANGERS_PACKAGES")
         cc.run("CONNECTED_BLOCKS")
```
khk-globus, line 48: An opaque-string protocol; makes sense this was overlooked for so long.  Yeesh.

khk-globus (approved):

